PR #6, opened 2017-03-01
Session refactor

mayfield:
This is a review seeking PR for @smtakeda (and @blandonnimrat)

It does test well for me on python3.5 and python2.7 but I would consider it as part of ongoing work to increase download concurrency safely.  Please feel free to provide feedback and criticism here and/or merge if you are happy enough with the intent.

I took some liberty in renaming existing mechanisms where it helped me make sense of the various pieces falling together but I can backtrack on them if they're undesired (named `access_url` -> `fetch`).

There is more overview in the commit logs as well.

### DIFF
--- a/network.py
+++ b/network.py
@@ -3,6 +3,8 @@
 #
 # Copyright (c) 2012-2017 Snowflake Computing Inc. All right reserved.
 #
+import collections
+import contextlib
 import copy
 import gzip
 import json
@@ -16,7 +18,7 @@ from threading import Thread
 
 import OpenSSL
 from botocore.vendored import requests
-from botocore.vendored.requests.adapters import (HTTPAdapter, DEFAULT_POOLSIZE)
+from botocore.vendored.requests.adapters import HTTPAdapter
 from botocore.vendored.requests.auth import AuthBase
 from botocore.vendored.requests.exceptions import (ConnectionError, SSLError)
 from botocore.vendored.requests.packages.urllib3.exceptions import (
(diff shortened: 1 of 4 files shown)


smtakeda:
Thanks @mayfield! I'm OOO this week, so I'll check out by next week.

network.py
```diff
@@ -181,7 +188,17 @@ def close(self):
             del self._token
         if hasattr(self, u'_master_token'):
             del self._master_token
-        self._session = None
+        sessions = list(self._active_sessions)
+        if sessions:
+            self.logger.warn("Closing %d active sessions" % len(sessions))
```
smtakeda: Let's use parameterized logging, e.g., ```self.logger.warn("Closing %s active sessions", len(sessions))```

mayfield: I only avoid that syntax because it only works for logging functions and I sometimes switch back and forth with other printers like print().  But I can use printf style formatting from now on.

network.py
```diff
@@ -641,25 +604,40 @@ def _post_request(self, url, headers, body, token=None,
...
-                   is_raw_binary_iterator=True,
-                   use_ijson=False, is_single_thread=False):
-        logger = getLogger(__name__)
-
-        connection_timeout = timeout[0:2]
-        request_timeout = timeout[2]  # total request timeout
-        request_thread_timeout = 60  # one request thread timeout
-
-        def request_thread(result_queue):
+    def fetch(self, method, full_url, headers, *, data=None, timeout=None, **kwargs):
```
mayfield: Interesting, I was pretty sure I tested that particular signature on 2.7, but I do recall that keyword only arguments handling changed in py3k.  (Disclaimer: I rarely use python2)

network.py
```diff
@@ -889,11 +867,18 @@ def request_thread(result_queue):
                 })
         return return_object
 
+    def make_requests_session(self):
+        s = requests.Session()
+        s.mount(u'http://', HTTPAdapter(max_retries=REQUESTS_RETRY))
+        s.mount(u'https://', HTTPAdapter(max_retries=REQUESTS_RETRY))
+        s._reuse_count = itertools.count()
```
mayfield: Yeah, I was using it before these commits were pushed.  I'm probably going to implement session garbage collection and may use reuse_count as a mechanism for deciding when to close out old sessions.  But if not I'll yank before next PR.  Thanks!

PR #7, opened 2017-03-04
Remove redef of settimeout in ssl code.

mayfield:
settimeout was defined twice.


smtakeda:
yeah, I noticed a while back... thanks for the report. Will merge soon.

PR #8, opened 2017-03-04
Py2k syntaxerror fix

mayfield:
I can't get py2k installs to work but I validated the `python -m py_compile *.py` in the project root is now clean.


smtakeda:
Thanks. I already patched this in master. Please sync with it.
So skipping this merge request.

PR #12, opened 2017-03-15
Row perf

mayfield:
Micro optimizations for hot path code used to convert JSON raw data to native python data structures.  Most of the optimizations come from reducing function call overhead (by way of closures) and by more efficient use of `datetime.datetime` factories like `fromtimestamp()`. 

EDIT: Benchmarks.

In a micro benchmark with heavy datetime data sets and S3 downloads cached I see the following numbers..

Stock - Py3.5: 21.9s
Stock - Py3.6: 17.0s
This PR - Py3.5: 12.7s (72% faster)
This PR - Py3.6: 9.5s (79% faster)

These number are for 500k rows and include startup times.  The actual row processing speeds are over 100% faster.

### DIFF
--- a/chunk_downloader.py
+++ b/chunk_downloader.py
@@ -259,8 +259,6 @@ class SnowflakeChunkDownloader(object):
             self._connection._connect_timeout,
             DEFAULT_REQUEST_TIMEOUT
         )
-        data = self._connection.rest.fetch(u'get', url, headers,
+        return self._connection.rest.fetch(u'get', url, headers,
             timeouts=timeouts, is_raw_binary=True,
-            is_raw_binary_iterator=False, use_ijson=self._use_ijson)
-        return iter([self._cursor.row_to_python(x) if x is not None else None
-                     for x in data])
+            is_raw_binary_iterator=True, use_ijson=self._use_ijson)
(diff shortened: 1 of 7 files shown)


smtakeda:
Timestamp match issue is very likely due to http://bugs.python.org/issue23517
We use Python 3.4.3 and 2.7.10 internally. Here is a test code:

```
from datetime import datetime

ts='58510.989102000'
ret = datetime.utcfromtimestamp(float(ts)).time()
print('   ' + str(ts))
print(str(ret))
```
The results are:
```
>python2.6 test_time_precision.py
   58510.989102000
16:15:10.989102
>python2.7 test_time_precision.py  
   58510.989102000
16:15:10.989102
>python3.3 test_time_precision.py 
   58510.989102000
16:15:10.989101
>python3.4 test_time_precision.py 
   58510.989102000
16:15:10.989101
>python3.5 test_time_precision.py 
   58510.989102000
16:15:10.989102
>python3.6 test_time_precision.py 
   58510.989102000
16:15:10.989102
```

smtakeda:
I would like a have Python 3.4.3 special converter to workaround http://bugs.python.org/issue23517.

smtakeda:
3.5.0 has the same issue. 3.6.0 has the fix.

https://docs.python.org/3.4/whatsnew/changelog.html#python-3-4-4
https://docs.python.org/3.5/whatsnew/changelog.html#python-3-5-1-final
https://docs.python.org/3.6/whatsnew/changelog.html#python-3-6-0

PR #18, opened 2017-05-11
SNOW-27327 Added support for brotli and zstd in put statements

smtakeda:
Adding zstd and brotli compression type support for PUT.

This cannot be merged into master until the server enables this feature.


smtakeda:
Confirmed the features are enabled. Investigating the test failures.

PR #19, opened 2017-06-06
setup.py: use less restrictive cryptography>=1.7,<1.8

mikek:
Sometimes there is no need or possibility (because of the package or setup-specific issues) to change the cryptography minor versions. This allows us to be less restrictive.


smtakeda:
Yeah, this should work... I'll verify the change later this week. Thanks.

smtakeda:
This will be included in 1.3.18 this Thursday. Thanks.

PR #25, opened 2017-07-11
SNOW-31396: Attempt to fix large PUT and GET tests.

smtakeda:
A couple of major changes were made:
- Skip scanning all of the existing files before PUT to detect the duplicated files. Instead, check the existence of file on each PUT command to reduce overhead.
- Right after PUT a file complete, it checks if the file really accessible. 

Other than the above changes, a colleague added a new internal flag DISABLE_PUT_AND_GET_ON_EXTERNAL_STAGE for new customers to disable PUT and GET commands against the external stages. This is mainly because the external stages are not guaranteed to be secure and encrypted.

### DIFF
--- a/chunk_downloader.py
+++ b/chunk_downloader.py
@@ -260,6 +260,7 @@ class SnowflakeChunkDownloader(object):
             self._connection._connect_timeout,
             DEFAULT_REQUEST_TIMEOUT
         )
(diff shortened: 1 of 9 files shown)


smtakeda:
This was directly merged in master.

PR #46, opened 2018-01-12
Changed column names to lower case on Cursor column descriptors only.

pete0877:
@smtakeda after making these changes all the tests passed. Next we will try to use this version of the connector in our code base to see if everything magically starts working for us. 

Just wanted to have you take a very quick look at this fork / PR and see if you have any comments. 

All tests passed ("195 passed, 32 ignored")

### DIFF
--- a/cursor.py
+++ b/cursor.py
@@ -558,9 +558,9 @@ class SnowflakeCursor(object):
         self._column_idx_to_name = {}
         self._column_converter = []
         for idx, column in enumerate(data[u'rowtype']):
-            self._column_idx_to_name[idx] = column[u'name']
+            self._column_idx_to_name[idx] = column[u'name'].lower()
             type_value = FIELD_NAME_TO_ID[column[u'type'].upper()]
-            self._description.append((column[u'name'],
+            self._description.append((column[u'name'].lower(),
                                       type_value,
                                       None,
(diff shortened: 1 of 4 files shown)


smtakeda:
Hi @pete0877, sorry but maybe my explanation was clear.

First of all, I cannot accept this PR as the change will obviously break the compatibility for the existing customers. Secondly, my suggestion was either you make changes in your framework to translate the cases like SQLAlchemy does or folk this code for you to maintain on your end.

Again Snowflake connector itself does right behavior, and any difference to the DB should be handled outside of the connector itself. That's the whole point of SQLAlchemy dialect for Snowflake: https://github.com/snowflakedb/snowflake-sqlalchemy

Hope this clarifies, but if we still need discussions, please let us know. Thank you.

smtakeda:
Anyway it is good to know that's all changes you need to make. If that's the case, I bet changing your framework would not be too tough.

smtakeda:
Ok, I just emails that your intention was not PR but code review. The change looks good to me.

pete0877:
Shige, thanks - that's clear. Yeah, the idea would be for Pixability to just maintain its fork and rebase once in a while.  Sorry about the PR / merge confusion. I selected wrong repo as the target to show the diffs.

PR #48, opened 2018-01-18
Bump pyasn upper bound to 0.5.0

baxen:
The old upper bound of 0.4.0 conflicted with several google-cloud packages. But I don't see any reason why it shouldn't be compatible through 0.4.2.

I tested much of the Connector functionality with more recent versions of pyasn1, and did not run into any errors. However I didn't have access to run the tests in this repo (parameters.py was required and is encrypted).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,8 @@ setup(
         'pyOpenSSL>=16.2.0,<18.0.0',
         'cffi>=1.9',
         'cryptography>=1.7.2,<2.2',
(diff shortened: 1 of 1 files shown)


smtakeda:
Sure, I'll incorporate your change in the next set of changes.
Actually I'm removing the dependency of `pyasn1` by rewriting OCSP stuff with `asn1crypto`.
In either case, the next version will keep `pyasn1` anyway.

PR #49, opened 2018-01-18
Started replacing pyasn1 with asn1crypto. Not activated yet

smtakeda:
Refactored ocsp_pyopenssl using `asn1crypto`


smtakeda:
Fixing #48 as part of this change.

PR #79, opened 2018-04-03
Removes unnecessary session request types

tjj5036:
These aren't used anywhere in the driver, I think it's safe to remove them.


smtakeda:
yeah, I agree.

PR #86, opened 2018-05-30
Removes username restriction for OAuth.

tjj5036:
Usernames should not be required when authenticating with an OAuth Access Token. This changeset removes the username restriction in the driver for when the OAuth Authenticator is used.

Tested via the following:

```
import snowflake.connector
cnx = snowflake.connector.connect(
    account='testaccount',
    host='https://myaccount.snowflakecomputing.com',
    authenticator='OAUTH',
    token=oauth_access_token)
cnx.close()
```

### DIFF
--- a/connection.py
+++ b/connection.py
@@ -543,7 +543,8 @@ class SnowflakeConnection(object):
             if u'protocol' not in kwargs:
                 setattr(self, u'_protocol', u'https')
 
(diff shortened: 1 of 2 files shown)


smtakeda:
ok, working on the merge.

smtakeda:
done in #89

PR #109, opened 2018-08-25
Update Boto3 to <1.9.0

Fokko:
Hi all,

I'd like to loosen the  version requirement of Boto3 since we're using 1.8.1 in the Apache Airflow project.

Cheers, Fokko


smtakeda:
will work on it.

smtakeda:
botocore seems to have the update.

PR #117, opened 2018-09-19
Fix multiline double quote expressions

bensowden:
Previously the code would drop all contents of a multi-line double quote expression when remove comments is true, this fix means that all contents within the double quotes get propagated.

For example this SQL:
```
CREATE FUNCTION mean(a FLOAT, b FLOAT)
  RETURNS FLOAT LANGUAGE JAVASCRIPT AS $$
  var c = a + b;
  return(c / 2);
  $$;
```
would be returned as:
```
CREATE FUNCTION mean(a FLOAT, b FLOAT)
  RETURNS FLOAT LANGUAGE JAVASCRIPT AS $$
  $$;
```


smtakeda:
Thanks for PR. Will look into...

Issue #37, opened 2017-09-27
maximum number of expressions in a list exceeded, expected at most 16,384, got 190,800


JDFagan:
Snowflake connector seems to have limitation of accepting large sets at once (> 16,384 items).  Would be nice to substantially increase this limit.  My code is simply trying to upload a large (190k) csv file via use of Agate and SQLAlchemy.  This code works fine when targeting against RDS MySQL 5.6.

```
Creating transformer_loader for /Users/jdfagan/tmp/property_ledger/LIONSTONEGRP/transform/GLSummary.csv ...
Loading source: /Users/jdfagan/tmp/property_ledger/LIONSTONEGRP/transform/transform/GLSummary.csv ...
database table: LIONSTONEGRP_GLSummary
| column     | data_type |
| ---------- | --------- |
| AccountID  | Number    |
| AcctNum    | Text      |
| EntityID   | Text      |
| Department | Text      |
| Basis      | Text      |
| Period     | Text      |
| PeriodDate | Date      |
| Balfor     | Text      |
| Activity   | Number    |

FAILED
etl/mri/test_property_ledger_etl.py:109 (test_MriPropertyLedgerETL_load[LIONSTONEGRP])
self = <sqlalchemy.engine.base.Connection object at 0x10fe42438>
dialect = <snowflake.sqlalchemy.base.SnowflakeDialect object at 0x10fe42080>
constructor = <bound method DefaultExecutionContext._init_compiled of <class 'snowflake.sqlalchemy.base.SnowflakeExecutionContext'>>
statement = 'INSERT INTO "LIONSTONEGRP_GLSummary" ("AccountID", "AcctNum", "EntityID", "Department", "Basis", "Period", "PeriodDat...countID)s, %(AcctNum)s, %(EntityID)s, %(Department)s, %(Basis)s, %(Period)s, %(PeriodDate)s, %(Balfor)s, %(Activity)s)'
parameters = ({'AccountID': Decimal('1001000'), 'AcctNum': '1001-000', 'Activity': Decimal('-99.37'), 'Balfor': 'B', ...}, {'Accoun......}, {'AccountID': Decimal('1001000'), 'AcctNum': '1001-000', 'Activity': Decimal('300.00'), 'Balfor': 'B', ...}, ...)
[...]

PR #118, opened 2018-09-19
individually resolve Snowflake type for each list element

rhlahuja:
Without resolving the Snowflake type for each element in the list comp individually, passing a list of dicts to [`bulk_insert_mappings`](http://docs.sqlalchemy.org/en/latest/orm/session_api.html#sqlalchemy.orm.session.Session.bulk_insert_mappings) using `snowflake-sqlalchemy` with `snowflake.connector.paramstyle = u'qmark'` (to get around the [16384 row limit](https://github.com/snowflakedb/snowflake-connector-python/issues/37)) always results in the `snowflake_type` being set to `TEXT` [over here](https://github.com/snowflakedb/snowflake-connector-python/blob/6abc56c970cdb698a833b7f6ac9cbe7dfa667abd/connection.py#L779-L780) because `v` is always a list and [lists resolve to TEXT](https://github.com/snowflakedb/snowflake-connector-python/blob/6abc56c970cdb698a833b7f6ac9cbe7dfa667abd/converter.py#L60).

When inserting into a `TIMESTAMP_NTZ(9)` column, for example, we encounter:
```
snowflake.connector.errors.ProgrammingError: 255001: Binding datetime object with Snowflake data type TEXT is not supported.
```

### DIFF
--- a/connection.py
+++ b/connection.py
@@ -795,7 +795,7 @@ class SnowflakeConnection(object):
                 if isinstance(v, list):
                     vv = [
                         self.converter.to_snowflake_bindings(
-                            snowflake_type, v0) for v0 in v]
+                            self.converter.snowflake_type(v0), v0) for v0 in v]
                 else:
                     vv = self.converter.to_snowflake_bindings(
                         snowflake_type, v)


smtakeda:
Thanks for PR. Would you give a test case? Even a small code snippet would be helpful.

rhlahuja:
@smtakeda I couldn't get the tests to run locally, but added a test case which should cover this. The existing tests weren't sufficient because they only tested with STRING columns which doesn't trigger this bug. Let me know if I can further clarify anything.

smtakeda:
Perfect!

Issue #124, opened 2018-10-02
Optional dependencies for azure and boto?


mihaico:
As user not intending to use boto or azure modules, e.g. in case of using the connector for simple queries to Snowflake, is it be possible to relax the dependency requirements for these particular modules?

smtakeda:
Not in the current todo list. I would appreciate if you submit your suggestion by PR.

PR #128, opened 2018-11-01
Move to optional dependencies

theianrobertson:
(partial) fix for #124.

Before merge I wanted to start a discussion about exceptions too - we'd need to raise some exceptions on import when packages are missing and necessary. Happy to take a crack at it, but just wanted to confirm: I imagine you'd want to try/except on azure/boto imports, would you want a separate error class or just raise something like `NotSupportedError`?

Would need to tweak docs here too: https://docs.snowflake.net/manuals/user-guide/python-connector-install.html#step-1-install-the-connector, something like:

To install the connector, execute one of the following commands:

- `pip install --upgrade snowflake-connector-python` (just basic functionality)
- `pip install --upgrade snowflake-connector-python[aws]` (To load/unload when hosted on AWS)
- `pip install --upgrade snowflake-connector-python[azure]` (To load/unload when hosted on Azure)

### DIFF
--- a/azure_util.py
+++ b/azure_util.py
@@ -5,12 +5,17 @@ import os
 from collections import namedtuple
 from logging import getLogger
 import json
-from azure.storage.blob import BlockBlobService
-from azure.common import (AzureMissingResourceHttpError, AzureHttpError)
-from azure.storage.blob.models import ContentSettings
+
+try:
+    from azure.storage.blob import BlockBlobService
+    from azure.common import (AzureMissingResourceHttpError, AzureHttpError)
+    from azure.storage.blob.models import ContentSettings
+except ImportError:
+    from .errors import NotSupportedError
+    raise NotSupportedError('To access PUT/GET for Azure please install with [azure] tag.')
+
 from .constants import (SHA256_DIGEST, ResultStatus, FileHeader)
 from .encryption_util import (EncryptionMetadata)
-
 
 """
 Azure Location: Azure container name + path
--- a/file_transfer_agent.py
+++ b/file_transfer_agent.py
@@ -35,7 +35,6 @@ from .local_util import SnowflakeLocalUtil
 from .remote_storage_util import (SnowflakeFileEncryptionMaterial,
                                   SnowflakeRemoteStorageUtil,
                                   )
-from .s3_util import SnowflakeS3Util
 
 S3_FS = u'S3'
 AZURE_FS = u'AZURE'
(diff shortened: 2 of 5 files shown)


smtakeda:
Thank you for PR! Honestly I didn't know the setup tools Extras can be utilized.
yes, `NotSupportedError` would work.

One question is what does basic functionality mean?
```
pip install --upgrade snowflake-connector-python
```
No PUT or GET command?

theianrobertson:
I think that's what it would mean yeah, no PUT or GET.  The S3- and Azure-specific functionality looks like it's limited mostly to file transfer stuff.  Couple of questions though, that may be preference-based:

- In `file_transfer_agent.py` there's `SnowflakeS3Util.DATA_SIZE_THRESHOLD:  # TODO fli: s3 specific?` --> That looks like it'll be called whether it's Azure or AWS, but sounds like there's maybe a question about whether it applies for Azure?  If so, it probably belongs outside of the `SnowflakeS3Util` class.
- Looks like you're using the requests from botocore in e.g. `network.py` - I just bumped botocore back to a strong dependency, either that or you could add in `requests` itself, not sure what your preference is.

I was thinking that the checking for boto3/azure deps could be done in `s3_util.py`/`azure_util.py` respectively, then just move the imports to right where the functionality is called.  Thoughts?

Happy to poke around with test cases for this if you'd like, let me know.

smtakeda:
hmm, it is preferable that `snowflake-connector-python` includes both AWS and Azure packages by default to keep the current behavior, and optionally it could remove either one or both if needed.

Azure part is being cleaned up at the moment. Thanks for catch anyway.

Re: botocore's request dependency is somewhat required until we find a better solution of OCSP patch. It is currently monkey patched in `urllib3`.

theianrobertson:
Cool, yeah that definitely makes sense.  Unfortunately I can't figure out an easy way to do that - I think `extras_require` can only add, not remove dependencies, and I couldn't get anything to work to short-cut `setup.py`.

I suppose the best option for people right now if they want to not install Azure or AWS dependencies would be to install the non-provider-specific dependencies themselves manually, then `pip install snowflake-connector-python --no-deps` to ignore the `install_requires` in `setup.py`.  I think that'd probably work for my purposes, maybe that could be the response to #124?  If it gets unruly, especially if you're supporting other backends, maybe the package itself could be split.

I'll close this PR for now, maybe someone else has a slick way of solving this.  Thanks for taking the time to respond!

smtakeda:
Thanks for PR anyway. One way is split packages into 4 at least, and `snowflake-connector-python` would become a wrapper of all three packages... but I would like to do more investigations.

PR #149, opened 2019-01-11
Added Mac and Windows tests. Added Azure Tests

smtakeda:
This is the snapshot of Python Connector change that enabled Mac and Windows tests on Travis and Appveyor.


smtakeda:
@ankit-bhatnagar167 I would like to push this change earlier so that we can run tests on Mac and Windows test results.

ankit-bhatnagar167:
@smtakeda i see the AppVeyor build is failing above

smtakeda:
@ankit-bhatnagar167 this specific case is ok, because `branch` build worked fine.
This is an Appveyor's limitation:
https://www.appveyor.com/docs/how-to/secure-files/
```
Note that file won’t be decrypted on Pull Request builds as secure variables are not set during PR build.
```

PR #167, opened 2019-04-16
This changeset moves the client from validating database, schema, and…

tjj5036:
… warehouse to the server.

Currently, validating the schema, database, and warehouse occurs at the
client level when `validate_default_parameters` is set to true in the
connection properties used by the `connect` method. As part of an
upcoming change, which is currently guarded by a feature parameter, we
will be performing this validation server side as opposed to client
side. However, this feature is not enabled for all accounts at the
moment - it is expected to be enabled within the next week or so. If any
clients are affected, we can keep the customers on the old behavior so
that there are no regressions.


smtakeda:
@tjj5036 will the server raise an exception at connection time if the new parameter is true and the given db, schema or warehouse doesn't exist? Which version of server have this change?

tjj5036:
When the new parameter is enabled, the server will raise an exception at connection time during the initial request to `/v1/login-request`. The sever changes were done several releases ago, but not yet enabled. We can enable it anytime, and selectively disable if anyone runs into issues with the Python connector and existing behavior.

smtakeda:
This was fixed in v1.8.7

PR #179, opened 2019-06-11
Urllib3 version

mknister:
I updated the urllib3 library restrictions by requiring the version to be between 1.21.1 and 1.25. I experienced errors when trying to specify that the version restrictions were only for Python 3.4, so I took out that part. Is that okay?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,7 @@ setup(
         'pyasn1>=0.4.0,<0.5.0;python_version<"3.0"',
         'pyasn1-modules>=0.2.0,<0.3.0;python_version<"3.0"',
         'enum34;python_version<"3.4"',
+        'urllib3>=1.21.1,<1.25;python_version="3.4"',
     ],
 
     namespace_packages=['snowflake'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -53,6 +53,7 @@
         'pyasn1>=0.4.0,<0.5.0;python_version<"3.0"',
         'pyasn1-modules>=0.2.0,<0.3.0;python_version<"3.0"',
         'enum34;python_version<"3.4"',
+        'urllib3>=1.21.1,<1.25',
```
smtakeda: can you limit only for Python 3.4 or less?

smtakeda: Test the package locally before sending the change.

```
cd ~/snowflake/Python
./build_connector.sh -c
python3.4 -m venv /tmp/p
source /tmp/p/bin/activate
pip install dist/snowflake_connector*
```
No error or warning should show up.

smtakeda: would not this work?
```
'urllib3>=1.21.1,<1.25;python_version<"3.5"',
```

PR #186, opened 2019-07-29
Ensure that the cython components are present in sdist

mariusvniekerk:
Currently the cython components are not present in the source distribution of this library.

In https://github.com/conda-forge/snowflake-connector-python-feedstock/pull/19 we've attempted to build a conda-forge package for this, but failed due to missing source files.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,4 @@
-include *.rst *.py
+include *.rst *.py *.pyx
+include LICENSE.txt
+recursive-include cpp *.cpp *.hpp
 prune test


smtakeda:
Thanks for PR, but we have not included the cython code in the package yet. That's future extension, and the current package won't depend on it.

howryu:
@mariusvniekerk Thanks for your pr. Have you tested the change in conda-forge and does it succeed? I have not tested it yet.

mariusvniekerk:
Conda forge builds from official releases, thus we can't package the cython extension without this.

howryu:
@mariusvniekerk Ok cool, I will include those files. One more question, does cython generated .cpp file needs to be included in source dist? Or conda forge will run cython compiler at build time? Also, can you commit to branch prep-2.0.2? That's the branch we will use for next release.

mariusvniekerk:
@howryu Conda-forge will just run the setup.py (potentially with a magic env var or two) which will update these files.  For other end users they may want he generated cpp files as part of the sdist.

mariusvniekerk:
@howryu updated the pr.

howryu:
Merged to prep-2.0.2, which is will be released next week. @mariusvniekerk Thanks for your contribution.

PR #188, opened 2019-08-06
make a tzinfo class at the module level instead of inlining (fixes #81)

### LINKED ISSUE #81: TZ objects not pickleable
https://github.com/snowflakedb/snowflake-connector-python/blob/6c3e666ef6c388a3bde2b190256f88a615259196/converter.py#L155

Hello,

those dynamically-defined classes are not pickleable because they are not attached to the module:
```
PicklingError: Can't pickle <class 'snowflake.connector.converter.GMTN0400'>: attribute lookup snowflake.connector.converter.GMTN0400 failed
```

I wonder if we can cache them in the module's namespace so that they can be accessed and pickled.

beckjake:
Fixes #81 by creating a generic class with a name and an offset, and using instances of that for timezones.

I'm not clear on why snowflake uses `type` inline to create a fresh class for each timezone, but it breaks picking of results, which is very annoying and means we have to wrap all results with code that does basically this if we want to pickle anything.

This pattern comes straight from the python docs, check out the "FixedOffset" class defined [on this page](https://docs.python.org/2/library/datetime.html#datetime.datetime.tzinfo)

### DIFF
--- a/converter.py
+++ b/converter.py
@@ -128,6 +128,24 @@ def _adjust_fraction_of_nanoseconds(value, max_fraction, scale):
         return int(TO_UNICODE(max_fraction - frac) + ZERO_FILL[:9 - scale])
 
 
+class OffsetTimezone(tzinfo):
+    def __init__(self, name, tzoffset_minutes):
+        self.name = name
+        self.tzoffset = timedelta(minutes=tzoffset_minutes)
+
+    def utcoffset(self, dt, is_dst=False):
+        return self.tzoffset
+
+    def tzname(self, dt):
+        return self.name
+
+    def dst(self, dt):
+        return ZERO_TIMEDELTA
+
+    def __repr__(self):
+        return self.name
+
+
 def _generate_tzinfo_from_tzoffset(tzoffset_minutes):
     """
     Generates tzinfo object from tzoffset.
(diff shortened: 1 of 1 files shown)


drewbanin:
Is anyone able to take a look at this one?

smtakeda:
no worry, this is in the queue.

smtakeda:
Writing a test case:
```
def test_pickle_timestamp_tz(tmpdir, conn_cnx):
    tmp_dir = str(tmpdir.mkdir('pickles'))
    output = os.path.join(tmp_dir, 'tz.pickle')
    expected_tz = None
    with conn_cnx() as con:
        for rec in con.cursor().execute("select '2019-08-11 01:02:03.123 -03:00'::TIMESTAMP_TZ"):
            expected_tz = rec[0]
            with open(output, 'wb') as f:
                pickle.dump(expected_tz, f)
                print(expected_tz)
                print(type(expected_tz))

    with open(output, 'rb') as f:
        read_tz = pickle.load(f)
        print(read_tz)
        assert expected_tz == read_tz
```
But got this error:
```
>           read_tz = pickle.load(f)
E           TypeError: __init__() missing 2 required positional arguments: 'name' and 'tzoffset_minutes'
```

smtakeda:
This looks like working well.
https://github.com/stub42/pytz/blob/master/src/pytz/__init__.py#L391
```
pytz.FixedOffset
```

smtakeda:
@beckjake and @drewbanin can you review #191 ?

drewbanin:
Hey @smtakeda - this looks really great! I didn't know about `pytz.FixedOffset`. I gave this patch a spin locally and it appears to work really well :)

smtakeda:
working on other branch.

PR #189, opened 2019-08-08
bump urllib3 version

mparker3:
I'd to use urllib v1.25.3 in my project, however, this repo limits the maximum version to 1.24.x. Per the [changelog](https://github.com/urllib3/urllib3/blob/master/CHANGES.rst), it doesn't look like there are any breaking changes between the two versions. Thanks for your time!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -159,7 +159,7 @@ setup(
         'pyasn1>=0.4.0,<0.5.0;python_version<"3.0"',
         'pyasn1-modules>=0.2.0,<0.3.0;python_version<"3.0"',
         'enum34;python_version<"3.4"',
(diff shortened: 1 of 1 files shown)


smtakeda:
This check required to mitigate Python 3.4 + requests 2.21.0 issue, and now we are dropping Python 3.4 support. Are you using Python 2.7?

mparker3:
I am. For reference, which issue are you referring to – [this one](https://github.com/psf/requests/issues/5067)?

smtakeda:
will update this when the python connector completely remove Python 3.4 support.

mparker3:
👍 Thank you!

PR #200, opened 2019-09-13
make change for 3.7 pep 479

jmathis-jc:
I had to change the raise exception to make this work with 3.7 per pep [479](https://www.python.org/dev/peps/pep-0479/). Added the __from for backwards compatibility but didn't test anything. Not sure if you all are aware of this but didn't see anything in the docs.

### DIFF
--- a/json_result.py
+++ b/json_result.py
@@ -3,7 +3,7 @@
 #
 # Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
 #
-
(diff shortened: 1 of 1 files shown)


smtakeda:
Thanks for PR, but it appeared the change doesn't work for Python2 based package...

jmathis-jc:
oh shoot, no worries it works for now. Maybe reconsider after january 2020. Thanks!

PR #204, opened 2019-09-25
SNOW-99281: disabled OCSP revoked certificate test for now

smtakeda:
@ankit-bhatnagar167 this is to fix the main travisCI test failure.


smtakeda:
merging this change to prep-2.0.0

PR #214, opened 2019-10-08
Make pyarrow an optional dependency again

nk-tyler:
The fix in 2.0.0 for SNOW-93406 moved the dependency for pyarrow from
optional to required. This commit reverses the setup.py changes from that
and makes the dependency optional again.

Pyarrow is a difficult dependency to support on systems that aren't
manylinux1 compatible, such as alpine linux (uses musl instead of glibc).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -174,7 +174,6 @@ setup(
         'pyasn1>=0.4.0,<0.5.0;python_version<"3.0"',
         'pyasn1-modules>=0.2.0,<0.3.0;python_version<"3.0"',
         'enum34;python_version<"3.4"',
-        'pyarrow>=0.14.0;python_version>"3.4"',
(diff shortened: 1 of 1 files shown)


smtakeda:
Thanks for PR. We are making pyarrow Package optional in the next release.

Georgzw:
@smtakeda When is next release planned to make pyarrow optional?
We are using the snowflake-connector-python package in an AWS lambda function and the size of the pyarrow package is not compatible with the maximum code size for a lambda function, so this breaks our entire setup.

smtakeda:
will be released this week.

howryu:
@nk-tyler 2.0.2 has been released and pyarrow is an optional dependency. Closed this one.

PR #253, opened 2019-12-25
Fix comment error

ptc-rhandler:
Fix a copy paste error from `Schema` to `Warehouse`


keller00:
This was accepted and merged.
Thank you!

Issue #257, opened 2020-01-09
uppercaseing authenticator breaks Okta URL which may include case-sensitive elements


levahim:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?

3.8.1

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?

Linux-5.4.8-arch1-1-x86_64-with-glibc2.2.5

3. What are the component versions in the environment (`pip list`)? 

Package                    Version   
-------------------------- ----------
asn1crypto                 1.3.0     
attrs                      19.3.0    
azure-common               1.1.24    
azure-storage-blob         2.1.0     
azure-storage-common       2.1.0     
bidict                     0.18.3    
boto3                      1.10.47   
botocore                   1.13.47   
certifi                    2019.11.28
cffi                       1.13.2    
chardet                    3.0.4     
cryptography               2.8       
docutils                   0.15.2    
future                     0.18.2    
idna                       2.8       
ijson                      2.5.1     
jmespath                   0.9.4     
more-itertools             8.0.2     
oscrypto                   1.1.1     
packaging                  20.0      
pip                        19.3.1    
pluggy                     0.13.1    
py                         1.8.1     
pycparser                  2.19      
pycryptodomex              3.9.4     
PyJWT                      1.7.1     
pyOpenSSL                  19.1.0    
pyparsing                  2.4.6     
pytest                     5.3.2     
python-dateutil            2.8.1     
pytz                       2019.3    
requests                   2.22.0    
s3transfer                 0.2.1     
setuptools                 41.2.0    
six                        1.13.0    
snowflake-connector-python 2.1.3     
urllib3                    1.25.7    
wcwidth                    0.1.8     
wheel                      0.33.6    

4. What did you do?

`connection.py` uppercases the authenticator option in line 765. In case of the Okta authenticator, the authenticator parameter value is the Okta URL. Sometimes, the URL may include case-sensitive elements (even though URLs are supposed to be generally case-insensitive, this is real life). For example, something like `https://someaccount.okta.com/snowflake/oO56fExYCGnfV83/2345`. Such URLs commonly appear when multiple applications are managed by Okta for a business account. Uppercasing the URL breaks the authentication. The bug affects `snowsql` tool as well since it uses the Python connector internally.

5. What did you expect to see?

If the authenticator parameter is an Okta URL, it should not be uppercase (or changed in any other way).

6. What did you see instead?

[...]

PR #258, opened 2020-01-09
Do not uppercase authenticator if it is Okta URL

levahim:
This is a fix for issue #257 .


keller00:
We're accepting this, thank you for your pr!
You will see this change in the next release.

PR #182, opened 2019-06-28
Make AWS & Azure optional dependencies

borgstrom:
**This is still WIP**

This PR builds upon #172 and aims to make AWS & Azure optional dependencies that must be installed via setuptools “extras”.

TODO:

* [ ] Update documentation with instructions on how to install the optional dependencies

Closes #172 
Closes #124

PR #260, opened 2020-01-28
Creates Extras per Cloud provider

jperkelens:
As an AWS user, I do not want to install Azure modules to run snowflake and vice versa. This PR isolates each cloud provider into its own extra (GCP appears to acquire no extra modules).


keller00:
I don't think doing just this is enough.
There needs to be error user friendly error messages, the connector should know if it has access to dependencies and such.
You could take a look at: https://github.com/snowflakedb/snowflake-connector-python/blob/master/options.py

snowstakeda:
@jperkelens we are still investigating the impact of these type of changes. Hope we come up something in the next couple of months.

jperkelens:
@keller00 happy to implement something like that!

keller00:
Thank you, @jperkelens 
Let me know if I could be of help 😄

scripts/install.sh
```diff
@@ -34,15 +34,15 @@ if [ "$TRAVIS_OS_NAME" == "osx" ]; then
...
 else
     if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-         pip install .[pandas,development]
+         pip install .[pandas,development,azure,s3]
     else
         pv=${TRAVIS_PYTHON_VERSION/./}
         $THIS_DIR/build_inside_docker.sh $pv
         CONNECTOR_WHL=$(ls $THIS_DIR/../dist/docker/repaired_wheels/snowflake_connector_python*cp${PYTHON_ENV}*.whl | sort -r | head -n 1)
-        pip install -U ${CONNECTOR_WHL}[pandas,development]
+        pip install -U ${CONNECTOR_WHL}[pandas,development,azure,s3]
```
keller00: This reminds me, `tox.ini` has to be updated too.

keller00:
Here's someone else's PR to go off of:
https://github.com/snowflakedb/snowflake-connector-python/pull/182/files
It also reminded me about the scripts in the `scripts` directory

jperkelens:
@keller00 alright, i believe this covers the requested changed (tho there may be some build/test/install artifacts im missing)

keller00:
There's some more stuff missing, some tests directly import from `botocore` and others.
Did you want me to commit straight onto your branch to help out with this?

jperkelens:
@keller00 if you don't mind, that's fine. QQ though, wouldn't that just be a matter of adding the dependency to the `development` extra then?

keller00:
Sure that would run for now, but if we want to run testing without `aws` and `azure` extras separately then it'll be a problem.
Sorry, I'm going to be busy next week, so please don't worry I'm working on it and didn't forget about you! :)

Issue #248, opened 2019-12-12
Cannot use DictCursor with `execute_string` 


drop-saad:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?
3.6.9
2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
OSX
3. What are the component versions in the environment (`pip list`)? 
snowflake-connector-python=2.1.0
snowflake-sqlalchemy=1.1.16
apache-airflow=1.9.0

4. What did you do?
Something like this,
```
conn = snowflake.connector.connect(**conn_config)
cursor = conn.cursor(DictCursor)
cur = cursor.execute_string("select * from test LIMIT 2;")
```

5. What did you expect to see?

A list of cursor objects

6. What did you see instead?

AttributeError: 'DictCursor' object has no attribute 'execute_string'

drop-saad:
It could be fixed by adding cursor_class as an argument to the function and by befault it can use Snowflake cursor. 
```
    def execute_string(
        self,
        sql_text,
        remove_comments=False,
        return_cursors=True,
        cursor_class=SnowflakeCursor,
        **kwargs
    ):
        """
        Executes a SQL text including multiple statements.
        This is a non-standard convenient method.
        cursor.fetch* methods would return a dict object.

        Returns a list of Cursor Objects.
        """
        ret = []
        stream = StringIO(sql_text)
        for sql, is_put_or_get in split_statements(stream, remove_comments=remove_comments):
            cur = self.cursor(cursor_class=cursor_class)
            if return_cursors:
                ret.append(cur)
            cur.execute(sql, _is_put_get=is_put_or_get, **kwargs)
        return ret
```
Then anyone calling  `execute_string` with `DictCursor` class can get the desired behaviour. 
```
execute_string(
        sql_text,
        remove_comments=False,
        return_cursors=True,
        cursor_class=DictCursor,
    )
```

drop-saad:
If this seems reasonable I can clean up and submit a PR for this issue?

keller00:
I'm okay with this change as it would not change existing behaviour.
If you could create a PR that would be greatly appreciated!
**You could even be a rock star ⭐️  by including a test in the PR**

drop-saad:
@keller00 Sounds good will send a PR. Question about testing is there any DB that you are running tests against or just mocking the tests.
Also are there any special considerations for submitting a PR?

[...]

PR #262, opened 2020-02-01
Adds cursor_class as an argument to the execute_string function

Saad-K:
Fixes issue highlighted in this thread https://github.com/snowflakedb/snowflake-connector-python/issues/248

**Issue:** execute_string function was only returning a list of tuples, we wanted it to keep the default behaviour and add the ability to return a list of dictionaries in cursors.

**This pull request contains the following changes:**
- Adds `cursor_class` as an argument to the function `SnowflakeConnection.execute_string`
- Sets the default value for `cursor_class` to `SnowflakeCursor` to keep the existing functionality
- Write a test case to test the out of database cursors

**How is it tested:**
It is tested by creating a trial snowflake account and following the instruction to run the test cases from here https://github.com/snowflakedb/snowflake-connector-python/tree/master/test

### DIFF
--- a/connection.py
+++ b/connection.py
@@ -573,7 +573,7 @@ class SnowflakeConnection(object):
     def execute_string(self, sql_text,
                        remove_comments=False,
                        return_cursors=True,
-                       cursor_class=SnowflakeCursor
+                       cursor_class=SnowflakeCursor,
                        **kwargs):
         """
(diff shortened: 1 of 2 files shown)


drop-saad:
@keller00 pinging here for review.

keller00:
Thanks, @Saad-K! This has been accepted.

PR #263, opened 2020-02-04
SNOW-135600 add workflows and actions to OSS repo

marcmacintyre:
@snowstakeda these are the actions to support jira <--> issue integration.


marcmacintyre:
@snowstakeda I'm creating these Jiras in the Eng-Internal-Infra area, and that's not the right area.  Where should I be putting them?

snowstakeda:
I believe it is Eng-Internal-Infra area as this change is related to the build automation.

marcmacintyre:
So all jiras created as issues on the snowflake connector repo should be assigned to Eng-Infra?  That doesn't seem right.

PR #289, opened 2020-03-12
It is possible for the client_prefetch_threads to not be an int when …

mach10:
…compared. Prevent TypeError: '<=' not supported between instances of 'str' and 'int' errors

This PR prevents an error which occured when using sqlalchemy with the snowflake connector and setting the `client_prefetch_threads` config value.  That was being passed as a `str` hence failed the checks against `0` and `MAX_CLIENT_PREFETCH_THREADS` value.  

It gets cast to an int later anyway.

### DIFF
--- a/connection.py
+++ b/connection.py
@@ -1121,9 +1121,9 @@ class SnowflakeConnection(object):
         return self.client_session_keep_alive_heartbeat_frequency
 
     def _validate_client_prefetch_threads(self):
-        if self.client_prefetch_threads <= 0:
+        if int(self.client_prefetch_threads) <= 0:
             self._client_prefetch_threads = 1
-        elif self.client_prefetch_threads > MAX_CLIENT_PREFETCH_THREADS:
+        elif int(self.client_prefetch_threads) > MAX_CLIENT_PREFETCH_THREADS:
             self._client_prefetch_threads = MAX_CLIENT_PREFETCH_THREADS
         self._client_prefetch_threads = int(
             self.client_prefetch_threads)


sfc-gh-mkeller:
Sorry for taking so long to get to this @mach10 
I'd rather cast it here: https://github.com/snowflakedb/snowflake-connector-python/blob/b2a3a2974a4cb91d141d5c2ff5d9a88a234f97dd/connection.py#L345

Did you want to update your PR?

sfc-gh-mkeller:
Ahh, sorry @mach10 I now got around to reproduce the issue. I assume that you did something like this:
```python
    from sqlalchemy import create_engine
    engine = create_engine(
        'snowflake://{user}:{password}@{account}/{database}/{schema}'.format(...),
        connect_args={
            'client_prefetch_threads': '42',
        }
    )
    try:
        results = engine.execute('select current_version()').fetchone()
        assert results is not None
    finally:
        engine.dispose()
```
In this case my previous comment is invalid as the initial setting bypasses the setter (maybe we should change this behaviour).
However; we had a similar conversation when I updated `validate_default_parameters` and we decided not to silently cast things without the user's knowledge.
The issue can easily fixed by changing the value in the dictionary to an `int`.

I'd also recommend using the aforementioned `validate_default_parameters` like this:
```python
    from sqlalchemy import create_engine
    engine = create_engine(
        'snowflake://{user}:{password}@{account}/{database}/{schema}'.format(...),
        connect_args={
            'client_prefetch_threads': '42',
            'validate_default_parameters': True,
        }
    )
    try:
        results = engine.execute('select current_version()').fetchone()
        assert results is not None
    finally:
        engine.dispose()
```
This way the connector would have given you the following warning:
```python
snowflake/connector/connection.py:668: UserWarning: 'client_prefetch_threads' connection parameter should be of type 'int', but is a 'str'
```

Issue #287, opened 2020-03-09
SNOW-143578: Unable to use on AWS Lambda


valkolovos:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)? 3.8

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)? AWS Lambda Python3.8 runtime

3. What are the component versions in the environment (`pip list`)? 
```
Package         Version
--------------- -------
boto3           1.10.34
botocore        1.13.34
docutils        0.15.2
jmespath        0.9.4
pip             19.2.3
python-dateutil 2.8.0
rapid-client    0.0.0
s3transfer      0.2.1
setuptools      41.2.0
six             1.13.0
urllib3         1.25.7
```

4. What did you do?
Created a lambda. Tried to execute SQL using snowflake connector (created connector, got cursor from connector, executed SQL).

5. What did you expect to see? Results of SQL query

6. What did you see instead?
```
[ERROR] OSError: [Errno 38] Function not implemented
Traceback (most recent call last):
  File "/var/task/app.py", line 34, in lambda_handler
    load_student_transactions(txn_students),
  File "/var/task/app.py", line 61, in load_student_transactions
    cs.execute(
  File "/var/task/snowflake/connector/cursor.py", line 580, in execute
    self._init_result_and_meta(data, _use_ijson)
  File "/var/task/snowflake/connector/cursor.py", line 633, in _init_result_and_meta
    self._result = self._json_result_class(data, self, use_ijson)
  File "/var/task/snowflake/connector/json_result.py", line 23, in __init__
    self._init_from_meta(raw_response, use_ijson)
  File "/var/task/snowflake/connector/json_result.py", line 59, in _init_from_meta
    self._chunk_downloader = self._connection._chunk_downloader_class(
  File "/var/task/snowflake/connector/chunk_downloader.py", line 101, in __init__
    self._pre_init(chunks, connection, cursor, qrmk, chunk_headers,
  File "/var/task/snowflake/connector/chunk_downloader.py", line 89, in _pre_init
    self._pool = ThreadPool(self._effective_threads)
  File "/var/lang/lib/python3.8/multiprocessing/pool.py", line 922, in __init__
    Pool.__init__(self, processes, initializer, initargs)
  File "/var/lang/lib/python3.8/multiprocessing/pool.py", line 196, in __init__
    self._change_notifier = self._ctx.SimpleQueue()
[...]

PR #304, opened 2020-05-05
SNOW-162294 When process pools cannot be used, fall back to thread pools

jberkenbilt:
In some environments, such as AWS Lambda, _multiprocessing.SemLock is
not implemented, which prevents process pools from working. Detect
this and fall back to using concurrent.futures.ThreadPoolExecutor in
this case.

This pull request is to address #287. With this change, I am able to use the snowflake connector (2.2.5) from lambda.

I am not able to run py.test in my environment for unknown reasons -- all the tests fail with py.test errors about writing to a closed file handle, possibly related to https://github.com/pytest-dev/pytest/issues/5502. I have manually tested the behavior of the thread pool in general and have tested with the connector using chunk_downloader. I have not tested the changes to file_transfer.py.

### DIFF
--- a/chunk_downloader.py
+++ b/chunk_downloader.py
@@ -10,7 +10,6 @@ from collections import namedtuple
 from gzip import GzipFile
 from io import BytesIO
 from logging import getLogger
-from multiprocessing.pool import ThreadPool
 from threading import Condition, Lock
 
 from snowflake.connector.gzip_decoder import decompress_raw_data
@@ -20,6 +19,7 @@ from .arrow_context import ArrowConverterContext
 from .errorcode import ER_CHUNK_DOWNLOAD_FAILED
 from .errors import Error, OperationalError
 from .time_util import get_time_millis, DecorrelateJitterBackoff
+from .thread_pool import ThreadPool
 
 DEFAULT_REQUEST_TIMEOUT = 7
 
--- a/file_transfer_agent.py
+++ b/file_transfer_agent.py
@@ -12,7 +12,6 @@ import sys
 import tempfile
 import threading
 from logging import getLogger
-from multiprocessing.pool import ThreadPool
 from time import sleep, time
 
 import botocore.exceptions
@@ -38,6 +37,7 @@ from .gcs_util import SnowflakeGCSUtil
 from .local_util import SnowflakeLocalUtil
 from .remote_storage_util import SnowflakeFileEncryptionMaterial, SnowflakeRemoteStorageUtil
 from .s3_util import SnowflakeS3Util
+from .thread_pool import ThreadPool
 
 S3_FS = u'S3'
 AZURE_FS = u'AZURE'
--- a/thread_pool.py
+++ b/thread_pool.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
+#
+from multiprocessing.pool import ThreadPool as p_ThreadPool
+from concurrent.futures import ThreadPoolExecutor as t_ThreadPool
+from logging import getLogger
+
+logger = getLogger(__name__)
+
+# This implements a wrapper around either
+# multiprocessing.pool.ThreadPool or
+# concurrent.futures.ThreadPoolExecutor, favoring process pools if
+# they work. This is not a complete wrapper. It only wraps interfaces
+# actually used by the snowflake connector. The behavior is not
+# identical in that there is no equivalent of terminate(), but the
+# behavior is close and should work fine based on the usage patterns
+# in the code. In any case, without this, the connector is unusable in
+# environments in which the process pool doesn't work, such as AWS
+# lambda.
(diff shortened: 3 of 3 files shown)


jberkenbilt:
Also, I think possibly test/README.rst may have omitted mention of pendulum in the list of things to pip install to run tests.

sfc-gh-mkeller:
I will look at this soon, thank you!

> Also, I think possibly test/README.rst may have omitted mention of pendulum in the list of things to pip install to run tests.

Please note our `development` optional dependency, I should document this into the README. 😅

stevesuh:
Any updates on this issue?   Otherwise my team is gonna use Node instead.

antonb-via:
Guys, this is a pressing issue for us. Can you please issue this fix so we can use the python connector?

sfc-gh-mkeller:
> @sfc-gh-mkeller
> Guys, this is a pressing issue for us. Can you please issue this fix so we can use the python connector?

Please contact Snowflake support in that case, they should be able to help you out.
We definitely have customers use our connector in Lambda.

sfc-gh-mkeller:
Unfortunately testing doesn't pass with the changes as they are right now.
I'll look into fixing it up.

jberkenbilt:
@sfc-gh-mkeller For what it's worth, when I opened my pull request, I had wanted to get something with passing tests, but I wasn't able to get the tests to run even before my changes. Nearly all the tests failed, suggesting there was something wrong with my testing environment. I'd suggest going through the READMEs and making sure it is possible to run the tests in a clean environment. This would likely improve the quality of future contributions. Certainly I would have ensured the tests passed in my PR if I could have gotten the tests to work prior to my change.

My code only does a few operations, mostly just direct SQL statements against the data and has low coverage, but it's been running without issue many times per minute continuously since my PR.

Please let me know if you would like me to do anything on the PR, particularly if the instructions for running tests are updated.

sfc-gh-mkeller:
I'll beef up the READMEs, sorry about that @jberkenbilt .
I feel like I owe an explanation why this is. It is because main development takes place in our internal repository.

Now that said we are working on changing this, when that transition happens we'll be able to accept PRs finally and updating the READMEs will be less of a hassle.

sfc-gh-mkeller:
After a few hours reading documentations I think I arrived to what I think is the right solution here.
According to https://aws.amazon.com/blogs/compute/parallel-processing-in-python-with-aws-lambda/ the process pools are not allowed in Lambda because of the memory sharing between processes required for queues used by process pools.

But `multiprocessing.pool.ThreadPool` is a mostly undocumented code that uses the same architecture as a process pool, with a process looking wrapper around threads.
I now think that the best course of action would be to just use `ThreadPoolExecutor` exclusively instead of introducing another layer of abstraction.

What do you think @jberkenbilt ?

jberkenbilt:
@sfc-gh-mkeller Actually, my first set of changes was to replace the process pool with `ThreadPoolExecutor`. Then I went and made it interface compatible with the existing code to try to minimize the change. So I think using `ThreadPoolExecutor` is the way to go.

Of course, you have to be mindful of the normal differences between threads and processes, all of which apply here:

* You can't forcibly terminate a thread like you can with a process. There is some code that relies on calling `terminate()` on the process pool. You will have to check that code to make sure the right thing happens with threads. In particular, a process won't normally exit (unless forced to do so) until all the threads have exited. The current code terminates the process pool, which makes it pretty much irrelevant and stops things in flight. If you switch to `ThreadPoolExecutor`, you won't have that. You may have to have another way to ensure the threads exit if that's important.
* Threads share memory; processes don't. So if you have any code that assumes it can modify things locally and not affect other processes, that will be different with threads. I looked specifically for that case and didn't see it, so I think you're fine there, but it would be worth a second look.

I still think `ThreadPoolExecutor` is the way to go and that the code changes will be small -- smaller than my pull request. Just make sure the code that is currently relying on terminate() does the right thing. Let me know if you want me to update the pull request. I'm happy to do so.

sfc-gh-mkeller:
Thank you @jberkenbilt for all your insight and your PR, I'll talk this through with the team to see if they had any input, I'll also need to run some benchmarks before merging this.

I think it'll be okay to use `ThreadPoolExecutor` instead, as I said `multiprocessing.pool.ThreadPool` is a wrapper to make Threads look like Processes. If you look at this code https://github.com/python/cpython/blob/1cba1c9abadf76f458ecf883a48515aa3b534dbd/Lib/multiprocessing/pool.py#L723
Terminate seems to not do anything more than join threads in a `ThreadPool`.

Our code also assumes that Threads change dictionaries passed to them. I think we only use this old `ThreadPool` because the new one was not available when the code was written.

I now have a branch that passes testing and uses `ThreadPoolExecutor` instead of `ThreadPool`.
Please allow me to close this PR as we could use the issue you linked to keep track of my progress with this change. Although the hard part is over now!

Thank you again! 🍻

jberkenbilt:
I see 2.2.7 is out and has the changes. I'm going to redeploy my lambda functions with 2.2.7 and without my patch and see it how it goes. Thanks!

PR #309, opened 2020-05-18
SNOW-160665 add wss

sfc-gh-stakeda:
Add whitesource scan


sfc-gh-stakeda:
will merge this change to the mono repo later.

PR #312, opened 2020-05-22
register numpy.str_ in converter.py:PYTHON_TO_SNOWFLAKE_TYPE

mattyHawthorn:
This simply adds a registration for `numpy.str_` in `converter.py`.

I hit an obscure error - if one by some path one ends up with a pandas column or numpy array of object type (`dtype('O')`), _containing instances of `numpy.str_`, which _is a proper subclass of `str`_, then the snowflake sqlalchemy dialect fails on trying to render those values to params for a parameterized query. The inserted line here fixes the problem.

### DIFF
--- a/converter.py
+++ b/converter.py
@@ -40,6 +40,7 @@ PYTHON_TO_SNOWFLAKE_TYPE = {
     u'decimal': u'FIXED',
     u'float': u'REAL',
     u'str': u'TEXT',
+    u'str_': u'TEXT',
     u'unicode': u'TEXT',
     u'bytes': u'BINARY',
     u'bytearray': u'BINARY',


sfc-gh-mkeller:
Hey @mattyHawthorn is there any chance that you could add a test for the behaviour that you are describing. This would accelerate me looking at your PR greatly.
If you don't have time to figure out how our testing works, or set it up then a small code snippet to reproduce the issue would be super helpful for me. We want to make sure that we are proactively testing for this issue in the future.
Note: someone pointed it out that our testing documentation is not good at all, so if you do want to run our tests, make sure to install the connector with the `development` optional dependency for all the extra stuff we use for tests.

mattyHawthorn:
Yeah honestly it seemed like it was going to take a little work to get the full test suites running, and I think I had a permissions error even when I was able to get some of them running. I would have to defer to you all on actually adding the test, I just don't have time.

A relevant-looking test appears to be [here](https://github.com/Snagajob/snowflake-connector-python/blob/87c77da3aaf7dfa0267914725fd2b239a93599d6/test/test_numpy_binding.py#L7-L46) but this uses percent-style params, and we're using qmark-style params. If that test case were using qmark, I would just add a column to that test table with `TEXT` type and add an insertion of a `numpy.str_` value into it.

My particular use case is:
If I have a connection/engine - in this case a `sqlalchemy.engine.Engine` called `engine`, and a `pandas.DataFrame` called `df`, I should be able to 

```
import numpy as np

# this would be the column that fails:
df['test_column'] = np.str_("just a string")

df.to_sql(
   'some_table',
   engine,
)
```

But at present I get a `ProgrammingError`, raised by [these lines](https://github.com/Snagajob/snowflake-connector-python/blob/12cf7f72bd0c543e6eada0a25534cc72c923729f/connection.py#L1001-L1015), which call to [this method](https://github.com/Snagajob/snowflake-connector-python/blob/2d386d494fd205d329c34e418baf75e8167dabd2/converter.py#L358-L364), which ultimately looks for the key `'str_'` (the name of the type of values in column `'test_column'`) in `PYTHON_TO_SNOWFLAKE_TYPE`.

Again I'll note that it appears that you all have already attempted to handle other special cases of this nature, such as `numpy.bool_`, which behaves most meaingful respects as builtin `bool`.

sfc-gh-mkeller:
Thank you, this is super helpful. I'll add a test soon and run tests.

PR #323, opened 2020-06-17
Fix master wss script

sfc-gh-stakeda:


sfc-gh-stakeda:
I'm still testing... and will get approval from @sfc-gh-jfan

PR #354, opened 2020-08-11
SNOW-182731 fix put issue

sfc-gh-mkeller:
Please see ticket for description!

### DIFF
--- a/file_transfer_agent.py
+++ b/file_transfer_agent.py
@@ -346,9 +346,9 @@ class SnowflakeFileTransferAgent(object):
                 if len(retry_meta) == 0:
                     # no new AWS token is required
                     break
+                client = self.renew_expired_client()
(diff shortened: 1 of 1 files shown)


file_transfer_agent.py
```diff
@@ -346,9 +346,9 @@ def _upload_files_in_parallel(self, file_metas):
                 if len(retry_meta) == 0:
                     # no new AWS token is required
                     break
+                client = self.renew_expired_client()
```
sfc-gh-stakeda, line 349: hmm,  this renew token happens only if `result_status` is ResultStatus.RENEW_TOKEN?

sfc-gh-mkeller: `RENEW_TOKEN` and `RENEW_PRESIGNED_URL`, the line just before breaks if nothing needs to be retried

sfc-gh-stakeda: ah, ok

Issue #361, opened 2020-08-17
SNOW-183907: Loading data with write_pandas/to_sql does not honor column defaults


kyleflan:
1. What version of Python are you using (`python --version`)? Python 3.8.2

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)? Linux-4.19.84-microsoft-standard-x86_64-with-glibc2.29

3. What are the component versions in the environment (`pip list`)? 
```
Package                    Version    
-------------------------- -----------
asn1crypto                 1.3.0      
astroid                    2.4.2      
attrs                      19.3.0     
azure-common               1.1.25     
azure-core                 1.7.0      
azure-storage-blob         12.3.2     
backcall                   0.2.0      
boto3                      1.14.29    
botocore                   1.17.29    
certifi                    2020.6.20  
cffi                       1.14.1     
chardet                    3.0.4      
contextlib2                0.6.0.post1
cryptography               2.9.2      
cycler                     0.10.0     
datarobot                  2.21.3     
decorator                  4.4.2      
docutils                   0.15.2     
idna                       2.9        
ipykernel                  5.3.4      
ipython                    7.16.1     
ipython-genutils           0.2.0      
isodate                    0.6.0      
isort                      4.3.21     
jedi                       0.17.2     
jmespath                   0.10.0     
jupyter-client             6.1.6      
jupyter-core               4.6.3      
kiwisolver                 1.2.0      
lazy-object-proxy          1.4.3      
matplotlib                 3.3.0      
mccabe                     0.6.1      
msrest                     0.6.17     
numpy                      1.19.1     
oauthlib                   3.1.0      
oscrypto                   1.2.0      
pandas                     1.0.5      
parso                      0.7.1      
pexpect                    4.8.0      
pickleshare                0.7.5      
Pillow                     7.2.0      
pip                        20.0.2     
pkg-resources              0.0.0      
prompt-toolkit             3.0.5      
ptyprocess                 0.6.0      
pyarrow                    0.17.1     
pycparser                  2.20       
pycryptodomex              3.9.8      
Pygments                   2.6.1      
PyJWT                      1.7.1      
pylint                     2.5.3      
pyOpenSSL                  19.1.0     
pyparsing                  2.4.7      
python-dateutil            2.8.1      
pytz                       2020.1     
PyYAML                     5.3.1      
pyzmq                      19.0.1     
requests                   2.23.0     
requests-oauthlib          1.3.0      
[...]

Issue #329, opened 2020-06-23
SNOW-170139: CASE_SENSITIVE in write_pandas causes silent failures


bmwilly:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?

```
Python 3.8.2
```

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?

```
macOS-10.15.5-x86_64-i386-64bit
```

3. What are the component versions in the environment (`pip list`)? 

```
Package                      Version             Location
---------------------------- ------------------- -------------------------------------------------------
agate                        1.6.1
alembic                      1.4.2
amazon-kclpy                 2.0.1
ansiwrap                     0.8.4
appdirs                      1.4.3
appnope                      0.1.0
arrow                        0.15.6
arviz                        0.8.3
asn1crypto                   1.3.0
async-generator              1.10
atpublic                     1.0
attrs                        19.3.0
autograd                     1.3
autograd-gamma               0.4.2
autopep8                     1.5.1
aws-mfa                      0.0.12
awscli                       1.18.75
azure-common                 1.1.25
azure-core                   1.6.0
azure-nspkg                  3.0.2
azure-storage-blob           12.3.2
azure-storage-common         0.37.1
azure-storage-nspkg          3.1.0
b2                           1.4.2
b2sdk                        0.1.8
Babel                        2.8.0
backcall                     0.2.0
binaryornot                  0.4.4
black                        19.10b0
bleach                       3.1.5
boto                         2.49.0
boto3                        1.11.17
botocore                     1.14.17
Brotli                       1.0.7
brotlipy                     0.7.0
cached-property              1.5.1
cachetools                   4.1.0
catboost                     0.22
certifi                      2020.4.5.2
cffi                         1.13.2
cftime                       1.1.3
chardet                      3.0.4
click                        7.1.1
cloudpickle                  1.4.1
cmdstanpy                    0.4.0
colorama                     0.4.3
conda                        4.8.3
conda-package-handling       1.6.0
configobj                    5.0.6
configparser                 5.0.0
convertdate                  2.2.1
convoys                      0.1.7
cookiecutter                 1.7.0
cryptography                 2.9.2
cycler                       0.10.0
Cython                       0.29.20
cytoolz                      0.10.1
dash                         1.12.0
dash-bootstrap-components    0.10.1
dash-core-components         1.10.0
dash-daq                     0.5.0
[...]

PR #362, opened 2020-08-18
Fix 329 361

kyleflan:
* Fixes issue where using MATCH_BY_COLUMN_NAME led to case sensitivity issues. Replaced with a `quote_identifier` parameter to `write_pandas`, which defaults to False, but if specified as True will quote all identifiers when passing to Snowflake.
* Fixes using `pd_writer` as a method to `DataFrame.to_sql` via `write_pandas` where default columns in a table pre-defined in Snowflake were not having their default values populated

### DIFF
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -164,7 +164,8 @@ def write_pandas(conn: 'SnowflakeConnection',
 def pd_writer(table: pandas.io.sql.SQLTable,
               conn: Union['sqlalchemy.engine.Engine', 'sqlalchemy.engine.Connection'],
               keys: Iterable,
-              data_iter: Iterable) -> None:
+              data_iter: Iterable,
+              quote_identifiers: bool = True) -> None:
     """This is a wrapper on top of write_pandas to make it compatible with to_sql method in pandas.
 
         Example usage:
@@ -173,12 +174,19 @@ def pd_writer(table: pandas.io.sql.SQLTable,
 
             sf_connector_version_df = pd.DataFrame([('snowflake-connector-python', '1.0')], columns=['NAME', 'NEWEST_VERSION'])
             sf_connector_version_df.to_sql('driver_versions', engine, index=False, method=pd_writer)
+            
+            # to use quote_identifiers=False
+            from functools import partial
+            sf_connector_version_df.to_sql(
+                'driver_versions', engine, index=False, method=partial(pd_writer, quote_identifiers=False))
 
     Args:
         table: Pandas package's table object.
         conn: SQLAlchemy engine object to talk to Snowflake.
         keys: Column names that we are trying to insert.
         data_iter: Iterator over the rows.
+        quote_identifiers: if True (default), quote identifiers passed to Snowflake. If False, identifiers are not
+            quoted (and typically coerced to uppercase by Snowflake)
     """
     sf_connection = conn.connection.connection
     df = pandas.DataFrame(data_iter, columns=keys)
@@ -186,4 +194,5 @@ def pd_writer(table: pandas.io.sql.SQLTable,
                  df=df,
                  # Note: Our sqlalchemy connector creates tables case insensitively
                  table_name=table.name.upper(),
-                 schema=table.schema)
+                 schema=table.schema,
+                 quote_identifiers=quote_identifiers)
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -36,7 +36,8 @@ def write_pandas(conn: 'SnowflakeConnection',
                  chunk_size: Optional[int] = None,
                  compression: str = 'gzip',
                  on_error: str = 'abort_statement',
-                 parallel: int = 4
+                 parallel: int = 4,
+                 quote_identifiers: bool = False
                  ) -> Tuple[bool, int, int,
                             Sequence[Tuple[str, str, int, int, int, int, Optional[str], Optional[int],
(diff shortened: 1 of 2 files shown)


kyleflan:
@sfc-gh-mkeller do the remaining tests have to be run by your team? Or did I miss a configuration option for the `$PARAMETERS_SECRET` environment variable?

sfc-gh-mkeller:
Hi, @kyleflan 
Thank you for the PR! I'll be more than happy to work with you on this PR.

However; please understand that this is a [behavior change](https://community.snowflake.com/s/article/Pending-Behavior-Change-Log), so I need to get some approvals and announcements before this can be accepted and released.

> @sfc-gh-mkeller do the remaining tests have to be run by your team? Or did I miss a configuration option for the `$PARAMETERS_SECRET` environment variable?

So the `$PARAMETERS_SECRET` is only available to people with write permission to the repository. I need to figure out how to run tests for incoming PRs still, as we just recently moved full development from a private to this public repository.

It is stored in the repository's secrets, so I think after careful inspection of incoming PRs someone with write permission should be able to trigger tests.

If anyone had any input on how to do this properly, I would appreciate the help!

Could you first fix the PR by rebasing onto master? We just introduced the `src` directory.

kyleflan:
Understood about the behavior change. Rebased and pushed.

src/snowflake/connector/pandas_tools.py
```diff
@@ -36,7 +36,8 @@ def write_pandas(conn: 'SnowflakeConnection',
                  chunk_size: Optional[int] = None,
                  compression: str = 'gzip',
                  on_error: str = 'abort_statement',
-                 parallel: int = 4
+                 parallel: int = 4,
+                 quote_identifiers: bool = False
```
sfc-gh-mkeller, line 40: By making this `False` by default is the part that makes this a breaking change. Unfortunately I have still not received approval for this, which is why I have been MIA from your PR @kyleflan

test/pandas/test_pandas_tools.py
```diff
@@ -63,14 +64,100 @@ def test_write_pandas(conn_cnx: Callable[..., Generator['SnowflakeConnection', N
             cnx.execute_string("DROP TABLE IF EXISTS {}".format(table_name))
 
 
+@pytest.mark.parametrize('chunk_size', [5, 4, 3, 2, 1])
+@pytest.mark.parametrize('compression', ['gzip', 'snappy'])
+# Note: since the file will to small to chunk, this is only testing the put command's syntax
+@pytest.mark.parametrize('parallel', [4, 99])
+def test_write_pandas(conn_cnx: Callable[..., Generator['SnowflakeConnection', None, None]],
```
sfc-gh-mkeller, line 71: In general I'd rather have you add one more parametrized argument to the tests than copy paste the whole test just to change one parameter passed down.

test/pandas/test_pandas_tools.py
```diff
@@ -63,14 +64,100 @@ def test_write_pandas(conn_cnx: Callable[..., Generator['SnowflakeConnection', N
             cnx.execute_string("DROP TABLE IF EXISTS {}".format(table_name))
 
 
+@pytest.mark.parametrize('chunk_size', [5, 4, 3, 2, 1])
+@pytest.mark.parametrize('compression', ['gzip', 'snappy'])
+# Note: since the file will to small to chunk, this is only testing the put command's syntax
```
sfc-gh-mkeller, line 69: Oof, I don't know what I was thinking when I wrote this...
```suggestion
# Note: since the file will be too small to chunk, this is only testing the put command's syntax
```

sfc-gh-mkeller (changes requested):
@kyleflan I'd suggest breaking this change into 2 parts. One to deal with #361 and another to deal with #329 
Then we could test and merge the #329 even today to unblock you.
If you don't have time to do this please let me know and I can separate things

kyleflan:
@sfc-gh-mkeller how would you like me to separate them? Two new PRs?

sfc-gh-mkeller:
Both that and transforming this into one of them works for me. I’ll let you decide! 🙂

kyleflan:
What if we set `quote_identifiers=True` by default? That will bring it in line with the current behavior, but offer the opportunity for developers to utilize `write_pandas` without quoted identifiers. 

I still think, long term, it makes more sense for the function to operate similar to SQL clients. At that point, it's just a one-line change if we go this route.

sfc-gh-mkeller:
> I still think, long term, it makes more sense for the function to operate similar to SQL clients. At that point, it's just a one-line change if we go this route.

I agree! Yeah, could we flip the default around.
Also please let's merge the 2 tests, you could add a new parametrized parameter to it to have it run with both of the quote_identifiers turned on and off.
I'm sorry for not responding earlier, I'm a bit spread thin nowadays! 😓

kyleflan:
Done. One thing I added to the docstring of `pd_writer` was a description of how to use the `quote_identifiers` arg when using pandas' `DataFrame.to_sql()`. 

I added a test for that: `test_pandas_to_sql()`. However, it requires the user testing it to have the `snowflake-sqlalchemy` package installed, so I check that the user has it installed before running the test.

sfc-gh-mkeller:
Thank you so much @kyleflan .
I like the changes to the code.
But I would like to add some more tests myself to this PR and remove our SQLAlchemy library from the testing for the connector. I can move your test over to our SQLAlchemy repo and then modify your test here not to not test `snowflake-sqlalchemy`.
If I just close this PR don't be alarmed, I'll pull your changes until now into a local branch and then add the test changes there. This procedure will preserve you as the author of these commits, but you should see a mention on this PR anyways.

PR #366, opened 2020-08-19
SNOW-184712 add sdist test

sfc-gh-mkeller:
In light of the MANIFEST issue from last night I'm adding a test to build and install sdists.

This also makes it so not all tests wait for all build jobs to finish before starting, e.g.: linux tests will only wait for linux build jobs.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -23,12 +23,35 @@ jobs:
           python-version: '3.x'
       - name: Display Python version
         run: python -c "import sys; import os; print(\"\n\".join(os.environ[\"PATH\"].split(os.pathsep))); print(sys.version); print(sys.executable);"
-      - name: Upgrade setuptools, pip and wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/build_test.yml
```diff
@@ -126,20 +149,47 @@ jobs:
...
+        run: python -c "import sys; print(sys.version)"
+      - name: Upgrade setuptools and pip
+        run: python -m pip install -U setuptools pip
+      - name: Install sdist
+        run: pip install dist/*
+
   test:
     name: Test ${{ matrix.os.download_name }}-${{ matrix.python-version }}-${{ matrix.cloud-provider }}
-    needs: [build-manylinux, build-macos, build-windows]
+    needs: ${{ matrix.os.needs }}
```
sfc-gh-mkeller, line 178: Unfortunately it seems like this is not possible.
See: https://github.community/t/matrix-variable-does-not-work-in-needs-keyword/16972/2
Going to roll this back.

PR #374, opened 2020-08-24
manually syncing requirements files

sfc-gh-mkeller:
manually syncing newest requirements files for release

### DIFF
--- a/tested_requirements/requirements_35.txt
+++ b/tested_requirements/requirements_35.txt
@@ -1,19 +1,21 @@
-asn1crypto==1.3.0
+# Generated on: Python 3.5.9
+# With snowflake-connector-python version: 2.3.0
+asn1crypto==1.4.0
 azure-common==1.1.25
 azure-storage-blob==2.1.0
 azure-storage-common==2.1.0
-boto3==1.13.26
-botocore==1.16.26
-certifi==2020.4.5.2
-cffi==1.14.0
+boto3==1.14.47
+botocore==1.17.47
+certifi==2020.6.20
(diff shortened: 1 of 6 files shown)


tested_requirements/requirements_35.txt
```diff
@@ -1,25 +1,26 @@
-asn1crypto==1.3.0
+# Generated on: Python 3.5.9
+# With snowflake-connector-python version: 2.3.0
```
sfc-gh-mkeller, line 2: I changed how this files looks a bit, now customers should be able to use this as a requirements file straight away after downloading it.

sfc-gh-mkeller: Except that it doesn't have `snowflake-connector-python` in the list, that's what this comment is for

sfc-gh-wshangguan: Just to confirm, this is to tell customers they need to have `snowflake-connector-python` already installed?

sfc-gh-mkeller: Take a look at how it looked before: https://github.com/snowflakedb/snowflake-connector-python/blob/master/tested_requirements/requirements_35.txt
I wanted to make this easier to use for customers.
So now if you `wget` this file or copy-paste it, you can do `pip install -r file.txt` and it will work, but I don't think that it would work with the older files.
This just tells what version of the connector was used to create this requirements file.

sfc-gh-wshangguan: That sounds good!

PR #352, opened 2020-08-09
Added proxies to SnowflakeRestfuls requests session instead of env vars

Adaptive-Alon:
Added proxies to SnowflakeRestfuls requests session.

What this PR did:

1. Changed `set_proxies` method to not set the proxies as system wide environment variables, and renamed it `format_proxies`
2. Added a `proxies` param to `SnowflakeRestful`, and passed the formatted proxy dict from 'SnowflakeConnection.__open_connection()`.
3. Added the proxy dict to the reqests Session created in `SnowflakeRestful.make_requests_session()`

PR #379, opened 2020-08-25
Use requests proxies object

adrianoesch:
The [proxies](https://requests.readthedocs.io/en/master/user/advanced/#proxies) object from `requests` is more versatile and also allows for the use of [socks](https://requests.readthedocs.io/en/master/user/advanced/#socks) proxies.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -93,10 +93,7 @@ DEFAULT_CONFIGURATION = {
     'host': ('127.0.0.1', str),  # standard
     'port': (8080, (int, str)),  # standard
     'database': (None, (type(None), str)),  # standard
-    'proxy_host': (None, (type(None), str)),  # snowflake
-    'proxy_port': (None, (type(None), str)),  # snowflake
-    'proxy_user': (None, (type(None), str)),  # snowflake
-    'proxy_password': (None, (type(None), str)),  # snowflake
+    'proxies' : (None, (type(None), dict)),  # standard
     'protocol': (u'http', str),  # snowflake
     'warehouse': (None, (type(None), str)),  # snowflake
     'region': (None, (type(None), str)),  # snowflake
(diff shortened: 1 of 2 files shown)


sfc-gh-mkeller:
Seems like a duplicate of #352

sfc-gh-mkeller:
@adrianoesch 

Please understand that this is a [behavior change](https://community.snowflake.com/s/article/Pending-Behavior-Change-Log), so I need to get some approvals and announcements before this can be accepted and released, if we were to proceed with this.
I think in this case adding support for reading out proxy settings from the previously used environmental variables would be easier and better to do.

Would you be able to provide code for that?

src/snowflake/connector/proxy.py
```diff
@@ -39,4 +40,7 @@ def set_proxies(proxy_host, proxy_port, proxy_user=None, proxy_password=None):
         }
         os.environ['HTTP_PROXY'] = proxies['http']
         os.environ['HTTPS_PROXY'] = proxies['https']
+    warn("Arguments `proxy_host`, `proxy_port`, `proxy_user`"
```
adrianoesch, line 43: Maybe like this? Keep the old method but remove from docs and raise a warning?

sfc-gh-mkeller: No, sorry. I meant that we instruct customers to possibly set Proxies like this: https://docs.snowflake.com/en/user-guide/python-connector-example.html#using-a-proxy-server

For this change to not be a behavior change we should read out these 2 settings out of the environmental variables and then pass them to `requests`.

Because if someone set up a proxy with env variables months ago then the new connector version should still respect those settings.

adrianoesch:
`requests` supports both proxy methods (via [env or args](https://2.python-requests.org/en/master/user/advanced/#proxies)), so this should not break anything. my issue with environment variables is that they're global and I only want the snowflake requests to go through a proxy.

sfc-gh-mkeller:
Sorry @adrianoesch kicking off tests now!

src/snowflake/connector/proxy.py
sfc-gh-mkeller: Ahh, the reason why the tests are failing is because this function is called every time a connection is created and whenever this function is called even with `None`s it will raise a warning.
I'd need an actual approval to introduce a deprecation warning, would it be possible to remove this warning for now and put the information about the `proxies` connection parameter into `SnowflakeConnection`'s docstring?

adrianoesch: warning removed

src/snowflake/connector/connection.py
```diff
@@ -170,10 +171,7 @@ class SnowflakeConnection(object):
         host: The host name the connection attempts to connect to.
         port: The port to communicate with on the host.
         region: Region name if not the default Snowflake Database deployment.
-        proxy_host: The hostname used proxy server.
-        proxy_port: Port on proxy server to communicate with.
-        proxy_user: User name to login with on the proxy sever.
-        proxy_password: Password to be used to authenticate with proxy server.
+        proxies: Proxies attribute to use in requests.
```
adrianoesch, line 174: And the docstring reflects changes already.

Issue #377, opened 2020-08-25
SNOW-186747: Release 2.3.0 breaks requests for other modules


dodijk:
In upgrading to the latest release of the Snowflake Python connector, we have seen that using the requests library from other 
modules is broken. I think this is caused by the [monkey patching `urllib3` with `ssl_wrap_socket_with_ocsp`](https://github.com/snowflakedb/snowflake-connector-python/blob/master/src/snowflake/connector/ssl_wrap_socket.py#L74) and the [removal of a default value for `FEATURE_OCSP_MODE`](https://github.com/snowflakedb/snowflake-connector-python/commit/83dfe771c5404657f2008fec15bf0386185b6d70#diff-a8cb7f98804ae57600a6efa12d4b50dfL17). 

1. What version of Python are you using (`python --version`)?
Python 3.7.5

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
Darwin-19.6.0-x86_64-i386-64bit

3. What are the component versions in the environment (`pip list`)?
```
snowflake-connector-python     2.3.0`
requests                       2.24.0
requests-oauthlib              1.3.0
```
Partial list, I can add more where relevant.

4. What did you do?
I created a minimal example:
```
import snowflake.connector
import requests

requests.post('https://postman-echo.com')
```

5. What did you expect to see?
```
<Response [404]>
```

6. What did you see instead?
```
NameError: name 'FEATURE_OCSP_MODE' is not defined
```

7. Can you set logging to DEBUG and collect the logs?

```
import logging
import os

for logger_name in ['snowflake.sqlalchemy', 'snowflake.connector', 'botocore']:
    logger = logging.getLogger(logger_name)
    logger.setLevel(logging.DEBUG)
    ch = logging.StreamHandler()
    ch.setLevel(logging.DEBUG)
    ch.setFormatter(logging.Formatter('%(asctime)s - %(threadName)s %(filename)s:%(lineno)d - %(funcName)s() - %(levelname)s - %(message)s'))
    logger.addHandler(ch)
```
```
>>> import snowflake.connector
2020-08-25 11:23:12,973 - MainThread ssl_wrap_socket.py:73 - inject_into_urllib3() - DEBUG - Injecting ssl_wrap_socket_with_ocsp
2020-08-25 11:23:12,975 - MainThread auth.py:47 - <module>() - DEBUG - Failed to import keyring module. err=[No module named 'keyring']
[...]

PR #380, opened 2020-08-26
RETRO-246 adding extras tests

sfc-gh-mkeller:
Adding new set of tests to test some special cases with the connector.
This in specific is to catch the issue that caused 2.3.1 release.
Tests #377

### DIFF
--- a/test/extras/no_use_pycon.py
+++ b/test/extras/no_use_pycon.py
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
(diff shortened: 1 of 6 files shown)


test/extras/no_use_pycon.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
```
sfc-gh-stakeda, line 4: nit; 2020

sfc-gh-mkeller: Done, sorry I named the commit wrongly here but it's going to be squashed anyway

PR #381, opened 2020-08-26
SNOW-183542 add check-wheel-contents

sfc-gh-mkeller:
SNOW-183542
I recently learned of this tool and I wanted to see if this tool would help us with catching some issues.
Not sure if we're going use/merge this yet.


sfc-gh-mkeller:
`check-wheel-contents` does not support Python 3.5, but 3.5 is EOL next month, so I will just try this again once we drop support for it.

groodt:
It seems you've dropped 3.5 support now, so you could integrate this change. It does seem like it might be helpful.

PR #391, opened 2020-09-02
update snowflake.net links to snowflake.com

jameslamb:
When I clicked the docs links in the README, I noticed that I was redirected from `snowflake.net` to `snowflake.com`. I confirmed this with `curl`

```shell
curl -I https://docs.snowflake.net/
```

returns this

```text
HTTP/1.1 301 Moved Permanently
Content-Length: 0
Connection: keep-alive
Date: Wed, 02 Sep 2020 23:14:06 GMT
Location: /manuals/index.html
Server: AmazonS3
X-Cache: Miss from cloudfront
Via: 1.1 5971d213ff39e16c310a05523f08e121.cloudfront.net (CloudFront)
```

This PR proposes updating the existing documentation links that point to `.net` to `.com`. I tested these new links manually and all worked without redirects.

I didn't change the links in the test files, since I'm guessing that those are still valid.

image

## Why this PR is important

I have been in settings where browsers block redirects. This PR will help out any of your users that are in such settings, and might otherwise not be able to click through these links.

### DIFF
--- a/DESCRIPTION.rst
+++ b/DESCRIPTION.rst
@@ -2,7 +2,7 @@ This package includes the Snowflake Connector for Python, which conforms to the 
 https://www.python.org/dev/peps/pep-0249/
 
 Snowflake Documentation is available at:
-https://docs.snowflake.net/
+https://docs.snowflake.com/
 
 Source code is also available at: https://github.com/snowflakedb/snowflake-connector-python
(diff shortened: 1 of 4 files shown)


sfc-gh-mkeller (approved):
🚢 🚢

PR #393, opened 2020-09-03
SNOW-215479 remove unused variables

jameslamb:
Similar to my past few PRs (like #387 ), I've been spending a little time running linters over this project, looking for things I can help contribute.

Today I'm working with `pylint`. This turns up a bunch of things that should be left to maintainers or even ignored: style things like the using snake_case for variable names. However, there are a few things this linter found that I think should be addressed.

<details><summary>pylint command I ran</summary>

```shell
# pylint disables:
#   * C0301: line too long
#   * C0103: snake-case naming
#   * C0114: missing module docstring
#   * C0115: missing class docstring
#   * C0116: missing function docstring
#   * C0302: too many lines in module
#   * C0325: unnecessary parentheses after if / elif
#   * C0330: wrong hanging indent before block
#   * C0413: standard imports before third-part imports
#   * C0415: import outside top level
#   * E0401: unable to import
#   * R0201: method could be a function
#   * R0205: unnecessary object inheritance
#   * R0401: cyclic imports
#   * R0801: similar lines in multiple files
#   * R0902: too many instance attributes
#   * R0903: too few public methods
#   * R0904: too many public methods
#   * R0912: too many branches
#   * R0913: too many arguments
#   * R0914: too many local variables
#   * R0915: too many statements
#   * R1702: too many nested blocks
#   * R1705: unnecessary elif after return
#   * R1710: all return statements should return an expression or none should
#   * R1720: unecessary else after raise
#   * R1723: unnecessary elif
#   * R1724: unnecessary elif after return
#   * R1725: use Python 3 style super()
#   * W0105: string statement has no effect (https://twitter.com/gvanrossum/status/112670605505077248)
#   * W0107: unnecessary pass
#   * W0212: access to protected member
#   * W0511: TODO comments
#   * W0603: using the 'global' statement
#   * W0703: catching too general exception
#   * W1202: use lazy string formatting in logging
pylint \
    --disable=C0103,C0114,C0115,C0116,C0301,C0302,C0325,C0330,C0413,C0415,E0401,R0201,R0205,R0401,R0801,R0902,R0903,R0904,R0912,R0913,R0914,R0915,R1702,R1705,R1710,R1720,R1723,R1724,R1725,W0105,W0107,W0212,W0511,W0603,W0703,W1202 \
    src/snowflake/connector/ \
    > pylint.txt

```

</details>

This PR proposes fixing the first class of issue I found: unused variables.

```text
[...]

PR #394, opened 2020-09-03
SNOW-215481 fix string formatting in exceptions

jameslamb:
Similar to #393 , this proposes fixing a small bug caught with `pylint`. See that PR for the command I ran.

```text
src/snowflake/connector/ocsp_snowflake.py:415:16: W0715: Exception arguments suggest string formatting might be intended (raising-format-tuple)
src/snowflake/connector/remote_storage_util.py:150:16: W0715: Exception arguments suggest string formatting might be intended (raising-format-tuple)
src/snowflake/connector/remote_storage_util.py:248:16: W0715: Exception arguments suggest string formatting might be intended (raising-format-tuple)
```

A few exceptions in this library's Python code have a small mistake in the way they're handling string formatting. The code and screenshot below illustrate the issue:

```python
thing = "abc"

# will raise an exception with a tuple and the format string
raise Exception(
    "Unknown Error in downloading a file: %s",
    thing
)

# will raise an exception with the intended message
raise Exception(
    "Unknown Error in downloading a file: {}".format(thing)
)
```

image

I observed this behavior on Python version 3.7.6.

This PR proposes replacing `%s` with `.format()`, which will work in all versions of Python.

### How does this PR improve `snowflake-connector-python`?

This makes the messages in these exceptions slightly more accurate, removing stuff that might be misleading (like the presence of format strings like `%s`).

Thanks for your time and consideration!


jameslamb:
Does this need any other changes?

sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-mkeller:
Closes #396

sfc-gh-mkeller:
Merged through #396

PR #398, opened 2020-09-07
SNOW-80651 fire and forget/async query execution

sfc-gh-mkeller:
SNOW-80651
Adding Async query execution to the Python Connector.
Adds ability to retrieve data for at most day old queries.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1121,7 +1121,14 @@ class SnowflakeConnection(object):
 
         Args:
             sf_qid: Snowflake query id of interest.
+
+        Raises:
+            ValueError: if sf_qid is not a valid UUID string.
         """
+        try:
+            uuid.UUID(sf_qid)
+        except ValueError:
+            raise ValueError("Invalid UUID: '{}'".format(sf_qid))
         logger.debug('get_query_status sf_qid=\'{}\''.format(sf_qid))
 
         status = 'NO_DATA'
@@ -1145,6 +1152,9 @@ class SnowflakeConnection(object):
 
         Args:
             sf_qid: Snowflake query id of interest.
+
+        Raises:
+            ValueError: if sf_qid is not a valid UUID string.
         """
         status, status_resp = self._get_query_status(sf_qid)
         return status
@@ -1156,6 +1166,9 @@ class SnowflakeConnection(object):
 
         Args:
             sf_qid: Snowflake query id of interest.
+
+        Raises:
+            ValueError: if sf_qid is not a valid UUID string.
         """
         status, status_resp = self._get_query_status(sf_qid)
         queries = status_resp['data']['queries']
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -4,6 +4,7 @@
 # Copyright (c) 2012-2020 Snowflake Computing Inc. All right reserved.
 #
 
+import copy
 import logging
 import os
 import re
@@ -15,7 +16,7 @@ from io import StringIO
 from logging import getLogger
 from threading import Lock
 from time import strptime
-from typing import Callable, Dict, Generator, Iterable, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Generator, Iterable, List, Optional, Tuple, Union
 
 from . import errors, proxy
 from .auth import Auth
@@ -26,7 +27,7 @@ from .auth_oauth import AuthByOAuth
 from .auth_okta import AuthByOkta
 from .auth_webbrowser import AuthByWebBrowser
 from .chunk_downloader import DEFAULT_CLIENT_PREFETCH_THREADS, MAX_CLIENT_PREFETCH_THREADS, SnowflakeChunkDownloader
-from .compat import IS_LINUX, IS_WINDOWS, urlencode
+from .compat import IS_LINUX, IS_WINDOWS, quote, urlencode
 from .constants import (
     PARAMETER_AUTOCOMMIT,
     PARAMETER_CLIENT_PREFETCH_THREADS,
@@ -39,6 +40,7 @@ from .constants import (
     PARAMETER_SERVICE_NAME,
     PARAMETER_TIMEZONE,
     OCSPMode,
+    QueryStatus,
 )
 from .converter import SnowflakeConverter
 from .cursor import LOG_MAX_QUERY_LENGTH, SnowflakeCursor
@@ -203,6 +205,8 @@ class SnowflakeConnection(object):
         self._errorhandler = Error.default_errorhandler
         self._lock_converter = Lock()
         self.messages = []
+        self._async_sfqids = set()
+        self._done_async_sfqids = set()
         logger.info(
             "Snowflake Connector for Python Version: %s, "
             "Python Version: %s, Platform: %s",
@@ -429,7 +433,13 @@ class SnowflakeConnection(object):
             # close telemetry first, since it needs rest to send remaining data
             logger.info('closed')
             self._telemetry.close(send_on_close=retry)
-            self.rest.delete_session(retry=retry)
+            if self.safe_to_close():
+                logger.info('No async queries seem to be running, deleting session')
+                self.rest.delete_session(retry=retry)
+            else:
+                logger.info('There are {} async queries still running, not deleting session'.format(
+                    len(self._async_sfqids)
+                ))
             self.rest.close()
             self._rest = None
             del self.messages[:]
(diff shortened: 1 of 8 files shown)


.github/workflows/build_test.yml
```diff
@@ -230,6 +230,7 @@ jobs:
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
           cloud_provider: ${{ matrix.cloud-provider }}
+          PYTEST_ADDOPTS: --color=yes
```
sfc-gh-hchaturvedi, line 233: Nice.

src/snowflake/connector/connection.py
```diff
@@ -3,6 +3,7 @@
 #
 # Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
```
sfc-gh-hchaturvedi, line 4: Can you update this header?

.github/workflows/build_test.yml
sfc-gh-mkeller: You can see this in action on the Checks tab. 😄

src/snowflake/connector/connection.py
sfc-gh-mkeller: I'm always confused about these, are they supposed to start at the current year. Or what should I have written here?

sfc-gh-mkeller: Oh I guess it should be 2020, not 2019. Let me change that!

src/snowflake/connector/constants.py
```diff
@@ -164,3 +165,21 @@ class OCSPMode(Enum):
     FAIL_CLOSED = 'FAIL_CLOSED'
     FAIL_OPEN = 'FAIL_OPEN'
     INSECURE = 'INSECURE'
+
+
+@unique
```
sfc-gh-hchaturvedi, line 170: Nice!

src/snowflake/connector/cursor.py
```diff
@@ -911,6 +926,36 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        def wait_until_ready():
+            """Makes sure query has finished executing and once it has retrieves results."""
+            status = self.connection.get_query_status(sfqid)
+            no_data_counter = 0
+            no_data_max_retry = 24
+            while self.connection.is_still_running(status):
+                status = self.connection.get_query_status(sfqid)
+                if not self.connection.is_still_running(status) and status != QueryStatus.SUCCESS:
+                    raise DatabaseError(
```
sfc-gh-hchaturvedi, line 940: If this is an acceptable scenario do we really want to raise an error?

src/snowflake/connector/cursor.py
```diff
@@ -911,6 +926,36 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                        "Status of query '{}' is {}, results are unavailable".format(sfqid, status.name))
+                if status == QueryStatus.NO_DATA:
+                    no_data_counter += 1
+                    if no_data_counter > no_data_max_retry:
+                        raise DatabaseError("Cannot retrieve data on the status of this query. No information returned "
```
sfc-gh-hchaturvedi, line 945: same comment.

src/snowflake/connector/connection.py
sfc-gh-mkeller: I have also filed SNOW-194973 will do this separately from this PR tho.

src/snowflake/connector/cursor.py
sfc-gh-mkeller: I have copied this error from the JDBC PR.
But this would be triggered in the following case, you started a query and tried to get results while it was still running. Then you waited and all of the sudden it was done, but failed with an error.
I think raising an error in this case would be acceptable.

src/snowflake/connector/cursor.py
sfc-gh-mkeller: I have copied this error from the JDBC PR as well.
This would be triggered if someone wanted to wait for results on a still running query, but the query wasn't done in 25 * 5s = 125 s.
This is more to just be consistent with JDBC, maybe @sfc-gh-mknister could also comment if this was specifically requested by someone.

test/test_async.py
```diff
@@ -0,0 +1,160 @@
...
+            status = con.get_query_status(q_id)
+            assert status == QueryStatus.FAILED_WITH_ERROR
+            assert con.is_an_error(status)
+            with pytest.raises(ProgrammingError):
+                con.get_query_status_throw_if_error(q_id)
+            with pytest.raises(ProgrammingError):
+                cur.get_results_from_sfqid(q_id)
+
+
+def test_mix_sync_async(conn_cnx):
```
sfc-gh-wshangguan, line 62: A small question, what will happen if we start a sync query while there is already a running async query? I guess that's supported? If yes, then do we want to add a simple test for it?

src/snowflake/connector/cursor.py
```diff
@@ -876,6 +885,12 @@ def reset(self):
         self._total_rowcount = -1  # reset the rowcount
         if self._result is not None:
             self._result._reset()
+        if self._inner_cursor is not None:
+            self._inner_cursor.reset()
+            self._result = None
+            self._inner_cursor = None
+        if self._prefetch_hook is not None:
+            self._prefetch_hook = None
```
sfc-gh-wshangguan, line 893: nit: maybe the `if` can be removed?

test/test_async.py
sfc-gh-mkeller: One step ahead of you. Line 73 and 80 do that here.

src/snowflake/connector/cursor.py
sfc-gh-mkeller: Sure, good idea!

src/snowflake/connector/cursor.py
sfc-gh-wshangguan: Hi man @sfc-gh-mkeller , I was wondering if the pr you referenced was the right one. I guess you might also want to take a look at  https://github.com/snowflakedb/snowflake-jdbc/commit/eaed3362f837c399d95a3e4634a739fc37573fbd. 

So now the code has a problem that we might wait for another 5s even if the async query has finished successfully.  This was also in the original jdbc async pr. The pr I mentioned above is the new one to fix/optimize it, which is also the current async code for JDBC.

test/test_async.py
sfc-gh-wshangguan: Sorry, actually I mean to run a **sync** query when there is a still-running async query. But nvm, it looks like not an error-prone scenario to me now. I think we are good 😁

src/snowflake/connector/cursor.py
sfc-gh-mkeller: @sfc-gh-wshangguan I didn't know of this PR. Let me update the retry logic to mimic how JDBC retries.

src/snowflake/connector/connection.py
```diff
@@ -1094,3 +1104,101 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            if message is None:
+                message = ''
+            code = status_resp.get('code')
+            if code is None:
+                code = -1
+            sql_state = None
+            if 'data' in status_resp:
+                message += queries[0].get('errorMessage', '') if len(queries) > 0 else ''
+                sql_state = status_resp['data'].get('sqlState')
+            Error.errorhandler_wrapper(self, None,
```
sfc-gh-hchaturvedi, line 1166: @sfc-gh-mkeller It seems the message is meant to be exposed to the customer correct? I was just wondering if there's a possibility that the 'message' field has something that we don't want to show / log. Like some credentials / signatures?

sfc-gh-hchaturvedi (approved):
LGTM.

src/snowflake/connector/connection.py
sfc-gh-mkeller: I have made sure to raise Exceptions the same way as it is done when someone executes a synchronous query. So I hope that the way this is raising exceptions is appropriate 😅 

https://github.com/snowflakedb/snowflake-connector-python/blob/f4fe2277b82d98f2122478d5df712c59420ea0bd/src/snowflake/connector/cursor.py#L590

test/test_async.py
```diff
@@ -0,0 +1,159 @@
...
+def test_async_exec(conn_cnx):
+    """Tests whether simple async query execution works.
+
+    Runs a query that takes a few seconds to finish and then totally closes connection
+    to Snowflake. Then waits enough time for that query to finish, opens a new connection
+    and fetches results. It also tests QueryStatus related functionality too.
+    """
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            cur.execute('select count(*) from table(generator(timeLimit => 5))', exec_async=True)
```
sfc-gh-stakeda, line 27: could the method name be `execute_async` instead of having a parameter?

sfc-gh-mkeller: Would it work if I made a convenience function on top of the current implementation?
The function `execute_async` could just call `execute` with this parameter set to true.

sfc-gh-stakeda: Yes, that's exactly the point. Hide `exec_async` option by changing the name to `_exec_async`, though it won't be completely invisible.

sfc-gh-mkeller: Sure, can add that!

sfc-gh-stakeda (approved):

sfc-gh-wshangguan (approved):
LGTM!

sfc-gh-stakeda:
@sfc-gh-mkeller and @sfc-gh-sshankar, can we include this in the stability criteria?

sfc-gh-mkeller:
FYI, just rebased this onto master. If I made any errors during the process I will be pushing fixes

Issue #397, opened 2020-09-04
SNOW-194062: feature request: `fetch_dask_dataframe()`


jameslamb:
I've been working with this library for a few days. Thanks so much for maintaining it! It has made it really easy to work with data from Snowflake using PyData libraries.

I'd like to propose a feature that I think could improve the experience of using Snowflake results with other PyData libraries (like `dask`, `dask-ml`, and `scikit-learn`).

### Feature Request

Would you support a pull request that adds a `fetch_dask_dataframe()`? This method would have a similar API to `fetch_pandas_all()`, but would return a [Dask DataFrame](https://examples.dask.org/dataframe.html) instead of a `pandas` data frame.

I'd be happy to contribute this if you're interested in it.

### How this feature would improve `snowflake-connector-python`

Dask is a popular open source tool for working with larger-than-memory datasets in Python. Efficiently creating a Dask Dataframe from a Snowflake result today requires a lot of knowledge of Dask and Snowflake, and having a client method that does this would reduce a lot of friction for users trying to work with large query results in Python.

This feature would provide an easy path to working with large query results in Python.

### Overview 

Today, there are two ways to read the results of a query into a data frame in a Python session (documented [here](https://docs.snowflake.com/en/user-guide/python-connector-pandas.html#reading-data-from-a-snowflake-database-to-a-pandas-dataframe)):

* `fetch_pandas_all()`: pull the entire result into a single `pandas` data frame
* `fetch_pandas_batches()`: iterate over chunks of a query result one `pandas` data frame at a time

If your query result, once it's been converted to a `pandas` data frame, is too large to fit in memory on the machine where your code is running, `fetch_pandas_all()` will run out of memory and fail. One way to get around this is to iterate over smaller chunks one at a time with `fetch_pandas_batches()` and then do something with each batch, like writing it to a file.

Another option for working with a larger-than-memory dataset might be to create a [Dask DataFrame](https://examples.dask.org/dataframe.html). This would allow you to work with larger-than-memory data and have Dask handle the details of spilling to / reading from disk.

To do this, you'd have to template your query and create a series of `dask delayed` functions. Each function could get a piece of the query, execute it, and call `fetch_pandas_all()` on the result. That work will be spread across the Dask cluster (or executed as individual tasks on one machine, if you're using a Dask `LocalCluster`).

Something like this:

```python
from dask import delayed
import dask.dataframe as dd

query = """
    SELECT
[...]

PR #399, opened 2020-09-10
SNOW-195747 add more object files, core dumps to gitignore

jameslamb:
While exploring #397 today, I found that when I made mistakes working with Arrow, a core dump file was generated.

These are not ignored in this project's `.gitignore` today. This PR proposes that they should be ignored.

image

While touching `.gitignore`, I also noticed that some other common C object file extensions aren't currently ignored. This PR also proposes ignoring those.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,10 @@ __pycache__/
 *$py.class
 
 # C extensions
+*.a
+*.dll
+*.exe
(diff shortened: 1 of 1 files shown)


sfc-gh-mkeller (approved):
Thank you  @jameslamb 🚢 🚢

PR #400, opened 2020-09-11
[work-in-progress] SNOW-194062 add support for fetching a Dask dataframe (fixes #397)

### LINKED ISSUE #397: SNOW-194062: feature request: `fetch_dask_dataframe()`

jameslamb:
In #397 , I described a potential feature to work with large results: creating a Dask dataframe from a query result. Because Snowflake query results are already chunked, it's possible to create a Dask cluster and spread responsibility for pulling the chunks across all of the workers in that cluster. This allows the entire result to be in memory (spread over multiple machines) or to transparently spill to disk (useful if you're using Dask on a single machine).

I'm opening this draft PR to show a working implementation I've come up with, to aid in the discussion. I understand that this still needs a lot of work...logging, tests, error handling, and other things. But before I invest the time into those things or ask the maintainers here to invest the time into reviewing, I'd like to get buy-in on whether this type of feature is something the maintainers support.


jameslamb:
Sorry! I made some git mistakes in opening this PR. I tried to re-open it but GitHub won't let me. You can totally ignore this, I've opened #401

PR #407, opened 2020-09-15
SNOW-197039 add feature request issue template

jameslamb:
Thanks for this great project!

While contributing here, I've found that the issue template that shows up when you click `New Issue` has language that is specific to bug reports, such as:

> Can you set logging to DEBUG and collect the logs?

In this PR, I'd like to propose creating separate templates for bug reports and feature requests. I think this will help to provide some structure and avoid unnecessary information in feature reports.

With the setup proposed in this PR, when someone clicks `New Issue`, they'll see this choice:

image

You can read more about having multiple issue templates [in the GitHub docs](https://docs.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates).

Thanks for your time and consideration!

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -1,3 +1,8 @@
+---
+name: Bug Report 🐞
+about: Something isn't working as expected? Here is the right place to report.
+---
+
 Please answer these questions before submitting your issue. Thanks!
 
 1. What version of Python are you using (`python --version`)?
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -1,3 +1,8 @@
+---
+name: Bug Report 🐞
+about: Something isn't working as expected? Here is the right place to report.
+---
```
jameslamb, line 4: everything above the `---` in this PR came directly from the templates we use in `LightGBM` (https://github.com/microsoft/LightGBM/tree/master/.github/ISSUE_TEMPLATE)

sfc-gh-mkeller (approved):
This is great, thank you @jameslamb

sfc-gh-mkeller (approved):
This is great, thank you @jameslamb

PR #409, opened 2020-09-17
SNOW-195220 moved flake8 setting to tox.ini

sfc-gh-mkeller:
`flake8` supports reading config values straight from `tox.ini`

This could remove `setup.cfg` from our repository.


sfc-gh-kwagner (approved):
LGTM

PR #410, opened 2020-09-17
SNOW-194973 update copyright headers

sfc-gh-mkeller:
This PR introduces precommit hooks to require copyright notices in each cpp, hpp, py and pyx files.

This will also help us update them easily in the future, procedure description: https://github.com/Lucas-C/pre-commit-hooks#removing-old-license-and-replacing-it-with-a-new-one

### DIFF
--- a/src/snowflake/connector/arrow_iterator.pyx
+++ b/src/snowflake/connector/arrow_iterator.pyx
@@ -1,7 +1,3 @@
-#
-# Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
-#
-
 # distutils: language = c++
 # cython: language_level=3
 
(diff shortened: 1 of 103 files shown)


src/snowflake/connector/arrow_iterator.pyx
```diff
@@ -1,6 +1,4 @@
-#
-# Copyright (c) 2012-2019 Snowflake Computing Inc. All right reserved.
-#
+# Copyright (c) 2012-2020 Snowflake Computing Inc. All right reserved.
```
sfc-gh-kwagner, line 1: Nit: Can we keep the three lines for the copyright with this tool?

sfc-gh-kwagner (approved):
LGTM

src/snowflake/connector/arrow_iterator.pyx
sfc-gh-mkeller: Sure, I didn't think this tool would be smart enough, but I tried and it worked 🎉
But that made the C++ comments 5 lines long, so I switched those to `//` style instead, hope that's okay with you!

PR #418, opened 2020-09-21
SNOW-200637 support Arrow 2.0.x and ease building conda packages

xhochy:
The added environment variables would help us get rid of patching the package in the conda package.

PR #419, opened 2020-09-21
SNOW-200637 ease building conda packages

sfc-gh-mkeller:
Tests #418

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,8 @@ if _ABLE_TO_COMPILE_EXTENSIONS:
             current_dir = os.getcwd()
 
             if ext.name == 'snowflake.connector.arrow_iterator':
-                self._copy_arrow_lib()
+                if not os.environ.get("SF_NO_COPY_ARROW_LIB", False):
+                    self._copy_arrow_lib()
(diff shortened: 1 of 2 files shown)


xhochy:
Mh, weird.

I manually downloaded the MacOS Python 3.8 wheel and I can import `snowflake.connector.arrow_result`. So I'm a bit confused why it seems in the test setup, the import fails.

sfc-gh-mkeller:
The back-end seems to decide to not return Arrow format to the client and defaults to return JSON instead. I need to look into why this is.

xhochy:
This is still failing as it cannot import `snowflake.connect.arrow_result` in the CI. I'm not a tox expert, @sfc-gh-mkeller do you have an idea where we could add `python -c 'import snowflake.connect.arrow_result'` in the tests to see what is the issue?

xhochy:
🎉 looks like the issues are unrelated to my changes and most of CI is now green.

sfc-gh-mkeller:
@xhochy would your PR be okay to merge like this? I just rebased it onto master

xhochy (approved):
Yes, that looks fine/nice.

sfc-gh-kwagner (approved):
LGTM

PR #431, opened 2020-09-23
SNOW-197364 fixed assignee for Jiras generated

sfc-gh-mkeller:
New Jiras will have now the proper assignee.
Atlassian deprecated user names for account ids.


sfc-gh-mkeller:
Note: I have tested this manually with curl

sfc-gh-stakeda (approved):
Thanks for the fix!

sfc-gh-mkeller:
See #433 on Jira

Issue #156, opened 2019-03-07
KeyError: 'endpoint_resolver'


bt-wil:
I'm seeing intermittent errors when uploading files via 
* `PUT file://... ... AUTO_COMPRESS = TRUE PARALLEL = 8`

I see issues at the start of the script, but goes away after the a few passes of processing.

```
def process(self):
    with ThreadPoolExecutor(max_workers=32) as pool:
        files = [...]
        results = pool.map(self.load_file_to_snowflake, files)

def load_file_to_snowflake(self, fname):
    with SnowflakeConnection(...) as conn, conn.cursor() as curs:
        res = curs.execute(
            'PUT file://{0} {1} AUTO_COMPRESS = TRUE PARALLEL = 8'.format(fname, stage_table)
        )
        res.fetchall()
```

This exception that get raised.
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/<file>.py", line xyz, in <function>
    'PUT file://{0} {1} AUTO_COMPRESS = TRUE PARALLEL = 8'.format(..., ...)
  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/cursor.py", line 515, in execute
    sf_file_transfer_agent.execute()
  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/file_transfer_agent.py", line 212, in execute
    self._transfer_accelerate_config()
  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/file_transfer_agent.py", line 285, in _transfer_accelerate_config
    use_accelerate_endpoint=False)
  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/remote_storage_util.py", line 52, in create_client
    return util_class.create_client(stage_info, use_accelerate_endpoint=False)
  File "/usr/local/lib/python3.6/dist-packages/snowflake/connector/s3_util.py", line 75, in create_client
    config=config,
  File "/usr/local/lib/python3.6/dist-packages/boto3/__init__.py", line 100, in resource
    return _get_default_session().resource(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/boto3/session.py", line 389, in resource
    aws_session_token=aws_session_token, config=config)
  File "/usr/local/lib/python3.6/dist-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
[...]

PR #437, opened 2020-09-29
SNOW-216307 Make PUT statements threadsafe

jameskrach:
The [documentation](https://docs.snowflake.com/en/user-guide/python-connector-api.html#threadsafety) indicates that the connection can be shared by threads, but by using the [default boto3 session](https://github.com/snowflakedb/snowflake-connector-python/blob/14c96fbc279eca5a1440be927f5095dae8d50322/src/snowflake/connector/s3_util.py#L75), PUT statements are not threadsafe.

I believe this to be the cause of #156. Further docs of the non-threadsafety of boto3 can be seen [here](https://github.com/boto/botocore/issues/1246).

This PR simply creates a `boto3.Session` within `snowflake.connector.s3_util.SnowflakeS3Util.create_client()`. This will prevent multiple threads that share a `snowflake.connector.Connection` object from attempting to unsafely share the default boto3 connection when running a statement like `.execute("PUT ...")`.


jameskrach:
Any thoughts or questions on this? I’m happy to be helpful in getting this resolved however I can.

sfc-gh-mkeller:
No, sorry @jameskrach I just haven't gotten around to read all of the documentation yet.
I promise I'll get around to early next week.

sfc-gh-mkeller:
Hi @jameskrach right here in the docs boto3 says that sessions are not thread safe https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html#multithreading-and-multiprocessing
Are you sure that your code would make AWS PUTs thread safe here?

jameskrach:
I believe the `.create_client()` method is called by the connector whenever a `PUT` statement gets executed. So if the method is called within a thread (e.g. by a thread that's sharing a connection), and the session creation happens within the thread, then there shouldn't be any sharing of the session across threads.

I haven't run this (I can try it out tomorrow), but a a minimum example should look something like the code below. eventually that while loop should raise some kind of boto error (it might require adjusting the rowcount/number of concurrent uploads)

```
from concurrent.futures import ThreadPoolExecutor

import pandas as pd
from snowflake.connector import connect
from snowflake.connector.pandas_tools import write_pandas


kwargs = {...}
df = pd.DataFrame({"A": range(500), "B": range(500, 1000)})
tbl_nm = "bug_example"
uploads_at_a_time = 10


with connect(**kwargs) as conn, ThreadPoolExecutor(max_workers=5) as tp:
    with conn.cursor() as cur:
        _ = cur.execute(f"CREATE TABLE {tbl_nm} (a NUMBER, b NUMBER)")

    f = lambda n: write_pandas(conn, df, tbl_nm.upper(), parallel=1)
    results = []

    while True:
        results.append(tp.map(f, range(uploads_at_a_time)))
        print(results[-uploads_at_a_time:])
```

sfc-gh-mkeller:
I think the proper way to do this would be to return the Session as the client and then each worker thread could create their own Resource from that.

Does that sound like the proper way to fix the concurrency issue?

sfc-gh-stakeda:
According to https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html#multithreading-and-multiprocessing, each thread needs its own `session` as well as `resource`. So it msees the main thread should pass the credentials to the downloader threads to get authenticated individually.

jameskrach:
The docs are a bit ambiguous about whether or not sessions are threadsafe, but resources certainly aren’t.

Since (I assume) most PUT operations spend a fairly large amount of time uploading data relative to how long it takes to create a session, I’d imagine @sfc-gh-stakeda’s suggestion of creating a session as well is a good one

sfc-gh-stakeda:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/session.html#multithreading-or-multiprocessing-with-sessions
> Similar to Resource objects, Session objects are not thread safe and should not be shared across threads and processes.

I imagine the session is stateful that impacts a series of function calls.

sfc-gh-mkeller:
I guess I'll take over this ticket, unless you'd really like to implement this @jameskrach

jameskrach:
Seeing as how I missed the right spot to implement the change on the first go round I think I’d leave it to you

sfc-gh-mkeller:
Thank you for bringing this to our attention, @jameskrach

PR #413, opened 2020-09-18
SNOW-164802 resolve flaky tests

sfc-gh-mkeller:
Flaky tests are the worst, this PR tries to resolve them all.

These are going to be long running tests.

PR #441, opened 2020-09-29
SNOW-164802 resolve OCSP flakiness and add more logging

sfc-gh-mkeller:
From #413 
This fixes the most often failing flaky tests

It mostly just adds more logging and fixes the tests for the OCSP cache the right way.

### DIFF
--- a/test/test_ocsp.py
+++ b/test/test_ocsp.py
@@ -7,6 +7,7 @@
 import codecs
 import json
 import logging
+import os
 import tempfile
 import time
 from concurrent.futures.thread import ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


test/test_ocsp.py
```diff
@@ -212,8 +220,7 @@ def test_ocsp_with_bogus_cache_files(tmpdir):
 
     # forces to use the bogus cache file but it should raise errors
     SnowflakeOCSP.clear_cache()
-    ocsp = SFOCSP(
-        ocsp_response_cache_uri='file://' + cache_file_name)
```
sfc-gh-stakeda, line 216: nit; maybe we won't need this parameter any longer, then in the future change, can we remove it?

sfc-gh-stakeda (approved):

test/test_ocsp.py
sfc-gh-mkeller: I will give myself a ticket to remove it (SNOW-200287).

sfc-gh-hchaturvedi: Are we absolutely sure that the customer will not need it?

sfc-gh-hchaturvedi (approved):
Changes look good to me. I suggest we give it a stronger thought that the customer will never need to set an alternate location to the OCSP response cache file before removing that option. If we decide to keep it then we should be looking at fixing the connector to use the user specified location (if set) rather than reverting to a hardcoded location every time.

PR #451, opened 2020-10-01
Improve docs

PratikSawalkar:


sfc-gh-mkeller:
Hi @PratikSawalkar 
I'd say that this extra wording is not necessary. By not specifying a version of Windows in our current  README should mean that it applies to all versions of Windows.

PR #455, opened 2020-10-01
SNOW-196990 ocsp better error messages

sfc-gh-hchaturvedi:
The goal of this fix is to add more granularity to error messages / out of band telemetry that is generated by the driver. This will not only make the error causes more meaningful for the customers to discern themselves but also for cases where the customer cannot gather logs, will add more context to out of band telemetry for the support team to consume and debug in the absence of the logs. 

The new information is in the form of event-sub-type. Each event sub type will have a one-one mapping with documentation explaining what the error means and what are the potential solutions to look into.


src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -1083,7 +1099,7 @@ def validate_by_direct_connection(self, issuer, subject, telemetry_data, hostnam
                 logger.debug("using OCSP response cache")
 
             if not ocsp_response:
-                logger.debug('No OCSP URL is found.')
+                telemetry_data.set_event_sub_type('OCSPResponseUnavailable')
```
sfc-gh-stakeda, line 1102: can you make the sub type constant variables so that the application can refer the same value if required?

sfc-gh-stakeda:
what is your test plan?

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-hchaturvedi: Sure that makes sense.

sfc-gh-hchaturvedi:
> what is your test plan?

There is no feature addition in this jira. This was just meant to increase the granularity of the error information sent in the telemetry and in the log messages. If you're talking about the test coverage for OCSP in general then that is something we can take separately.

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -1100,6 +1116,20 @@ def validate_by_direct_connection(self, issuer, subject, telemetry_data, hostnam
...
+                    elif op_er.errno == ER_OCSP_RESPONSE_ATTACHED_CERT_INVALID:
+                        telemetry_data.set_event_sub_type("OCSPResponseAttachedCertInvalid")
+                    elif op_er.errno == ER_OCSP_RESPONSE_ATTACHED_CERT_EXPIRED:
+                        telemetry_data.set_event_sub_type("OCSPResponseAttachedCertExpired")
+                    elif op_er.errno == ER_OCSP_RESPONSE_INVALID_SIGNATURE:
+                        telemetry_data.set_event_sub_type("OCSPResponseSignatureInvalid")
+                    elif op_er.errno == ER_OCSP_RESPONSE_INVALID_EXPIRY_INFO_MISSING:
+                        telemetry_data.set_event_sub_type("OCSPResponseExpiryInfoMissing")
+                    elif op_er.errno == ER_OCSP_RESPONSE_EXPIRED:
+                        telemetry_data.set_event_sub_type("OCSPResponseExpired")
```
sfc-gh-stakeda, line 1132: Looks like a map from the errno to sub type would make sense.

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -1100,6 +1116,20 @@ def validate_by_direct_connection(self, issuer, subject, telemetry_data, hostnam
...
                     raise op_er
```
sfc-gh-stakeda, line 1133: what if no error is matched? no sub type?

sfc-gh-hchaturvedi: The goal was to populate telemetry data with the right event_sub_type. Part of the OCSP revocation check failed exceptions are raised in ocsp_asn1crypto.py. I had two choices, either expose telemetry data to ocsp_asn1crypto or use the err no of the revocation check exception raised in it to fill the telemetry data in the calling function itself. The point being I have added new error codes each of which corresponds to an event_sub_type in ocsp_snowflake.py. Removing the case of no error number is matched. In the future if we introduce new error numbers we will create a corresponding event_sub_type to maintain the granularity. Worse case if someone forgets to add a corresponding sub type, then it would show up as none which is the default.

sfc-gh-stakeda: ok, sound like the sub type is a string representation of errno for OOB. If you are concern about missing corresponding sub type for the errno, you could do this:
- Create a map/dict data structure from errno to sub type
- Create a test case to attempt to map all of the existing OCSP errno and ensure the sub type is returned. If not returned, it means no sub type is defined, so the developer will notice he/she has to add a subtype whenever a new OCSP errno is introduced.

sfc-gh-hchaturvedi: That's a good idea.  A few things: 
1. At first I did not think that all error numbers need to be mapped to event sub types since this mapping was just introduced to not have to send telemetry_data object to oscp_asn1crypto. So only the error numbers being used in ocsp_asn1crypto really needed to be mapped. That said I don't see any harm in mapping all ocsp related error number to an event sub type. 
2. About your idea to avoid a potential lapse by a developer. The developer will still need to update the map with the new error number after introducing it in the errorcode.py so that the test knows a new error number has been introduced.

sfc-gh-stakeda:
@sfc-gh-hchaturvedi please fix the test failures.

sfc-gh-hchaturvedi:
I am running into issues while trying to run tests in my local environment. Will send out a message once this is ready to review and has no outstanding errors. @sfc-gh-stakeda

test/test_unit_telemetry_oob.py
```diff
@@ -104,6 +106,15 @@ def test_telemetry_oob_http_log(telemetry_setup):
     assert telemetry.size() == 0
 
 
+def test_telemetry_oob_error_code_mapping():
+    """Tests that all OCSP error codes have a corresponding Telemetry sub event type."""
+    ec_dict = snowflake.connector.errorcode.__dict__
+    for ec in ec_dict:
+        if not ec.startswith('__'):
+            if 254000 <= ec_dict[ec] < 255000:
+                assert ec_dict in OCSPTelemetryData.ERROR_CODE_MAP
```
sfc-gh-mkeller, line 115: This is just an FYI, if you use `.items()` then you don't need to get the key and then retrieve the value with the key from the dict again.

```suggestion
    for key, value in ec_dict.items():
        if not key.startswith('__'):
            if 254000 <= value < 255000:
                assert key in OCSPTelemetryData.ERROR_CODE_MAP
```

Although I'm guessing what you wanted to to do with `ec_dict` in the last line, make sure that you wanted to check whether the whole dictionary was in the other dictionary.

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -62,6 +115,20 @@ def __init__(self):
         self.fail_open = False
         self.insecure_mode = False
 
+    def set_event_sub_type(self, event_sub_type):
```
sfc-gh-mkeller, line 118: I'd like for everyone to start adding type hints to new functions.
Let me help adding them here, but for more info you could read: https://docs.python.org/3/library/typing.html
You could just accept my suggestions and then you don't need to add these locally first.

```suggestion
    def set_event_sub_type(self, event_sub_type: str) -> None:
```

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -62,6 +115,20 @@ def __init__(self):
...
+        # There can be multiple event_sub_types
+        # in the case where there was a problem
+        # with the cache.
```
sfc-gh-mkeller, line 121: Maybe a doc string would be better here
```suggestion
        """Sets the event's sub type.
        
        There can be multiple event_sub_types in the case where there
        was a problem, with the cache. Either appends the sub type, or sets it.
        """
```

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -244,14 +312,16 @@ def download_cache_from_server(self, ocsp):
...
+                    raise RevocationCheckError(msg="OCSP Cache Server Unavailable.",
+                                               errno=ER_OCSP_RESPONSE_CACHE_DOWNLOAD_FAILED)
                 logger.debug("downloaded OCSP response cache file from %s",
                              self.CACHE_SERVER_URL)
                 logger.debug("# of certificates: %s", len(OCSPCache.CACHE))
             except RevocationCheckError as rce:
                 logger.debug("OCSP Response cache download failed. The client"
                              "will reach out to the OCSP Responder directly for"
                              "any missing OCSP responses %s\n" % rce.msg)
+                raise rce
```
sfc-gh-mkeller, line 324:
```suggestion
                raise
```

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -62,6 +115,20 @@ def __init__(self):
...
+        # in the case where there was a problem
+        # with the cache.
+        if self.event_sub_type is not None:
+            self.event_sub_type = "{}|{}".\
+                format(self.event_sub_type,
+                       event_sub_type)
+        else:
+            self.event_sub_type = event_sub_type
+
+    def set_ocsp_connection_method(self, ocsp_conn_method):
```
sfc-gh-mkeller, line 129:
```suggestion
    def set_ocsp_connection_method(self, ocsp_conn_method: str) -> None:
```

sfc-gh-mkeller (changes requested):
Most of my recommendations are code hygiene related. So they are not super urgent, but please consider adding them.
You can accept my changes if you agree just by batching them and committing them from the UI here.

Please check my comment about the unit test, I think there's a typo in that.

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-hchaturvedi: Thanks for the pointer. I know what type hints are :) , will remember to add them.

sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-stakeda (approved):

sfc-gh-hchaturvedi:
@sfc-gh-mkeller are you aware of any flaky behavior in Python 3.7? I got these three failures and all three of them have this requests not found error message 

`Traceback (most recent call last):
  File "test/extras/run.py", line 17, in <module>
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 444, in check_returncode
    self.stderr)
subprocess.CalledProcessError: Command '['python', 'test\\extras\\no_use_pycon.py']' returned non-zero exit status 1.
ERROR: InvocationError for command 'D:\a\snowflake-connector-python\snowflake-connector-python\.tox\py37-extras-ci\Scripts\python.EXE' test/extras/run.py (exited with code 1)`

sfc-gh-mkeller:
> @sfc-gh-mkeller are you aware of any flaky behavior in Python 3.7? I got these three failures and all three of them have this requests not found error message
> 
> `Traceback (most recent call last): File "test/extras/run.py", line 17, in <module> sub_process.check_returncode() File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\subprocess.py", line 444, in check_returncode self.stderr) subprocess.CalledProcessError: Command '['python', 'test\\extras\\no_use_pycon.py']' returned non-zero exit status 1. ERROR: InvocationError for command 'D:\a\snowflake-connector-python\snowflake-connector-python\.tox\py37-extras-ci\Scripts\python.EXE' test/extras/run.py (exited with code 1)`

Yes, it seems like it's a path issue. Tox is supposed to update `python` to be in the environment that it's using. However, this seems to consistently fail since yesterday on Windows with Python 3.7. I have been blaming this on GH actions.

Let's see if the issue resolves itself, or not today!

sfc-gh-hchaturvedi:
@sfc-gh-mkeller / @sfc-gh-abhatnagar Guys seems like there is some issue with the way tests are being built .. Can one of you please confirm? This has been open for a while and I haven't been able to commit due to tests, I would appreciate if I can merge this PR today unless of course there's problem with the new code I am adding.

sfc-gh-mkeller:
> @sfc-gh-mkeller / @sfc-gh-abhatnagar Guys seems like there is some issue with the way tests are being built .. Can one of you please confirm? This has been open for a while and I haven't been able to commit due to tests, I would appreciate if I can merge this PR today unless of course there's problem with the new code I am adding.

I wish I could tell you what happened, but since Thursday/Friday the tests have been not working the same as before 😓

sfc-gh-hchaturvedi:
@sfc-gh-stakeda can we merge this?

sfc-gh-stakeda:
Not yet. Restarted pr-head.

sfc-gh-hchaturvedi:
> Not yet. Restarted pr-head.

I have done that twice before already and I guess it's jinxed, hopefully it'll pass this time since you started it :)

sfc-gh-hchaturvedi:
@sfc-gh-stakeda it has failed again, notice the highlighted line : 

2020-10-20 15:10:36 Build step 'Groovy Postbuild' changed build result to NOT_BUILT
2020-10-20 15:10:36 [GitHub Checks] Causes for no suitable publisher found: 
2020-10-20 15:10:36 [GitHub Checks] Trying to resolve checks parameters from Git SCM...
**2020-10-20 15:10:36 [GitHub Checks] No GitHub app credentials found: 'b4f59663-ae0a-4384-9fdc-c7f2fe1c4fca'**
2020-10-20 15:10:36 [GitHub Checks] See: https://github.com/jenkinsci/github-branch-source-plugin/blob/master/docs/github-app.adoc
2020-10-20 15:10:36 [GitHub Checks] Trying to resolve checks parameters from GitHub SCM...
2020-10-20 15:10:36 [GitHub Checks] Job does not use GitHub SCM
2020-10-20 15:10:36 [Checks API] No suitable checks publisher found.
2020-10-20 15:10:36 [WS-CLEANUP] Deleting project workspace...
2020-10-20 15:10:36 [WS-CLEANUP] Skipped based on build state NOT_BUILT

sfc-gh-stakeda:
what is the link?

sfc-gh-hchaturvedi:
> what is the link?

https://ci44.int.snowflakecomputing.com/job/RT-PyConnector36-PC/101/console

sfc-gh-stakeda:
See the last on https://ci44.int.snowflakecomputing.com/job/PyConnector/job/snowflake-connector-python/job/PR-455/5/console Looks like they are running tests.

sfc-gh-hchaturvedi:
@sfc-gh-stakeda seems like it passed ..

PR #459, opened 2020-10-02
SNOW-201468 fix test names

sfc-gh-mkeller:
I tried changing some of our `tox` configuration to allow me locally to run a single test easily.
However, I forgot that we do something a little different in Jenkins that would break with that change, here in this PR I revert the part of that change that is causing the Jenkins failures, I will give more thought to how to properly resolve this in the future.
Sorry for the failures!

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -276,5 +276,6 @@ jobs:
       - name: Combine coverages
         run: python -m tox -e coverage
       - uses: codecov/codecov-action@v1
+        if: false
         with:
           file: .tox/coverage.xml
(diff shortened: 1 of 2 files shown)


sfc-gh-stakeda:
btw @sfc-gh-mkeller , could you disable Codecov?

sfc-gh-stakeda:
I'm curious why you didn't catch the last push broke jenkins?

sfc-gh-mkeller:
It's because lately there was so much noise in my e-mail caused by the checkout issues and by the time the PR was merged I even forgot that I changed some tox configuration and I thought it was some temporary issue.
I'm sorry, I'll keep up with the tests better.

sfc-gh-stakeda (approved):

PR #460, opened 2020-10-02
Fixed SecretDetector Bug for None String

sfc-gh-kwagner:
Fixed SecretDetector Bug for None String. Also got secret detector to 100% code cov

### DIFF
--- a/src/snowflake/connector/secret_detector.py
+++ b/src/snowflake/connector/secret_detector.py
@@ -13,6 +13,7 @@ It masks secrets that might be leaked from two potential avenues
 import logging
 import os
 import re
+from typing import Tuple
 
 MIN_TOKEN_LEN = os.getenv("MIN_TOKEN_LEN", 32)
 MIN_PWD_LEN = os.getenv("MIN_PWD_LEN", 8)
@@ -70,7 +71,7 @@ class SecretDetector(logging.Formatter):
         return SecretDetector.PRIVATE_KEY_DATA_PATTERN.sub('"privateKeyData": "XXXX"', text)
 
     @staticmethod
-    def mask_secrets(text: str) -> str:
+    def mask_secrets(text: str) -> Tuple[bool, str, str]:
         """Masks any secrets. This is the method that should be used by outside classes.
 
         Args:
@@ -80,7 +81,7 @@ class SecretDetector(logging.Formatter):
             The masked string.
         """
         if text is None:
-            return (True, None, None)
+            return (False, None, None)
 
         masked = False
         err_str = None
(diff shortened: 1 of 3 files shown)


sfc-gh-mkeller (approved):
Great work @sfc-gh-kwagner 🚢 🚢 
Make sure that @sfc-gh-wshangguan reviews before merging though

src/snowflake/connector/secret_detector.py
```diff
@@ -70,7 +71,7 @@ def mask_private_key_data(text):
         return SecretDetector.PRIVATE_KEY_DATA_PATTERN.sub('"privateKeyData": "XXXX"', text)
 
     @staticmethod
-    def mask_secrets(text: str) -> str:
+    def mask_secrets(text: str) -> Tuple[bool, str, str]:
```
sfc-gh-stakeda, line 74: `telemetry.py` looks like expect this returns a str.

sfc-gh-wshangguan: +1

sfc-gh-kwagner: Thats a bug then. The secret detector has always returned a tuple. I'll fix that in telemetry.py and look for other misuses

sfc-gh-wshangguan: That might be a wrong usage of telemetry.py. Could you please help to fix that in telemetry.py? @sfc-gh-kwagner

sfc-gh-kwagner: @sfc-gh-stakeda @sfc-gh-wshangguan Fixed

sfc-gh-wshangguan:
Would be better to get review from author as well. Invited @sfc-gh-hchaturvedi .

sfc-gh-stakeda (approved):

sfc-gh-wshangguan (approved):
LGTM, that's a nice catch!

sfc-gh-stakeda:
sounds like the python's type checking doesn't work for all cases...

sfc-gh-kwagner:
> sounds like the python's type checking doesn't work for all cases...

@sfc-gh-stakeda from what I understand, Python's type checking is really just a docstring. Doesn't cause any compilation or build issues but is more for the IDE and new developers using that method

sfc-gh-wshangguan:
Agree with @sfc-gh-kwagner . There are some tools to help do the static analysis, but what we have now is not enough for achieving that goal.

src/snowflake/connector/secret_detector.py
```diff
@@ -80,7 +81,7 @@ def mask_secrets(text: str) -> str:
             The masked string.
         """
         if text is None:
-            return None
+            return (False, None, None)
```
sfc-gh-hchaturvedi, line 84: <facepalm>. Good catch.

sfc-gh-hchaturvedi (approved):
LGTM!

PR #471, opened 2020-10-09
Test PR 362

sfc-gh-mkeller:
I added my own test to what was included in #362 but otherwise the same code changes are included.

So this PR fixes a couple of issues with `pandas_tools`.
- It makes it insert default/autoincrement column values.
- It allows customers to use a non-quoted syntax for column and table names.

Added integration tests to verify new code paths. It adds all of these while not causing a behavior change.


sfc-gh-mkeller:
This one is ready for review!

sfc-gh-kwagner (approved):
LGTM

PR #473, opened 2020-10-12
SNOW-200123: Added more test coverage for file_transfer_agent

sfc-gh-kwagner:
Added type hints for the progress updating code in file transfer as well as adding test coverage for that code as well

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -242,7 +242,7 @@ jobs:
             .tox/coverage.xml
 
   combine-coverage:
-    if: true
+    if: ${{ success() || failure() }}
     name: Combine coverage
     needs: test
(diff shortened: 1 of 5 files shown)


sfc-gh-mkeller (approved):
Thanks Kris, looks great!

On a side note we should look into enforcing type hints in the foreseeable future, as the Python runtime does not enforce function and variable type annotations, we need a third party type checker for this.

Maybe Mypy? https://mypy.readthedocs.io/en/stable/getting_started.html
I think this is where the whole type hinting Python design was merged from

sfc-gh-mkeller:
Also Kris please go ahead and rebase onto master to see if the tests pass now

sfc-gh-stakeda (approved):

PR #474, opened 2020-10-14
SNOW-207403 pin virtualenv until issue is fixed upstream

sfc-gh-mkeller:
This is the short term solution to fix the new behaviour by `virtualenv`

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -224,7 +224,7 @@ jobs:
       - name: Upgrade setuptools, pip and wheel
         run: python -m pip install -U setuptools pip wheel
       - name: Install tox
-        run: python -m pip install tox tox-external-wheels
+        run: python -m pip install tox tox-external-wheels virtualenv==20.0.33
       - name: Run tests
         run: python -m tox -e "py${PYTHON_VERSION/\./}{-extras,,-pandas,-sso}-ci"
         env:


sfc-gh-hchaturvedi:
It would be great if you can explain in a little more detail what is the new behavior and how it disrupts the test. Another thing, I'd like to believe that since it's only the virtualenv version we're pinning which essentially is used for test environments, it should not affect any community user unless they end up running tests and have dependency mismatch?

sfc-gh-mkeller:
> It would be great if you can explain in a little more detail what is the new behavior and how it disrupts the test. Another thing, I'd like to believe that since it's only the virtualenv version we're pinning which essentially is used for test environments, it should not affect any community user unless they end up running tests and have dependency mismatch?

I'm still investigating what caused this unexpected behaviour change. In the newest release they introduced a redirector script for Windows and Python 3.7 and above to align with how `venv` works. Docs for venv: https://docs.python.org/3/library/venv.html

This is a short term fix, you can follow my issue with virtualenv here: https://github.com/pypa/virtualenv/issues/1981

I will be sharing my findings with you once I figured out what is going on but until then this will unblock PRs.

sfc-gh-hchaturvedi (approved):
LGTM. As confirmed with Mark this should not affect the customers as the change is only within the test setup.

sfc-gh-sshankar (approved):
Approval

PR #480, opened 2020-10-19
SNOW-164802 separate unit and integration tests

sfc-gh-mkeller:
This PR should now eliminate flaky tests. It does a few things:
(Sorry, if I forget something I'll keep this list growing)

- Add some more retryable errors (`TimeoutError`, `OSError`) to SSL checking.
- Redefined `tox` test environments, they all use tags now instead of the exact folders. This allows us to mic and match them if necessary later on and cleans up our tox configuration quite a bit.
- I defined test as a package, so we can import things from other directories inside of it.
- Added a brand new `conftest.py` into integ with the extra fixtures that are necessary only for integration tests (so this means that if a dev is running unit tests locally, no Snowflake login is required).
- Separated out unit tests from integration tests, where an integration test is defined as a tests that needs connection to a Snowflake deployment and unit tests are the leftovers (notice how unit tests still might require internet connection).
- Our py.test code automatically tags tests with whatever folders they are in inside of the `test` directory.
- Fixed a bug where a thread tried to acquire the same lock that it was holding already by using a `RLock` instead of a `Lock`.
- From now on the default `cloud_provider` environmental variable will be `dev`, which will never be skipped. This is useful because since I added the test tagging for specific cloud providers I have been finding my default of AWS to be more annoying than anything. So if the env variable `cloud_provider` is not set than all tests should run instead of skipping.
- Updated `test/README.rst` with the newest information.
- Added a Lazy Variable class to our new test package, it is to replace variables defined outside of tests, to be used by multiple tests. It is necessary because if that variable needs an optional dependency to resolve then it should only be evaluated if one of the using tests is ran.


src/snowflake/connector/ssl_wrap_socket.py
```diff
@@ -421,15 +421,9 @@ def _openssl_connect(hostname, port=443, max_retry=20):
...
-            if isinstance(ex, OpenSSL.SSL.SysCallError) or \
-                    ex.__class__.__name__ == 'TimeoutError':
-                err = ex
-                sleeping_time *= 2
-                if sleeping_time > 16:
-                    sleeping_time = 16
-                time.sleep(sleeping_time)
-            else:
-                raise ex
+        except (OpenSSL.SSL.SysCallError, ConnectionRefusedError, TimeoutError, OSError) as ex:
```
sfc-gh-stakeda, line 424: Is `TimeoutError` an exception class? I'm not sure if that works, though.

sfc-gh-mkeller: 3.3 and above yes https://docs.python.org/3/library/exceptions.html#TimeoutError

sfc-gh-mkeller: I have observed all of these exception types

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -897,7 +897,7 @@ class SnowflakeOCSP(object):
     ROOT_CERTIFICATES_DICT = {}  # root certificates
 
     # root certificate cache lock
-    ROOT_CERTIFICATES_DICT_LOCK = Lock()
+    ROOT_CERTIFICATES_DICT_LOCK = RLock()
```
sfc-gh-kwagner, line 900: +1

src/snowflake/connector/pandas_tools.py
```diff
@@ -165,7 +165,7 @@ def write_pandas(conn: 'SnowflakeConnection',
             copy_results)
 
 
-def pd_writer(table: pandas.io.sql.SQLTable,
+def pd_writer(table: 'pandas.io.sql.SQLTable',
```
sfc-gh-kwagner, line 168: Whats the reason for this change?

sfc-gh-mkeller: Since I'm switching to tagging test, instead of ignoring a growing list of sub-directories we need to be more careful on how we add Type hints.
If we didn't put this type into the string then Python would need to resolve `pandas.io.sql.SQLTable` (in which case our `options` sub-package would throw an Exception when the pandas optional dependency is not installed).
For this exact reason Python allows the types to be put into a string, so PyCharm and `mypy` should have no problem with the string.

sfc-gh-kwagner (approved):

PR #481, opened 2020-10-20
SNOW-206821 updated from add-path to environment files

sfc-gh-mkeller:
More info: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


sfc-gh-sshankar (approved):
Approval

PR #487, opened 2020-10-22
Added testing for expired tokens for S3 Put/Get

sfc-gh-kwagner:
Added testing for expired tokens for S3 Put/Get


sfc-gh-mkeller (commented):
I just saw that this is not ready, but I added some comments already. So here they are

PR #489, opened 2020-10-23
SNOW-196978 add old driver test for python

sfc-gh-stakeda:
- Add `py37-olddriver-ci2` tox env to run the old driver test.

This change is only for the Github actions.

I'll work on the jenkins change in a separate PR.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -30,216 +30,216 @@ jobs:
       - name: Run fix_lint
         run: python -m tox -e fix_lint
 
-#  build-sdist:
-#    needs: lint
-#    name: Build sdist
-#    runs-on: ubuntu-latest
-#    steps:
-#      - uses: actions/checkout@v1
-#      - name: Set up Python
-#        uses: actions/setup-python@v2
-#        with:
-#          python-version: '3.6'
-#      - name: Display Python version
-#        run: python -c "import sys; print(sys.version)"
-#      - name: Upgrade setuptools and pip
-#        run: python -m pip install -U setuptools pip
-#      - name: Creating source distribution
-#        run: python setup.py sdist
-#      - name: Show sdist generated
-#        run: ls -lh dist
-#      - uses: actions/upload-artifact@v2
-#        with:
-#          name: sdist
-#          path: dist/
+  build-sdist:
+    needs: lint
+    name: Build sdist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Upgrade setuptools and pip
+        run: python -m pip install -U setuptools pip
+      - name: Creating source distribution
+        run: python setup.py sdist
+      - name: Show sdist generated
+        run: ls -lh dist
+      - uses: actions/upload-artifact@v2
+        with:
+          name: sdist
+          path: dist/
 
-#  build-manylinux:
-#    needs: lint
-#    name: Build linux-py${{ matrix.python-version }}
-#    runs-on: ubuntu-latest
-#    container: quay.io/pypa/manylinux1_x86_64
-#    strategy:
-#      matrix:
-#        python-version: [3.5, 3.6, 3.7, 3.8]
-#    steps:
-#      - uses: actions/checkout@v1
-#      - name: Updating path
-#        run: |
-#          echo "/opt/python/cp35-cp35m/bin/" >> $GITHUB_PATH
-#          echo "/opt/python/cp36-cp36m/bin/" >> $GITHUB_PATH
-#          echo "/opt/python/cp37-cp37m/bin/" >> $GITHUB_PATH
-#          echo "/opt/python/cp38-cp38/bin/" >> $GITHUB_PATH
-#      - name: Display Python version
-#        run: python${{ matrix.python-version }} -c "import sys; print(sys.version)"
-#      - name: Upgrade setuptools and pip
-#        run: python${{ matrix.python-version }} -m pip install -U setuptools pip
-#      - name: Generate wheel
-#        run: python${{ matrix.python-version }} -m pip wheel -v -w dist_tmp --no-deps .
-#      - name: Run auditwheel
-#        run: auditwheel repair --plat manylinux2010_x86_64 -L connector dist_tmp/snowflake_connector_python*.whl -w dist
-#      - name: Show wheels generated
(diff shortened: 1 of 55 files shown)


sfc-gh-stakeda:
@sfc-gh-mkeller , is this change valid for old driver test? Basically we want to run the subset of test case by the specified version of driver.

.github/workflows/build_test.yml
```diff
@@ -241,6 +241,47 @@ jobs:
...
+    needs: lint
+    runs-on: ${{ matrix.os.image_name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+         - image_name: ubuntu-latest
+           download_name: linux
+        python-version: [3.7]
+        driver-version: [1.9.1]
```
sfc-gh-stakeda, line 255: The old driver version.

sfc-gh-stakeda:
It appeared all integration tests except pandas tests are success for the old driver.

tox.ini
```diff
@@ -38,6 +38,7 @@ external_wheels =
 setenv =
     COVERAGE_FILE = {env:COVERAGE_FILE:{toxworkdir}/.coverage.{envname}}
     ci: SNOWFLAKE_PYTEST_OPTS = -vvv --color=yes
+    ci2: SNOWFLAKE_PYTEST_OPTS = -vvv --color=yes
```
sfc-gh-stakeda, line 41: This skips referring to `dist/*py37*` file.

sfc-gh-mkeller: Alternatively we could use `pip download` to get the wheel file and then call regular `ci` with it.

sfc-gh-mkeller:
> It appeared all integration tests except pandas tests are success for the old driver.

In that case what do you think about applying the tag automatically? Or we could do if `pytest ... -m "not pandas" test` for old tests.

Do you think that manually having the oldtest mark is beneficial?

tox.ini
sfc-gh-stakeda: Yeah, but in this case, the name `dist/*cp37*` won't match anyway because 1.9.1 doesn't have the python version specific wheel. Yet the name of env `ci2` could be better name?

sfc-gh-stakeda:
> In that case what do you think about applying the tag automatically? Or we could do if `pytest ... -m "not pandas" test` for old tests.
> 
> Do you think that manually having the oldtest mark is beneficial?
> 
> edit: Or maybe `skipoldtest` would be better than `oldtest`?

Let me try `skipoldtest`, which will be in pandas and unit tests.

sfc-gh-stakeda:
@sfc-gh-mkeller , I found the olddriver test installs `snowflake-connector-python-2.3.4.zip`...
```
2020-10-26T23:47:32.9470273Z .tox installdeps: tox-external-wheels>=0.1.6, tox >= 3.7
2020-10-26T23:47:37.3947009Z GLOB sdist-make: /home/runner/work/snowflake-connector-python/snowflake-connector-python/setup.py
2020-10-26T23:47:37.8521955Z py37-olddriver-ci2 create: /home/runner/work/snowflake-connector-python/snowflake-connector-python/.tox/py37-olddriver-ci2
2020-10-26T23:47:38.1935086Z py37-olddriver-ci2 installdeps: pip >= 19.3.1
2020-10-26T23:47:39.3794984Z py37-olddriver-ci2 inst: /home/runner/work/snowflake-connector-python/snowflake-connector-python/.tox/.tmp/package/1/snowflake-connector-python-2.3.4.zip
```
How can I skip this?

sfc-gh-stakeda:
I was wrong. More work is required.

test/conftest.py
```diff
@@ -21,6 +21,8 @@ def pytest_collection_modifyitems(items) -> None:
         relative_path = item_path.relative_to(top_test_dir)
         for part in relative_path.parts:
             item.add_marker(part)
+            if part in ('unit', 'pandas'):
+                item.add_marker('skipolddriver')
```
sfc-gh-stakeda, line 25: skipping old test driver if `unit` or `pandas` test.

In the next chance of old driver version change, `unit` still needs to be excluded but `pandas` may need to be added.

@sfc-gh-mkeller , which on do you like? this or adding `skipolddriver` to `pandas` modules individually?

test/conftest.py
```diff
@@ -30,9 +32,16 @@ def filter_log() -> None:
...
     import logging
     import pathlib
+    from snowflake.connector.secret_detector import SecretDetector
+
+    if not isinstance(SecretDetector, logging.Formatter):
+        # Override it if SecretDetector is not an instance of logging.Formatter
+        class SecretDetector(logging.Formatter):
+            def format(self, record: logging.LogRecord) -> str:
+                return super().format(record)
+
```
sfc-gh-stakeda, line 44: The old version of `SecretDetector` was not an instance of logging.Formatter.

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -12,7 +12,12 @@
 
 import pytest
 
-from snowflake.connector.options import installed_pandas, pandas
+try:
+    from snowflake.connector.options import installed_pandas, pandas  # NOQA
+except ImportError:
+    installed_pandas = False
+    pandas = None
```
sfc-gh-stakeda, line 19: Although this test won't run for the old driver, pytest runs this part to determine if the test is runnable.

.github/workflows/build_test.yml
```diff
@@ -241,6 +241,51 @@ jobs:
...
+      - name: Install tox
+        run: python -m pip install tox
+      - name: Install Old Python Connector
+        run: python -m pip install snowflake-connector-python==${{ matrix.driver-version }}
+      - name: Adjust the dependency
+        run: python -m pip install azure-storage-blob==2.1.0 pandas pendulum
+      - name: Install pytest
+        run: python -m pip install pytest==6.0.2 pytest-cov pytest-rerunfailures pytest-timeout mock
+      - name: Run tests
+        run: python -m tox --skip-pkg-install -e "py${PYTHON_VERSION/\./}-olddriver-ci2"
```
sfc-gh-stakeda, line 283: `--skip-pkg-install` will skip installing the current version.

test/conftest.py
sfc-gh-mkeller: I'd generally like not adding them one-by-one, but in that case we'd have to keep a good lists of which tags are not run on the old drivers.
However, I'm open to add them manually if you think that being more explicit would be beneficial.

sfc-gh-stakeda: I'm ok with either case. It is more or less up to you which is good for your maintenance.

sfc-gh-stakeda:
@sfc-gh-mkeller, good for you to review the change.

sfc-gh-mkeller (approved):
🚢 🚢

PR #503, opened 2020-10-28
Update cryptography dependency pin

ChristineTChen:
The issue fixed in this PR  https://github.com/pyca/cryptography/pull/5507/ was released in cryptography 3.2.0.


sfc-gh-mkeller:
Thanks @ChristineTChen but this is an exact duplicate of #500 
Testing is almost done over there, closing this one!

PR #511, opened 2020-10-29
SNOW-213931 Send all Python Connector exceptions to in-band or out-of-band telemetry

sfc-gh-kwagner:


src/snowflake/connector/errors.py
```diff
@@ -66,10 +73,68 @@ def __unicode__(self):
...
+            telemetry_data = self.generate_telemetry_exception_data(msg)
+            if cursor is not None:
+                self.send_exception_telemetry(cursor.connection, telemetry_data)
+            elif connection is not None:
+                self.send_exception_telemetry(connection, telemetry_data)
+            else:
+                self.send_exception_telemetry(None, telemetry_data)
+        except Exception:
+            # Do nothing but log if sending telemetry fails
+            logger.warning("Sending exception telemetry failed")
```
sfc-gh-stakeda, line 132: let's not use the warning level. Failing to send the metric is not critical.

src/snowflake/connector/errors.py
```diff
@@ -66,10 +73,68 @@ def __unicode__(self):
...
+            # Send with in-band telemetry
+            telemetry_data[TelemetryField.KEY_TYPE] = TelemetryField.SQL_EXCEPTION
+            telemetry_data[TelemetryField.KEY_EXCEPTION] = self.__class__.__name__
+            ts = get_time_millis()
+            try:
+                connection._log_telemetry(TelemetryData(telemetry_data, ts))
+            except AttributeError:
+                logger.warning(
+                    "Cursor failed to log to telemetry.",
+                    exc_info=True)
```
sfc-gh-stakeda, line 111: same here. Let's not use the warning log level.

src/snowflake/connector/errors.py
sfc-gh-kwagner: Okay I'll change them both to info

sfc-gh-mkeller (changes requested):
I think in general we should reserve string type hints for our own types, so that no forward declaration would be necessary, but builtin types should not be put into string and I think self should not be type hinted

src/snowflake/connector/errors.py
```diff
@@ -66,10 +73,68 @@ def __unicode__(self):
...
+            # Send with in-band telemetry
+            telemetry_data[TelemetryField.KEY_TYPE] = TelemetryField.SQL_EXCEPTION
+            telemetry_data[TelemetryField.KEY_EXCEPTION] = self.__class__.__name__
+            ts = get_time_millis()
+            try:
+                connection._log_telemetry(TelemetryData(telemetry_data, ts))
+            except AttributeError:
+                logger.info(
+                    "Cursor failed to log to telemetry.",
+                    exc_info=True)
```
sfc-gh-stakeda, line 111: `debug` is better as this log message is not for the customers, i.e., no action can be taken or not much value for them.

src/snowflake/connector/errors.py
```diff
@@ -66,10 +73,68 @@ def __unicode__(self):
...
+        if self.sqlstate:
+            telemetry_data[TelemetryField.KEY_SQLSTATE] = self.sqlstate
+        if msg:
+            telemetry_data[TelemetryField.KEY_REASON] = SecretDetector.mask_secrets(msg)
+        if self.errno:
+            telemetry_data[TelemetryField.KEY_ERROR_NUMBER] = str(self.errno)
+        if self.msg:
+            telemetry_data[TelemetryField.KEY_ERROR_MESSAGE] = SecretDetector.mask_secrets(self.msg)
+
+        telemetry_data[TelemetryField.KEY_STACKTRACE] = SecretDetector.mask_secrets(self.__traceback__)
```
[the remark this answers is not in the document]
sfc-gh-kwagner: Thats how you do it in Python 2, but in Python 3 the exception has that __traceback__ property. This was introduced as part of PEP 3134 and its recommended to get the traceback information from that property. https://www.python.org/dev/peps/pep-3134/#traceback-attribute

src/snowflake/connector/telemetry_oob.py
```diff
@@ -154,7 +155,7 @@ def __init__(self):
         self.num_of_retry_to_trigger_telemetry = DEFAULT_NUM_OF_RETRY_TO_TRIGGER_TELEMETRY
         self.context = dict()
         self.connection_params = dict()
-        self.deployment = None
+        self.deployment = TelemetryServerDeployments.PROD
```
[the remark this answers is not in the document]
sfc-gh-kwagner: __instance does get assigned in the init function. This is necessary because we may be sending OOB telemetry before we have the connection parameters and call update_context on this class

sfc-gh-stakeda (approved):

sfc-gh-kwagner:
Here is an example of a raw Python traceback and what I condense it to:

`  File "/home/kwagner/snowflake-connector-python/test/integ/test_errors.py", line 37, in test_error_telemetry
    ctx.cursor().execute("SELECT * FROOOM TEST")
  File "/home/kwagner/snowflake-connector-python/src/snowflake/connector/cursor.py", line 614, in execute
    errvalue)
  File "/home/kwagner/snowflake-connector-python/src/snowflake/connector/errors.py", line 216, in errorhandler_wrapper
    cursor.errorhandler(connection, cursor, error_class, error_value)
  File "/home/kwagner/snowflake-connector-python/src/snowflake/connector/errors.py", line 178, in default_errorhandler
    cursor=cursor)`


`  File "snowflake/connector/cursor.py", line 614, in execute
  File "snowflake/connector/errors.py", line 216, in errorhandler_wrapper
  File "snowflake/connector/errors.py", line 178, in default_errorhandler`

src/snowflake/connector/errors.py
```diff
@@ -65,10 +77,85 @@ def __unicode__(self):
...
+        if self.errno:
+            telemetry_data[TelemetryField.KEY_ERROR_NUMBER] = str(self.errno)
+
+        telemetry_data[TelemetryField.KEY_STACKTRACE] = SecretDetector.mask_secrets(self.telemetry_traceback)
+
+        return telemetry_data
+
+    def send_exception_telemetry(self,
+                                 connection: Optional['SnowflakeConnection'],
+                                 telemetry_data: Dict[str, str]):
```
sfc-gh-mkeller:
```suggestion
                                 telemetry_data: Dict[str, str]) -> None:
```

src/snowflake/connector/errors.py
```diff
@@ -65,10 +77,85 @@ def __unicode__(self):
...
+                    exc_info=True)
+        else:
+            # Send with out-of-band telemetry
+            telemetry_oob = TelemetryService.get_instance()
+            telemetry_oob.log_general_exception(self.__class__.__name__, telemetry_data)
+
+    def exception_telemetry(self,
+                            msg: str,
+                            cursor: Optional['SnowflakeCursor'],
+                            connection: Optional['SnowflakeConnection']):
```
sfc-gh-mkeller:
```suggestion
                            connection: Optional['SnowflakeConnection']) -> None:
```

src/snowflake/connector/errors.py
```diff
@@ -65,10 +77,85 @@ def __unicode__(self):
...
+        except Exception:
+            # Do nothing but log if sending telemetry fails
+            logger.debug("Sending exception telemetry failed")
+
     @staticmethod
     def default_errorhandler(connection: 'SnowflakeConnection',
                              cursor: 'SnowflakeCursor',
-                             error_class,
+                             error_class: Exception,
                              error_value: Dict[str, str]):
```
sfc-gh-mkeller, line 167:
```suggestion
                             error_value: Dict[str, str]) -> None:
```

src/snowflake/connector/errors.py
```diff
@@ -5,25 +5,34 @@
...
 
 
 class Error(BASE_EXCEPTION_CLASS):
     """Base Snowflake exception class."""
 
-    def __init__(self, msg=None, errno=None, sqlstate=None, sfqid=None,
-                 done_format_msg=False):
+    def __init__(self, msg: Optional[str] = None, errno: Optional[int] = None, sqlstate: Optional[str] = None,
+                 sfqid: Optional[str] = None, done_format_msg: Optional[bool] = None,
+                 connection: Optional[SnowflakeConnection] = None, cursor: Optional[SnowflakeCursor] = None):
```
sfc-gh-mkeller, line 35:
```suggestion
                 connection: Optional['SnowflakeConnection'] = None, cursor: Optional['SnowflakeCursor'] = None):
```
It's not  a problem in this file but putting these in strings can save us from some circular imports. I'd like to be consistent. If we type hint one of our own classes unless it's in the same file let's put them in strings if that makes sense to you.

sfc-gh-mkeller (approved):
🚢 🚢

src/snowflake/connector/connection.py
```diff
@@ -207,6 +207,7 @@ def __init__(self, **kwargs):
         self.messages = []
         self._async_sfqids = set()
         self._done_async_sfqids = set()
+        self.telemetry_enabled = False
```
sfc-gh-hchaturvedi, line 210: So customers will have to turn it on manually right?

sfc-gh-kwagner: Yes. If they have telemetry enabled then this will send the exception information to Snowflake. If it is disabled then it won't get sent. However, we don't have flags for sending specific types of telemetry. Its either all telemetry info is sent or no telemetry info is sent

sfc-gh-hchaturvedi: Seems reasonable for now. If we start adding too much different types of telemetry it might be worth while to explore some form of layered approach in the design.

src/snowflake/connector/telemetry_oob.py
```diff
@@ -382,6 +383,27 @@ def log_http_request_error(self,
             # Do nothing on exception, just log
             logger.debug("Failed to log HTTP request error", exc_info=True)
 
+    def log_general_exception(self,
+                              event_name: str,
+                              telemetry_data: dict,
+                              tags: Optional[dict] = None,
+                              urgent: Optional[bool] = False):
+        """Sends any type of exception through OOB telemetry."""
```
sfc-gh-hchaturvedi, line 391: Assuming from the comment above """Send telemetry data by in-band telemetry if it is enabled, otherwise send through out-of-band telemetry."""
This will be send if in-band is disabled. Can you be a little more explicit about it? What does it mean for in-band telemetry to be enabled / disabled. Is there a separate connection parameter / backend parameter that flips the switch on inband telemetry. It might be helpful for someone who doesn't have context and looking at this for the first time.

sfc-gh-hchaturvedi (approved):
LGTM. Just some minor comments, else looks good. Good work resolving this tricky situation Kris!

src/snowflake/connector/telemetry_oob.py
sfc-gh-kwagner: Ah yes. So the intent is that if we hit an exception during connection setup we want telemetry on that exception which is where the OOB telemetry framework comes in. I just updated the logic so that we only send through OOB telemetry if the connection object is None. This way if the connection object exists and in-band telemetry is disabled we don't try to send it through OOB telemetry

sfc-gh-spandey (approved):

sfc-gh-nsfard (approved):
I can't comment on the logic but this looks safe from a security perspective

sfc-gh-sshankar:
@sfc-gh-mkeller feel free to merge this once you are ready.

PR #512, opened 2020-10-30
SNOW-200120 increase ssl coverage

sfc-gh-mkeller:
Deleting some unused code. Consequently removing a now unneeded dependency


src/snowflake/connector/ssl_wrap_socket.py
```diff
@@ -41,36 +34,6 @@
 """
 FEATURE_OCSP_RESPONSE_CACHE_FILE_NAME = None
 
-
```
sfc-gh-stakeda, line 44: Are these for python2? I forgot why I added these...

sfc-gh-mkeller: I'm not sure, let me dig around!

sfc-gh-mkeller: I can't find the reason of why they were added, they were part of the initial release.
But basically this whole commit came from noticing that `ssl_wrap_socket` function was unused and then deleting that exposed a lot of things that were only called from that function.

sfc-gh-stakeda: Looks like this removed the dependency of `ssl_wrap_socket` in this module.
https://github.com/snowflakedb/snowflake-connector-python/commit/6c84a997

sfc-gh-kwagner (approved):
Nice job!

sfc-gh-stakeda (approved):

PR #518, opened 2020-10-31
Update mock imports in tests 

viswan29:
Python 2 support has been disabled in master branch, `unittest.mock` can be used in place of `mock` library from pypi

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -243,7 +243,6 @@ setup(
             'pytest-timeout',
             'coverage',
             'pexpect',
-            'mock',
             'pytz',
             'pytzdata',
             'Cython',
--- a/test/integ/pandas/test_pandas_tools.py
+++ b/test/integ/pandas/test_pandas_tools.py
@@ -7,8 +7,8 @@
 import math
 from datetime import datetime
 from typing import Callable, Dict, Generator
+from unittest import mock
 
-import mock
 import pytest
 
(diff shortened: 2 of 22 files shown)


sfc-gh-mkeller:
Hi @viswan29 

I kept `mock` because I thought it would give us backports of newer features and bug fixes on all Python versions without new Python releases.

Do you not agree?

viswan29:
@sfc-gh-mkeller New feature or bug fixes in mock are hypothetical. It feels like an unnecessary dependency for a hypothetical need

viswan29:
@sfc-gh-mkeller Hey I have also observed few changes with respect to using `super` function and `future` import. Is that some thing i can work ?

sfc-gh-mkeller:
> @sfc-gh-mkeller New feature or bug fixes in mock are hypothetical. It feels like an unnecessary dependency for a hypothetical need

See https://mock.readthedocs.io/en/latest/changelog.html#b1
There's a ton of bug fixes in this release from earlier this year. I'd rather have an extra dependency than to debug why some mocked object is not working on our lowest supported Python version (3.5.1).


> @sfc-gh-mkeller Hey I have also observed few changes with respect to using `super` function and `future` import. Is that some thing i can work ?

Yes, of course. I'm sure we haven't touched those since we dropped 2.7 support 😅
Thank you!

PR #525, opened 2020-11-06
SNOW-216517 improve cursor coverage

sfc-gh-mkeller:
Raises `cursor.py` coverage to ~100%. Covering the leftover part would only be questionably useful.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -279,6 +279,7 @@ jobs:
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
           cloud_provider: ${{ matrix.cloud-provider }}
+          PYTEST_ADDOPTS: --color=yes
         shell: bash
 
   test-fips:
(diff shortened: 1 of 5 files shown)


.github/workflows/build_test.yml
```diff
@@ -279,6 +279,7 @@ jobs:
         env:
           PYTHON_VERSION: ${{ matrix.python-version }}
           cloud_provider: ${{ matrix.cloud-provider }}
+          PYTEST_ADDOPTS: --color=yes
```
sfc-gh-kwagner, line 282: What is this for?

sfc-gh-kwagner (approved):

.github/workflows/build_test.yml
sfc-gh-mkeller: It allows py.test to use colored output. Previously I set this in our tox file, but that also made Jenkins try to use colored output. Jenkins cannot deal with that, so I moved the setting into GH actions. Now we can keep colored output here and not have it on Jenkins.

PR #531, opened 2020-11-12
Allow 1.16.X series of boto3 dependencies

jperkelens:
We have dependency conflicts across several of our libraries due to this pin.


jperkelens:
I see this is a duplicate of https://github.com/snowflakedb/snowflake-connector-python/pull/508 so im going to close. This and #529 have been preventing us from a successful build for two days, would appreciate your attention to them. Thanks.

PR #535, opened 2020-11-13
SNOW-175900: Username pwd mfa token cache

sfc-gh-wshangguan:
NOTE!
This pr won't be merged until a thorough test with backend changes is done.

The whole context:

> For users who are using username-pwd plus MFA authentication, we plan to add a temporary mfa token cache. This mfa token cache helps prevent multiple MFA authentication with a short perioud of time, which is similar to previous ID_TOKEN. In other words, if user's connection request is sent with a correct pwd (yes, pwd is still needed) and a valid mfa token, then users don't need to do mfa authentication again.

What this pr does:

1. Add a new authenticator USERNAME_PASSWORD_MFA. CLIENT_REQUEST_MFA_TOKEN parameter is added.
2. Support both ID_TOKEN and MFATOKEN in Local Secure Storage.
3. Change the way we name local cache of ID_TOKEN.
4. For Linux local file cache, support partially deletion of temporary credential file.

`test/unit/test_mfa_cache.py` can be a good start for learning use cases


src/snowflake/connector/auth.py
```diff
@@ -91,8 +95,7 @@ def base_auth_data(user, account, application,
                        internal_application_name,
                        internal_application_version,
                        ocsp_mode, login_timeout,
-                       network_timeout=None,
-                       store_temp_cred=None):
```
sfc-gh-wshangguan, line 95: `store_temp_cred` is redundant here for id_token.

src/snowflake/connector/auth.py
```diff
@@ -336,19 +331,20 @@ def post_request_wrapper(self, url, headers, body):
                          'NULL')
             logger.debug('id_token = %s',
                          '******' if ret['data'].get(
-                             'id_token') is not None else
```
sfc-gh-wshangguan, line 339: An old typo.

src/snowflake/connector/auth.py
```diff
@@ -361,80 +357,100 @@ def post_request_wrapper(self, url, headers, body):
 
         return session_parameters
 
-    def read_temporary_credential(self, host, account, user, session_parameters):
+    def _read_temporary_credential(self, host, user, cred_type):
```
sfc-gh-wshangguan, line 360: Support `cred_type` now.

src/snowflake/connector/auth.py
```diff
@@ -361,80 +357,100 @@ def post_request_wrapper(self, url, headers, body):
...
-    if not id_token:
-        logger.debug("no ID token is given when try to store temporary credential")
-        return
-    if IS_MACOS or IS_WINDOWS:
-        if not installed_keyring:
-            logger.debug("Dependency 'keyring' is not installed, cannot cache id token. You might experience "
-                         "multiple authentication pop ups while using ExternalBrowser Authenticator. To avoid "
-                         "this please install keyring module using the following command : pip install "
-                         "snowflake-connector-python[secure-local-storage]")
+    def _write_temporary_credential(self, host, user, cred_type, cred):
```
sfc-gh-wshangguan, line 388: Support cred_type.

src/snowflake/connector/auth.py
```diff
@@ -490,16 +506,32 @@ def unlock_temporary_credential_file():
...
-            logger.debug("Failed to delete credential in the keyring: err=[%s]",
-                         ex)
-    elif IS_LINUX and store_temporary_credential:
-        delete_temporary_credential_file()
+            logger.debug("Failed to delete credential in the keyring: err=[%s]", ex)
+    elif IS_LINUX:
+        temporary_credential_file_delete_password(host, user, cred_type)
+
+
+def temporary_credential_file_delete_password(host, user, cred_type):
```
sfc-gh-wshangguan, line 519: Add linux local file cache partial deletion function.

src/snowflake/connector/auth.py
```diff
@@ -516,6 +548,9 @@ def delete_temporary_credential_file():
...
 
-def convert_target(host, user):
-    return "{host}:{user}:{driver}".format(
-            host=host.upper(), user=user.upper(), driver=KEYRING_DRIVER_NAME)
+def build_temporary_credential_name(host, user, cred_type):
+    return "{host}:{user}:{driver}:{cred}".format(
+            host=host.upper(),
+            user=user.upper(),
+            driver=KEYRING_DRIVER_NAME,
+            cred=cred_type)
```
sfc-gh-wshangguan, line 556: New construction of token name.

src/snowflake/connector/connection.py
```diff
@@ -607,8 +613,13 @@ def __open_connection(self):
                 PARAMETER_CLIENT_STORE_TEMPORARY_CREDENTIAL] = \
                     self._client_store_temporary_credential if IS_LINUX else True
 
+        if self._authenticator == USR_PWD_MFA_AUTHENTICATOR:
+            self._session_parameters[
+                PARAMETER_CLIENT_REQUEST_MFA_TOKEN] = \
+                    self._client_request_mfa_token if IS_LINUX else True
```
sfc-gh-wshangguan, line 619: This feature is only open for Mac & Windows. Linux support is used for internal testing. And may become public in the future.

sfc-gh-stakeda (approved):
Ensure the code coverage is good enough.

src/snowflake/connector/auth.py
```diff
@@ -361,80 +357,100 @@ def post_request_wrapper(self, url, headers, body):
...
+            if not installed_keyring:
+                logger.debug("Dependency 'keyring' is not installed, cannot cache id token. You might experience "
+                             "multiple authentication pop ups while using ExternalBrowser Authenticator. To avoid "
+                             "this please install keyring module using the following command : pip install "
+                             "snowflake-connector-python[secure-local-storage]")
+                return
+            try:
+                keyring.set_password(build_temporary_credential_name(host, user, cred_type), user.upper(), cred)
+            except keyring.errors.KeyringError as ke:
+                logger.debug("Could not store id_token to keyring, %s", str(ke))
```
sfc-gh-mkeller, line 402: Should these not be error level logs?

src/snowflake/connector/auth.py
```diff
@@ -361,80 +357,100 @@ def post_request_wrapper(self, url, headers, body):
...
+    global TEMPORARY_CREDENTIAL
+    global TEMPORARY_CREDENTIAL_FILE
+    for _ in range(10):
+        if lock_temporary_credential_file():
+            break
+        time.sleep(1)
     else:
-        logger.debug("connection parameter client_store_temporary_credential not set or OS not support")
+        logger.debug("The lock file still persists. Will ignore and "
+                     "write the temporary credential file: %s",
```
sfc-gh-mkeller, line 428:
```suggestion
        logger.debug("The lock file still persists after the maximum wait time. "
                     "Will ignore it and write temporary credential file: %s",
```

src/snowflake/connector/auth.py
```diff
@@ -361,80 +357,100 @@ def post_request_wrapper(self, url, headers, body):
...
+                     TEMPORARY_CREDENTIAL_FILE)
+    try:
+        with codecs.open(TEMPORARY_CREDENTIAL_FILE, 'w',
+                         encoding='utf-8', errors='ignore') as f:
```
sfc-gh-mkeller, line 432:
```suggestion
        with open(TEMPORARY_CREDENTIAL_FILE, 'w',
                  encoding='utf-8', errors='ignore') as f:
```
https://stackoverflow.com/a/22288895

test/integ/sso/test_unit_mfa_cache.py
```diff
@@ -0,0 +1,157 @@
...
+
+MFA_TOKEN = "MFATOKEN"
+
+
+# Although this is an unit test, we put it under test/integ/sso, since it needs keyring package installed under ci environment
+@pytest.mark.skipif(delete_temporary_credential is None or IS_MACOS is None,
+                    reason="delete_temporary_credential or IS_MACOS is not available.")
+@patch('snowflake.connector.network.SnowflakeRestful._post_request')
+def test_mfa_cache(
+        mockSnowflakeRestfulPostRequest):
```
sfc-gh-mkeller, line 32:
```suggestion
def test_mfa_cache(mockSnowflakeRestfulPostRequest):
```

test/unit/test_linux_local_file_cache.py
```diff
@@ -0,0 +1,67 @@
...
+CRED_TYPE_1 = "MFA_TOKEN"
+
+
+def get_credential(sys, user):
+    return auth.TEMPORARY_CREDENTIAL.get(sys.upper(), {}).get(user.upper())
+
+
+def test_basic_store():
+    if not IS_LINUX:
+        return
```
sfc-gh-mkeller, line 29: Please use a skip here instead, we can attach a reason to that, so that'd be a little more verbose than this.

sfc-gh-mkeller (changes requested):

src/snowflake/connector/auth_usrpwdmfa.py
```diff
@@ -0,0 +1,48 @@
...
+        return "*********"
+
+    def __init__(self, password, mfa_token=None):
+        """Initializes and instance with a password and a mfa token."""
+        self._password = password
+        self.mfa_token = mfa_token
+
+    @property
+    def mfa_token(self):
+        return self._mfa_token
```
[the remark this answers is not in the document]
sfc-gh-wshangguan: Nice catch! I planed to add some logic here, but went with another one instead and forgot to remove the property. Will delete it!

test/unit/test_linux_local_file_cache.py
sfc-gh-wshangguan: Will do!

src/snowflake/connector/auth.py
```diff
@@ -361,80 +360,100 @@ def post_request_wrapper(self, url, headers, body):
...
+    for _ in range(10):
+        if lock_temporary_credential_file():
+            break
+        time.sleep(1)
     else:
-        logger.debug("connection parameter client_store_temporary_credential not set or OS not support")
+        logger.debug("The lock file still persists after the maximum wait time."
+                     "Will ignore it and write temporary credential file: %s",
+                     TEMPORARY_CREDENTIAL_FILE)
+    try:
```
sfc-gh-mkeller, line 433: This isn't super important, but the try-catch-finally block could go before the break statement.

sfc-gh-mkeller (changes requested):

src/snowflake/connector/network.py
```diff
@@ -326,9 +339,8 @@ def _renew_session(self):
 
     def _token_request(self, request_type):
         logger.debug(
-            'updating session. master_token: %s, id_token: %s',
-            '****' if self.master_token else None,
-            '****' if self.id_token else None)
+            'updating session. master_token: %s',
+            '****' if self.master_token else None)
```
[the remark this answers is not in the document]
sfc-gh-wshangguan: Got it

test/integ/sso/test_unit_mfa_cache.py
```diff
@@ -0,0 +1,197 @@
...
+            # a failed login will be forced by mocking for this connection
+            con = snowflake.connector.connect(
+                account=account,
+                user=user,
+                password=pwd,
+                host=host,
+                authenticator=authenticator,
+                client_request_mfa_token=True,
+            )
+        except DatabaseError:
```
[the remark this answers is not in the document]
sfc-gh-wshangguan: Good to know that!

test/integ/sso/test_unit_mfa_cache.py
```diff
@@ -0,0 +1,197 @@
...
+import pytest
+from mock import Mock, patch
+
+import snowflake.connector
+from snowflake.connector.errors import DatabaseError
+
+try:
+    from snowflake.connector.compat import IS_MACOS  # NOQA
+except ImportError:
+    IS_MACOS = None
```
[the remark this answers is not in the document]
sfc-gh-wshangguan: Okay, I guess we need the fallback for old driver test, otherwise it would fail. And I found similar fallback in test_unit_sso_connection.py, which I think is for the same goal. Maybe we can use another way to avoid explicitly fallback in every file later.

src/snowflake/connector/connection.py
```diff
@@ -607,8 +613,13 @@ def __open_connection(self):
                 PARAMETER_CLIENT_STORE_TEMPORARY_CREDENTIAL] = \
                     self._client_store_temporary_credential if IS_LINUX else True
 
+        if self._authenticator == USR_PWD_MFA_AUTHENTICATOR:
+            self._session_parameters[
```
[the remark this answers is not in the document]
sfc-gh-wshangguan: Logically no difference, but only one concern. In that way, we have to do additional judge of whether self._session_parameters is `None`, which is done later on line 585.

test/unit/test_mfa_no_cache.py
```diff
@@ -0,0 +1,105 @@
...
+import pytest
+from mock import patch
+
+import snowflake.connector
+
+try:
+    from snowflake.connector.compat import IS_LINUX  # NOQA
+    from snowflake.connector.options import installed_keyring
+except ImportError:
+    IS_LINUX = True
```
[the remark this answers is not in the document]
sfc-gh-wshangguan: Okay, I just remembered what's happening here. The fallback is here for the same reason as `IS_MACOS` (for old driver test). However, since `IS_LINUX` and `IS_MACOS` are not added into repo at the same time, i.e. `IS_MACOS` is relatively new, `IS_LINUX` may not break old driver test as `IS_MACOS`. I was doing the test in one of my test file and it turned out the `IS_LINUX` is old enough. Let me remove the fallback for `IS_LINUX` while keep `IS_MACOS` fallback.

src/snowflake/connector/auth.py
sfc-gh-wshangguan: I guess we'd better not do it. Because the original behavior is like after 10 times of tries, even if we still can't grab the lock for temporary_credential_file, we still do the operation anyway. And if we move try-catch-finally block right before break statement, then it won't work as usual under rare situations.

test/integ/sso/test_unit_mfa_cache.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # first connection, no mfa token cache
+        con = snowflake.connector.connect(
+            account=account,
+            user=user,
+            password=pwd,
+            host=host,
+            authenticator=authenticator,
+            client_request_mfa_token=True,
+        )
```
sfc-gh-mkeller, line 135: I think you could have achieved the same thing with a little less typing. But I'm only pointing this out since you said you wanted to learn a more Pythonic coding style.
```suggestion
        conn_cfg = {
            'account': 'testaccount'
            'user': 'testuser'
            'password': 'testpwd'
            'authenticator': 'username_password_mfa'
            'host': 'testaccount.snowflakecomputing.com'
        }

        delete_temporary_credential(host=host, user=user, cred_type=MFA_TOKEN)

        # first connection, no mfa token cache
        con = snowflake.connector.connect(
            client_request_mfa_token=True,
            **conn_cfg
        )
```

sfc-gh-mkeller (approved):
🚢 🚢

src/snowflake/connector/connection.py
```diff
@@ -620,6 +631,9 @@ def __preprocess_auth_instance(self, auth_instance):
         if type(auth_instance) is AuthByWebBrowser:
```
sfc-gh-hchaturvedi, line 631: Can you rename this to AuthByExtBrowser please. This has a potential to cause confusion.

src/snowflake/connector/auth_usrpwdmfa.py
```diff
@@ -0,0 +1,43 @@
...
+    def authenticate(self, authenticator, service_name, account, user, password):
+        """NOOP."""
+        pass
+
+    def update_body(self, body):
+        """Sets the password and mfa_token if available.
+
+        Don't set body['data']['AUTHENTICATOR'], since this is still snowflake default authenticator.
+        """
+        if self._password:
```
sfc-gh-hchaturvedi, line 40: Do you think the password might be missing at this stage? I think we should throw an exception here if the password is missing. By this stage, the password should be set already.

src/snowflake/connector/connection.py
sfc-gh-wshangguan: Created a ticket https://snowflakecomputing.atlassian.net/browse/SNOW-232776 to do it later!

test/integ/sso/test_unit_mfa_cache.py
sfc-gh-wshangguan: Appreciate the suggestion!

src/snowflake/connector/auth_usrpwdmfa.py
sfc-gh-wshangguan: Actually we have a prechecking, before this update_body function get called. And this code is only to be align with the code in auth_default.py. So basically, if no password is provided with username_password_mfa authenticator, then an exception will be thrown before this part of code is reached. Does this sound acceptable?

sfc-gh-hchaturvedi: update_body function call is made on the auth_instance only. There's only one auth_instance that exists at once, so if you have chosen the username_password_mfa authenticator you will only initialize the mfa auth instance and none of the checks performed in the default auth instance would apply. When you say there's already a check, do you mean check in the config parser? Can you point out where's the check you're referring to?

sfc-gh-wshangguan: Oh yeah, I referred to the check in config parser https://github.com/snowflakedb/snowflake-connector-python/blob/28682e91b15477b22ab8941a8ab4723c37fddd53/src/snowflake/connector/connection.py#L734.

sfc-gh-hchaturvedi: If that's the case, why's there an if clause, you can just assign it right? If you think the password will always be set then there's not need of the if condition, if you want to account for the case that it might not be set then you should raise an exception if it is not set.

sfc-gh-wshangguan: Ahh okay, because I found https://github.com/snowflakedb/snowflake-connector-python/blob/28682e91b15477b22ab8941a8ab4723c37fddd53/src/snowflake/connector/auth_default.py#L26-L29 for normal snowflake username password login, which is almost the same as username password mfa login to some extent. And if this always works, I guess not change that is better for a huge code base, since there could be some logic we are not aware of that require it as is.

sfc-gh-hchaturvedi: Change has to start somewhere right. Please add the exception clause if you want to be safe.

sfc-gh-wshangguan: Okay, let me add that

sfc-gh-hchaturvedi (approved):
LGTM!

sfc-gh-wshangguan:
I will merge this pr by the end of today, please let me know if there is any concern!

PR #542, opened 2020-11-19
SNOW-206907 GCS Downscope Token

sfc-gh-kphan:
Description
- Implement update to work with downscoped accesstoken instead of using
presigned url. To use this new feature, SNOW-206907 must be reviewed and merged
on the server side first.
- The upload and download proccess is unchanged, XML api URL is built and used
with the access token set in header when access token is available.

Testing
- All put/get test for gcs test with parameter GCS_USE_DOWNSCOPED_CREDENTIAL turn
both on and off

Most important change is in `gs_utils.py`, change to `remote_storage_util.py` is minor

Concern:
- The implementation for overwrite here is different from jdbc, current python connector use get metadata to check for file existence, but PUT request should not have GET permission (current implementation for AWS and AZURE both allow GET request on WRITE only credential). Should this be changed to use LIST to check for file existence?

### DIFF
--- a/src/snowflake/connector/gcs_util.py
+++ b/src/snowflake/connector/gcs_util.py
@@ -6,7 +6,10 @@
 
 import json
 import os
+import xml.etree.ElementTree as ElementTree
+from collections import namedtuple
 from logging import getLogger
+from urllib.parse import quote
 
 import requests
 
@@ -21,6 +24,15 @@ GCS_FILE_HEADER_DIGEST = 'gcs-file-header-digest'
 GCS_FILE_HEADER_CONTENT_LENGTH = 'gcs-file-header-content-length'
 GCS_FILE_HEADER_ENCRYPTION_METADATA = 'gcs-file-header-encryption-metadata'
 CONTENT_CHUNK_SIZE = 10 * 1024
+
+"""
+Gcs Location: Gcs bucket name + path
+"""
+GcsLocation = namedtuple(
+    "GcsLocation", [
+        "bucket_name",  # Gcs bucket name
+        "path"  # Gcs path name
+    ])
 
 
 class SnowflakeGCSUtil:
@@ -44,9 +56,8 @@ class SnowflakeGCSUtil:
 
         if security_token:
             logger.debug("len(GCS_ACCESS_TOKEN): %s", len(security_token))
-            logger.debug("GCS operations with an access token are currently "
-                         "unsupported")
-            client = None
+            logger.debug("Access token is saved as client for renew")
+            client = security_token
 
         else:
             logger.debug("No access token received from GS, constructing "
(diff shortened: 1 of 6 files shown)


sfc-gh-ppaulus:
We should ensure the Python driver can continue to work against either presigned URLs or tokens coming from GS. I'm not very familiar with the Python driver, but please test that they both continue to work.

src/snowflake/connector/gcs_util.py
```diff
@@ -286,22 +308,61 @@ def get_file_header(meta, filename):
...
+            path = stage_location[stage_location.index('/') + 1:]
+            if path and not path.endswith('/'):
+                path += '/'
+
+        return GcsLocation(
+            bucket_name=container_name,
+            path=path)
+
+    @staticmethod
+    def generate_file_url(stage_location, filename):
```
sfc-gh-ppaulus, line 348: Why do we generate the file URL? Are we not getting the file URLs from GS?

sfc-gh-kphan: GS only give us the access token in cased of downscope, so we have to build the url.

sfc-gh-kphan:
> We should ensure the Python driver can continue to work against either presigned URLs or tokens coming from GS. I'm not very familiar with the Python driver, but please test that they both continue to work.

Already covered in test.

src/snowflake/connector/gcs_util.py
```diff
@@ -44,9 +56,8 @@ def create_client(stage_info,
 
         if security_token:
             logger.debug("len(GCS_ACCESS_TOKEN): %s", len(security_token))
-            logger.debug("GCS operations with an access token are currently "
-                         "unsupported")
-            client = None
+            logger.debug("Access token is saved as client for renew")
+            client = security_token
```
sfc-gh-ssrinivasa, line 60: I can understand why, but I'm wondering if we have other alternatives, like directly setting the token on meta['access_token'] or something like that..
@sfc-gh-mkeller any thoughts here?

sfc-gh-ssrinivasa (approved):
LGTM, thanks Khang! 
I'd prefer @sfc-gh-mkeller to make the final call.

sfc-gh-ppaulus:
It looks like code coverage fell with this change. @sfc-gh-kphan can you make sure we're not decreasing code coverage?

sfc-gh-kphan:
> It looks like code coverage fell with this change. @sfc-gh-kphan can you make sure we're not decreasing code coverage?

@sfc-gh-ppaulus The test for gcs is failing because the parameter is not in LIVE yet, coverage should be better then. There only a part I think of that need some code coverage, I can work on thatt.

test/integ/test_put_get_with_gcp_account.py
```diff
@@ -40,6 +42,12 @@ def test_put_get_with_gcp(tmpdir, conn_cnx, db_parameters):
 
     with conn_cnx() as cnx:
         with cnx.cursor() as csr:
+            try:
+                csr.execute(f'ALTER SESSION SET GCS_USE_DOWNSCOPED_CREDENTIAL = {enable_gcs_downscoped}')
+            except ProgrammingError:
+                # skip test if parameter is not available (server not updated)
+                if enable_gcs_downscoped:
+                    return
```
sfc-gh-mkeller, line 50: Don't swallow this error, please. Remove this and mark the test as `xfail`.

test/integ/test_put_get_with_gcp_account.py
```diff
@@ -92,6 +107,10 @@ def run(csr, sql):
             ratio number(6,2))
             """)
             try:
+                statement = "put file://{files} @%{name}"
+                if enable_gcs_downscoped:
+                    statement += " overwrite = true"
+
                 all_recs = run(csr, "put file://{files} @%{name}")
```
sfc-gh-mkeller, line 114:
```suggestion
                all_recs = run(csr, statement)
```

test/integ/test_put_get_with_gcp_account.py
```diff
@@ -322,3 +361,133 @@ def test_auto_compress_off_gcp(tmpdir, conn_cnx, db_parameters):
...
+
+                with mock.patch('snowflake.connector.cursor.SnowflakeFileTransferAgent',
+                                side_effect=mocked_file_agent):
+                    with mock.patch('requests.put', side_effect=mocked_put):
+                        with mock.patch('requests.head', side_effect=mocked_head):
+                            csr.execute("put file://{} @%{} auto_compress=true parallel=30".format(fname, table_name))
+                    if error_code == 401:
+                        assert mocked_file_agent.agent.renew_expired_client.call_count == 2
+                assert csr.fetchone()[6] == 'UPLOADED'
+                csr.execute("copy into {}".format(table_name))
```
sfc-gh-mkeller, line 424: You could use `PURGE=True` here instead of RMing the file manually in the next execute

test/integ/test_put_get_with_gcp_account.py
```diff
@@ -322,3 +361,133 @@ def test_auto_compress_off_gcp(tmpdir, conn_cnx, db_parameters):
...
+                except ProgrammingError:
+                    # skip test if parameter is not available (server not updated)
+                    return
+                with mock.patch.object(cur, '_init_result_and_meta', wraps=cur._init_result_and_meta) as mock_result:
+                    cur.execute("PUT file://{} @~/test_put_overwrite".format(test_data))
+                    assert mock_result.call_args[0][0]['rowset'][0][-2] == 'UPLOADED'
+                with mock.patch.object(cur, '_init_result_and_meta', wraps=cur._init_result_and_meta) as mock_result:
+                    cur.execute("PUT file://{} @~/test_put_overwrite".format(test_data))
+                    assert mock_result.call_args[0][0]['rowset'][0][-2] == 'SKIPPED'
+                with mock.patch.object(cur, '_init_result_and_meta', wraps=cur._init_result_and_meta) as mock_result:
```
sfc-gh-mkeller, line 485: Why do you mock these? You could just actually get the results

sfc-gh-mkeller (changes requested):
Please remove all of the return statements from the tests, and mark them `xfail` if they need to be, this way we will forget about updating these tests once the feature gets released.

src/snowflake/connector/gcs_util.py
```diff
@@ -286,22 +307,94 @@ def get_file_header(meta, filename):
...
+        # split stage location as bucket name and path
+        if '/' in stage_location:
+            container_name = stage_location[0:stage_location.index('/')]
+            path = stage_location[stage_location.index('/') + 1:]
+            if path and not path.endswith('/'):
+                path += '/'
+
+        return GcsLocation(
+            bucket_name=container_name,
+            path=path)
```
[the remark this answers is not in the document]
sfc-gh-mkeller: This is just formatting

src/snowflake/connector/gcs_util.py
```diff
@@ -73,134 +83,65 @@ def upload_file(data_file, meta, encryption_metadata, max_concurrency):
...
+        upload_url = meta.get('presigned_url', None)
+        access_token = None
 
-        Args:
-            meta: File meta object (contains credentials and remote location).
-            full_dst_file_name: Local path of the file to download to.
-            max_concurrency: Not applicable to GCS.
+        if not upload_url:
+            upload_url = SnowflakeGCSUtil.generate_file_url(meta['stage_info']['location'],
+                                                            meta['dst_file_name'].lstrip('/'))
```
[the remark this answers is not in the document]
sfc-gh-kphan: I believe everything in meta is parsed from GS.

test/integ/test_put_get_with_gcp_account.py
sfc-gh-kphan: Updated, I copied this from s3, did not look at it much

src/snowflake/connector/gcs_util.py
```diff
@@ -286,22 +307,94 @@ def get_file_header(meta, filename):
...
+        stage_location = os.path.expanduser(stage_location)
+        container_name = stage_location
+        path = ''
+
+        # split stage location as bucket name and path
+        if '/' in stage_location:
+            container_name = stage_location[0:stage_location.index('/')]
+            path = stage_location[stage_location.index('/') + 1:]
+            if path and not path.endswith('/'):
+                path += '/'
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Then what is `os.path.expanduser` for?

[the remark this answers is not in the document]
sfc-gh-kphan: I copy this from S3/Azure which have this in it. Didn't check much. Maybe I will remove it both for azure and gcs.

test/integ/test_put_get_with_gcp_account.py
sfc-gh-kphan: Updated

test/unit/test_gcs_util.py
```diff
@@ -23,16 +23,18 @@
     SnowflakeGCSUtil = None
 
 
+@pytest.mark.xfail(reason='Newer version support access token')
```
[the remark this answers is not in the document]
sfc-gh-kphan: Added version

test/unit/test_gcs_util.py
```diff
@@ -23,16 +23,18 @@
...
 def test_create_client(caplog):
     """Creates a GCSUtil with an access token."""
     client = SnowflakeGCSUtil.create_client({'creds': {'GCS_ACCESS_TOKEN': 'fake_token'}})
-    assert client is None
+    assert client is not None
     assert all([log in caplog.record_tuples for log in [
```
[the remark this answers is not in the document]
sfc-gh-kphan: I changed this test so it should be fine now.

src/snowflake/connector/gcs_util.py
sfc-gh-stakeda: I have no preference. I'll let @sfc-gh-mkeller choose the style.

src/snowflake/connector/gcs_util.py
```diff
@@ -6,10 +6,14 @@
 
 import json
 import os
+from collections import namedtuple
 from logging import getLogger
+from typing import Any, Dict
 
 import requests
 
+from snowflake.connector.compat import quote
```
sfc-gh-mkeller, line 15: I know I suggested this, but I was mistaken import this like: `from .compat import quote` but then run `fix_lint` locally first.
I'm sorry for giving you a wrong suggestion previously!

src/snowflake/connector/gcs_util.py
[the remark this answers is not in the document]
sfc-gh-mkeller: Let's not touch AWS and Azure for now, but it seems to be safe to remove it from here!

test/integ/test_put_get_with_gcp_account.py
```diff
@@ -308,12 +343,20 @@ def mocked_file_agent(*args, **kwargs):
     assert original_contents == contents, 'Output is different from the original file'
 
 
-def test_auto_compress_off_gcp(tmpdir, conn_cnx, db_parameters):
+@pytest.mark.parametrize('enable_gcs_downscoped', [False, pytest.param(True, marks=pytest.mark.xfail(
+    reason="Server need to update with merged change. Expected release version: 4.41.0 "))])
+def test_auto_compress_off_gcp(tmpdir, conn_cnx, db_parameters, enable_gcs_downscoped):
     """[gcp] Puts and Gets a small text using gcp with no auto compression."""
```
sfc-gh-mkeller, line 349: I just realized that this will not allow Jenkins tests to test the new behavior even though they have access to `master` code. We can work around this and mark the tests expected to fail only when running against a PROD deployment by using `is_public_test` fixture. Something like this:
```suggestion
@pytest.mark.parametrize('enable_gcs_downscoped', [True, False])
def test_auto_compress_off_gcp(tmpdir, conn_cnx, db_parameters, is_public_test, enable_gcs_downscoped):
    """[gcp] Puts and Gets a small text using gcp with no auto compression."""
    if enable_gcs_downscoped and is_public_test:
        pytest.xfail("Server need to update with merged change. Expected release version: 4.41.0 ")
```

sfc-gh-mkeller (changes requested):
This is really good! I made a couple mistakes in reviewing your changes before, sorry!
I also dusted off my Windows vm and verified the path changes. 👍 
On top of my 3 comments please let's remove all mocking from integration tests.

sfc-gh-mkeller (approved):
🚢 🚢

PR #543, opened 2020-11-21
SNOW-226899 removed some logging messages

sfc-gh-mkeller:
This PR standardizes how logs are collected, INFO and above in py.test outputs and DEBUG and above into `snowflake_ssm_rt.log` when running tests.


sfc-gh-kwagner (approved):

sfc-gh-wshangguan (approved):
LGTM!

Issue #544, opened 2020-11-21
Support converting VARIANT to python object 


roi-meir:
## What is the current behavior?

When querying a variant (or array and object) column the results are returned as dumped json string. 
In the past before the Arrow format support we used the `converter_class` parameter to pass our own subclass of `SnowflakeConverter` that calls `json.loads` on the returned string. 
But when using the Arrow format this parameter has no affect

## What is the desired behavior?

To be able to control if the results should be a string or a python object (list or dict)

## How would this improve `snowflake-connector-python`?
Either have a way to pass a costume converter class or have another parameter that call `json.loads`

I will be happy to open a pull request that adds the parameter `convert_arrow_variant_to_object` that if enabled, call `json.loads` on variant results.

shahargl:
idk if it's relevant for you because it's more of a sqlalchemy thing, but we (@anecdotes.ai) managed to work with `DELETE/UPDATE/INSERT/SELECT` to VARIANT type columns using sqlalchemy's events mechanism (specifically `before_cursor_execute`).

jonashaag:
Any news on this? As a workaround we currently have a hardcoded list of columns to `json.parse` in the query "postprocessing".

salexln:
any update on this issue? we are facing the same problems and currently cannot use Arrow because of this

sfc-gh-mkeller:
Server-side binding for VARIANT and ARRAY is actively in the works. We didn't want to add support for retrieving them until binding them wasn't supported.

srausser:
any updates/eta's on when we can expect converting VARIANT/ARRAY to python object?
can't use snowflake as my [feast](feast.dev) offline store if the type conversion doesn't happen

YoniMelki:
any update on this issue? We are facing the same problems and currently cannot use Arrow as well

sfc-gh-achandrasekaran:
We will look into prioritizing this work in Q2.

KantorSerhiy:
Hi guys!
any update on this issue?

vanducng:
hello team, any updates?

sfc-gh-dszmolka:
hi all - to my best knowledge server-side thankfully supports `VARIANT` and `ARRAY` now, so next step would be the client side ;) at this moment I don't have any timeline to share for this planned enhancement, but will keep this thread posted with every new information . Thank you so much all for bearing with us !

timgraham:
[...]

PR #545, opened 2020-11-21
SNOW-228474: Support converting VARIANT to python object

roi-meir:
What this pr does:

Adds the convert_arrow_variant_to_object parameter to SnowflakeConnection. 
If this parameter value is `True` columns of the types VARIANT, ARRAY and OBJECT are retuned as python object and not string. 

To support this feature I added
- The `convert_arrow_variant_to_object` parameter and property. The default value of this parameter is False to preserve the old behavior.
- The method `VARIANT_to_python` in `ArrowConverterContext` that uses `json.loads` if the flag is True.
- A cpp `VariantConverter` class that call the method `VARIANT_to_python`.
- Some test to tests the new functionality.

Why this change is necessary: 
I wrote about it in #544 :
> When querying a variant (or array and object) column the results are returned as dumped json string.
> In the past before the Arrow format support one could've used the converter_class parameter to pass its own subclass of `SnowflakeConverter` that calls `json.loads` on the returned string.
> But when using the Arrow format this parameter has no affect.

I hope my changes make sense in the project and I would be happy to fix any comments you have.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,6 +111,7 @@ if _ABLE_TO_COMPILE_EXTENSIONS:
                                 os.path.join(ARROW_ITERATOR_SRC_DIR, 'FloatConverter.cpp'),
                                 os.path.join(ARROW_ITERATOR_SRC_DIR, 'IntConverter.cpp'),
                                 os.path.join(ARROW_ITERATOR_SRC_DIR, 'StringConverter.cpp'),
+                                os.path.join(ARROW_ITERATOR_SRC_DIR, 'VariantConverter.cpp'),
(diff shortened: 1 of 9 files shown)


roi-meir:
I see that some checks are failing but it doesn't seem to be related to my changes (looks like a missing environment variable). 
Should I change anything in order to resolve it ?

sfc-gh-mkeller:
> I see that some checks are failing but it doesn't seem to be related to my changes (looks like a missing environment variable).
> Should I change anything in order to resolve it ?

No, don't worry about those. Once someone from the team reviews your code for malicious code (as an example dumping password) we will kick of the same tests for you with our secrets added.

PR #552, opened 2020-12-04
SNOW-230346 sped up arrow chunk rebuilding

sfc-gh-mkeller:
SNOW-230346
Previously when returning NUMBERs with a non-zero scales need to be converted into doubles and switched out in the RecordBatches. Every time `AddColumn` and `DeleteColumn` were called they had to do a lot of resizing of hashmaps and copying a lot of values into new objects.
This PR removes the need to use `AddColumn` and `DeleteColumn` and instead operates on the underlying data and rebuilds RecordBatches only if necessary.

The query provided to me by support went from:
```
48.83139857795322
0.8859019490191713
```

to

```
2.7106516549829394
0.9343068830203265
```

Note that I got a lot more consistent numbers previously, I think my computer is just busier right now.
I was getting numbers close to `0.9` and `0.7` seconds.

### DIFF
--- a/src/snowflake/connector/arrow_iterator.pyx
+++ b/src/snowflake/connector/arrow_iterator.pyx
@@ -69,6 +69,16 @@ cdef extern from "arrow/api.h" namespace "arrow" nogil:
         c_bool IsIndexError()
         c_bool IsSerializationError()
 
+    cdef cppclass CResult "arrow::Result"[T]:
+        CResult()
+        CResult(CStatus status)
+
+        c_string ToString()
+        c_string message()
+
+        c_bool ok()
+        const CStatus& status()
+        T& ValueOrDie()
 
     cdef cppclass CBuffer" arrow::Buffer":
         CBuffer(const uint8_t* data, int64_t size)
@@ -83,8 +93,7 @@ cdef extern from "arrow/ipc/api.h" namespace "arrow::ipc" nogil:
     cdef cppclass CRecordBatchStreamReader \
             " arrow::ipc::RecordBatchStreamReader"(CRecordBatchReader):
         @staticmethod
-        CStatus Open(const InputStream* stream,
-                     shared_ptr[CRecordBatchReader]* out)
+        CResult[shared_ptr[CRecordBatchReader]] Open(const InputStream* stream)
 
 
 cdef extern from "arrow/io/api.h" namespace "arrow::io" nogil:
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/arrow_iterator.pyx
```diff
@@ -154,20 +163,23 @@ cdef class PyArrowIterator(EmptyPyArrowIterator):
...
                   object numpy):
         cdef shared_ptr[InputStream] input_stream
-        cdef shared_ptr[CRecordBatchReader] reader
         cdef shared_ptr[CRecordBatch] record_batch
+        cdef CStatus ret
         input_stream.reset(new PyReadableFile(py_inputstream))
-        cdef CStatus ret = CRecordBatchStreamReader.Open(input_stream.get(), &reader)
-        if not ret.ok():
+        cdef CResult[shared_ptr[CRecordBatchReader]] readerRet = CRecordBatchStreamReader.Open(input_stream.get())
+        #cdef  ret = CRecordBatchStreamReader.Open(input_stream.get(), &reader)
```
sfc-gh-wshangguan, line 170: nit: I guess we can remove this line

src/snowflake/connector/arrow_iterator.pyx
```diff
@@ -154,20 +163,23 @@ cdef class PyArrowIterator(EmptyPyArrowIterator):
     def __cinit__(self, object cursor, object py_inputstream, object arrow_context, object use_dict_result,
                   object numpy):
         cdef shared_ptr[InputStream] input_stream
-        cdef shared_ptr[CRecordBatchReader] reader
         cdef shared_ptr[CRecordBatch] record_batch
+        cdef CStatus ret
         input_stream.reset(new PyReadableFile(py_inputstream))
-        cdef CStatus ret = CRecordBatchStreamReader.Open(input_stream.get(), &reader)
-        if not ret.ok():
+        cdef CResult[shared_ptr[CRecordBatchReader]] readerRet = CRecordBatchStreamReader.Open(input_stream.get())
```
sfc-gh-wshangguan, line 169: May I ask the reason we go with the new `Open` api here?

src/snowflake/connector/arrow_iterator.pyx
sfc-gh-mkeller: Sure, sorry

src/snowflake/connector/arrow_iterator.pyx
sfc-gh-mkeller: Arrow changed its API and deprecated the old one. This change now gets rid of warnings and allows us to use versions where this old API was removed.

sfc-gh-wshangguan: Sounds great!

src/snowflake/connector/cpp/ArrowIterator/CArrowTableIterator.cpp
```diff
@@ -154,45 +188,45 @@ std::shared_ptr<ReturnVal> CArrowTableIterator::next()
...
+    const std::shared_ptr<arrow::Array>& newColumn,
+    std::vector<std::shared_ptr<arrow::Field>>& futureFields,
+    std::vector<std::shared_ptr<arrow::Array>>& futureColumns,
+    bool& needsRebuild)
 {
   // replace the targeted column
-  std::shared_ptr<arrow::RecordBatch> currentBatch = (*m_cRecordBatches)[batchIdx];
-  arrow::Status ret = currentBatch->AddColumn(colIdx+1, newField, newColumn, &currentBatch);
-  if(!ret.ok())
+  if (needsRebuild == false)
```
sfc-gh-wshangguan, line 200: It looks like we will do the replace anyway. Maybe no need for this judge?

sfc-gh-wshangguan (approved):
Overall looks good to me, great job man!
nit: if we can separate out the changes for reformat that would be better!

sfc-gh-mkeller:
> Overall looks good to me, great job man!
> nit: if we can separate out the changes for reformat that would be better!

This is probably a good idea. Let me drop that commit

src/snowflake/connector/cpp/ArrowIterator/CArrowTableIterator.cpp
sfc-gh-mkeller: Okay, so the idea behind this if then clause is the following.

We have already existing batches that we can turn into a table for `pyarrow`, but if the batches have some data in them that needs modification then we do it at this state.
So `replace_column` is called every time we do one of these modifications, so if this is the first time we are replacing a column then we need to make a deep copy of the the chunk first and then modify that copy and at the end replace the whole chunk.
Making deep copies should not be done in 2 cases. 1. when we have already done it, or 2. if we don't need to modify any of the columns. So this if then clause does it only the one time.

sfc-gh-wshangguan: Gotcha, thanks for the clarification!

sfc-gh-kwagner (approved):
LGTM

PR #557, opened 2020-12-08
SNOW-201003 vendoring requests and urllib3

sfc-gh-mkeller:
This PR vendors `requests` and `urllib3`, as we monkey patch into `urllib3`, but we need the `requests` we use to call the vendored `urllib3`.

Please note that most of [`492e544` (#557)](https://github.com/snowflakedb/snowflake-connector-python/pull/557/commits/492e54442bb910ff42815dc8b861da7ae996c609) has the source files added, I'd suggest looking that over quickly only.

Note that I will not squash all the commits together, so that we can have 1 separate commit that adds the source files in git history.

A new `NOTICE` file was added to point out the newly vendored libraries. More info on these files is available [here](https://www.apache.org/licenses/LICENSE-2.0.html#redistribution).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -195,8 +195,8 @@ setup(
         'azure-common<2.0.0',
         'azure-storage-blob>=12.0.0,<13.0.0',
         'boto3>=1.4.4,<2.0.0',
-        'requests<2.24.0',
-        'urllib3>=1.20,<1.26.0',
+        'requests<3.0.0',
+        'urllib3<2.0.0',
         'certifi<2021.0.0',
         'pytz<2021.0',
         'pycryptodomex>=3.2,!=3.5.0,<4.0.0',
@@ -210,16 +210,29 @@ setup(
         # required. Python 3.6 was released at the end of 2016. setuptools 34.0.0 was released
         # in early 2017, so we pick this version as a reasonably modern base.
         'setuptools>34.0.0',
+        # requests requirements
+        'chardet>=3.0.2,<4',
+        'idna>=2.5,<3',
+        'certifi>=2017.4.17',
     ],
 
     namespace_packages=['snowflake'],
     packages=[
         'snowflake.connector',
         'snowflake.connector.tool',
+
+        'snowflake.connector.vendored',
+        'snowflake.connector.vendored.requests',
+        'snowflake.connector.vendored.urllib3',
+        'snowflake.connector.vendored.urllib3.packages',
+        'snowflake.connector.vendored.urllib3.packages.ssl_match_hostname',
+        'snowflake.connector.vendored.urllib3.packages.backports',
+        'snowflake.connector.vendored.urllib3.contrib',
+        'snowflake.connector.vendored.urllib3.contrib._securetransport',
+        'snowflake.connector.vendored.urllib3.util',
     ],
     package_dir={
-        'snowflake.connector': os.path.join('src', 'snowflake', 'connector'),
-        'snowflake.connector.tool': os.path.join('src', 'snowflake', 'connector', 'tool'),
+        '': 'src',
     },
     package_data={
         'snowflake.connector': ['*.pem', '*.json', '*.rst', 'LICENSE.txt'],
--- a/src/snowflake/connector/ssl_wrap_socket.py
+++ b/src/snowflake/connector/ssl_wrap_socket.py
@@ -16,16 +16,17 @@ import time
 from functools import wraps
 from inspect import getfullargspec as get_args
 from socket import socket
+from typing import Optional
 
 import certifi
 import OpenSSL.SSL
-import requests.packages.urllib3.connection as connection_
-import requests.packages.urllib3.util.ssl_ as ssl_
 from urllib3.contrib.pyopenssl import PyOpenSSLContext
 
 from .constants import OCSPMode
 from .errorcode import ER_OCSP_RESPONSE_CERT_STATUS_REVOKED
 from .errors import OperationalError
+from .vendored.urllib3 import connection as connection_
+from .vendored.urllib3.util import ssl_ as ssl_
 
 FEATURE_OCSP_MODE = OCSPMode.FAIL_OPEN
 
@@ -98,7 +99,10 @@ def ssl_wrap_socket_with_ocsp(*args, **kwargs):
     return ret
 
 
-def _openssl_connect(hostname, port=443, max_retry=20):
+def _openssl_connect(hostname: str,
+                     port: int = 443,
+                     max_retry: int = 20,
+                     timeout: Optional[int] = None) -> OpenSSL.SSL.Connection:
     """The OpenSSL connection without validating certificates.
 
     This is used to diagnose SSL issues.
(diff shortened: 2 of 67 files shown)


potiuk:
Hey @sfc-gh-mkeller  - can we help somehow with moving this forward :) ? 

The library is quite problematic for us without those  vendored in. This is quite a blocker for Airflow to both: 

* release the "snowflake" provider package - we had to `yank` it https://pypi.org/project/apache-airflow-providers-snowflake/ right after release because it breaks many other providers. 
* finish making our requirements fully consistent and passing `pip check` - we had to explicily disable 'apache-beam' from our CI tests because it conflicts with snowflake connector `requests` .

We'd love to help if possible.

sfc-gh-mkeller:
Hi @potiuk 

Your input and review is welcome

sfc-gh-mkeller:
Hi @potiuk 

Your input and review is welcome (please feel free to post it here), but I'll need other maintainers to review this change too.
Let me complete the description, because I think this is ready for review.

potiuk:
> Hi @potiuk 
> 
> Your input and review is welcome (please feel free to post it here), but I'll need other maintainers to review this change too.
> Let me complete the description, because I think this is ready for review. 

Happy to take a look later this week ! I add it to my list of todos :).

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -1265,7 +1266,7 @@ def _lazy_read_ca_bundle(self):
                     self.read_cert_bundle(ca_bundle)
                 else:
                     import sys
-                    from requests import certs
+                    from .vendored.requests import certs
```
sfc-gh-stakeda, line 1269: `generic_requests` ?

sfc-gh-stakeda: or if not then, the above `from .vendored import requests as generic_requests` is required?

sfc-gh-stakeda (approved):
overall LGTM. Ensure all tests pass on all supported platforms.

sfc-gh-kwagner (approved):
LGTM, this should make a lot of people happy!

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-hchaturvedi: I think this was meant to reach out to the ocsp responders directly. @sfc-gh-mkeller we'd want to use regular requests outside the vendor-d package for this one.

sfc-gh-mkeller: Oops, thank you @sfc-gh-hchaturvedi

sfc-gh-hchaturvedi (approved):
Super Awesome! Just one comment about fixing the generic requests but looks good beyond that. Thanks for doing this Mark!

sfc-gh-hchaturvedi:
> Super Awesome! Just one comment about fixing the generic requests but looks good beyond that. Thanks for doing this Mark!

Never mind the comment - I see it has been fixed.

src/snowflake/connector/ssl_wrap_socket.py
```diff
@@ -39,6 +40,7 @@
 
 def inject_into_urllib3():
     """Monkey-patch urllib3 with PyOpenSSL-backed SSL-support and OCSP."""
+    # TODO replace this with the code in vendored
```
sfc-gh-wshangguan, line 43: nit: still TODO?

sfc-gh-mkeller: I'll remove the TODO, I was going to move this into the vendored code, but since I decided not to do this. Having it separate like this makes it easier to update the vendored library. Thanks for catching this, I'll look through the changes to make sure I don't add more TODOs.

sfc-gh-wshangguan (approved):
Overall LGTM!

setup.py
```diff
@@ -195,8 +195,8 @@ def _get_arrow_lib_as_linker_input(self):
         'azure-common<2.0.0',
         'azure-storage-blob>=12.0.0,<13.0.0',
         'boto3>=1.4.4,<2.0.0',
-        'requests<2.24.0',
-        'urllib3>=1.20,<1.26.0',
+        'requests<3.0.0',
```
potiuk, line 198: Do you still need those? I think you can safely remove them since you added both requests and urlib as vendored library as well as added the requests dependencies below, you do not need (or even should not have) any dependencies directly to requests/urllib.

Unless there is a good reason to keep them?

test/integ/test_put_get_with_gcp_account.py
```diff
@@ -23,6 +22,13 @@
 from ..generate_test_files import generate_k_lines_of_n_files
 from ..randomize import random_string
 
+try:
+    from snowflake.connector.vendored import requests  # NOQA
+    vendored_request = True
+except ImportError:  # pragma: no cover
+    import requests
```
potiuk, line 29: Where do you think you need to handle the Import Error? I belive the code will always run with the vendored requests in. this is a test code, so not that this is a problem. but I think this try/except might be not needed?

Especially that below patching explicitly patches the 'snowflake.connector.vendored.requests'. That might lead to some problems if for any reasons this import does not work and might be hard to find problems.

test/unit/test_gcs_util.py
```diff
@@ -22,6 +20,13 @@
 except ImportError:
     SnowflakeGCSUtil = None
 
+try:
+    from snowflake.connector.vendored import requests  # NOQA
+    vendored_request = True
+except ImportError:  # pragma: no cover
+    import requests
+    vendored_request = False
+
```
potiuk, line 29: Same here. I do not see a good reason to keep the try/except :)

test/unit/test_retry_network.py
```diff
@@ -29,6 +27,13 @@
 from snowflake.connector.errors import DatabaseError, InterfaceError, OtherHTTPRetryableError
 from snowflake.connector.network import STATUS_TO_EXCEPTION, RetryRequest, SnowflakeRestful
 
+try:
+    from snowflake.connector.vendored import requests  # NOQA
+    from snowflake.connector.vendored import urllib3  # NOQA
+except ImportError:  # pragma: no cover
```
potiuk, line 33: Same here .

potiuk (commented):
A little late to the party, but I think a small follow-up might be needed I wll prepare PR shortly.

test/integ/test_put_get_with_gcp_account.py
potiuk: Update. I see the if's below, still not sure why would you want to need that (I do not see any case where importing snoflake.connector.vendored would fail (I have not seen any exclusion while preparing packages).

test/unit/test_gcs_util.py
sfc-gh-mkeller: We need these for our `OldDriver` tests. We run most up to date tests with the oldest supported driver version

potiuk: OK.Fair enough. Happy to add this comment instead of removing it.

setup.py
sfc-gh-mkeller: Yes, when we send OCSP queries to the responders, only then we don't want to do OCSP checks.
So unfortunately we still need to keep `requests` as a dependency.

For the future we could make each PUT/GET call decide if they wanted OCSP or not.

test/integ/test_put_get_with_gcp_account.py
sfc-gh-mkeller: The tox environment called `olddriver` runs the newest set of our tests with an older version of the connector to make sure that we don't break compatibility, but the old version will not have the vendored `requests` and `urllib3` in it. So maybe we don't actually need to set them to `None`, but I do this trick often for type hints.

PR #573, opened 2020-12-15
SNOW-208979 Add option to trigger Github Action Build/Test manually

sfc-gh-abhatnagar:
This will add a button for build-test workflow to run it manually . This will be useful for Releng team to run python tests on master  or any other branch as part of pre-release of server release is done to ensure it does not break python connector


sfc-gh-mkeller (approved):
🚢 🚢

PR #591, opened 2021-01-06
SNOW-257201 Explain why requests, urllib3 are still used 

potiuk:
This PR explains why urllib3 and requests are
still used. The PR #557 vendored in urrlib3
and request libraries but some usages of the
original libraries remained (but were not
explained)

This was not clear from the code why urllib and requests
are still used but after explanations the comments
about it have been placed in the code.

Also some inconsistencies have been removed where
the old urlib3 and requests were directly referred to,
while the vendored version should be used insted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -195,8 +195,8 @@ setup(
         'azure-common<2.0.0',
         'azure-storage-blob>=12.0.0,<13.0.0',
         'boto3>=1.4.4,<2.0.0',
+        # While requests is vendored, we use regular requests to perform OCSP checks
         'requests<3.0.0',
-        'urllib3<2.0.0',
         'certifi<2021.0.0',
         'pytz<2021.0',
         'pycryptodomex>=3.2,!=3.5.0,<4.0.0',
--- a/test/extras/no_use_pycon.py
+++ b/test/extras/no_use_pycon.py
@@ -4,7 +4,6 @@
 # Copyright (c) 2012-2020 Snowflake Computing Inc. All right reserved.
 #
 # This test tests for SNOW-186747
-
 import requests
 
 import snowflake.connector  # NOQA
--- a/tox.ini
+++ b/tox.ini
@@ -161,7 +161,7 @@ include_trailing_comma = True
 force_grid_wrap = 0
 line_length = 120
 known_first_party =snowflake,parameters,generate_test_files
-known_third_party =Cryptodome,OpenSSL,asn1crypto,azure,boto3,botocore,certifi,chardet,cryptography,dateutil,idna,jwt,mock,ntlm,numpy,pendulum,pkg_resources,pytest,pytz,requests,setuptools,urllib3
+known_third_party =Cryptodome,OpenSSL,asn1crypto,azure,boto3,botocore,certifi,chardet,cryptography,dateutil,idna,jwt,mock,ntlm,numpy,pendulum,pkg_resources,pytest,pytz,requests,setuptools
 
 [flake8]
 ignore=F821,W504,E501,E402,E122,E127,E126,E128,E131,E731,E125,E722x,D100,D101,D102,D103,D104,D105,D107,D401,D204
--- a/src/snowflake/connector/ocsp_snowflake.py
+++ b/src/snowflake/connector/ocsp_snowflake.py
@@ -23,6 +23,9 @@ from threading import Lock, RLock
 from time import gmtime, strftime
 
 import jwt
+# We use regular requests and urlib3 when we reach out to do OCSP checks, basically in this very narrow
+# part of the code where we want to call out to check for revoked certificates,
+# we don't want to use our hardened version of requests.
 import requests as generic_requests
 
 from snowflake.connector.compat import OK, urlsplit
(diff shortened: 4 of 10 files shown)


potiuk:
I do not know the library well enough, yet, but I believe those vendored in dependencies can and should completely replace the original ones, and it was not the case. This is not a big problem for test code, but the library code also used 'generic' library for no apparent reason. Not mentioning that the vendored urllib3 library itself used the non-vendored one (importing urllib3 directly). 

This caused completely unnecessary requirements to urllib3 and requests remained in the install_requires, but since they were vendored in, it is not really necessary (and I think actually harmful because it prevents other libraries from using even higher versions of those when they are released).

Again - I might not realise all the limitations (so please correct me and explain if I do not understand something.  Happy to iterate on it and improve it if there are some hidden requirements/expectations I am not aware of :)

potiuk:
I just noticed you use pre-commit :). Cool. Love it and use it for years :).

potiuk:
I do not have all the secrets for the PR coming from a fork, but seems all the non-secret bound tests are passing !

sfc-gh-mkeller:
Hi @potiuk 

We use regular `requests` and `urlib3` when we reach out to do OCSP checks, basically in this very narrow part of the code where we want to call out to check for revoked certificates, we don't want to use our hardened version of `requests`. The other import that depends on these libraries is to import certificates from them, we would like to have these as up to date as possible.

Therefore while we could remove these dependencies completely, we would like to keep them (with very loose pinning). We have setup dependabot on this repo, so that bot will let us know if these pins should be updated.

CC @sfc-gh-hchaturvedi for security insight.

potiuk:
Ok. Understood. Makes sense. 

I will modify it tomorrow then to leave only the vendored library change (but also fix the test import exceptions as i think those are bogus).

The urllib3 vendored library used urllib3 import rather than vendored one - that's where it started :)

potiuk:
@sfc-gh-mkeller I added explanations that I got from you on both accounts - this PR #591 and the related one #557. 

I commented out (following the comments I received) the areas that needed comment (as they were unclear why they were left) and I fixed all the other places, where I thought still the vendored library should be used instead of the generic one. Looking forward to incrporating this one.

BTW. Do you know what is the expected timeline for release of those changes? We are really looking forward to this one in Apache Airflow to be able to release much more stable set of requirements!

test/extras/no_use_pycon.py
```diff
@@ -5,9 +5,8 @@
 #
 # This test tests for SNOW-186747
 
-import requests
-
 import snowflake.connector  # NOQA
+from snowflake.connector.vendored import requests
```
sfc-gh-mkeller, line 9: This test makes sure that the monkey patching didn't affect the use cases where our connector was imported, but it never used. This has burned us in the past.
Please don't change it even though the monkey patching should really not affect other libraries anymore it won't hurt to have this one extra test for a bit longer 😄

setup.py
```diff
@@ -195,6 +195,13 @@ def _get_arrow_lib_as_linker_input(self):
         'azure-common<2.0.0',
         'azure-storage-blob>=12.0.0,<13.0.0',
         'boto3>=1.4.4,<2.0.0',
+        # We use regular requests and urlib3 when we reach out to do OCSP checks, basically in this very
+        # narrow part of the code where we want to call out to check for revoked certificates,
+        # we don't want to use our hardened version of requests. Also we import certificates from them,
+        # we would like to have these as up to date as possible.
+        # Therefore while we could remove these dependencies completely, we would like to keep them
+        # (with very loose pinning). We have setup dependabot on this repo, so that bot will let us
+        # know if these pins should be updated.
```
sfc-gh-mkeller, line 204:
```suggestion
        # While these are vendored we use regular requests and urllib3 to perform OCSP checks
```

tox.ini
```diff
@@ -161,7 +161,7 @@ include_trailing_comma = True
 force_grid_wrap = 0
 line_length = 120
 known_first_party =snowflake,parameters,generate_test_files
-known_third_party =Cryptodome,OpenSSL,asn1crypto,azure,boto3,botocore,certifi,chardet,cryptography,dateutil,idna,jwt,mock,ntlm,numpy,pendulum,pkg_resources,pytest,pytz,requests,setuptools,urllib3
+known_third_party =Cryptodome,OpenSSL,asn1crypto,azure,boto3,botocore,certifi,chardet,cryptography,dateutil,idna,jwt,mock,ntlm,numpy,pendulum,pkg_resources,pytest,pytz,requests,setuptools
```
sfc-gh-mkeller, line 164: Ohh, this might mean that `urllib3` is never imported anymore. I guess that last import that you fixed was the only one. In that case we can remove the dependency on `urllib3` 😄

sfc-gh-mkeller (commented):
After I typed up that last comment I realized that if `urllib3` was not imported more then we could remove it from our dependencies. I took the liberty of applying my suggestions to a branch and kicked off testing. Let's see if we can at least get rid of our `urllib3` dependency :)

tox.ini
potiuk: Yep. It looks like :)

potiuk:
All fixed!

sfc-gh-mkeller:
> All fixed!

There's an import statement that was removed by mistake and I want to retire the `test_another_site` tests after all. It has been acting up since we merged the vendoring change 😓 
If it's okay with you I'll pull your current changes into #592 make these changes myself and merge that once tests pass.
Merging the code that way should also maintain your contribution!

potiuk:
Sure! No problem @sfc-gh-mkeller !

sfc-gh-mkeller:
Merged through #591 
Thank you for your help @potiuk

PR #592, opened 2021-01-07
SNOW-257201 Test PR 591

sfc-gh-mkeller:
Tests PR #591

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -195,8 +195,8 @@ setup(
         'azure-common<2.0.0',
         'azure-storage-blob>=12.0.0,<13.0.0',
         'boto3>=1.4.4,<2.0.0',
+        # While requests is vendored, we use regular requests to perform OCSP checks
         'requests<3.0.0',
-        'urllib3<2.0.0',
         'certifi<2021.0.0',
         'pytz<2021.0',
         'pycryptodomex>=3.2,!=3.5.0,<4.0.0',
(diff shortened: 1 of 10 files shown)


sfc-gh-kwagner (approved):

setup.py
```diff
@@ -195,8 +195,8 @@ def _get_arrow_lib_as_linker_input(self):
         'azure-common<2.0.0',
         'azure-storage-blob>=12.0.0,<13.0.0',
         'boto3>=1.4.4,<2.0.0',
+        # While requests is vendored, we use regular requests to perform OCSP checks
```
sfc-gh-hchaturvedi, line 198: Did we not decide to _not_ pin requests to a specific version?

sfc-gh-mkeller: I pinned it to next major release just in case to be safe. I was thinking that we can always relax it later if we want to, but it's still pinned currently. However, there hasn't been many major releases for `requests`.

Do you feel like we could unpin it safely?

sfc-gh-hchaturvedi (approved):
Just had the concern about pinning requests and running into customer issues where the driver does not auto upgrade to latest request. Post conversations with Mark we decided that we want to continue holding the driver to a stable major release and we do a major release update ourselves. Given that major releases have major changes and are not that frequent I think it's reasonable to do that. Hence :shipit:

PR #595, opened 2021-01-08
SNOW-257781 adding more documentation and files necessary to build custom GH actions

sfc-gh-mkeller:
I was going to add documentation on how to build the connector in the most up to date way as per @sfc-gh-hchaturvedi s request, but I found that I wrote a fairly good one already at: https://github.com/snowflakedb/snowflake-connector-python/blob/master/test/README.rst

But then I found that out custom GH actions are not documented at all, so I checked in those files and wrote a small tutorial on how to compile them again if need be.

I've also updated some gitignores to allow for multiple venvs and parameter files. This is something that I started doing and would like to have these extra files not show up in my `git status`s from now on.

### DIFF
--- a/.github/actions/.gitignore
+++ b/.github/actions/.gitignore
@@ -1,0 +1,1 @@
+**/node_modules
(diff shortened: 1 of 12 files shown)


sfc-gh-kwagner (approved):
LGTM

PR #602, opened 2021-01-19
SNOW-265864 adding CODEOWNERS

sfc-gh-mkeller:
SNOW-265864

Adding CODEOWNERS

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @snowflakedb/Client
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @snowflakedb/Client
+* @sfc-gh-mkeller


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @snowflakedb/Client
```
sfc-gh-kwagner, line 1: Won't this add everyone on the team as a code owner and therefore as a code reviewer on PRs?

sfc-gh-mkeller: Yes. I thought that wasn't the best idea either, but this is what has been done in our other repos.

Do you think that we should add individual owners to repositories?

sfc-gh-kwagner: Yea so it adds a lot of noise. Emails are already noisy and so I use https://github.com/pulls/review-requested to be able to really tell which reviews I need to take a serious look at. With this change, I'm getting included on all sorts of reviews and its not super clear which ones I really need to take a look at. Having individuals as code owners could work and when an individual picks up a driver they can send a change to this file to add their name. Similarly when they leave they can send a PR to remove their name from this list. Its slightly more work, but it would be distributed to the individual and this shouldn't get updated too often

sfc-gh-kwagner (approved):

Issue #284, opened 2020-02-27
SNOW-141545: Request - Reduce package dependencies chaos


manugarri:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?
`python 3.7`
2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
`Linux-5.3.0-40-generic-x86_64-with-debian-buster-sid`

3. What are the component versions in the environment (`pip list`)? 
```
Package                    Version   
-------------------------- ----------
ansible                    2.8.6     
asn1crypto                 1.2.0     
attrs                      19.3.0    
awscli                     1.16.265  
azure-common               1.1.23    
azure-storage-blob         2.1.0     
azure-storage-common       2.1.0     
backcall                   0.1.0     
bleach                     3.1.0     
blis                       0.4.1     
bokeh                      1.4.0     
boto                       2.49.0    
boto3                      1.10.50   
botocore                   1.13.50   
certifi                    2019.11.28
cffi                       1.13.1    
chardet                    3.0.4     
Click                      7.0       
cloudpickle                1.2.2     
colorama                   0.4.1     
conda                      4.8.0     
conda-package-handling     1.6.0     
credsman                   0.0.1     
croniter                   0.3.31    
cryptography               2.7       
cymem                      2.0.2     
dask                       2.6.0     
datasketch                 1.4.10    
dateparser                 0.7.0     
decorator                  4.4.0     
defusedxml                 0.6.0     
distributed                2.6.0     
docutils                   0.15.2    
dxtrack                    0.6.1     
elasticsearch              7.1.0     
entrypoints                0.3       
fastjsonschema             2.14.1    
fastparquet                0.3.2     
fsspec                     0.6.2     
future                     0.18.1    
gspread                    3.1.0     
HeapDict                   1.0.1     
idna                       2.8       
ijson                      2.5.1     
importlib-metadata         0.23      
ipykernel                  5.1.3     
ipython                    7.8.0     
ipython-genutils           0.2.0     
ipywidgets                 7.5.1     
jedi                       0.15.1    
jellyfish                  0.7.2     
Jinja2                     2.10.3    
jmespath                   0.9.4     
joblib                     0.14.0    
json5                      0.8.5     
jsonschema                 3.1.1     
jupyter-client             5.3.3     
jupyter-core               4.5.0     
jupyterlab                 1.1.4     
jupyterlab-server          1.0.6     
[...]

PR #604, opened 2021-01-20
SNOW-270639 pyjwt major bump and compatibility

### LINKED ISSUE #586: SNOW-252627: connect call fails from pyjwt == 2.0.0
>>> c = snowflake.connector.connect(
...             warehouse=warehouse,
...             account=CONFIGURATION.account,
...             user=CONFIGURATION.username,
...             private_key=CONFIGURATION.key,
...             client_session_keep_alive=True
...         )
  File "<stdin>", line 6, in <module>
  File "/opt/zme-ues/venv/lib/python3.6/site-packages/snowflake/connector/__init__.py", line 33, in Connect
  File "/opt/zme-ues/venv/lib/python3.6/site-packages/snowflake/connector/connection.py", line 187, in __init__
    self.connect(**kwargs)
  File "/opt/zme-ues/venv/lib/python3.6/site-packages/snowflake/connector/connection.py", line 484, in connect
    self.__open_connection()
  File "/opt/zme-ues/venv/lib/python3.6/site-packages/snowflake/connector/connection.py", line 702, in __open_connection
    self.__authenticate(auth_instance)
  File "/opt/zme-ues/venv/lib/python3.6/site-packages/snowflake/connector/connection.py", line 915, in __authenticate
    password=self._password,
  File "/opt/zme-ues/venv/lib/python3.6/site-packages/snowflake/connector/auth_keypair.py", line 78, in authenticate
    algorithm=self.ALGORITHM).decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
>>>

jameslamb:
`pyjwt` 2.0.0 was released [about a month ago](https://github.com/jpadilla/pyjwt/releases/tag/2.0.0). In that release, the behavior of `jwt.encode()` changed. It now returns a string instead of `bytes`, which is what caused the bug reported in #586 .

This PR proposes removing any version requirements on `pyjwt` in `snowflake-connector-python`. I think with the minor changes proposed in this PR, it's possible for `snowflake-connector-python` to work with both `pyjwt` 1.x and `pyjwt` 2.x.

## Changes in this PR

* makes the one use of `jwt.encode()` in this project compatible with `pyjwt` 1.x and 2.x
* removes version constraint on `pyjwt`

## How this improves `snowflake-connector-python`

This change would make `snowflake-connector-python` easier to install alongside other packages, because it relaxes a constraint placed on a popular and central package (`pyjwt`).

I think this could be a small step towards alleviating some of the dependency pain expressed in #284.

## Notes for reviewers

I looked for uses of `pyjwt` in this project with `git grep -E "jwt\."`. I think that the uses of `get_unverified_header()` and `.decode()` will work without modification, with both 1.x and 2.x.

`get_unverified_header()` was not changed between versions.

2.0.1 (https://github.com/jpadilla/pyjwt/blob/3993ce1d3503b58cf74699a89ba9e5c18ef9b556/jwt/api_jws.py)

```python
def get_unverified_header(self, jwt):
    """Returns back the JWT header parameters as a dict()
    Note: The signature is not verified so the header parameters
    should not be fully trusted until signature verification is complete
    """
    headers = self._load(jwt)[2]
    self._validate_headers(headers)

    return headers

def _validate_headers(self, headers):
    if "kid" in headers:
        self._validate_kid(headers["kid"])

def _validate_kid(self, kid):
    if not isinstance(kid, str):
        raise InvalidTokenError("Key ID header parameter must be a string")
```

1.7.1 (https://github.com/jpadilla/pyjwt/blob/b65e1ac6dc4d11801f3642eaab34ae6a54162c18/jwt/api_jws.py)

```python
def get_unverified_header(self, jwt):
    """Returns back the JWT header parameters as a dict()
    Note: The signature is not verified so the header parameters
    should not be fully trusted until signature verification is complete
    """
    headers = self._load(jwt)[2]
    self._validate_headers(headers)

    return headers

def _validate_headers(self, headers):
    if 'kid' in headers:
        self._validate_kid(headers['kid'])

def _validate_kid(self, kid):
    if not isinstance(kid, string_types):
        raise InvalidTokenError('Key ID header parameter must be a string')

```

`.decode()` in 2.x still supports algorithm RS512

https://github.com/jpadilla/pyjwt/blob/9dca8eaa0ec43bcf54ac645872078ecdb0ca8d96/jwt/algorithms.py#L80

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -203,7 +203,7 @@ setup(
         'pyOpenSSL>=16.2.0,<20.0.0',
         'cffi>=1.9,<2.0.0',
         'cryptography>=2.5.0,<4.0.0',
-        'pyjwt',
+        'pyjwt<3.0.0',
         'oscrypto<2.0.0',
         'asn1crypto>0.24.0,<2.0.0',
         # A functioning pkg_resources.working_set.by_key and pkg_resources.Requirement is
--- a/setup.py
+++ b/setup.py
@@ -203,7 +203,7 @@ setup(
         'pyOpenSSL>=16.2.0,<20.0.0',
         'cffi>=1.9,<2.0.0',
         'cryptography>=2.5.0,<4.0.0',
-        'pyjwt<2.0.0',
+        'pyjwt',
         'oscrypto<2.0.0',
         'asn1crypto>0.24.0,<2.0.0',
         # A functioning pkg_resources.working_set.by_key and pkg_resources.Requirement is
--- a/src/snowflake/connector/auth_keypair.py
+++ b/src/snowflake/connector/auth_keypair.py
@@ -74,8 +74,18 @@ class AuthByKeyPair(AuthByPlugin):
             self.EXPIRE_TIME: self._jwt_token_exp
         }
 
-        self._jwt_token = jwt.encode(payload, private_key,
-                                     algorithm=self.ALGORITHM).decode('utf-8')
+        _jwt_token = jwt.encode(
+            payload,
+            private_key,
+            algorithm=self.ALGORITHM
+        )
+
+        # jwt.encode() returns bytes in pyjwt 1.x and a string
+        # in pyjwt 2.x
+        if isinstance(_jwt_token, bytes):
+            self._jwt_token = _jwt_token.decode('utf-8')
+        else:
+            self._jwt_token = _jwt_token
 
         return self._jwt_token
 


setup.py
```diff
@@ -203,7 +203,7 @@ def _get_arrow_lib_as_linker_input(self):
         'pyOpenSSL>=16.2.0,<20.0.0',
         'cffi>=1.9,<2.0.0',
         'cryptography>=2.5.0,<4.0.0',
-        'pyjwt<2.0.0',
+        'pyjwt',
```
sfc-gh-mkeller, line 206:
```suggestion
        'pyjwt<3.0.0',
```
Let's add this, so that we take a look at changelog upon the next major release and verify that the API changes don't affect us before approving it, please!

sfc-gh-mkeller (changes requested):
Thank you for making the code work with both the older and the newer version of `pyjwt`.
Please address my one comment and then I can kick of tests for you 🚀

setup.py
jameslamb: In my opinion, major-version ceilings are a good practice for applications, but should be avoided for libraries unless you know for sure that your library is incompatible with that new version.

Imagine that you merge this PR and release `snowflake-connector-python` today, and someone installs it. Tomorrow, `pyjwt` 3.0.0 comes out.

Here's how that could play out, based on whether or not `snowflake-connector-python` is compatible with `pyjwt` 3.0.0.

**without a ceiling**

*  incompatible --> need to pin to older `pyjwt` or open an issue asking for a new `snowflake-connector-python` release
*  compatible --> no need to pin `pyjwt`, no action required by snowflake

**with a ceiling**

* incompatible --> need to pin to older `pyjwt` or open an issue asking for a new `snowflake-connector-python` release
* compatible --> need to pin to older `pyjwt` or open an issue asking for a new `snowflake-connector-python` release

Not putting a ceiling in place gives this project's users a chance of not needing to upgrade `snowflake-connector-python` if they require `pyjwt` (which might come from them requiring some other package that requires `pyjwt>3.0.0`).

jameslamb: all that said...if you read that and disagree, I'd be happy to follow your suggestion and add the ceiling

sfc-gh-mkeller: I agree with you James and this is something that many customers have been asking for (there are many other issues open with this request already).

Short answer of why this is that historically our library has been treated as an application, so now many of our customers demand that we minimize the possibility of errors, but this is only possible if we control dependency versions.

I wasn't here when this decision was made, so I can't comment on the specifics of why, but if I had to take a guess our library broke way too often. I have definitely seen this myself many times. For example this new `pyjwt` release had the possibility of breaking customer deployments as you found. Not once did we have to get online over the weekends to pin against certain newly released dependencies and do an unscheduled release of the connector to fix things. Over the years this lead to the state we are in now. I'm trying to make a better effort to keep things up to date, but I'd lying if I said that I'm on top of everything, this is why we introduced dependabot. It kicks off tests every dependency bump before actually moving our pin.

My other view is that if there are people that would like our library with less tightly coupled dependencies, they can really easily build it for themselves. This is the specific reason why I have introduced PEP-517 build and I was hoping that having production grade build openly documented in our GitHub actions would further aid users of this group.

I hope that this insight makes sense to you and that you find it to make at least some sense, that being said please add the next major pin, we should test at least breaking public API changes.

jameslamb: > This is the specific reason why I have introduced PEP-517 build and I was hoping that having production grade build openly documented in our GitHub actions would further aid users of this group.

Can you link me to the docs on this? I wasn't aware of an alternative build of `snowflake-connector-python` that has looser constraints on dependencies.

I don't see an alternative wheel / sdist tarball at https://pypi.org/project/snowflake-connector-python/#files or on [the latest release](https://github.com/snowflakedb/snowflake-connector-python/releases/tag/v2.3.8). That isn't mentioned in https://docs.snowflake.com/en/user-guide/python-connector-install.html or https://docs.snowflake.com/en/user-guide/python-connector-dependencies.html.

> please add the next major pin

Sure, thanks for hearing me out. Added in https://github.com/snowflakedb/snowflake-connector-python/pull/604/commits/e8ff9a2d2dcec0f1dd75de2814b9665ab809808a

sfc-gh-mkeller: I'm sorry, I didn't mean that we release that version. What I meant is that if you want to have a build like that then you could patch our `setup.py` file and copy what we do here: https://github.com/snowflakedb/snowflake-connector-python/blob/e5e4979ca373fb6a639f5981e63e04c3aea18c0b/.github/workflows/build_test.yml#L87
and that would yield you the same caliber of wheels that we release on PyPi with your patched dependency pins.

sfc-gh-mkeller (approved):
Thank you @jameslamb

PR #606, opened 2021-01-20
SNOW 266042 add logging for request id guid and query

sfc-gh-stan:
[SNOW-266042](https://snowflakecomputing.atlassian.net/browse/SNOW-266042)
**Description**
Added logging for Query ID before ping-pong of query result retrieval. This made debugging easier for the case when query result retrieval fails after a query succeeded.
Added logging for Request ID and Request GUID.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (c) 2012-2021 Snowflake Computing Inc. All right reserved.
-
 #
 
 import copy
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -830,7 +830,7 @@ class SnowflakeConnection(object):
             )
 
         ret = self.rest.request(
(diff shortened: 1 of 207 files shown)


src/snowflake/connector/connection.py
```diff
@@ -829,12 +829,10 @@ def cmd_query(self,
                 is_file_transfer
             )
 
-        url_parameters = {REQUEST_ID: request_id}
-
         ret = self.rest.request(
-            '/queries/v1/query-request?' + urlencode(url_parameters),
+            '/queries/v1/query-request?',
```
sfc-gh-mkeller, line 833: If we leave the `?` at the end we end up with urls like `{host}/queries/v1/query-request?&request_id=asd&request_guid=def`. Let's remove the extra `?`.
```suggestion
            '/queries/v1/query-request',
```

sfc-gh-mkeller (changes requested):
Other than the one comment, the change looks good!

sfc-gh-mkeller (approved):
🚢 🚢

PR #614, opened 2021-01-28
SNOW-266042 Add logging for request id, guid and move logging for query id before ping pong

sfc-gh-stan:
[SNOW-266042](https://snowflakecomputing.atlassian.net/browse/SNOW-266042)

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -561,6 +561,9 @@ class SnowflakeRestful(object):
                 return self._post_request(
                     url, headers, body, token=self.token, timeout=timeout)
 
+        if ret.get('data') and ret['data'].get('queryId'):
+            logger.debug('Query id: {}'.format(ret['data']['queryId']))
+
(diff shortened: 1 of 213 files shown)


src/snowflake/connector/network.py
```diff
@@ -561,6 +561,9 @@ def _post_request(self, url, headers, body, token=None,
                 return self._post_request(
                     url, headers, body, token=self.token, timeout=timeout)
 
+        if ret.get('data') and ret['data'].get('queryId'):
+            logger.debug('Query id: {}'.format(ret['data']['queryId']))
```
sfc-gh-mkeller, line 565: Hmmm, should we come up with a standard when to use f-strings and when to do formats. Maybe should we just only only use one of them, or just leave the connector to be the wild west of these 2 options? Interested in your input @sfc-gh-stan @sfc-gh-kwagner @sfc-gh-wshangguan @sfc-gh-hchaturvedi @sfc-gh-stakeda

PR #617, opened 2021-02-01
SNOW-274789: Bumped up PythonConnector PATCH version from 2.3.9 to 2.…

sfc-gh-abhatnagar:
…3.10

@noreview - This is an automated process. No review is required


sfc-gh-abhatnagar (commented):
Change

PR #621, opened 2021-02-03
SNOW-274851 removing pytz pin as it doesn't follow semver anyway

sfc-gh-mkeller:
This PR removes the upper version pin of `pytz`, as this library doesn't follow semantic versioning I don't see a reason to maintain an upper pin. Secondly it has not broken testing yet, so I'm going to give it the benefit of the doubt and remove the upper pin until it breaks something in our testing pipeline.


sfc-gh-kwagner (approved):

sfc-gh-stan (approved):
🚢 🚢

PR #565, opened 2020-12-11
SNOW-240336 adding Python 3.9 support

sfc-gh-mkeller:
SNOW-240336
Adding support and testing for Python 3.9 for both Jenkins and GH actions.
`pyarrow` dropped releasing `manylinux1` wheels, so I did some digging and it appears as it's okay for us to do the same as well.

PR #624, opened 2021-02-09
SNOW-281257 pin cryptodomex

sfc-gh-mkeller:
The newest `cryptodomex` release `3.10.1` is causing issues, making our 3.6 tests to hang indefinitely.
While we investigate the culprit pin to the last working version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -199,7 +199,7 @@ setup(
         'requests<3.0.0',
         'certifi<2021.0.0',
         'pytz',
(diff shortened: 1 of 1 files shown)


sfc-gh-mkeller:
Moving this into #565 as this branch still uses `manylinux1` to build in and `cryptography` requires rust to build from source now (they also stopped releasing for `manylinux1` at the same time).

PR #625, opened 2021-02-11
SNOW-281257 relaxing pycryptodomex pin from SNOW-281257 and fixing hanging test

sfc-gh-mkeller:
SNOW-281257
Originally I solved this issue by pinning back `pycryptodomex` to last release, this caused our Mac tests to start hanging. I am still unable to find the cause of Mac Python 3.6 hanging issue, but I found that `manylinux2014` was causing the newest version to hang. So by changing the Docker container used from `manylinux2014` to `manylinux2010` we can remove the pin that I introduced and have all the Jenkins tests pass.

Here's the successful Mac Python 3.6 run: http://172.16.20.10:8080/view/Python%20Connector/job/BuildPyConnector-Mac/194/ and the successful Linux 3.6 is running right now.

### DIFF
--- a/ci/build_linux.sh
+++ b/ci/build_linux.sh
@@ -41,7 +41,7 @@ for PYTHON_VERSION in ${PYTHON_VERSIONS}; do
     # Use new PEP-517 build
     ${PYTHON} -m pip wheel -w ${BUILD_DIR} --no-deps .
     # On Linux we should repair wheel(s) generated
-    auditwheel repair --plat manylinux2010_x86_64 -L connector ${BUILD_DIR}/*.whl -w ${REPAIRED_DIR}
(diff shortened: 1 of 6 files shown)


sfc-gh-stakeda (approved):
manylinux2010 makes sense. Maybe manylinux2014 use is too early.

sfc-gh-mkeller:
No, nevermind. Apparently I only fixed the issue locally even though the tests run inside of Docker

sfc-gh-mkeller:
After a lot of testing it seems like the hanging was because of `tempfile`, changing that to a `pytest` fixture fixed the hanging. The exact cause is still unknown, I'll file a Jira to investigate this at a later date, SNOW-285147.

When I was able to attach `gdb` to a Docker process it looked like we were waiting on a lock, or subprocess, there might be a deadlock inside of `tempfile` on Python 3.6 possibly? 👀

PR #627, opened 2021-02-16
SNOW-136583 adding multipart PUT threshold support

sfc-gh-mkeller:
Adds support for the upcoming `THRESHOLD` keyword in `PUT` statements. While adding this I found that file metadatas were these very dynamic dictionaries, so it was pretty complicated to see where they went when I added them. So I well-defined them with `dataclasses`.

I also did some smaller code cleanups, like updated logging messages, and changed around some class names to be PEP-8 compliant.

### DIFF
--- a/src/snowflake/connector/errors.py
+++ b/src/snowflake/connector/errors.py
@@ -10,8 +10,6 @@ import traceback
 from logging import getLogger
 from typing import TYPE_CHECKING, Dict, Optional, Type, Union
 
-from snowflake.connector.constants import UTF8
-
 from .compat import BASE_EXCEPTION_CLASS
 from .description import CLIENT_NAME, SNOWFLAKE_CONNECTOR_VERSION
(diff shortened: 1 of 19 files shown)


sfc-gh-stan (approved):
LGTM 👍

sfc-gh-kwagner (approved):

src/snowflake/connector/errors.py
```diff
@@ -236,10 +248,7 @@ def errorhandler_wrapper(connection: 'SnowflakeConnection',
             raise error_class(error_value)
 
 
-Error.__str__ = lambda self: self.__unicode__()
-
-
-class Warning(BASE_EXCEPTION_CLASS):
+class _Warning(BASE_EXCEPTION_CLASS):
```
asottile, line 251: this class rename is an api break, was this intentional?

PR #628, opened 2021-02-16
SNOW-270946 Progress percentage computation to handle the case when f…

sfc-gh-cshi:
…ile size is zero

### DIFF
--- a/src/snowflake/connector/file_transfer_agent.py
+++ b/src/snowflake/connector/file_transfer_agent.py
@@ -171,7 +171,12 @@ def _update_progress(
     return progress == 1.0
 
 
-class SnowflakeProgressPercentage:
+def percent(seen_so_far: int, size: float) -> float:
+    return 1.0 if seen_so_far >= size or size <= 0\
+                else float(seen_so_far / size)
+
+
+class SnowflakeProgressPercentage():
     """Built-in Progress bar for PUT commands."""
 
     def __init__(
@@ -208,7 +213,7 @@ class SnowflakeS3ProgressPercentage(SnowflakeProgressPercentage):
         with self._lock:
             if self._output_stream:
                 self._seen_so_far += bytes_amount
-                percentage = float(self._seen_so_far / self._size)
+                percentage = percent(self._seen_so_far, self._size)
                 if not self._done:
                     self._done = _update_progress(
                         self._filename, self._start_time,
@@ -234,7 +239,7 @@ class SnowflakeAzureProgressPercentage(SnowflakeProgressPercentage):
         with self._lock:
             if self._output_stream:
                 self._seen_so_far = current
-                percentage = float(self._seen_so_far / self._size)
+                percentage = percent(self._seen_so_far, self._size)
                 if not self._done:
                     self._done = _update_progress(
                         self._filename, self._start_time,
--- a/test/unit/test_put_get.py
+++ b/test/unit/test_put_get.py
@@ -11,7 +11,12 @@ from mock import MagicMock
 
 from snowflake.connector.compat import IS_WINDOWS
 from snowflake.connector.errors import Error
-from snowflake.connector.file_transfer_agent import SnowflakeFileTransferAgent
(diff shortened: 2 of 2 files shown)


sfc-gh-kwagner (approved):

test/unit/test_put_get.py
```diff
@@ -74,3 +78,16 @@ def test_put_error(tmpdir):
...
+    """Tests for raise_put_get_error flag (now turned on by default) in SnowflakeFileTransferAgent."""
+    tmp_dir = str(tmpdir.mkdir('putfiledir'))
+    file_name = path.join(tmp_dir, 'zero_file1')
+    from pathlib import Path
+    Path(file_name).touch()
+    func_callback = SnowflakeS3ProgressPercentage(file_name, 0)
+    func_callback(1)
+
+    func_callback = SnowflakeAzureProgressPercentage(file_name, 0)
+    func_callback(1)
```
sfc-gh-mkeller, line 93: Check out [tmp_path fixture](https://docs.pytest.org/en/stable/tmpdir.html#the-tmp-path-fixture), it'll cut out a bunch of your code. Also please be aware of our existing type hints. The percentage classes expect a `str` path.
```suggestion
def test_percentage(tmp_path):
    """Tests for raise_put_get_error flag (now turned on by default) in SnowflakeFileTransferAgent."""
    file = tmp_path / 'zero_file1'
    file.touch()
    func_callback = SnowflakeS3ProgressPercentage(str(file), 0)
    func_callback(1)

    func_callback = SnowflakeAzureProgressPercentage(str(file), 0)
    func_callback(1)
```

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -107,6 +107,11 @@ def _update_progress(
     return progress == 1.0
 
 
+def percent(seen_so_far, size):
```
sfc-gh-mkeller, line 110: Please verify these
```suggestion
def percent(seen_so_far:int , size: float) -> float:
```

sfc-gh-mkeller (changes requested):

test/unit/test_put_get.py
```diff
@@ -74,3 +79,19 @@ def test_put_error(tmpdir):
...
+def test_percentage(tmpdir):
+    """Tests for ProgressPercentage classes."""
+    assert 1.0 == percent(0, 0)
+    assert 1.0 == percent(20, 0)
+    assert 1.0 == percent(40, 20)
+    assert 0.5 == percent(14, 28)
+
+    file_path = tmpdir / 'zero_file1'
+    from pathlib import Path
+    Path(file_path).touch()
```
sfc-gh-mkeller, line 93: Please take a look at https://docs.pytest.org/en/stable/tmpdir.html `tmpdir` and `tmp_path` are not the same.
```suggestion
def test_percentage(tmp_path):
    """Tests for ProgressPercentage classes."""
    assert 1.0 == percent(0, 0)
    assert 1.0 == percent(20, 0)
    assert 1.0 == percent(40, 20)
    assert 0.5 == percent(14, 28)

    file_path = tmpdir / 'zero_file1'
```

sfc-gh-mkeller (changes requested):

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -208,7 +213,7 @@ def __call__(self, bytes_amount: int):
         with self._lock:
             if self._output_stream:
                 self._seen_so_far += bytes_amount
-                percentage = float(self._seen_so_far / self._size)
+                percentage = percent(self._seen_so_far, self._size)
```
sfc-gh-mkeller, line 216: `self._size` is initialized as `float(filesize)` so you are not respecting the type hints of `percent`

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):
🚢 🚢

PR #629, opened 2021-02-17
Add bug fix for SNOW-266042 Improve Query/Request ID logging

sfc-gh-stan:
The previous code changes raises an exception when client receives "log received" confirmation from telemetry because `ret['data']` is a str in this case, the exception is caught by upper level try clause.
This fix adds a type check to avoid the exception.


sfc-gh-mkeller (approved):
🚢 🚢

PR #631, opened 2021-02-19
Add ability for externalbrowser authentication mechanism to optionally use a UNIX pipe for SAML response

r-d-w:
Our Data Science team likes to use Jupyter notebooks inside of a locally running Docker container to access our Snowflake instance.  We also want to leverage SSO for all human based authentication to Snowflake.  This was not possible with the hard dependency in the existing code on accessing the local webbrowser (as the host is inaccessible within the docker context).

This PR adds the functionality to enable docker use with externalbrowser auth by providing an environment variable based flag to instruct the connector use a unix socket file that can be shared with another process to proxy host browser SAML responses back into the container.

Changes
- Added 3 environment variables, all optional. One enables the functionality and the other two override defaults
- Added logic to use unix socket instead of tcp socket when flag is set
- Added print of sso_url when using UNIX socket flag

None of the existing behavior is altered and all new code is behind optional flag.

Additional non-code bits we are using to make this work us:
- docker-compose (just the specific pieces to this)
```
version: '3.6'
services:
  analytics:
    environment:
      - SNOWFLAKE_EXTERNALBROWSER_USE_UNIX_SOCKET=true
      - SNOWFLAKE_EXTERNALBROWSER_UNIX_SOCKET=/var/run/snowflake/sock
    depends_on: 
      - "proxy"
  proxy:
    volumes: 
      - shared:/var/run/snowflake
    ports:
      - "4433:4433"
volumes:
  shared:
```
- proxy service Dockerfile
```
FROM nginx
COPY nginx.conf /etc/nginx/nginx.conf
```
- proxy service nginx.conf
```
user www-data;
include /etc/nginx/modules/*.conf;
events {
  worker_connections 1024;
}

http {
    access_log /dev/stdout main;
    error_log stderr;
    server {
        listen 4433 default_server;
        charset     utf-8;        
        location / {
            proxy_pass  http://snowflake;
        }
    }
    upstream snowflake {
        server unix:/var/run/snowflake/sock;
    }
}
```

### DIFF
--- a/src/snowflake/connector/result_batch.py
+++ b/src/snowflake/connector/result_batch.py
@@ -49,9 +49,22 @@ if TYPE_CHECKING:  # pragma: no cover
     from .vendored.requests import Response
 
 if installed_pandas:
-    from pyarrow import Table
+    from pyarrow import DataType, Table
+    from pyarrow import binary as pa_bin
+    from pyarrow import bool_ as pa_bool
+    from pyarrow import date64 as pa_date64
+    from pyarrow import field
+    from pyarrow import float64 as pa_flt64
+    from pyarrow import int64 as pa_int64
+    from pyarrow import schema
+    from pyarrow import string as pa_str
+    from pyarrow import time64 as pa_time64
+    from pyarrow import timestamp as pa_ts
 else:
-    Table = None
+    DataType, Table = None, None
+
+# emtpy pyarrow type array corresponding to FIELD_TYPES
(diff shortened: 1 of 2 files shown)


feluelle:
Hi @r-d-w,

Not sure if this is really needed. There is a PR out giving you the possibility to retrieve the token outside of docker environment and just passing back in the token you received when you authenticate through the browser on the host.

See: https://github.com/snowflakedb/snowflake-connector-python/pull/883

PR #633, opened 2021-02-22
SNOW-281305 implement bulk array binding

sfc-gh-stan:
[SNOW-281305](https://snowflakecomputing.atlassian.net/browse/SNOW-281305?atlOrigin=eyJpIjoiYzY4MTg1MTJkOTIwNDFjMGExZjQ4NTEzOGEyZTJmZDEiLCJwIjoiaiJ9)


src/snowflake/connector/cursor.py
```diff
@@ -436,7 +440,7 @@ def interrupt_handler(*_):  # pragma: no cover
 
     def execute(self,
                 command: str,
-                params: Union[List, Tuple, None] = None,
+                params: Union[List, Tuple, None, str] = None,
```
sfc-gh-mkeller, line 443: Split this into a private argument instead as per our conversation.
In that case will this work with `pyformat` too?

test/integ/test_bindings.py
```diff
@@ -287,33 +306,40 @@ def test_binding_bulk_insert(conn_cnx, db_parameters):
...
 drop table if exists {name}
 """.format(name=db_parameters['name']))
 
 
+@pytest.mark.skipolddriver
+def test_bulk_insert_binding_fallback(conn_cnx, db_parameters):
+    """When stage creation fails, bulk inserts falls back to server side binding and disables stage optimization."""
+    with conn_cnx(paramstyle='qmark') as cnx, cnx.cursor() as csr:
+        query = 'insert into {name}(c1,c2) values(?,?)'.format(
+            name=db_parameters['name'])
```
sfc-gh-mkeller, line 328: Use randomized name generator instead

test/unit/test_bind_upload_agent.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
```
sfc-gh-mkeller, line 2:
```suggestion
```

sfc-gh-mkeller (changes requested):

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -0,0 +1,64 @@
...
+        self.rows = rows
+        self._stream_buffer_size = stream_buffer_size
+        self.stage_path = f"@{_STAGE_NAME}/{uuid.uuid4().hex}"
+
+    def _create_stage(self):
+        self.cursor.execute(_CREATE_STAGE_STMT)
+
+    def upload(self):
+        try:
+            self._create_stage()
```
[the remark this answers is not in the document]
sfc-gh-stan: I separated this into two functions for testing, where I patch the stage creation to fail but does not wrap the upload logic.

src/snowflake/connector/cursor.py
sfc-gh-stan: I don't think so, `pyformat` always uses client side binding, I'm not sure if GS will parse the query correctly and do `COPY INTO` even if we upload bindings as files.

sfc-gh-mkeller (approved):
🚢 🚢

PR #634, opened 2021-02-23
SNOW 264606 client session keep alive fix

sfc-gh-hchaturvedi:
This is a fix for SNOW-264606. 

Up until now client_session_keep_alive client side parameter would only be communicated to the backend if it is set to True. The default value is False ergo if the value is set to False it is not communicated to the backend. This can confuse the  user. Users might want to configure the parameter to False or True at session level irrespective of the account / user level backend configuration. 

This fix updates the default value to None. If the user sets any non default value for the session it would be communicated to the backend and would be given precedence over the account or user level configuration.


src/snowflake/connector/connection.py
```diff
@@ -127,7 +127,7 @@ def DefaultConverterClass():
     'inject_client_pause': (0, int),  # snowflake internal
     'session_parameters': (None, (type(None), dict)),  # snowflake session parameters
     'autocommit': (None, (type(None), bool)),  # snowflake
-    'client_session_keep_alive': (False, bool),  # snowflake
+    'client_session_keep_alive': (None, bool),  # snowflake
```
sfc-gh-mkeller, line 130:
```suggestion
    'client_session_keep_alive': (None, (type(None), bool)),  # snowflake
```

src/snowflake/connector/connection.py
```diff
@@ -323,7 +323,10 @@ def client_session_keep_alive(self):
 
     @client_session_keep_alive.setter
     def client_session_keep_alive(self, value):
-        self._client_session_keep_alive = True if value else False
+        if value is None:
+            self._client_session_keep_alive = None
+        else:
+            self._client_session_keep_alive = value
```
sfc-gh-mkeller, line 329:
```suggestion
        self._client_session_keep_alive = value
```

test/integ/test_session_parameters.py
```diff
@@ -25,3 +25,72 @@ def test_session_parameters(db_parameters):
...
+        host=db_parameters['host'],
+        port=db_parameters['port'],
+        database=db_parameters['database'],
+        schema=db_parameters['schema'],
+        client_session_keep_alive=val
+    )
+    return connection
+
+
+def set_backend_client_session_keep_alive(db_parameters: object, val: bool) -> None:
```
sfc-gh-mkeller, line 80: I'd change that this function takes a connection from the outside, the less round trips we need to do to Snowflake the better

test/integ/test_session_parameters.py
```diff
@@ -25,3 +25,72 @@ def test_session_parameters(db_parameters):
...
+def test_client_session_keep_alive(db_parameters):
+    """Tests client_session_keep_alive setting.
+
+    Ensures that client's explicit config for client_session_keep_alive
+    session parameter is always honored and given higher precedence over
+    user and account level backend configuration.
+    """
+    # Ensure backend parameter is set to False
+    set_backend_client_session_keep_alive(db_parameters, False)
+    connection = create_client_connection(db_parameters, True)
```
sfc-gh-mkeller, line 39: I think this extra function is not necessary. Try something like this instead: `with conn_cnx(client_session_keep_alive=True) as connection:`
You'll need to add `conn_cnx` argument to the test and indent the next 3 lines, but this will also get rid of the `close` call. That fixture inserts the necessary data like username and then puts the extra arguments you give it on top of the default ones. You can follow it here: https://github.com/snowflakedb/snowflake-connector-python/blob/2b41eaa9ce6a57a3a46ad683b9a89fc7ce408b8e/test/integ/conftest.py#L221

src/snowflake/connector/connection.py
```diff
@@ -1103,9 +1107,16 @@ def _set_parameters(self, ret, session_parameters):
...
+                # Only set if the local config is None.
+                # Always give preference to user config.
+                if self.client_session_keep_alive is None:
+                    self.client_session_keep_alive = value
+                else:
+                    client_session_keep_alive_set = True
             elif PARAMETER_CLIENT_SESSION_KEEP_ALIVE_HEARTBEAT_FREQUENCY == \
-                    name:
+                    name and self.client_session_keep_alive is True\
+                    and client_session_keep_alive_set is False:
```
sfc-gh-mkeller, line 1118: What happens if this parameter is iterated through first and then the keepalive? Should we force this parameter to be set after client keepalive?

sfc-gh-mkeller (changes requested):

test/integ/test_session_parameters.py
sfc-gh-hchaturvedi: I tried this and it didn't work. The parameter isn't being communicated it seems.

test/integ/test_session_parameters.py
```diff
@@ -25,3 +32,75 @@ def test_session_parameters(db_parameters):
...
+        user=db_parameters['sf_user'],
+        password=db_parameters['sf_password'],
+        host=db_parameters['sf_host'],
+        port=db_parameters['sf_port']
+    )
+
+    # Ensure backend parameter is set to False
+    set_backend_client_session_keep_alive(db_parameters, admin_cnxn, False)
+    with conn_cnx(client_session_keep_alive=True) as connection:
+        connection = create_client_connection(db_parameters, True)
```
sfc-gh-mkeller, line 60: `connection` is already defined in the line above
```suggestion
```

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-hchaturvedi:
PR Head failing due to backend issue. All tests passed - merging.

PR #635, opened 2021-02-23
SNOW-216307 Thread safety for parallel upload

sfc-gh-cshi:
SNOW-216307

### DIFF
--- a/src/snowflake/connector/file_transfer_agent.py
+++ b/src/snowflake/connector/file_transfer_agent.py
@@ -14,10 +14,10 @@ import tempfile
 import threading
 from concurrent.futures.thread import ThreadPoolExecutor
 from logging import getLogger
+from queue import Queue
 from time import sleep, time
 from typing import IO, TYPE_CHECKING, Any, Dict, List, Optional, Type, Union
 
-import botocore.exceptions
 from boto3.session import Session
 
 from .compat import GET_CWD, IS_WINDOWS, dataclass, field
@@ -84,6 +84,14 @@ def result_fixed_column_desc(name):
 
 
 @dataclass
+class SFResourceMeta:
+    storage_util: Optional[Type[Union['SnowflakeRemoteStorageUtil', 'SnowflakeLocalUtil']]] = None
+    stage_info: Optional[Dict[str, Any]] = field(default_factory=dict)  # TODO could be strongly defined if need be
+    use_accelerate_endpoint: bool = False
+    cloud_client: Union['Session.resource', 'BlobServiceClient', str, None] = None
+
+
+@dataclass
 class SnowflakeFileMeta:
     """Class to keep track of information necessary for file operations."""
     name: str
@@ -91,7 +99,8 @@ class SnowflakeFileMeta:
     stage_location_type: str
     result_status: Optional['ResultStatus'] = None
 
-    client: Optional[Union['Session.resource', 'BlobServiceClient', Optional[str]]] = None
+    # client_meta holds cloud client object and the metadata needed to create such client
+    client_meta: 'SFResourceMeta' = SFResourceMeta(cloud_client=None)
     self: Optional['SnowflakeFileTransferAgent'] = None
     put_callback: Optional[Type['SnowflakeProgressPercentage']] = None
     put_azure_callback: Optional[Type['SnowflakeProgressPercentage']] = None
@@ -116,7 +125,6 @@ class SnowflakeFileMeta:
     gcs_file_header_content_length: Optional[int] = None
     gcs_file_header_encryption_metadata: Optional[Dict[str, Any]] = None
 
-    stage_info: Dict[str, Any] = field(default_factory=dict)  # TODO could be strongly defined if need be
     encryption_material: Optional['SnowflakeFileEncryptionMaterial'] = None
     # Specific to Uploads only
     src_file_size: int = 0
@@ -345,17 +353,15 @@ class SnowflakeFileTransferAgent(object):
         for result in self._results:
             result.result_status = result.result_status.value
 
-    def upload(self, large_file_metas: List['SnowflakeFileMeta'], small_file_metas: List[SnowflakeFileMeta]):
-        storage_client = SnowflakeFileTransferAgent.get_storage_client(
+    def upload(self, large_file_metas: List['SnowflakeFileMeta'], small_file_metas: List['SnowflakeFileMeta']):
+        storage_util = SnowflakeFileTransferAgent.get_storage_client(
             self._stage_location_type)
-        client = storage_client.create_client(
-            self._stage_info,
-            use_accelerate_endpoint=self._use_accelerate_endpoint
-        )
+        client_meta = SFResourceMeta(storage_util, self._stage_info, self._use_accelerate_endpoint)
+
         for meta in small_file_metas:
-            meta.client = client
+            meta.client_meta = client_meta
         for meta in large_file_metas:
-            meta.client = client
+            meta.client_meta = client_meta
 
         if len(small_file_metas) > 0:
             self._upload_files_in_parallel(small_file_metas)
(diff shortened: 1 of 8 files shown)


src/snowflake/connector/file_transfer_agent.py
```diff
@@ -348,14 +357,15 @@ def execute(self):
...
             self._stage_location_type)
-        client = storage_client.create_client(
-            self._stage_info,
-            use_accelerate_endpoint=self._use_accelerate_endpoint
-        )
+        client_meta = SFResourceMeta(storage_client, self._stage_info, self._use_accelerate_endpoint)
+        # client = storage_client.create_client(
+        #    self._stage_info,
+        #    use_accelerate_endpoint=self._use_accelerate_endpoint
+        # )
```
sfc-gh-mkeller, line 364: Now everything else will still use `client` except this parallel uploading (this uses `cln_meta`). Either change all of them, or non of them, the file meta class is already bulky

sfc-gh-mkeller: Also please don't leave commented out code there

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -388,59 +398,73 @@ def _transfer_accelerate_config(self):
                 'use_accelerate_endpoint: %s',
                 self._use_accelerate_endpoint)
 
-    def _upload_files_in_parallel(self, file_metas: List['SnowflakeFileMeta']) -> None:
+    @staticmethod
+    def _upload_files_in_queue_thread(qtask: Queue, qnext: Queue, results: list, triggers: list):
```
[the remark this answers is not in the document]
sfc-gh-mkeller:
```suggestion
    def _upload_files_in_queue_thread(qtask: Queue['SnowflakeFileMeta'], qnext: Queue['SnowflakeFileMeta'], results: List['SnowflakeFileMeta'], triggers: List['SnowflakeFileMeta']) -> None:
```
Why don't you use the return of this function for anything? One of the lists could be returned by the function.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -388,59 +398,73 @@ def _transfer_accelerate_config(self):
...
+                if result_meta.result_status == ResultStatus.RENEW_PRESIGNED_URL:
                     self._update_file_metas_with_presigned_url()
-                target_meta = retry_meta
 
-            if end_of_idx == len_file_metas:
-                break
-            idx += self._parallel
+            while not qnext.empty():
+                qtask.put(qnext.get())
+            retry_round = retry_round + 1
```
sfc-gh-mkeller, line 466:
```suggestion
            retry_round += 1
```

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -388,59 +398,73 @@ def _transfer_accelerate_config(self):
...
+            logger.debug("retry {} round for {} items".format(retry_round, qnext.qsize()))
```
sfc-gh-mkeller, line 467: I don't think that I have told you about f-strings yet, they are kinda cool.
This is not a hard requirement, but I want you  to know what this would look like as an f-string
```suggestion
            logger.debug(f"retry {retry_round} round for {qnext.qsize()} items")
```

sfc-gh-mkeller (changes requested):

src/snowflake/connector/file_transfer_agent.py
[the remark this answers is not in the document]
sfc-gh-cshi: > Put this as an internal function inside of `_upload_files_in_parallel` as it won't be used by any other function.

Noticed this comment did not get published. Seems threading has some problem using internal functions.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -367,80 +373,76 @@ def _transfer_accelerate_config(self):
...
-                    ret and 'Status' in ret and \
-                    ret['Status'] == 'Enabled'
-            except botocore.exceptions.ClientError as e:
-                if e.response['Error'].get('Code', 'Unknown') == \
-                        'AccessDenied':
-                    logger.debug(e)
-                else:
-                    # unknown error
-                    logger.debug(e, exc_info=True)
+            self._use_accelerate_endpoint = SnowflakeS3Util.transfer_accelerate_config(
```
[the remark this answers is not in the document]
sfc-gh-cshi: The removed code was using s3 SDK directly(get_bucket_accelerate_configuratio, botocore.exceptions. etc) . So here I moved them to S3Util to hide these S3 details from this layer. While we could add another layer like this "SnowflakeRemoteStorageUtil. transfer_accelerate_config  --> SnowflakeS3Util.transfer_accelerate_config" to totally move S3 dependency under SnowflakeRemoteStorageUtil, ideally we should move cloud specific config like 'transfer_accelerate_config' to S3 specific layer first. And I think we could do it in another task.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -367,80 +373,76 @@ def _transfer_accelerate_config(self):
...
-                        for result_meta in results]):
+        qtask = Queue()
+        for meta in file_metas:
+            qtask.put(meta)
+        retry_round = 0
+        while not qtask.empty():
+            len_file_metas = qtask.qsize()
+            thread_number = len_file_metas if len_file_metas < self._parallel else self._parallel
+            logger.debug(f'start {retry_round} round for {qtask.qsize()} items with {thread_number} threads')
+            pool = ThreadPoolExecutor(thread_number)
```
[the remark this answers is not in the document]
sfc-gh-cshi: we are reusing the threads, only we hit refresh_presigned, then we have to modify metas and restart

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -367,80 +373,76 @@ def _transfer_accelerate_config(self):
...
-                            file_metas[idx0].client = client
-                if any([result_meta.result_status == ResultStatus.RENEW_PRESIGNED_URL
-                        for result_meta in results]):
+        qtask = Queue()
+        for meta in file_metas:
+            qtask.put(meta)
+        retry_round = 0
+        while not qtask.empty():
+            len_file_metas = qtask.qsize()
+            thread_number = len_file_metas if len_file_metas < self._parallel else self._parallel
```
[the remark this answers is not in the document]
sfc-gh-cshi: calling function is considered higher cost in python. But here is not the critical path(code path run against each row, for example) meaning min(,) is more readable.

test/integ/test_put_get.py
```diff
@@ -63,7 +63,7 @@ def fin():
 
     class TestData(object):
         def __init__(self):
-            self.test_data_dir = (pathlib.Path(__file__).parent / 'data').absolute()
+            self.test_data_dir = (pathlib.Path(__file__).parent.parent / 'data').absolute()
```
sfc-gh-mkeller, line 66: Nice catch

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -367,80 +373,75 @@ def _transfer_accelerate_config(self):
...
+            if not triggers.empty():
                 break
-            idx += self._parallel
+
+            meta = qtask.get()
+            # initialize resource if it is not initialized yet
+            cln_meta = meta.client_meta
+            if thread_client is None:
+                thread_client = cln_meta.storage_util.create_client(cln_meta.stage_info,
+                                                                      cln_meta.use_accelerate_endpoint)
```
sfc-gh-mkeller, line 430: The indent seems to be wrong here, please fix that

sfc-gh-mkeller (changes requested):

src/snowflake/connector/file_transfer_agent.py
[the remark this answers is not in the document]
sfc-gh-cshi: pool can not schedule new future after shutdown().  So will start a new pool for next round. And since this only occur when errors happened, won't expect this to be regular and often during one PUT so the performance should be fine.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -367,80 +373,77 @@ def _transfer_accelerate_config(self):
...
+                if not presigned_url_handled and result_meta.result_status == ResultStatus.RENEW_PRESIGNED_URL:
                     self._update_file_metas_with_presigned_url()
-                target_meta = retry_meta
+                    presigned_url_handled = True
 
-            if end_of_idx == len_file_metas:
+            retry_round += 1
+
+    @staticmethod
+    def _upload_files_in_queue_thread(qtask: Queue, results: list, triggers: Queue):
```
sfc-gh-mkeller, line 411:
```suggestion
    def _upload_files_in_queue_thread(qtask: Queue['SnowflakeFileMeta`], results: List['SnowflakeFileMeta`], triggers: Queue['SnowflakeFileMeta`]):
```

sfc-gh-mkeller (approved):
Other than the one type issue it looks 🚢 🚢

PR #646, opened 2021-03-09
SNOW-298303 improve .pre-commit-config.yaml

asottile:
apologies for the drive-by, was looking at this repo for another reason

- pre-commit only runs on checked-in files, so I fixed `exclude`
- `seed-isort-config` is unnecessary for isort >= 5
- upgraded all the hooks
- `flake8` has its own pre-commit config now (and is removed from
  pre-commit/pre-commit-hooks)

you might also be interested in https://pre-commit.ci :)


sfc-gh-mkeller (approved):
Thank you so much, this looks good to me but let me make sure that tests complete successfully before merging this! 🚀

PR #653, opened 2021-03-13
SNOW-298604 making a flaky test more resilient

sfc-gh-mkeller:
As I describe in the test's comment:
```
This test tends to hang longer than expected when the testing warehouse is overloaded.
Manually looking at query history reveals that when a full GH actions + Jenkins test load hits one warehouse
it can be queued for 15 seconds, so for now we wait 5 seconds before checking and then we give it another 25
seconds to finish.
```


sfc-gh-stan (approved):
LGTM 👍

Issue #654, opened 2021-03-14
SNOW-298756: executemany - INSERT regex wrong


shahargl:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using (`python --version`)?
Python 3.7.3

2. What operating system and processor architecture are you using (`python -c 'import platform; print(platform.platform())'`)?
Darwin-19.5.0-x86_64-i386-64bit

3. What are the component versions in the environment (`pip freeze`)?

```
alembic==1.5.5
appnope==0.1.2
asn1crypto==1.4.0
attrs==20.3.0
azure-common==1.1.26
azure-core==1.11.0
azure-storage-blob==12.7.1
backcall==0.2.0
boto3==1.17.15
botocore==1.20.15
cachetools==4.2.1
certifi==2020.12.5
cffi==1.14.5
chardet==3.0.4
click==7.1.2
clickclick==20.10.2
connexion==2.7.0
cryptography==3.2.1
curlify==2.2.1
decorator==4.4.2
Flask==1.1.1
Flask-Cors==3.0.10
Flask-Testing==0.8.0
google-api-core==1.22.0
google-api-python-client==1.9.3
google-auth==1.27.0
google-auth-httplib2==0.0.4
google-cloud-core==1.6.0
google-cloud-kms==1.4.0
google-cloud-pubsub==2.4.0
google-cloud-scheduler==2.1.0
google-cloud-secret-manager==1.0.0
google-cloud-storage==1.29.0
google-cloud-tasks==1.5.0
google-resumable-media==0.5.1
googleapis-common-protos==1.52.0
graphql-core==3.1.3
greenlet==1.0.0
grpc-google-iam-v1==0.12.3
grpcio==1.35.0
gunicorn==20.0.4
httplib2==0.19.0
hypothesis==5.49.0
hypothesis-graphql==0.3.3
hypothesis-jsonschema==0.19.0
idna==2.8
importlib-metadata==1.7.0
inflection==0.5.1
iniconfig==1.1.1
ipython==7.21.0
ipython-genutils==0.2.0
isodate==0.6.0
itsdangerous==1.1.0
jedi==0.18.0
Jinja2==2.11.3
jmespath==0.10.0
jsonpatch==1.28
jsonpointer==2.0
jsonschema==3.2.0
junit-xml==1.9
jwcrypto==0.8
keyring==22.0.1
libcst==0.3.17
Mako==1.1.4
marisa-trie==0.7.5
[...]

PR #657, opened 2021-03-16
Fixing the regex pattern of fmt fetching in INSERT

shahargl:
Old behavior 
Buggy - fmt contains the second `)` which closes the `(SELECT`
```
In [46]: c = 'INSERT INTO "MESSAGES" (raw) (SELECT PARSE_JSON(column1) as raw from values (%(raw)s))'

In [47]: c = 'INSERT INTO "MESSAGES" (raw) (SELECT PARSE_JSON(column1) as raw from values (%(raw)s))'

In [48]: INSERT_SQL_VALUES_RE = re.compile(r'.*VALUES\s*(\(.*\)).*', re.IGNORECASE | re.MULTILINE | re.DOTALL)

In [49]: m = INSERT_SQL_VALUES_RE.match(c)

In [50]: m.group(1)
Out[50]: '(%(raw)s))'
```
**New behavior**
The fmt is correct
```
In [51]: INSERT_SQL_VALUES_RE = re.compile(r'.*VALUES\s*(\(\%\(.*\)s\)).*', re.IGNORECASE | re.MULTILINE | re.DOTALL)

In [52]: c = 'INSERT INTO "MESSAGES" (raw) (SELECT PARSE_JSON(column1) as raw from values (%(raw)s))'

In [53]: m = INSERT_SQL_VALUES_RE.match(c)

In [54]: m.group(1)
Out[54]: '(%(raw)s)'
```

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -134,7 +134,7 @@ class SnowflakeCursor(object):
     INSERT_SQL_RE = re.compile(r"^insert\s+into", flags=re.IGNORECASE)
     COMMENT_SQL_RE = re.compile(r"/\*.*\*/")
     INSERT_SQL_VALUES_RE = re.compile(
-        r".*VALUES\s*(\(.*\)).*", re.IGNORECASE | re.MULTILINE | re.DOTALL
+        r".*VALUES\s*(\(\%\(.*\)s\)).*", re.IGNORECASE | re.MULTILINE | re.DOTALL
     )
     ALTER_SESSION_RE = re.compile(
         r"alter\s+session\s+set\s+(.*)=\'?([^\']+)\'?\s*;",


sfc-gh-stan:
Hi @shahargl , thank you so much for submitting a PR and helping us improve this open source repo! I looked at the changes, while this fix works perfectly with the `%(VARNAME)s` convention as you noted, it overlooks some alternative conventions we want to support. 

For example, when the connection uses [ANSI C printf format](https://www.python.org/dev/peps/pep-0249/#paramstyle), this regex does not match queries like `INSERT INTO TEST (col1, col2) VALUES (%s, %d )`. Another case is partial binding and partial embedded queries like `INSERT INTO TEST (col1, col2) VALUES (2021, %(raw)s)`. 

In [SNOW-298756](https://github.com/snowflakedb/snowflake-connector-python/issues/654), you mentioned that the Regex should not be greedy, which leads me to the idea of matching balanced parentheses after the word "VALUES". I first thought of using recursive Regex to match a single pair of parentheses with any text in between, as long as the text has balanced parentheses. However, this will break for case like `INSERT INTO TEST (col1, col2) VALUES ('))((((, '%s )`. So we decided to write a small parsing function that only counts parentheses outside of a string, I opened a PR in #659 , please take a look and let me know what you think about this approach 😃 !

shahargl:
@sfc-gh-stan 
Thanks for the heads up! I'll get to it later and I'll review it for sure!
Anyway, I'm glad that you've put that effort into this!

PR #659, opened 2021-03-17
Add parsing logic to match VALUES clause in nested insertion queries

sfc-gh-stan:
This is a fix for [SNOW-299744](https://snowflakecomputing.atlassian.net/browse/SNOW-299744), where we extract "a single pair of parentheses with any text in between, as long as they contain balanced parenthesis" from the `VALUES clause ` (i.e. ` (<possibly_nested_subqueries>)` in ` INSERT INTO <table> VALUES (<possibly_nested_subqueries>)`). The parser function counts `(` and `)` outside of single quote strings and returns the first balanced substring it finds, i.e. Input: `(%s, %d, '))((' )()`, Output: `(%s, %d, '))((' )`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -218,7 +218,6 @@ setup(
         "idna>=2.5,<3",
         "certifi>=2017.4.17",
         'dataclasses<1.0;python_version=="3.6"',
-        "regex",
     ],
     namespace_packages=["snowflake"],
     packages=[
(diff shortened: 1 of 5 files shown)


sfc-gh-mkeller (changes requested):
I have a fun counter example:
```
INSERT INTO table FROM VALUES('complicated string', $$
multi line string
$$)
```

Unfortunately this won't work with this code, other than multi line strings please address my other comments anyway.
By chance it looks like I'm going to need to write a tokenizer, that would be really helpful for this PR.
Let's put this on the back-burner until then!

PR #664, opened 2021-03-25
reverted SNOW-264606 client session keep alive fix

sfc-gh-hchaturvedi:
This reverts the client session keep alive fix.  This will be put back in after the preannouncement.


sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-mkeller:
Reverts #634 for easier housekeeping

PR #668, opened 2021-03-29
Mkeller snow 302368 number to decimal conversion arrow

sfc-gh-stan:
**TODO**


sfc-gh-mkeller:
Going to re-open this from my account, so @sfc-gh-stan could review

PR #673, opened 2021-04-07
SNOW-302368 new Arrow NUMBER to Decimal converter option

sfc-gh-mkeller:
SNOW-302368
This PR adds a way to turn `NUMBER` columns into `decimal.Decimal` when retrieved to keep infinite precision while using Arrow. When compared to #668 I needed to turn the function argument into a connection parameter, because of the case when Arrow is used automatically by the back-end then there's no way to pass an argument to the `__next__` function of the iterator through the results.

While I don't exactly like putting all options as connection parameters I think we should rework our setting system to have most of them configurable on all of the following levels:
* connection
* cursor
* statement (execution)

I have filed myself the following ticket SNOW-302368 to come up with a good system for adding and handling these settings on all 3 of these levels at the same time.

### DIFF
--- a/test/integ/pandas/test_arrow_pandas.py
+++ b/test/integ/pandas/test_arrow_pandas.py
@@ -3,13 +3,14 @@
 #
 # Copyright (c) 2012-2021 Snowflake Computing Inc. All right reserved.
 #
-
+import decimal
 import itertools
 import random
 import time
 from datetime import datetime
 from decimal import Decimal
 
+import numpy
 import pytest
 
 try:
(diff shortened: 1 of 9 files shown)


sfc-gh-stan (approved):
LGTM 🚢 🚢

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -908,3 +909,45 @@ def test_query_resultscan_combos(conn_cnx, query_format, resultscan_format):
         if isinstance(scanned_results, pandas.DataFrame):
             scanned_results = [tuple(e) for e in scanned_results.values.tolist()]
         assert results == scanned_results
+
+
+@pytest.mark.parametrize(
```
sfc-gh-stan, line 914: Alternative:
Parameterize `use_decimal` and `expected_type` together instead of an if-else case with two asserts.
e.g.
```
@pytest.mark.parametrize("use_decimal,expected", [(False, numpy.float64), (True, decimal.Decimal)])
def test_number_fetchall_retrieve_type(conn_cnx, use_decimal, expected):
```

sfc-gh-mkeller: Good idea! I'll make this change 👍

PR #677, opened 2021-04-13
SNOW-303944 Add utility function to python connector to retrieve JWT token from privatekey file and pwd

sfc-gh-mknister:
SNOW-303944
This change creates a static function **get_token_from_private_key()** that can be called by Snowsql to get a jwt token.


src/snowflake/connector/auth_keypair.py
```diff
@@ -45,6 +44,10 @@ def __init__(self, private_key):
         self._private_key = private_key
         self._jwt_token = ""
         self._jwt_token_exp = 0
+        self._lifetime = timedelta(seconds=120)
```
sfc-gh-mkeller, line 47: I'd rather make this an argument for the `__init__` function. Could we do that instead?
It could be something like:
```
def __init__(self, private_key, lifetime_in_seconds: int = 120):
...
self._lifetime = timedelta(seconds=lifetime_in_seconds)
```

src/snowflake/connector/auth_keypair.py
```diff
@@ -45,6 +44,10 @@ def __init__(self, private_key):
...
+
+    def change_token_lifetime(self, lifetime_in_seconds):
```
sfc-gh-mkeller, line 49: Please use type hints for every function that you touch. [More info](https://docs.python.org/3/library/typing.html)
For this function it should be like:
```suggestion
    def change_token_lifetime(self, lifetime_in_seconds: int) -> None:
```

src/snowflake/connector/auth.py
```diff
@@ -463,6 +472,25 @@ def read_temporary_credentials(self, host, user, session_parameters):
                 host, user, MFA_TOKEN
             )
 
+    @staticmethod
+    def get_token_from_private_key(user, account, privatekey_path, key_password):
```
sfc-gh-mkeller, line 476: If this function doesn't use the class itself please don't make it a static function.
In Python not everything has to be in a class and if you wanted to use it from the outside it'd still be `snowflake.connector.auth.get_token_from_pricate_key`
Also add type hints please!

src/snowflake/connector/auth.py
```diff
@@ -463,6 +472,25 @@ def read_temporary_credentials(self, host, user, session_parameters):
...
+        encoded_password = key_password.encode() if key_password is not None else None
```
sfc-gh-mkeller, line 477: Type hints could help with this. If you declare `def get_token_from_private_key(user, account, privatekey_path, key_password: str):` then that says that `key_password` should not be None ever.

sfc-gh-mkeller (dismissed):

src/snowflake/connector/auth_keypair.py
sfc-gh-mknister: Makes sense! Added

src/snowflake/connector/auth_keypair.py
sfc-gh-mknister: Removed this function entirely since now it's set in the constructor for KeyPairAuth

src/snowflake/connector/auth.py
sfc-gh-mknister: Removed- now it's just at the bottom of the class but not using the class object

src/snowflake/connector/auth.py
sfc-gh-mknister: I added type hints for user, account, and privatekey_path, but it's okay if key_password is None so I didn't add the type hint here. This could happen if the user has an unencrypted private key file and so doesn't set the environment variable SNOWSQL_PRIVATE_KEY_PASSPHRASE, which will then be returned as None.

sfc-gh-mkeller: The right type hint for a nullable string is `Optional[str]`.

src/snowflake/connector/auth_keypair.py
```diff
@@ -29,14 +29,13 @@
...
 
-    LIFETIME = timedelta(seconds=120)
     ALGORITHM = "RS256"
     ISSUER = "iss"
     SUBJECT = "sub"
     EXPIRE_TIME = "exp"
     ISSUE_TIME = "iat"
 
-    def __init__(self, private_key):
+    def __init__(self, private_key, lifetime_in_seconds: int = 120):
```
sfc-gh-mkeller, line 38: Add a docstring description for `lifetime_in_seconds` please

sfc-gh-mkeller (dismissed):

src/snowflake/connector/auth.py
sfc-gh-mknister: Added!

src/snowflake/connector/auth.py
```diff
@@ -664,3 +673,23 @@ def build_temporary_credential_name(host, user, cred_type):
...
+        p_key = load_pem_private_key(
+            key.read(), password=encoded_password, backend=default_backend()
+        )
+
+    private_key = p_key.private_bytes(
+        encoding=Encoding.DER,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=NoEncryption(),
+    )
+    auth_instance = AuthByKeyPair(private_key, 1440 * 60)
```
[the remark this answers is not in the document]
sfc-gh-mknister: The precommit hook called 'black' caused this to fail because I think comments inside the parameter list is not acceptable, but I added a comment describing the behavior above this line.

src/snowflake/connector/auth_keypair.py
```diff
@@ -29,24 +29,25 @@
...
```
sfc-gh-stakeda, line 38:
```suggestion
    def __init__(self, private_key, lifetime_in_seconds: int = self.LIFETIME):
```

src/snowflake/connector/auth.py
```diff
@@ -664,3 +673,24 @@ def build_temporary_credential_name(host, user, cred_type):
...
+        )
+
+    private_key = p_key.private_bytes(
+        encoding=Encoding.DER,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=NoEncryption(),
+    )
+    # Create new AuthByKeyPair object and set its token validity to be 24 hours (1440 min * 60)
+    auth_instance = AuthByKeyPair(private_key, 1440 * 60)
+    return auth_instance.authenticate(
```
sfc-gh-stakeda, line 694: `authenticate` method should do the following:
```
if not '.global' in account:
    idx = account.find('.')
    if idx > 0:
        account = account[0:idx]
else:
    idx = account.find('-')
    if idx > 0:
        account = account[0:idx]
account = account.upper().replace('-', '_')
```

sfc-gh-stakeda: Please add the test to cover all cases.

src/snowflake/connector/auth.py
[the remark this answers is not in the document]
sfc-gh-mkeller: `black` will reformat your files, just run it again with the same commit message, or you can actually manually run `black` with `tox -e fix_lint` once you install `tox` with `python3 -m pip install tox`.

src/snowflake/connector/auth.py
```diff
@@ -664,3 +673,24 @@ def build_temporary_credential_name(host, user, cred_type):
     return "{host}:{user}:{driver}:{cred}".format(
         host=host.upper(), user=user.upper(), driver=KEYRING_DRIVER_NAME, cred=cred_type
     )
+
+
+def get_token_from_private_key(
```
sfc-gh-stakeda, line 678: Can you add tests for this method?

src/snowflake/connector/auth.py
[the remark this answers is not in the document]
sfc-gh-mknister: changed

src/snowflake/connector/auth_keypair.py
sfc-gh-mknister: changed

src/snowflake/connector/auth.py
sfc-gh-mknister: Added this logic, and new test case has array of accounts that go through these cases

sfc-gh-mknister: test added in test_key_pair_authentication.py

src/snowflake/connector/auth.py
sfc-gh-mknister: test added in test_key_pair_authentication.py

test/integ/test_key_pair_authentication.py
```diff
@@ -5,13 +5,84 @@
...
+    load_pem_private_key,
+)
 
 import snowflake.connector
+from snowflake.connector.auth import get_token_from_private_key
+from snowflake.connector.auth_keypair import AuthByKeyPair
+
+
+@pytest.mark.skipolddriver
+def test_get_token_from_private_key():
```
sfc-gh-mknister, line 32: This test case generates a jwt token using the new utility function.
It then decodes the token and asserts that the fields (iss, sub, exp) contain the correct values.

I do not see a way to make a python connection with the token without changing a lot of the python connector logic, which seems too risky just for a test case. However, manual testing of the tokens using the sqlapi interface shows that they work. The test cases for making a connection with the token will be in the t_sqlapi regression test library.

sfc-gh-stakeda (approved):
LGTM

sfc-gh-stakeda:
@sfc-gh-mknister , please get an approval from @sfc-gh-mkeller before merge.

test/integ/test_key_pair_authentication.py
```diff
@@ -14,6 +17,46 @@
...
+    public_key_fingerprint = snowflake.connector.auth.get_public_key_fingerprint(
+        private_key_file_path, private_key_password
+    )
+    # Create a dict containing input accounts and their corrected account fields used to generate the correct token
+    accounts = {
+        "s3testaccount.global": "S3TESTACCOUNT.GLOBAL",
+        "acct-with-dashes": "ACCT_WITH_DASHES",
+        "testaccount.extra": "TESTACCOUNT",
+    }
+    for a in accounts:
```
sfc-gh-mkeller, line 37: You should use [parametrized tests](https://docs.pytest.org/en/stable/parametrize.html). We use them all over the code-base.

test/integ/test_key_pair_authentication.py
```diff
@@ -14,6 +17,46 @@
 import snowflake.connector
 
 
+@pytest.mark.skipolddriver
+def test_get_token_from_private_key():
+    test_user = "python_test_keypair_user_" + str(uuid.uuid4()).replace("-", "_")
+    current_dir = path.dirname(path.realpath(__file__))
+    private_key_file_path = path.join(
+        current_dir, "../data/rsa_keys", "rsa_key_encrypted.p8"
```
sfc-gh-mkeller, line 25: The separator embedded like that won't work on Windows
```suggestion
        current_dir, "..", "data", "rsa_keys", "rsa_key_encrypted.p8"
```

sfc-gh-mkeller (changes requested):

sfc-gh-stan (approved):
LGTM 👍

test/integ/test_key_pair_authentication.py
```diff
@@ -14,6 +17,53 @@
...
+    decoded_token = jwt.decode(jwt_token, options={"verify_signature": False})
+    # Assert "sub" field matches {corrected account}.{user}
+    assert expected_account + "." + test_user.upper() == decoded_token.get("sub")
+    # Assert "iss" field matches {corrected account}.{user}.{public key fingerprint}
+    assert (
+        expected_account
+        + "."
+        + test_user.upper()
+        + "."
+        + public_key_fingerprint.upper()
```
sfc-gh-mkeller, line 55: This is more of an FYI, don't do a new commit for this only.
There are these cool new features called [f-string](https://www.python.org/dev/peps/pep-0498/)
```suggestion
        f"{expected_account}.{test_user.upper()}.{public_key_fingerprint.upper()}"
```

sfc-gh-mkeller (approved):
🚢 🚢

PR #682, opened 2021-04-20
Uses s3 regional URL in private links when a param is set

sfc-gh-vreddy:
When the parameter ENABLE_STAGE_S3_PRIVATELINK_FOR_US_EAST_1 is set to true then invoke a different API to generate S3 regional url. 

The parameter will be set by CSE (And not customer) on AWS private link us-east-1 deployments, so that private link customers can use the s3 regional URL to upload files.

To reviewers:
The default value will be false for this parameter
This parameter has no affect on AZURE, GCP and non us-east-1 aws deployments.
The test case SHOULD run only on dev/reg environments


test/integ/test_put_get_user_stage.py
```diff
@@ -60,6 +60,63 @@ def test_put_get_large_data_via_user_stage(
...
+    number_of_files=1,
+    number_of_lines=1,
+    from_path=True,
+):
+    try:
+        with conn_cnx(
+            user=db_parameters["user"],
+            account=db_parameters["account"],
+            password=db_parameters["password"],
+            role="accountadmin",
```
sfc-gh-mkeller, line 100: You only need to give `conn_cnx` what you want it to overwrite on top of the defaults
```suggestion
            role="accountadmin",
```

test/integ/test_put_get_user_stage.py
```diff
@@ -60,6 +60,63 @@ def test_put_get_large_data_via_user_stage(
...
+    "from_path", [True, pytest.param(False, marks=pytest.mark.skipolddriver)]
+)
+def test_put_small_data_use_s3_regional_url(
+    is_public_test, tmpdir, conn_cnx, db_parameters, from_path
+):
+    """[s3] Puts Small Data via User Stage using regional url."""
+    if is_public_test:
+        pytest.skip(
+            "This test requires to change the internal parameter and requires account admin privileges"
+        )
```
sfc-gh-mkeller, line 74: We should have a `mark` named `internal` that does this. I know that the code has both of these. I think I'd prefer to keep the `internal` tag instead of this if statement being copied around.

test/integ/test_put_get_user_stage.py
```diff
@@ -60,6 +60,63 @@ def test_put_get_large_data_via_user_stage(
...
+
+def _put_get_user_stage_s3_regional_url(
+    tmpdir,
+    conn_cnx,
+    db_parameters,
+    number_of_files=1,
+    number_of_lines=1,
+    from_path=True,
+):
+    try:
```
sfc-gh-mkeller, line 95: This try should be after the first execute completes

test/integ/test_put_get_user_stage.py
```diff
@@ -60,6 +60,63 @@ def test_put_get_large_data_via_user_stage(
...
+            )
+        _put_get_user_stage(
+            tmpdir, conn_cnx, db_parameters, number_of_files, number_of_lines, from_path
+        )
+    finally:
+        with conn_cnx(
+            user=db_parameters["user"],
+            account=db_parameters["account"],
+            password=db_parameters["password"],
+            role="accountadmin",
```
sfc-gh-mkeller, line 113:
```suggestion
            role="accountadmin",
```

src/snowflake/connector/s3_util.py
```diff
@@ -53,7 +53,9 @@ class SnowflakeS3Util:
     """S3 Utility class."""
 
     @staticmethod
-    def create_client(stage_info, use_accelerate_endpoint=False) -> Session.resource:
+    def create_client(
+        stage_info, use_accelerate_endpoint=False, use_s3_regional_url=False
```
sfc-gh-mkeller, line 57: I don't know exactly what `stage_info` is could you please verify and add that too?
```suggestion
        stage_info, use_accelerate_endpoint: bool = False, use_s3_regional_url: bool = False
```

src/snowflake/connector/remote_storage_util.py
```diff
@@ -58,12 +58,16 @@ def get_for_storage_type(_type):
             return None
 
     @staticmethod
-    def create_client(stage_info, use_accelerate_endpoint=False):
+    def create_client(
+        stage_info, use_accelerate_endpoint=False, use_s3_regional_url=False
```
sfc-gh-mkeller, line 62: Type hints here, please!

src/snowflake/connector/connection.py
```diff
@@ -186,6 +187,10 @@ def DefaultConverterClass():
     ),  # only use openssl instead of python only crypto modules
     # whether to convert Arrow number values to decimal instead of doubles
     "arrow_number_to_decimal": (False, bool),
+    "enable_stage_s3_privatelink_for_us_east_1": (
```
sfc-gh-mkeller, line 190: Add getter and maybe a setter for `enable_stage_s3_privatelink_for_us_east_1` to return `_enable_stage_s3_privatelink_for_us_east_1`

src/snowflake/connector/connection.py
```diff
@@ -251,6 +257,7 @@ def __init__(self, **kwargs):
         self._done_async_sfqids = set()
         self.telemetry_enabled = False
         self._session_parameters: Dict[str, Union[str, int, bool]] = {}
+        self.enable_stage_s3_privatelink_for_us_east_1 = False
```
sfc-gh-mkeller, line 260: This is gonna be unnecessary once you add the getter for `enable_stage_s3_privatelink_for_us_east_1`

sfc-gh-mkeller (changes requested):

sfc-gh-stan:
Seems like the branch name has the wrong JIRA ticket number, should be SNOW-334404 instead.

src/snowflake/connector/s3_util.py
```diff
@@ -71,6 +73,11 @@ def create_client(stage_info, use_accelerate_endpoint=False) -> Session.resource
             ("https://" + stage_info["endPoint"]) if stage_info["endPoint"] else None
         )
 
+        # If FIPS endpoint is in us-east-1 this will still work as we are testing for stage_info['endPoint'] which
+        # is populated only in the FIPS gov deployments.
+        if use_s3_regional_url and not stage_info["endPoint"]:
```
sfc-gh-stan, line 78: On line 89, we specify the `region_name` parameter in boto3 session creation, which will be used to construct url for every client connection like what the code does here, regardless of the value of `use_s3_regional_url`. I'm curious why we need an extra parameter here?

sfc-gh-stan (changes requested):

src/snowflake/connector/s3_util.py
sfc-gh-vreddy: @sfc-gh-stan This changes the way S3 url is generated by aws sdk. 
For us-east-1 regions the endpoint is built as stage.s3.amazonaws.com and this is a global URL for customers in private link environments this global URL cannot be resolved (or the customers do not want to use it). So if use_s3_regional_url is set to true then we make modification to the endpoint so that the generated S3 url is a regional URL which has region name in it and will be resolved locally.

sfc-gh-stan: I see, the local changes I have (to remove S3 SDK dependency) constructs all url with the region name if given, so I guess this parameter might be deprecated in the future?

sfc-gh-vreddy: Yeah, in that case this will be deprecated. But currently we need this for the private client customers.

sfc-gh-vreddy: @sfc-gh-stan If you are changing all puts regional URL then that might be a breaking change. As long as we publish the change and give notice to customers that should be fine

sfc-gh-stan: Could you explain why would that be a breaking change? I will approve this change and let's schedule a separate call to discuss that!

sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-stan (approved):
LGTM!

PR #688, opened 2021-04-24
SNOW-296993 reworking internals for result_set and result_chunks

sfc-gh-mkeller:
SNOW-296993

This PR adds picklable `ResultChunk`s to the project. As an outcome please see the following codesnippet:
```
with connect(**connection_config) as conn:
    with conn.cursor() as cur:
        cur.execute('select seq4() as n from table(generator(rowcount => 100000));')
		partitions = cur.get_result_partitions()

for chunk in partitions:
	for row in partitions:
		print(row)
```
Note that partitions don't need a `SnowflakeConnection` anymore to download and parse their own results, but they do need `snowflake-connector-python` installed. They can also be iterated over multiple times. When you call `iter(partition)` it downloads its data, parses it and returns an `iterator` over the final data. This way things can be garbage collected when possible while working nicely with `ResultSet`s.

`ResultSet`s provide a way for us to manage `ResultChunk`s they handle pre-fetching the 2nd-5th chunks of data, like how we used to pre-fetch data historically and it also provides us a way to collect metrics. Which are important, since a `ResultChunk` doesn't have a connection to Snowflake, while they collect their own numbers while working `ResultSet` is what actually send them to us when they finish iterating. Their one trick is their special chain function that allows used up `ResultChunk` iterators to fall out of usage to allow them to be garbage collected.

Review notes:
I'd start byreviewing `result_chunk.py`, then moving onto `result_set.py` those two files are most of the changes. The rest is just integrating them instead of `chunk_downloader.py`, `arrow_result.pyx`, `json_result.py`

Behavior changes:
* `cursor.rownumber` was incorrect before https://www.python.org/dev/peps/pep-0249/#rownumber and https://github.com/snowflakedb/snowflake-connector-python/blob/e6137a05d184ec095bd6be50ff1bbd4e131064f4/test/integ/test_cursor.py#L933 Moved to SNOW-349276

Things still needing to be done:
* [x] Review all docstrings
* [x] Review API names: Going to switch all name from `result` to `chunk` to be consistent with `fetch_pandas_chunk` API we already had
* [x] Review behavior change: Moving into own Jira
* [x] Figure out what's the issue with `codecov`: Nothing apparently. I must have missed the comment getting updated by the bot


src/snowflake/connector/result_batch.py
```diff
@@ -0,0 +1,549 @@
...
+        self, **kwargs
+    ) -> Union[
+        Iterator[Union[Dict, Exception]],
+        Iterator[Union[Tuple, Exception]],
+        Iterator[Table],
+    ]:
+        iter_unit = kwargs.pop("iter_unit", ROW_UNIT)
+        if self._local:
+            return self._from_data(self._data, iter_unit)
+        sleep_timer = 1
```
sfc-gh-cshi, line 518: next 30 lines are the same as the code in JsonResultBatch._download and might be able to use a common method

src/snowflake/connector/cursor.py
```diff
@@ -1106,11 +1121,18 @@ def wait_until_ready():
         self._sfqid = sfqid
         self._prefetch_hook = wait_until_ready
 
+    def get_result_batches(self) -> Optional[List["ResultBatch"]]:
```
sfc-gh-cshi, line 1124: This is the main exposed interface, let us have some comments here.

test/integ/test_cursor.py
```diff
@@ -929,7 +935,9 @@ def test_rownumber(conn):
     """Checks whether rownumber is returned as expected."""
     with conn() as cnx:
         with cnx.cursor() as cur:
-            assert cur.execute("select * from values (1), (2)").fetchone() == (1,)
+            assert cur.execute("select * from values (1), (2)")
+            assert cur.rownumber is None
```
sfc-gh-cshi, line 939: we want to test if the rownumber is None before fetchone?

test/integ/test_cursor.py
```diff
@@ -1195,3 +1203,74 @@ def test__log_telemetry_job_data(conn_cnx, caplog):
...
+        ("arrow", ArrowResultBatch),
+    ),
+)
+def test_resultbatch(
+    conn_cnx,
+    result_format,
+    expected_chunk_type,
+):
+    """This test checks the following things:
+    1. After executing a query can we pickle the result batches
```
sfc-gh-cshi, line 1222: Maybe there is one more thing to test or demo: pre_pickle does not contain the actual data, the lazy fetching happens as expected(when iterating)

test/integ/test_cursor.py
```diff
@@ -1195,3 +1203,65 @@ def test__log_telemetry_job_data(conn_cnx, caplog):
...
+
+@pytest.mark.skipolddriver(reason="new feature in v2.5.0")
+@pytest.mark.parametrize(
+    "result_format,expected_chunk_type",
+    (
+        ("json", JSONResultChunk),
+        ("arrow", ArrowResultChunk),
+    ),
+)
+def test_resultchunk_pickling(
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Sure, I'll add a new test for this

src/snowflake/connector/cursor.py
sfc-gh-mkeller: Oops, good catch

src/snowflake/connector/result_batch.py
```diff
@@ -0,0 +1,549 @@
...
+    def _download(
+        self, **kwargs
+    ) -> Union[Iterator[Union[Dict, Exception]], Iterator[Union[Tuple, Exception]]]:
+        if self._local:
+            return iter(self._data)
+        sleep_timer = 1
+        backoff = DecorrelateJitterBackoff(1, 16)
+        for retry in range(MAX_DOWNLOAD_RETRY):
+            try:
+                with TimerContextManager() as download_metric:
```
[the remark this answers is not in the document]
sfc-gh-mkeller: I do not think so, I don't think that's how it used to work before either

src/snowflake/connector/result_batch.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+def create_chunks_from_response(
+    cursor: "SnowflakeCursor",
+    _format: str,
+    data: Dict[str, Any],
+) -> List["ResultBatch"]:
+    column_converters: List[Tuple[str, "SnowflakeConverterType"]] = []
+    arrow_context: Optional["ArrowConverterContext"] = None
+    rowtypes = data["rowtype"]
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Sure, I can add explanation of this into `ResultSet` docstring

test/integ/test_cursor.py
sfc-gh-mkeller: Why not? It's the expected behavior as per https://www.python.org/dev/peps/pep-0249/#rownumber

sfc-gh-cshi: Nice!

test/integ/test_cursor.py
sfc-gh-mkeller: > pre_pickle does not contain the actual data

`ResultBatch` es never contain data (other than the embedded first chunk), not even after you fetch them once.

Let me see how hard testing the lazy fetching would be 👍 I can probably do some `Mock` magic to verify it automatically.

src/snowflake/connector/chunk_downloader.py
```diff
@@ -315,7 +272,7 @@ def get_timings(self):
         return self._timings
 
 
-class RawBinaryDataHandler:
+class RawBinaryDataHandler(ABC):
     """Abstract class being passed to network.py to handle raw binary data."""
 
     def to_iterator(self, raw_data_fd, download_time):
```
sfc-gh-cshi, line 278: Do you mean to use '@abstractmethod' for to_iterator with introducing 'ABC'?

sfc-gh-cshi (approved):
LGTM. Still some redundant code can be shared between Json and Arrow. Also the logic in chunk downloader is to prefetch ahead and this can be done in this result_set/result_batch implementation. We should monitor the performance difference related to this logic.

test/integ/test_cursor.py
sfc-gh-mkeller: > pre_pickle does not contain the actual data

A `ResultBatch` doesn't cache its results, because they way we process them can differ. For example see the code for `fetch_arrow_batches` and `fetchone` when the result is in Arrow format.

Added a test for checking lazy fetching correctness!

src/snowflake/connector/chunk_downloader.py
sfc-gh-mkeller: > LGTM. Still some redundant code can be shared between Json and Arrow. Also the logic in chunk downloader is to prefetch ahead and this can be done in this result_set/result_batch implementation. We should monitor the performance difference related to this logic.

I'm sorry I'm working through your comments one-by-one and that's the last thing that I left. I agree with you and am working on reducing the redundant code. Sorry, I maybe should have been more upfront with you and told you not to check the PR every time I push. I often do that to verify that I haven't broken any tests.

sfc-gh-mkeller: > Do you mean to use '@AbstractMethod' for to_iterator with introducing 'ABC'?

I'm sorry, I made this change in the very early days of me working on this. I'll roll it back. The while `chunk_downloader.py` is getting deleted.

sfc-gh-mkeller:
> We should monitor the performance difference related to this logic.

@sfc-gh-cshi so there's a good reason for this. As I mentioned in one of my previous comments in the case of iterating through an `ArrowResultBatch` and going through the logic that `fetch_pandas_batches` uses we only know at fetching time how we would like to process the downloaded data, but this decision also eliminated the need for having a non pre-fetching `ResultSet` implementation.

Also, we already have this logging in execute: https://github.com/snowflakedb/snowflake-connector-python/blob/6db699dff5bbc32f31c2bde51d6d51c2c7024555/src/snowflake/connector/cursor.py#L624
So we might be able to use the difference between those telemetry entries.

However, I'm still not sure whether this extra logging would be helpful. We're just moving the time it takes to complete pre-fetching from `execute` to the first `fetch*` function.

src/snowflake/connector/result_batch.py
sfc-gh-mkeller: I de-duplicated the code snippets

sfc-gh-cshi (approved):

sfc-gh-cshi:
Also the obsolete code in chunk_downloader.py, json_result.py, arrow_result.py and their correspondent unit tess can be removed too.

sfc-gh-mkeller:
> Also the obsolete code in chunk_downloader.py, json_result.py, arrow_result.py and their correspondent unit tess can be removed too.

In this PR? Okay, sure

sfc-gh-mkeller:
Our unit tests seem to be hanging for some reason on Jenkins, but I can reproduce the hanging locally. I'm gonna investigate what could be causing that before merging

PR #694, opened 2021-04-27
SNOW-349276 fix rownumber

sfc-gh-mkeller:
SNOW-34276
Fixing bug in `rownumber`.

As per https://www.python.org/dev/peps/pep-0249/#rownumber

> The next fetch operation will fetch the row indexed by .rownumber in that sequence.

So now if you take a look at: https://github.com/snowflakedb/snowflake-connector-python/blob/e6137a05d184ec095bd6be50ff1bbd4e131064f4/test/integ/test_cursor.py#L933

You can see that `rownumber` returns what we fetched in the previous fetch operation.

To be merged after #688

### DIFF
--- a/src/snowflake/connector/arrow_context.py
+++ b/src/snowflake/connector/arrow_context.py
@@ -8,6 +8,7 @@ import decimal
 import time
 from datetime import datetime, timedelta
 from logging import getLogger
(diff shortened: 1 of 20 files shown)


sfc-gh-mkeller:
Closing for now, will come back to this later

PR #696, opened 2021-04-29
SNOW-351142 explicitly set python version to 3.6 for fix_lint

sfc-gh-mkeller:
SNOW-351142

Hotfixing `fix_lint` step for GH actions, it also fixes a linting issue in `master`, but to see it skipping beforehand see this example: https://github.com/snowflakedb/snowflake-connector-python/runs/2460860144?check_suite_focus=true and the run on this PR


sfc-gh-stan (approved):
🚢 🚢

PR #699, opened 2021-05-03
SNOW-336959 Add telemetry for pandas fetch functions

sfc-gh-cshi:
To collect usage informaion for cursor.fetch_pandas_(all|batches) functions


src/snowflake/connector/cursor.py
```diff
@@ -849,6 +849,7 @@ def fetch_pandas_batches(self, **kwargs):
         self.check_can_use_pandas()
         if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
             raise NotSupportedError
+        self._log_telemetry_job_data(TelemetryField.PANDAS_FETCH_BATCHES, 1)
```
sfc-gh-kwagner, line 852: What does the 1 refer to in this case? Is it a count?

sfc-gh-mkeller: I used this before, I think I'm to blame for doing this originally.
The reason is that `_log_telemetry_job_data` does `int(value)`, so instead of doing `True` and expecting `True` in the telemetry data, I explicitly did the conversion, as
```pycon
>>> int(True)
1
```

sfc-gh-mkeller (changes requested):
Please add a unit/integration test

src/snowflake/connector/cursor.py
sfc-gh-kwagner: Can we turn this into a constant that represents a True value in telemetry

sfc-gh-cshi: Mark explained it. We just need occurrence of the function call for now, so a dummy '1' is fine.

sfc-gh-kwagner: We shouldn't have magic numbers. Someone down the line will have the same question that I did, and if we have a named constant its obvious what that 1 represents

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -567,6 +569,118 @@ def test_timestampltz(conn_cnx, scale, timezone):
...
+                        assert abs(c_case - c_new) < epsilon, (
+                            "{} row, {} column: original value is {}, "
+                            "new value is {}, epsilon is {} \
+                        values are not equal".format(
+                                i, j, cases[i], c_new, epsilon
+                            )
+                        )
+
+
 def validate_pandas(
```
sfc-gh-mkeller, line 684: Let's get rid of this function if it adds no value

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -591,99 +705,54 @@ def validate_pandas(
...
-                            )
-                        else:
-                            assert abs(c_case - c_new) < epsilon, (
-                                "{} row, {} column: original value is {}, "
-                                "new value is {}, epsilon is {} \
-                            values are not equal".format(
-                                    i, j, cases[i], c_new, epsilon
-                                )
-                            )
+def validate_pandas_with_telemetry(
```
sfc-gh-mkeller, line 727: This function doesn't really do anything either.
What if we had the `capture_sf_telemetry` in the test and added an assert in the test instead of passing things down to helper functions instead

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -984,3 +1053,40 @@ def test_resultbatches_pandas_functionality(conn_cnx):
...
+        conn_cnx,
+        capture_sf_telemetry,
+        sql_text,
+        cases,
+        1,
+        "batch",
+        telemetry_to_check=TelemetryField.PANDAS_FETCH_BATCHES,
+    )
+
+    finish(conn_cnx, table)
```
sfc-gh-mkeller, line 1092: I really like parametrized tests as it can tell us easily which one is failing when there are 2 tests that are super similar. Let me show you how it'd work in this case.
```suggestion
@pytest.mark.skipif(
    not installed_pandas or no_arrow_iterator_ext,
    reason="arrow_iterator extension is not built, or pandas is missing.",
)
@pytest.mark.parametrize("fetch_method,expected_telemetry_type",
    (
        ("one", TelemetryField.PANDAS_FETCH_ALL),
        ("batch", TelemetryField.PANDAS_FETCH_BATCHES),
    )
)
def test_pandas_telemetry(conn_cnx, capture_sf_telemetry, fetch_method, expected_telemetry_type):
    cases = ["NULL", 0.11, -0.11, "NULL", 1.27, -1.28, "NULL"]
    table = "test_pandas_telemetry"
    column = "(a number(5,2))"
    values = (
        "(" + "),(".join([f"{i}, {c}" for i, c in enumerate(cases)]) + ")"
    )
    init(conn_cnx, table, column, values)
    sql_text = f"select a from {table} order by s"

    validate_pandas_with_telemetry(
        conn_cnx,
        capture_sf_telemetry,
        sql_text,
        cases,
        1,
        fetch_method,
        telemetry_to_check=expected_telemetry_type,
    )

    finish(conn_cnx, table)
```

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -591,99 +705,54 @@ def validate_pandas(
         epsilon: For comparing double values (Default value = None).
         scale: For comparing time values with scale (Default value = 0).
         timezone: For comparing timestamp ltz (Default value = None).
+        telemetry_to_check: shall check if all strings in this array appear in telemetry data.
```
sfc-gh-mkeller, line 708: This is in the wrong spot

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -567,6 +569,118 @@ def test_timestampltz(conn_cnx, scale, timezone):
...
+        total_rows = df_new.shape[0]
+    else:
+        for df_new in cursor_table.fetch_pandas_batches():
+            total_rows += df_new.shape[0]
+            total_batches += 1
+
+    if telemetry_to_check is not None and telemetry_test is not None:
+        assert any(
+            t.message["type"] == telemetry_to_check for t in telemetry_test.records
+        )
```
sfc-gh-mkeller, line 607: I think this could live inside of the tests and then we wouldn't necessarily need the helper function refactor. Other than this change this helper function looks the same as `validate_pandas`.

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -567,6 +569,118 @@ def test_timestampltz(conn_cnx, scale, timezone):
...
+    end_time = time.time()
```
sfc-gh-mkeller, line 608: This used to be right after the fetch calls. Should we put this before looking through the telemetry records?

sfc-gh-mkeller (changes requested):
Please look into if we could do what @sfc-gh-kwagner asked for and reconsider the helper function refactoring

test/integ/pandas/test_arrow_pandas.py
sfc-gh-cshi: The reason to use a do_validate_pandas is, when we call to it, the connection and telemetry objects are ready. Before that, there is no good way to get these two object out of fixture context manager. So keep the code to load fixture outside in validate_pandas_with_telemetry and validate_pandas, and have the actual meat in do_validate_pandas I think is the clean way to use fixture.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-cshi: > I really like parametrized tests as it can tell us easily which one is failing when there are 2 tests that are super similar. Let me show you how it'd work in this case.

It is nice but with higher cost, e.g. call init and finish twice.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-cshi: I think this issue is one cons in using fixture. connection object was initialized not directly in test_** so the conn_cnx has to be passed into validate_pandas function where we get the actual connection object out of the fixture. Then telemetry can be initialized upon this connection. But if we want to use the telemetry object some where else far away from where we initialize these fixtures, we has to pass around. Like this one, we could do the assert outside in test, but the telemetry object initialized in validate_pandas also has to return to the caller of validate_pandas.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-cshi: end_time? Agree.

src/snowflake/connector/cursor.py
sfc-gh-cshi: I made this parameter optional so when there is no value to provide, callers do not have to explicitly provide dummy value.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-cshi: fixed

src/snowflake/connector/connection.py
```diff
@@ -439,6 +439,14 @@ def telemetry_enabled(self):
...
         self._telemetry_enabled = True if value else False
 
+    @property
+    def telemetry(self):
+        return self._telemetry
+
+    @telemetry.setter
+    def telemetry(self, value):
+        self._telemetry = value
+
```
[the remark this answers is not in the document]
sfc-gh-cshi: removed

test/integ/pandas/test_arrow_pandas.py
sfc-gh-mkeller: What's more important when running tests?

I'd say it's not the 5s that we save, but the error report being easier for the developers.

On the flip side, if I need to run the tests locally I can't run one without the other one.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-mkeller: I don't see how that's the fixture's fault, it sounds like to me that it's the helper function's fault. We could have the helper accept `SnowflakeConnection` instead of the `Callable[..., "SnowflakeConnection"]` that `conn_cnx` is and then use that connection in the helper function.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-mkeller: You could instead use the fixtures in the test itself and pass the created objects down into the helper functions instead of passing the fixtures down.

I think it'd be helpful for you to trace out what `conn_cnx` really does here https://github.com/snowflakedb/snowflake-connector-python/blob/641204933c4f7e125f3af064b67354199943d61c/test/integ/conftest.py#L221

The fixtures have nothing to do with anything once you're in a test, they pass in whatever the `fixture` function returns.

src/snowflake/connector/cursor.py
```diff
@@ -1054,9 +1056,13 @@ def __cancel_query(self, query):
...
                 self._connection._cancel_query(query, self._request_id)
 
-    def _log_telemetry_job_data(self, telemetry_field, value):
+    def _log_telemetry_job_data(self, telemetry_field, value=None):
         """Builds an instance of TelemetryData with the given field and logs it."""
-        obj = {"type": telemetry_field, "query_id": self._sfqid, "value": int(value)}
+        obj = {
+            "type": telemetry_field,
+            "query_id": self._sfqid,
+            "value": 1 if value is None else int(value),
```
sfc-gh-mkeller, line 1064: I like Kris' idea more of introducing a constant variable `1` for `True` over this.
If you do it this way there's no way to report `False` (which we could also declare as a constant `0`).

test/integ/pandas/test_arrow_pandas.py
sfc-gh-cshi: Providing SnowflakeConnection to helper functions, e.g. init, finish, validate_pandas means unnecessarily rewrite all tests that call these these functions and put these calls under "with...". The reason to use help functions like what I am doing right now is to maintain the same validate_pandas for other supposedly untouched tests to use, then only the newly added test needs to look at telemetry data. So the added complexity is contained only in new code. I feel this is simpler. What do you think?

src/snowflake/connector/cursor.py
sfc-gh-cshi: int(False) is 0, int(True) is 1, these values will be recorded correctly if they are passed into the call. But when value does not matter(using default 'None'), then we just fill the value with something, and it is '1' in this case. I think defining a dummy constant is not necessary when a telemetry query on this type shall just simply ignore the 'value'.

sfc-gh-kwagner (approved):

PR #701, opened 2021-05-05
SNOW-296993 fixing a ResultBatch telemetry bug

sfc-gh-mkeller:
SNOW-296993

SnowSQL builds have been broken since Friday because the telemetry `_first_chunk_time` is not always recorded.
This PR skips reporting that Telemetry when it needs to and it also introduces a fixture that we can use to easily capture Telemetry messages during testing (it can also be used to disable telemetry by not propagating the messages to the real telemetry objects).

### DIFF
--- a/src/snowflake/connector/result_set.py
+++ b/src/snowflake/connector/result_set.py
@@ -99,10 +99,13 @@ class ResultSet(Iterable[List[Any]]):
         This includes TIME_CONSUME_LAST_RESULT, TIME_DOWNLOADING_CHUNKS and
         TIME_PARSING_CHUNKS in that order.
         """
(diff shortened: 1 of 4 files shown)


sfc-gh-kwagner (approved):
LGTM

sfc-gh-cshi (approved):
I think we still want to create a _first_chunk_time when there is none. I think this metrics will be missing too for distributed workers.

sfc-gh-mkeller:
> I think we still want to create a _first_chunk_time when there is none.

What do you suggest we should report then in the telemetry?

> I think this metrics will be missing too for distributed workers.

We knew this from the start of this project. I thought this was an accepted consequence, as you cannot send telemetry without having a connection to Snowflake.

PR #716, opened 2021-05-13
SNOW-303685 fixing segfault when using DictCursor and arrow result format

sfc-gh-mkeller:
SNOW-303685
Fixing segfault error when using `DictCursor` with Arrow result format

### DIFF
--- a/src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
@@ -442,8 +442,11 @@ void DictCArrowChunkIterator::createRowPyObject()
   for (int i = 0; i < m_currentSchema->num_fields(); i++)
   {
     PyObject* value = m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch);
-    PyDict_SetItemString(m_latestReturnedRow.get(), m_currentSchema->field(i)->name().c_str(), value);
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
```diff
@@ -442,8 +442,11 @@ void DictCArrowChunkIterator::createRowPyObject()
   for (int i = 0; i < m_currentSchema->num_fields(); i++)
   {
     PyObject* value = m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch);
-    PyDict_SetItemString(m_latestReturnedRow.get(), m_currentSchema->field(i)->name().c_str(), value);
```
sfc-gh-mkeller, line 445: In the previous line `toPyObject` returns `null` when an exception has been encountered.
In which case `PyDict_SetItemString` segfaults

sfc-gh-kzaveri (approved):
LGTM. One option for the if statement is maybe doing `if(value)` but fine either way, your way is more clear. 👍

sfc-gh-mkeller:
> LGTM. One option for the if statement is maybe doing `if(value)` but fine either way, your way is more clear. 👍

Ahh, good point! I think I still slightly prefer this one (because of being explicit), but I'm open to changing it if the others think that we should.

sfc-gh-cshi (approved):

PR #718, opened 2021-05-14
SNOW-343175 capture more azure library's exception(cover ServiceRespo…

sfc-gh-cshi:
…nseTimeout) and retry

Add mock test to throw the previously missing exception


src/snowflake/connector/azure_util.py
```diff
@@ -10,7 +10,12 @@
 from logging import getLogger
 from typing import TYPE_CHECKING, Any, Dict
 
-from azure.core.exceptions import HttpResponseError, ResourceNotFoundError
+from azure.core.exceptions import (
+    HttpResponseError,
+    ResourceNotFoundError,
+    ServiceRequestError,
+    ServiceResponseError,
```
sfc-gh-stakeda, line 17: Are these new exception modules available only newer azure package or the existing ones since the oldest supported package but were not handled before?

sfc-gh-stakeda: If they are new, you may need to bump up the minimum supported Azure package version.

PR #723, opened 2021-05-20
add method to check if arrow is part of the current query result

lababidi:
Currently `fetch_pandas_all()` fails if the query result isn't an arrow data. This new method allows a user to check if the data is available to be fetched using `fetch_pandas_all` without calling and failing. This allows a user to use `fetchall` if the result isn't arrow compatible.

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -808,6 +808,14 @@ class SnowflakeCursor(object):
                 },
             )
 
+    def arrow_avail(self):
+        """
+        Check to see if current query result is an arrow compatible and if pyarrow is installed.
+        Use to safely call fetch_pandas_X() methods
+        """
+        global pyarrow
+        return pyarrow is not None and self._query_result_format == "arrow"
+
     def query_result(self, qid):
         url = "/queries/{qid}/result".format(qid=qid)
         ret = self._connection.rest.request(url=url, method="get")


lababidi:
@sfc-gh-stakeda Hi, Could you please take a look at this? Let me know if it looks good

sfc-gh-mkeller:
Hi @lababidi 
I think what you are adding here already exist in our code base, please take a look at here: https://github.com/snowflakedb/snowflake-connector-python/blob/7f47adae03039b82d437f41715af20a11deeb8d3/src/snowflake/connector/options.py#L64

jdanielmyers:
Hey folks -- we are taking a look at this now and will have run through our internal workflows. Stay tuned!

lababidi:
@sfc-gh-mkeller Thank you! this is very similar. The only issue is that the method you showed is a private method. I want a public method to I can choose what `fetch` to use. Thank you

sfc-gh-mkeller:
> @sfc-gh-mkeller Thank you! this is very similar. The only issue is that the method you showed is a private method. I want a public method to I can choose what `fetch` to use. Thank you

What you want exactly is the public variable that function creates https://github.com/snowflakedb/snowflake-connector-python/blob/7f47adae03039b82d437f41715af20a11deeb8d3/src/snowflake/connector/options.py#L123

Just try importing `snowflake.connector.options.installed_pandas`, no methods necessary and it has been part of the code base for a while so you probably already have it :)

PR #724, opened 2021-05-21
SNOW-169178 Replace pointers with UniqueRefs

sfc-gh-kzaveri:
This PR addresses changes requested in [SNOW-169178](https://snowflakecomputing.atlassian.net/browse/SNOW-169178).

Replaces raw PyObject*s with UniqueRefs (a wrapper for PyObject*). When the local variable goes out of scope, the destructor of UniqueRef is called, ultimately calling XDECREF on the pointer.

Verified that memory leaks are not happening using sample code: https://gist.github.com/schitlurharidas/25b66e6220923597fb019bffaf6749a5

### DIFF
--- a/src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
@@ -441,11 +441,13 @@ void DictCArrowChunkIterator::createRowPyObject()
   m_latestReturnedRow.reset(PyDict_New());
   for (int i = 0; i < m_currentSchema->num_fields(); i++)
   {
-    PyObject* value = m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch);
-    if (value != nullptr)
+    py::UniqueRef value = py::UniqueRef(m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch));
+    if (!value.empty())
     {
-      PyDict_SetItemString(m_latestReturnedRow.get(), m_currentSchema->field(i)->name().c_str(), value);
-      Py_DECREF(value);
+      PyDict_SetItemString(
+          m_latestReturnedRow.get(),
+          m_currentSchema->field(i)->name().c_str(),
+          value.get());
     }
   }
   return;
--- a/src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
@@ -89,6 +89,7 @@ void CArrowChunkIterator::createRowPyObject()
   m_latestReturnedRow.reset(PyTuple_New(m_columnCount));
   for (int i = 0; i < m_columnCount; i++)
   {
+    // PyTuple_SET_ITEM steals a reference to the PyObject returned by toPyObject below
     PyTuple_SET_ITEM(
         m_latestReturnedRow.get(), i,
         m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch));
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
```diff
@@ -441,11 +441,13 @@ void DictCArrowChunkIterator::createRowPyObject()
...
+    py::UniqueRef value = py::UniqueRef(m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch));
+    if (!value.empty())
     {
-      PyDict_SetItemString(m_latestReturnedRow.get(), m_currentSchema->field(i)->name().c_str(), value);
-      Py_DECREF(value);
+      PyDict_SetItemString(
+          m_latestReturnedRow.get(),
+          m_currentSchema->field(i)->name().c_str(),
+          value.get());
     }
```
sfc-gh-kzaveri, line 451: `Py_DECRED` called when `value` goes out of scope

src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
```diff
@@ -441,11 +441,13 @@ void DictCArrowChunkIterator::createRowPyObject()
   m_latestReturnedRow.reset(PyDict_New());
   for (int i = 0; i < m_currentSchema->num_fields(); i++)
   {
-    PyObject* value = m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch);
-    if (value != nullptr)
+    py::UniqueRef value = py::UniqueRef(m_currentBatchConverters[i]->toPyObject(m_rowIndexInBatch));
+    if (!value.empty())
```
sfc-gh-kzaveri, line 445: using `empty` from `UniqueRef` class

src/snowflake/connector/cpp/ArrowIterator/CArrowTableIterator.cpp
```diff
@@ -168,9 +168,8 @@ m_context(context),
 m_pyTableObjRef(nullptr),
 m_convert_number_to_decimal(number_to_decimal)
 {
-  PyObject* tz = PyObject_GetAttrString(m_context, "_timezone");
-  PyArg_Parse(tz, "s", &m_timezone);
-  Py_XDECREF(tz);
```
sfc-gh-kzaveri, line 173: Another occurrence of Py_XDECREF that I found

src/snowflake/connector/cpp/ArrowIterator/CArrowChunkIterator.cpp
sfc-gh-cshi: should the change apply to all usage of toPyObject? E.g. in CArrowChunkIterator::createRowPyObject?

sfc-gh-kzaveri: Since [PyTuple_SET_ITEM](https://docs.python.org/3/c-api/tuple.html#c.PyTuple_SetItem) steals a reference, it is okay. In the case of _SetItemString it does not steal a reference and hence the need for DECREF'ing / UniqueRefs to properly manage the memory.

sfc-gh-mkeller (approved):
I still think `toPyObject` should return `py::UniqueRef` for safety reasons, but you're right that the impact with the 2 usages would be so minimal that doing is might not be worth it...
At the same time your comments will help us if we do decide to change them anyway one day.
Please don't forget to file a Jira to re-evaluate whether to make this change, or not and then let's revisit that separately in a month, or so.
🚢🔥

sfc-gh-kzaveri:
For future reference, the JIRA to re-evaluate changing `toPyObject`: https://snowflakecomputing.atlassian.net/browse/SNOW-363399

sfc-gh-kwagner (approved):
LGTM

sfc-gh-kzaveri:
A passing Jenkins build - https://ci70.int.snowflakecomputing.com/job/RT-PyConnector37-PC/844/ (Ankit helped me kick this off), there is a known issue being investigated: https://snowflakecomputing.atlassian.net/browse/SNOW-364651 that is causing the failing PR head

PR #725, opened 2021-05-22
SNOW-300951 removing cloud SDKs

### LINKED ISSUE #284: SNOW-141545: Request - Reduce package dependencies chaos

sfc-gh-mkeller:
SNOW-300951

# Introduction 

Back when we started on this project it was mainly to decrease the number of our dependencies and the overall size of snowflake-connector-python, as we can disregard all other capabilities other than blob storage.

Since the start we ran into a few other issues with the cloud SDKs, like some more exotic proxies not being supported by the cloud SDKs and not supporting asyncio.

# Design

## Initial design

Different blob storage services work mostly the same, but there are subtle differences. For example consider that on S3 we need to keep track of the block numbers and the returned etags for all the pieces for multipart upload, while on Azure we specify the certain block IDs when putting the pieces and then we send the list of those ids to an end-point, because of this we quickly arrived to a class design like this:

simplified sdkless class UML

While performing PUT and GET operation most of the time is spent on the network calls, so we knew from the start that we needed to perform the network call in parallel.
Previously our method to parallelize execution was to use `ThreadPoolExecutor`s (I’ll be referring to them as `tpe`s from now on), which already handles some concurrency hurdles for us, but during this project their limitations became obvious.

simple sdkless transfer flow

Notice that the idea was that the main thread would be doing most of the work and then it would let the `network_tpe` take care of network calls.

Once the implementation was semi complete we ran benchmark and found that performance suffered greatly at least on machines with limited computing power when compared to their network speed (such as cheaper EC2 machines).

As it turned out when only the network calls were parallelized compression became the most expensive operation. So after some back-and-forth we arrived to the final design.

## Final design

sdkless transfer flow(1)

The new idea in this iteration is that we have separate thread pools for CPU and network bound tasks.

When designing the flow of files we wanted to make sure that if a chunk needed to be retried the queue to the `network_tpe` wasn’t too long so we introduced a conditional variable where each file chunk will wait to submit it’s chunk to the network tpe until the queue’s length is over 2 * max_concurrency. Once all of the chunks of a file have been processed by the `network_tpe` during a download gets submitted to the `postprocess_tpe` to be decrypted. In case of an upload, the last chunk to finish in the network_tpe makes a second call to finalize the file on blob storage. Note that the only difference between `preprocess_tpe` and `postprocess_tpe` is that we did not want to pollute the queue to the network_tpe with tasks that we can basically guaranteed to work.

# Issues with ThreadPoolExecutors

1. **Callbacks might be executed synchronously:** This is a gotcha of the tpes, if a thread submits a job and if that job is done by the time the thread is attaching a callback onto the future, then the callback will be executed by the thread that is attaching the callback to the future. For this reason we introduced the fn function_and_callback_wrapper that makes sure that the callback is called concurrently from the tpe.

2. **Error handling in callbacks is non-existent in tpes:** We need to be careful how we handle errors inside of tpes. While calling `.result()` on a tpe raises the same exception that was raised in the callable that was executed there’s no such logic for Exceptions raised inside of callback function. For this reason we introduced manually track errors and handle them inside of the `SnowflakeFileTransferAgent.result`.
Closes #284


src/snowflake/connector/file_transfer_agent.py
```diff
@@ -380,507 +390,269 @@ def execute(self):
...
             )
-            pool = ThreadPoolExecutor(thread_number)
-            triggers = Queue()
-            for _ in range(thread_number):
-                pool.submit(
-                    SnowflakeFileTransferAgent._upload_files_in_queue_thread,
-                    qtask,
-                    self._results,
-                    triggers,
+            with cv_chunk_process:
```
sfc-gh-cshi, line 491: try to lock cv_chunk_process here when a chunk is done, but cv_chunk_process in preprocess_done_cb is locked for a big loop, and that preprocess_done_cb is triggered in multi thread manner I am afraid some early done transfer chunks might be blocked here (might not be too long still) until the loop in preprocess_done_cb releases this Condition lock.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -380,507 +390,269 @@ def execute(self):
...
+                                function_and_callback_wrapper,
+                                # Work fn
+                                done_client.download_chunk,
+                                # Callback fn
+                                _callback,
+                                file_meta,
+                                # Arguments for work fn
+                                _chunk_id,
+                            )
+                        transfer_metadata.chunks_in_queue += 1
```
sfc-gh-cshi, line 477: once submit, chunks_in_queue adds one. A large file with many chunks(e.g. 1000 chunks) will cross the 2 * max_concurrency limit (e.g. 32) a lot, in this case, every transfer_done_cb will trigger  cv_chunk_process.notify() and cause the while loop (at line 444) of some other threads beging awaked and then wait again, maybe a semaphore could help reducing the cost here.

       while transfer_metadata.chunks_in_queue > 2 * max_concurrency:
                        cv_chunk_process.wait()"

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -380,507 +390,269 @@ def execute(self):
...
+                else:
+                    done_client.finish_download()
+                notify_file_completed()
+            elif done_client.meta.result_status == ResultStatus.SKIPPED:
+                # this case applies to upload only
+                notify_file_completed()
+            else:
+                logger.debug(f"Finished preparing file {done_client.meta.name}")
+                with cv_chunk_process:
+                    while transfer_metadata.chunks_in_queue > 2 * max_concurrency:
```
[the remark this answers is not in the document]
sfc-gh-mkeller: No, `max_concurrency` number of them are transferred, but `2 * max_concurrency` is waiting in the queue ready to go

src/snowflake/connector/storage_client.py
```diff
@@ -0,0 +1,416 @@
...
+                os.unlink(self.full_dst_file_name)
+            logger.exception(f"Failed to download a file: {self.full_dst_file_name}")
+            meta.dst_file_size = -1
+            meta.result_status = ResultStatus.ERROR
+
+    def upload_chunk(self, chunk_id: int) -> None:
+        fd = (
+            self.meta.real_src_stream
+            or self.meta.src_stream
+            or open(self.data_file, "rb")
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Not without buffering data in memory, which lead before to really high memory usage.
I think nowadays most people would be using SSDs, so I don't see the point in optimizing for HDDs.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -380,507 +390,269 @@ def execute(self):
...
+        max_concurrency = self._parallel
+        network_tpe = ThreadPoolExecutor(max_concurrency)
+        preprocess_tpe = ThreadPoolExecutor(min(len(metas), os.cpu_count()))
+        postprocess_tpe = ThreadPoolExecutor(min(len(metas), os.cpu_count()))
+        logger.debug(f"Chunk ThreadPoolExecutor size: {max_concurrency}")
+        cv_main_thread = threading.Condition()  # to signal the main thread
+        cv_chunk_process = (
+            threading.Condition()
+        )  # to get more chunks into the chunk_tpe
+        files = [self._create_file_transfer_client(m) for m in metas]
```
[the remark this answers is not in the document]
sfc-gh-mkeller: I think this should be handled as an enhancement after we release this feature. Created SNOW-398380 to track

src/snowflake/connector/file_transfer_agent.py
sfc-gh-mkeller: Good point, a semaphore is better fitted here!
The use-case is not something that we recommend though, do you think that we could treat this as an enhancement later instead of delaying this feature?

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -380,507 +390,269 @@ def execute(self):
...
-        retry_round = 0
-        presigned_url_handled = False
-
-        while not qtask.empty():
-            len_file_metas = qtask.qsize()
-            thread_number = min(len_file_metas, self._parallel)
+    def transfer(self, metas: List["SnowflakeFileMeta"]) -> None:
+        max_concurrency = self._parallel
+        network_tpe = ThreadPoolExecutor(max_concurrency)
+        preprocess_tpe = ThreadPoolExecutor(min(len(metas), os.cpu_count()))
```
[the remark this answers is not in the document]
sfc-gh-mkeller: We ran a benchmark and found that `os.cpu_count()` was better for files with size around 200MB (which is what we suggest customers to use), but this number seems to be too high for very large files.

src/snowflake/connector/file_transfer_agent.py
sfc-gh-mkeller: Let me link the line where this cv is used: https://github.com/snowflakedb/snowflake-connector-python/blob/9ad4b67e21fa4a7919c56d05426c705b443158ca/src/snowflake/connector/file_transfer_agent.py#L447

I think it's safe to say that the code on this spot is non-blocking, and doesn't do much checks a variable and sleeps on the CV. And there is at most `cpu_count` of threads in this cv section. I think it is safe to leave it as is.

I could add extra logging that would indicate that this is an issue. Is that okay with you?

src/snowflake/connector/azure_storage_client.py
```diff
@@ -0,0 +1,255 @@
+#
+# Copyright (c) 2012-2021 Snowflake Computing Inc. All right reserved.
+#
+
+from __future__ import division
+
+import json
+import os
+import xml.etree.cElementTree as ET
```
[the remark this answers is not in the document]
sfc-gh-mkeller: While you are right in this file, we do use `ET` for parsing responses from blob storage services: https://github.com/snowflakedb/snowflake-connector-python/blob/a8e8c423b69ff1e62bc1f692a0e2296e3c0b16cd/src/snowflake/connector/s3_storage_client.py#L187
Given that we need to keep it I'd like to keep our usage consistent across the driver (where we deal with XMLs we should keep using `ET`).

src/snowflake/connector/constants.py
```diff
@@ -193,3 +231,8 @@ class QueryStatus(Enum):
     RESTARTED = 10
     BLOCKED = 11
     NO_DATA = 12
+
+
+kilobytes = 1024
+megabytes = kilobytes * 1024
+gigabytes = megabytes * 1024
```
sfc-gh-kwagner, line 238: nit: rename these variables to be singular since it represents only one kilobyte, not multiple kilobytes

src/snowflake/connector/encryption_util.py
```diff
@@ -63,7 +48,7 @@ def encrypt_stream(
         encryption_material: "SnowflakeFileEncryptionMaterial",
         src: IO[bytes],
         out: IO[bytes],
-        chunk_size: int = block_size * 4 * 1024,
+        chunk_size: int = 64 * megabytes,  # block_size * 4 * 1024,
```
sfc-gh-kwagner, line 51: Shouldn't you use `kilobyte` here since you're replacing a 1024 constant?

src/snowflake/connector/encryption_util.py
```diff
@@ -184,8 +169,8 @@ def encrypt_file(
     def decrypt_file(
         metadata,
         encryption_material,
-        in_filename,
-        chunk_size=block_size * 4 * 1024,
+        in_filename: str,
+        chunk_size: int = block_size * 4 * 1024,
```
sfc-gh-kwagner, line 173: Why aren't you using the same formula to determine chunk size for both this decrypt method and the encrypt method above?

src/snowflake/connector/feature.py
```diff
@@ -3,7 +3,17 @@
...
 feature_use_pyopenssl = True  # use pyopenssl API or openssl command
+
+
+# use the new code-path without using SDK to put and get files
+def feature_sdkless_put() -> bool:
+    return os.environ.get("SF_SDKLESS_PUT", "").lower() == "true"
+
+
+def feature_sdkless_get() -> bool:
+    return os.environ.get("SF_SDKLESS_GET", "").lower() == "true"
```
sfc-gh-kwagner, line 19: Why is this needed?

src/snowflake/connector/file_util.py
```diff
@@ -34,7 +34,7 @@ def get_digest_and_size(src: IO[bytes]) -> Tuple[str, int]:
             Tuple of src's digest and src's size in bytes.
         """
         use_openssl_only = os.getenv("SF_USE_OPENSSL_ONLY", "False") == "True"
-        CHUNK_SIZE = 16 * 4 * 1024
+        CHUNK_SIZE = 1024 * 1024
```
sfc-gh-kwagner, line 37: Why did the chunk size change here? Can you add a comment. Also if this is the new chunk size can you replace this value with the megabyte constant?

src/snowflake/connector/feature.py
sfc-gh-mkeller: Because we are not removing the old code paths right away, so these environmental variables will switch to the new code paths separately for PUTs and GETs

src/snowflake/connector/encryption_util.py
sfc-gh-mkeller: From benchmarks and testing different chunk sizes I found that more smaller reads lead to more lock contention and overall lower performance. I found that 64 MB worked well

src/snowflake/connector/encryption_util.py
sfc-gh-mkeller: This was left over from testing even bigger block sizes, I'll change this to match the other

src/snowflake/connector/file_util.py
sfc-gh-mkeller: This was left over from testing even bigger block sizes, I'll change this back

src/snowflake/connector/encryption_util.py
sfc-gh-mkeller: Oops, I meant kilobyte and not megabyte

sfc-gh-mkeller: So this was meant to be just a simplification. I'll switch it to `kilobyte`

src/snowflake/connector/azure_storage_client.py
```diff
@@ -0,0 +1,255 @@
...
+
+logger = getLogger(__name__)
+
+"""
+Azure Location: Azure container name + path
+"""
+AzureLocation = namedtuple(
+    "AzureLocation",
+    ["container_name", "path"],  # Azure container name  # Azure path name
+)
```
sfc-gh-kwagner, line 34: nit: Use the same class inheritance structure for namedtuples that you used in constants.py

src/snowflake/connector/azure_storage_client.py
```diff
@@ -0,0 +1,255 @@
...
+SFCDIGEST = "x-ms-meta-sfcdigest"
+ENCRYPTION_DATA = "x-ms-meta-encryptiondata"
+MATDESC = "x-ms-meta-matdesc"
+
+
+class SnowflakeAzureRestClient(SnowflakeStorageClient):
+    def __init__(
+        self,
+        meta: "SnowflakeFileMeta",
+        credentials,
```
sfc-gh-kwagner, line 49: nit: Add type hints here

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+
+S3_FS = "S3"
+AZURE_FS = "AZURE"
+GCS_FS = "GCS"
+LOCAL_FS = "LOCAL_FS"
+CMD_TYPE_UPLOAD = "UPLOAD"
+CMD_TYPE_DOWNLOAD = "DOWNLOAD"
+FILE_PROTOCOL = "file://"
+
+MB = 1024.0 * 1024.0
```
sfc-gh-kwagner, line 67: Let's use `megabyte` from constants.py and get rid of this constant here

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+def _update_progress(
+    file_name: str,
+    start_time: float,
+    total_size: float,
+    progress: Union[float, int],
+    output_stream: Optional[IO] = sys.stdout,
+    show_progress_bar: Optional[bool] = True,
+) -> float:
+    bar_length = 10  # Modify this to change the length of the progress bar
+    total_size /= MB
```
sfc-gh-kwagner, line 167: Replace with `megabyte`

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+        status = "Done ({elapsed_time:.3f}s, {throughput:.2f}MB/s).\r\n".format(
+            elapsed_time=elapsed_time, throughput=throughput
+        )
+    if not status and show_progress_bar:
+        status = "({elapsed_time:.3f}s, {throughput:.2f}MB/s)".format(
+            elapsed_time=elapsed_time, throughput=throughput
+        )
+    if status:
+        block = int(round(bar_length * progress))
+        text = "\r{file_name}({size:.2f}MB): [{bar}] {percentage:.2f}% {status}".format(
```
sfc-gh-kwagner, line 190: nit: replace all these str.format() calls with f-strings

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+        idx = 0
+        len_file_metas = len(file_metas)
+        while idx < len_file_metas:
+            end_of_idx = (
+                idx + self._parallel
+                if idx + self._parallel <= len_file_metas
+                else len_file_metas
+            )
+
+            logger.debug("downloading files idx: {} to {}".format(idx, end_of_idx))
```
sfc-gh-kwagner, line 702: f-string

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+                self._ret["data"]["localLocation"]
+            )
+            if not os.path.isdir(self._local_location):
+                # NOTE: isdir follows the symlink
+                Error.errorhandler_wrapper(
+                    self._cursor.connection,
+                    self._cursor,
+                    ProgrammingError,
+                    {
+                        "msg": "The local path is not a directory: {}".format(
```
sfc-gh-kwagner, line 1131: f-string

test/integ/conftest.py
```diff
@@ -153,21 +156,8 @@ def init_test_schema(request, db_parameters) -> None:
...
-            password=ret1["password"],
-            host=ret1["host"],
-            port=ret1["port"],
-            database=ret1["database"],
-            account=ret1["account"],
-            protocol=ret1["protocol"],
-        ) as con1:
-            con1.cursor().execute("DROP SCHEMA IF EXISTS {}".format(TEST_SCHEMA))
-
-    request.addfinalizer(fin)
```
sfc-gh-kwagner, line 170: Why get rid of the finalizer method? Just curious because I copied this piece of code to snowpark-python and wondering if I should also make this change here

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+                    index_list += [index]
+
+            file_path_end_index = min(index_list) if index_list else -1
+
+            if file_path_end_index > file_path_begin_index:
+                file_path = command[file_path_begin_index:file_path_end_index]
+            elif file_path_end_index == -1:
+                file_path = command[file_path_begin_index:]
+
+        return file_path
```
sfc-gh-kwagner, line 1411: Need newline at the end of the file

test/integ/test_put_get.py
```diff
@@ -643,36 +658,24 @@ def test_put_overwrite(tmpdir, db_parameters, from_path):
...
-        schema=db_parameters["schema"],
-        account=db_parameters["account"],
-        protocol=db_parameters["protocol"],
-    ) as con:
+    with conn_cnx() as con:
         with con.cursor() as cur:
-            cur.execute("create temporary stage {}".format(stage_name))
+            cur.execute(f"create temporary stage {stage_name}")
             cur.execute(
                 "PUT 'file://{}' @{}".format(
```
sfc-gh-kwagner, line 669: f-string

test/integ/test_put_get_with_aws_token.py
```diff
@@ -69,8 +93,6 @@ def test_put_get_with_aws(tmpdir, conn_cnx, db_parameters, from_path):
                     "get @%{table_name} file://{}".format(
```
sfc-gh-kwagner, line 93: f-string

test/integ/test_put_get_with_aws_token.py
```diff
@@ -88,125 +110,119 @@ def test_put_get_with_aws(tmpdir, conn_cnx, db_parameters, from_path):
...
-                client.meta.client.upload_fileobj(
-                    file_stream, s3location.bucket_name, s3path
+        with conn_cnx() as cnx:
+            try:
+                cnx.cursor().execute(
+                    "create or replace table {} (a int, b string)".format(table_name)
+                )
+                ret = cnx.cursor()._execute_helper(
+                    "put file://{} @%{}".format(fname, table_name)
+                )
```
sfc-gh-kwagner, line 129: f-string

test/unit/test_cursor.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.mark.parametrize(
+    "sql,_type",
+    (
+        ("PUT file:///tmp/data/mydata.csv @my_int_stage;", FileTransferType.PUT),
+        ("GET @%mytable file:///tmp/data/;", FileTransferType.GET),
+        ("select 1;", None),
+    ),
+)
+def test_get_filetransfer_type(sql, _type):
+    assert SnowflakeCursor.get_file_transfer_type(sql) == _type
```
sfc-gh-kwagner, line 30: newline

src/snowflake/connector/gcs_storage_client.py
```diff
@@ -0,0 +1,392 @@
...
+logger = getLogger(__name__)
+
+GCS_METADATA_PREFIX = "x-goog-meta-"
+GCS_METADATA_SFC_DIGEST = GCS_METADATA_PREFIX + "sfc-digest"
+GCS_METADATA_MATDESC_KEY = GCS_METADATA_PREFIX + "matdesc"
+GCS_METADATA_ENCRYPTIONDATAPROP = GCS_METADATA_PREFIX + "encryptiondata"
+GCS_FILE_HEADER_DIGEST = "gcs-file-header-digest"
+GCS_FILE_HEADER_CONTENT_LENGTH = "gcs-file-header-content-length"
+GCS_FILE_HEADER_ENCRYPTION_METADATA = "gcs-file-header-encryption-metadata"
+CONTENT_CHUNK_SIZE = 10 * 1024
```
sfc-gh-kwagner, line 36: Replace 1024 with kilobyte

test/unit/test_gcs_client.py
```diff
@@ -0,0 +1,478 @@
...
+    else:
+        storage_credentials = Mock()
+        storage_credentials.creds = {}
+        stage_info = Mock()
+        connection = Mock()
+        client = SnowflakeGCSRestClient(
+            meta, storage_credentials, stage_info, connection, ""
+        )
+        file_header = client.get_file_header(meta.name)
+        assert file_header is None
```
sfc-gh-kwagner, line 478: newline

src/snowflake/connector/s3_storage_client.py
```diff
@@ -0,0 +1,474 @@
...
+            url, "GET", _resource, retry_id
+        )
+        if response.status_code == 200:
+            config = ET.fromstring(response.text)
+            use_accelerate_endpoint = (
+                config.find("Status") and config.find("Status").text == "Enabled"
+            )
+            logger.debug(f"use_accelerate_endpoint: {use_accelerate_endpoint}")
+            return use_accelerate_endpoint
+        return False
```
sfc-gh-kwagner, line 474: newline

src/snowflake/connector/storage_client.py
```diff
@@ -0,0 +1,422 @@
...
+        """Deletes the tmp_dir and closes the source stream belonging to this client.
+        This function is idempotent."""
+        if os.path.exists(self.tmp_dir):
+            logger.debug(f"cleaning up tmp dir: {self.tmp_dir}")
+            shutil.rmtree(self.tmp_dir)
+        if self.meta.real_src_stream and not self.meta.real_src_stream.closed:
+            self.meta.real_src_stream.close()
+
+    def __del__(self):
+        self.delete_client_data()
```
sfc-gh-kwagner, line 422: newline

src/snowflake/connector/s3_storage_client.py
```diff
@@ -0,0 +1,474 @@
...
+
+EXPIRED_TOKEN = "ExpiredToken"
+ADDRESSING_STYLE = "virtual"  # explicit force to use virtual addressing style
+
+"""
+S3 Location: S3 bucket name + path
+"""
+S3Location = namedtuple(
+    "S3Location", ["bucket_name", "path"]  # S3 bucket name  # S3 path name
+)
```
sfc-gh-kwagner, line 50: Use same namedtuple class structure

src/snowflake/connector/azure_storage_client.py
sfc-gh-mkeller: @sfc-gh-kwagner please take a look at SNOW-366822
I'll opening a follow up PR for these things, it'll be big again so I thought I'd separate it
If you'd rather have it in this then I can merge it into this PR, please let me know

src/snowflake/connector/file_transfer_agent_sdk.py
```diff
@@ -0,0 +1,1411 @@
...
+
+def _update_progress(
+    file_name: str,
+    start_time: float,
+    total_size: float,
+    progress: Union[float, int],
+    output_stream: Optional[IO] = sys.stdout,
+    show_progress_bar: Optional[bool] = True,
+) -> float:
+    bar_length = 10  # Modify this to change the length of the progress bar
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Sure, but let's split it into a separate Jira, I'd rather not put more things into this release

test/integ/conftest.py
sfc-gh-mkeller: It's because the new way of doing this is to yield what the fixture would have returned before and everything after the yield will be executed at tear-down time, see documentation here: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

src/snowflake/connector/file_transfer_agent_sdk.py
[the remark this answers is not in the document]
sfc-gh-kwagner: Yea that works, just create another JIRA for this

sfc-gh-kwagner:
@sfc-gh-mkeller The code coverage stats look good with significant improvement in the sdkless files. The only one that has a code coverage dip are the S3 tests. Particularly the S3 multipart-upload

sfc-gh-kwagner (approved):

PR #727, opened 2021-05-25
SNOW-361263: Retrieve metadata/schema without executing the query

sfc-gh-jdu:
Snowpark python needs a method in python connector, which can retrieve the metadata of a result set without executing the query. It already exists in JDBC. https://github.com/snowflakedb/snowflake/blob/f7b1d327d0b222be962c0b2d825ff286f70c3231/GlobalServices/src/main/java/com/snowflake/core/Query.java#L60


src/snowflake/connector/cursor.py
```diff
@@ -704,6 +710,11 @@ def execute_async(self, *args, **kwargs):
         kwargs["_exec_async"] = True
         return self.execute(*args, **kwargs)
 
+    def describe(self, *args, **kwargs):
+        """Obtain the schema/description of the result without executing the query"""
+        kwargs["_describe_only"] = kwargs["_is_internal"] = True
```
sfc-gh-jdu, line 715: mark it as `is_internal` as well to be consistent with JDBC. https://github.com/snowflakedb/snowflake-jdbc/blob/69f560b9370cbe5de112f259600911ccd74333bf/src/main/java/net/snowflake/client/core/SFStatement.java#L135

test/integ/test_cursor.py
```diff
@@ -1354,3 +1354,18 @@ def test_out_of_range_year(conn_cnx, result_format, cursor_type):
...
+
+def test_describe(conn_cnx):
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            cur.describe(
+                "select * from VALUES(1, 3.1415926, 'snow', TO_TIMESTAMP('2021-01-01 00:00:00'))"
+            )
+            assert cur.description is not None
+            column_types = [column[1] for column in cur.description]
+            assert constants.FIELD_ID_TO_NAME[column_types[0]] == "FIXED"
```
sfc-gh-okostakis, line 1367: These are the logical types. Shouldn't we return the SQL types?
https://snowflakecomputing.atlassian.net/wiki/spaces/EN/pages/3637294/Data+Type+Definitions

sfc-gh-jdu: I think there is no "sql types" in python connector (please correct me if I'm wrong @sfc-gh-mkeller). So I'm a little bit worried about the conversion in snowpark. But JDBC also gets the logical types first and convert them to sql types https://github.com/snowflakedb/snowflake-jdbc/blob/69f560b9370cbe5de112f259600911ccd74333bf/src/main/java/net/snowflake/client/jdbc/SnowflakeUtil.java#L129. 

So for scala we have Logical types -> sql types -> snowpark types, and for python we can simply have Logical types -> snowpark types, but it might be a little bit tricky to make the behavior consistent.

sfc-gh-okostakis: OK, then we should have a function with the same functionality as JDBC's `extractColumnMetadata` that you pointed to. Then we'll deal the transformation in snowpark on a per-scenario basis.

sfc-gh-jdu: Why do we need this? Why don't we just convert Logical types to snowpark types directly in python snowpark? AFAIK, sql types are only used in ServerConnection.scala https://github.com/snowflakedb/thundersnow/search?q=java.sql.Types

sfc-gh-okostakis: 0. We can't go from FIXED to the exact type in Snowpark unless we have all the information that's available in the connector. This is evident in the scala function you shared. So the best place for this functionality is close to the source.
1. This is a publicly facing app, so the output should conform to the information (i.e. the data types) that we publish in our docs (there is no mention of FIXED [here](https://docs.snowflake.com/en/sql-reference/intro-summary-data-types.html)
2. Parity among JDBC and the python connector.

sfc-gh-jdu: I feel it would be a non-trivial change to python connector and we might need a further thought.

src/snowflake/connector/cursor.py
```diff
@@ -704,6 +710,11 @@ def execute_async(self, *args, **kwargs):
         kwargs["_exec_async"] = True
         return self.execute(*args, **kwargs)
 
+    def describe(self, *args, **kwargs):
+        """Obtain the schema/description of the result without executing the query"""
```
sfc-gh-mkeller, line 714: I'm not a fan of using `*args, **kwargs` on public APIs.
It would require a really good docstring that explains default values, description and type hints, but there's no automatic tool that can check this for us. So in general I don't allow them on public APIs, but since we want this function to automatically work with any arguments we add to `execute` let's do it with one addition.
Also return type-hint.
```suggestion
    def describe(self, *args, **kwargs) -> "SnowflakeCursor":
        """Obtain the schema/description of the result without executing the query.
        
        
        This function takes the same arguments as execute, please refer to that function
        for documentation.
        """
```

src/snowflake/connector/cursor.py
```diff
@@ -704,6 +710,11 @@ def execute_async(self, *args, **kwargs):
...
+        kwargs["_describe_only"] = kwargs["_is_internal"] = True
+        return self.execute(*args, **kwargs)
```
sfc-gh-mkeller, line 716: Should we not return the schema explicitly instead of an empty resultset?

test/integ/test_cursor.py
```diff
@@ -1354,3 +1354,18 @@ def test_out_of_range_year(conn_cnx, result_format, cursor_type):
...
+            assert constants.FIELD_ID_TO_NAME[column_types[1]] == "FIXED"
+            assert constants.FIELD_ID_TO_NAME[column_types[2]] == "TEXT"
+            assert constants.FIELD_ID_TO_NAME[column_types[3]] == "TIMESTAMP_NTZ"
+            assert not cur.fetchall()
```
sfc-gh-mkeller, line 1371: Let's be a bit more explicit
```suggestion
            assert len(cur.fetchall()) == 0
```

sfc-gh-mkeller (changes requested):

src/snowflake/connector/cursor.py
sfc-gh-jdu: Yes, let's return the schema directly.

sfc-gh-jdu:
@sfc-gh-mkeller could you please take a look again? Thanks!

src/snowflake/connector/cursor.py
```diff
@@ -93,6 +94,19 @@
...
+    [
+        "name",
+        "type_code",
+        "display_size",
+        "internal_size",
+        "precision",
+        "scale",
+        "is_nullable",
+    ],
+)
```
sfc-gh-mkeller, line 108: You can actually add type hints to these by importing importing `NamedTuple` from `typing`.
For example see: https://docs.python.org/3/library/typing.html#typing.NamedTuple

sfc-gh-mkeller (changes requested):
I'm sorry, I should have warned you about the typed `NamedTuple` when I asked you to add it. 😓

src/snowflake/connector/cursor.py
sfc-gh-jdu: do you mean something like this?
```
ResultMetadata = NamedTuple(
    "ResultMetadata",
    [
        ("name", str),
        ("type_code", int),
        ("display_size", int),
        ("internal_size", int),
        ("precision", int),
        ("scale", int),
        ("is_nullable", bool),
    ],
)
```

sfc-gh-mkeller: Good on you for using the backwards compatible implementation, but we don't actually support <3.6 so I went ahead and updated the definition to the non-backwards compatible one

sfc-gh-mkeller (approved):
🔥 🔥

PR #731, opened 2021-06-01
SNOW-357321 Experiemental support for CI on linux of new architecture

sfc-gh-cshi:
This PR is basically moving Michael Dionne's change in ci-testing branch to 'ci/experimental' directory, to use right docker image. some other changes was just to make this directory change works.

### DIFF
--- a/ci/experimental/build_docker.sh
+++ b/ci/experimental/build_docker.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -e
+#
+# Build Snowflake Python Connector in Docker
+# NOTES:
+#   - To compile only a specific version(s) pass in versions like: `./build_docker.sh "3.5 3.6"`
+set -o pipefail
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source $THIS_DIR/set_base_image.sh
+CONNECTOR_DIR="$(dirname "$( dirname "${THIS_DIR}")")"
+echo "conndir= $CONNECTOR_DIR, thisdir=$THIS_DIR"
+mkdir -p $CONNECTOR_DIR/dist
+cd $THIS_DIR/docker/connector_build
+
+CONTAINER_NAME=build_pyconnector
+
+echo "[Info] Building docker image"
+docker build --pull -t ${CONTAINER_NAME}:1.0 --build-arg BASE_IMAGE=$BASE_IMAGE_MANYLINUX2014AARCH64 -f Dockerfile .
+
+echo "[Info] Building Python Connector"
+user_id=$(id -u ${USER})
+docker run \
+    -e TERM=vt102 \
+    -e PIP_DISABLE_PIP_VERSION_CHECK=1 \
+    -e LOCAL_USER_ID=${user_id} \
+    --mount type=bind,source="${CONNECTOR_DIR}",target=/home/user/snowflake-connector-python \
+    ${CONTAINER_NAME}:1.0 \
+    /home/user/snowflake-connector-python/ci/experimental/build_linux.sh $1
--- a/ci/experimental/docker/Dockerfile
+++ b/ci/experimental/docker/Dockerfile
@@ -1,0 +1,18 @@
+# We use manylinux1 base image because pyarrow_manylinux2010 has a bug and wheel failed to be audited
+ARG BASE_IMAGE=quay.io/pypa/manylinux2014_aarch64
+FROM $BASE_IMAGE
+
+# This is to solve permission issue, read https://denibertovic.com/posts/handling-permissions-with-docker-volumes/
+RUN curl -o /usr/local/bin/gosu -SL "https://github.com/tianon/gosu/releases/download/1.12/gosu-arm64"
+RUN chmod +x /usr/local/bin/gosu
+
+COPY scripts/entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
+
+WORKDIR /home/user
+RUN chmod 777 /home/user
+RUN git clone https://github.com/matthew-brett/multibuild.git && cd /home/user/multibuild && git checkout bfc6d8b82d8c37b8ca1e386081fd800e81c6ab4a
+
+ENV PATH="${PATH}:/opt/python/cp36-cp36m/bin:/opt/python/cp37-cp37m/bin:/opt/python/cp38-cp38/bin:/opt/python/cp39-cp39/bin"
+
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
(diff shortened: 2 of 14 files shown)


ci/experimental/docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# We use manylinux1 base image because pyarrow_manylinux2010 has a bug and wheel failed to be audited
+ARG BASE_IMAGE=quay.io/pypa/manylinux2014_aarch64
```
sfc-gh-cshi, line 2: change to aarch64 docker image

ci/experimental/build_docker.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
+#
+# Build Snowflake Python Connector in Docker
+# NOTES:
+#   - To compile only a specific version(s) pass in versions like: `./build_docker.sh "3.5 3.6"`
+set -o pipefail
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source $THIS_DIR/set_base_image.sh
+CONNECTOR_DIR="$(dirname "$( dirname "${THIS_DIR}")")"
```
sfc-gh-mkeller, line 10: Just looks a bit neater when the constants are all defined at the beginning in 1 block
```suggestion
CONNECTOR_DIR="$(dirname "$( dirname "${THIS_DIR}")")"
source $THIS_DIR/set_base_image.sh
```

ci/experimental/build_docker.sh
```diff
@@ -0,0 +1,28 @@
...
+echo "conndir= $CONNECTOR_DIR, thisdir=$THIS_DIR"
```
sfc-gh-mkeller, line 11: I don't think this adds any value, we use these constants everywhere, if they were wrong a lot of things would be broken.
I think we don't need to log them anymore

ci/experimental/build_docker.sh
```diff
@@ -0,0 +1,28 @@
...
+mkdir -p $CONNECTOR_DIR/dist
+cd $THIS_DIR/docker/connector_build
+
+CONTAINER_NAME=build_pyconnector
```
sfc-gh-mkeller, line 15: Wouldn't this clash with the other non aarch docker image?
```suggestion
CONTAINER_NAME="build_pyconnector_aarch"
```

ci/experimental/build_docker.sh
```diff
@@ -0,0 +1,28 @@
...
+# Build Snowflake Python Connector in Docker
+# NOTES:
+#   - To compile only a specific version(s) pass in versions like: `./build_docker.sh "3.5 3.6"`
+set -o pipefail
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source $THIS_DIR/set_base_image.sh
+CONNECTOR_DIR="$(dirname "$( dirname "${THIS_DIR}")")"
+echo "conndir= $CONNECTOR_DIR, thisdir=$THIS_DIR"
+mkdir -p $CONNECTOR_DIR/dist
```
sfc-gh-mkeller, line 12: This fails if the dist directory already exists, there's a similar code snippet in `build_linux.sh`

ci/experimental/docker/connector_build/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# We use manylinux1 base image because pyarrow_manylinux2010 has a bug and wheel failed to be audited
```
sfc-gh-mkeller, line 1: I think this file could go straight into `ci/docker/connector_build_aarch/Dockerfile`

ci/experimental/docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# We use manylinux1 base image because pyarrow_manylinux2010 has a bug and wheel failed to be audited
```
sfc-gh-mkeller, line 1: I think this file is the same as `ci/experimental/docker/connector_build/Dockerfile` I think we don't need this

ci/experimental/docker/connector_build/scripts/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
sfc-gh-mkeller, line 1: Instead of copying this file even more times, could we set soft links instead?
Could you look whether this is an option in git?

ci/experimental/docker/connector_build/Dockerfile
sfc-gh-mkeller: I think this comment isn't true anymore
```suggestion
```

ci/experimental/docker/connector_test/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# We use manylinux1 base image because pyarrow_manylinux2010 has a bug and wheel failed to be audited
```
sfc-gh-mkeller, line 1: This isn't true anymore
```suggestion
```

ci/experimental/docker/connector_test/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ARG BASE_IMAGE
```
sfc-gh-mkeller, line 2: There's no default here, is that on purpose?

ci/experimental/docker/connector_test/Dockerfile
sfc-gh-mkeller: Again, I think this could straight into `ci/docker/connector_test_aarch/Dockerfile`

ci/experimental/docker/connector_test_fips/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM centos:7
```
sfc-gh-mkeller, line 1: I think this could move straight into `ci/docker/connector_test_fips_aarch`

ci/experimental/docker/scripts/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
sfc-gh-mkeller, line 1: Please don't add `docker/scripts` directory, this script is in each of the docker directories already

ci/experimental/set_base_image.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
```
sfc-gh-mkeller, line 1: We could get around copying this by moving the `ci/experimental/docker` to `ci/docker` as I mentioned already

ci/experimental/test_docker.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -e
```
sfc-gh-mkeller, line 1: We should be able to reuse the already existing version of this

ci/experimental/test_linux.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash -e
```
sfc-gh-mkeller, line 1: Please, let's instead make `ci/test_linux.sh` work with aarch rather than copying so many shell scripts around

sfc-gh-mkeller (changes requested):

sfc-gh-cshi:
Will open a new PR for this.

PR #733, opened 2021-06-01
SNOW-177082 Adds `to_arrow` and `to_pandas`, refactors `result_set` and `result_batch`

sfc-gh-kzaveri:
This PR addresses changes requested in: SNOW-177082

To summarize the changes:
- the existing `ROW_UNIT` and `TABLE_UNIT` constants are replaced with an Enum. All usages of these constants are also replaced. In `arrow_iterator.pyx` I made sure to use the .value of the Enum to allow for string comparison
- `cursor.py` has 2 new functions, `fetch_arrow_batches()` and `fetch_arrow_all()`, which use the corresponding `result_set.py` function
- `result_set.py` no longer converts pyarrow.Tables to dataframes. Instead, `result_batch.py` will return a pyarrow.Table/dataframe where appropriate. This is specified by a keyword argument.
- `result_batch.py`:
    - added two new methods called `to_arrow` and `to_pandas`. These are implemented only by the `ArrowResultBatch`
    - moved the existing `create_iter()` in `ArrowResultBatch` to be a protected member called `_create_iter()`. Created a new function called `create_iter()` to be the interface that `result_set` uses.
    - created `get_pandas_iter()` and `get_arrow_iter()` which are both iterators as the name implies. `to_pandas()` and `to_arrow()` simply use these iterators

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -845,7 +845,7 @@ class SnowflakeCursor(object):
 
     def fetch_arrow_batches(self) -> Iterator[Table]:
         self.check_can_use_arrow_resultset()
-        if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
+        if self._query_result_format != "arrow":
             raise NotSupportedError
         self._log_telemetry_job_data(
             TelemetryField.ARROW_FETCH_BATCHES, TelemetryData.TRUE
@@ -854,7 +854,7 @@ class SnowflakeCursor(object):
 
     def fetch_arrow_all(self) -> Optional[Table]:
         self.check_can_use_arrow_resultset()
-        if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
+        if self._query_result_format != "arrow":
             raise NotSupportedError
         self._log_telemetry_job_data(TelemetryField.ARROW_FETCH_ALL, TelemetryData.TRUE)
         return self._result_set._fetch_arrow_all()
@@ -862,7 +862,7 @@ class SnowflakeCursor(object):
     def fetch_pandas_batches(self, **kwargs) -> Iterator["pandas.DataFrame"]:
         """Fetches a single Arrow Table."""
         self.check_can_use_pandas()
-        if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
+        if self._query_result_format != "arrow":
             raise NotSupportedError
         self._log_telemetry_job_data(
             TelemetryField.PANDAS_FETCH_BATCHES, TelemetryData.TRUE
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -62,6 +62,7 @@ except ImportError:
 
 try:
     from .arrow_iterator import PyArrowIterator  # NOQA
+
     CAN_USE_ARROW_RESULT_FORMAT = True
 except ImportError as e:  # pragma: no cover
     logger.debug(
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -62,7 +62,6 @@ except ImportError:
 
 try:
     from .arrow_iterator import PyArrowIterator  # NOQA
-
     CAN_USE_ARROW_RESULT_FORMAT = True
 except ImportError as e:  # pragma: no cover
     logger.debug(
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -844,7 +844,7 @@ class SnowflakeCursor(object):
         return self
 
     def fetch_arrow_batches(self) -> Iterator[Table]:
-        self.check_can_use_pandas()
+        self.check_can_use_arrow_resultset()
         if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
             raise NotSupportedError
         self._log_telemetry_job_data(
@@ -853,7 +853,7 @@ class SnowflakeCursor(object):
         return self._result_set._fetch_arrow_batches()
 
     def fetch_arrow_all(self) -> Optional[Table]:
-        self.check_can_use_pandas()
+        self.check_can_use_arrow_resultset()
         if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
             raise NotSupportedError
         self._log_telemetry_job_data(TelemetryField.ARROW_FETCH_ALL, TelemetryData.TRUE)
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -42,6 +42,7 @@ from .errors import (
     ProgrammingError,
 )
 from .file_transfer_agent import SnowflakeFileTransferAgent
+from .options import installed_pandas, pandas
 from .sqlstate import SQLSTATE_FEATURE_NOT_SUPPORTED
 from .telemetry import TelemetryData, TelemetryField
 from .time_util import get_time_millis
@@ -54,9 +55,9 @@ if TYPE_CHECKING:  # pragma: no cover
 
 logger = getLogger(__name__)
 
-try:
-    import pyarrow
-except ImportError:
+if installed_pandas:
+    from pyarrow import Table
+else:
     logger.debug("Failed to import pyarrow. Cannot use pandas fetch API")
     pyarrow = None
 
@@ -789,9 +790,7 @@ class SnowflakeCursor(object):
             )
 
     def check_can_use_pandas(self):
-        global pyarrow
-
-        if pyarrow is None:
+        if not installed_pandas:
             msg = (
                 "Optional dependency: 'pyarrow' is not installed, please see the following link for install "
                 "instructions: https://docs.snowflake.com/en/user-guide/python-connector-pandas.html#installation"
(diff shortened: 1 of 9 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -852,8 +870,7 @@ def fetch_pandas_batches(self, **kwargs):
         self._log_telemetry_job_data(
             TelemetryField.PANDAS_FETCH_BATCHES, TelemetryData.DUMMY_VALUE
         )
-        for df in self._result_set._fetch_pandas_batches(**kwargs):
-            yield df
+        return self._result_set._fetch_pandas_batches(**kwargs)
```
sfc-gh-kzaveri, line 873: We can simply return the iterator, rather than looping and yielding

src/snowflake/connector/result_batch.py
```diff
@@ -519,18 +556,57 @@ def from_data(
...
     ]:
-        iter_unit = kwargs.pop("iter_unit", ROW_UNIT)
+        """Create an iterator for the ResultBatch. Used by get_arrow_iter."""
         if self._local:
             return self._from_data(self._data, iter_unit)
         response = self._download()
         with TimerContextManager() as load_metric:
             loaded_data = self._load(response, iter_unit)
         self._metrics[DownloadMetrics.load.value] = load_metric.get_timing_millis()
         return loaded_data
```
sfc-gh-kzaveri, line 573: This is the original `create_iter`, now used by other functions

src/snowflake/connector/result_batch.py
```diff
@@ -519,18 +556,57 @@ def from_data(
...
+        self, **kwargs
+    ) -> Union[
+        Iterator[Union[Dict, Exception]],
+        Iterator[Union[Tuple, Exception]],
+        Iterator[Table],
+    ]:
+        """The interface used by ResultSet to create an iterator for this ResultBatch."""
+        iter_unit: IterUnit = kwargs.pop("iter_unit", IterUnit.ROW_UNIT)
+        if iter_unit == IterUnit.TABLE_UNIT:
+            structure = kwargs.pop("structure", "pandas")
```
sfc-gh-kzaveri, line 585: I chose to name this `structure` but please feel free to provide any suggestions

src/snowflake/connector/result_set.py
```diff
@@ -118,10 +118,7 @@ def _report_metrics(self) -> None:
                 metrics.get(DownloadMetrics.parse.value),
             )
 
-    def _fetch_arrow_batches(
-        self,
-    ) -> Iterator[Table]:
-        """Fetches a all the results as Arrow Tables, chunked by Snowflake back-end."""
+    def __can_create_arrow_iter(self):
```
sfc-gh-kzaveri, line 121: Made this a private helper function since it is used in multiple places

sfc-gh-mkeller (changes requested):

src/snowflake/connector/result_batch.py
```diff
@@ -519,18 +556,57 @@ def from_data(
...
+            return self._create_iter(iter_unit=iter_unit)
+
+    def get_pandas_iter(self, **kwargs) -> Iterator["pandas.DataFrame"]:
+        """Returns an iterator for this batch which yields a pandas DataFrame"""
+        self.check_can_use_pandas()
+        table = next(self.get_arrow_iter(), None)
+        if table:
+            yield table.to_pandas(**kwargs)
+        else:
+            yield pandas.DataFrame(columns=self._column_names)
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: This change might make the function yield 2 data frames, so we should keep the `else` clause

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -742,7 +741,8 @@ def test_empty(conn_cnx):
         df_count = 0
         for _ in cursor.fetch_pandas_batches():
             df_count += 1
-        assert df_count == 0
+        # an empty df is yielded
+        assert df_count == 1
```
sfc-gh-mkeller, line 745: This means that this would be a behavior change.
I fired up the latest master code and it does look like `fetch_pandas_batches` returns an empty iterator for empty results, while `fetch_pandas_all` returns an empty `pandas.DataFrame` with the column names.
Let's adhere to this old behavior

src/snowflake/connector/result_batch.py
```diff
@@ -519,18 +556,57 @@ def from_data(
...
+
+    def to_pandas(self) -> "pandas.DataFrame":
+        """Returns this batch as a pandas DataFrame"""
+        return next(self.get_pandas_iter())
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: Resolving based on our discussion on Friday -  rewrote `to_pandas` to contain the conversion logic

src/snowflake/connector/result_batch.py
```diff
@@ -304,6 +304,34 @@ def create_iter(
         """
         raise NotImplementedError()
 
+    def check_can_use_pandas(self):
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: Took a closer look and noticed that this function in `result_batch` should use `Error.errorhandler_make_exception` instead of the Error function used in `cursor.py` (`Error.errorhandler_wrapper`). 
So, I changed the code of this function and though it is still similar, putting them in one place wouldn't work as well as it would.

src/snowflake/connector/cursor.py
```diff
@@ -882,6 +883,24 @@ def query_result(self, qid):
             )
         return self
 
+    def fetch_arrow_batches(self):
+        self.check_can_use_pandas()
+        if self._query_result_format != "arrow":  # TODO: or pandas isn't imported
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: I gave this more thought. I changed the `check_can_use_pandas` to use the `installed_pandas` options.
This way, it enforces that both pyarrow and pandas is installed. Before, we only checked if `pyarrow` was installed. The `fetch_pandas_batches` and `fetch_pandas_all` use this new `check_can_use_pandas`.

The reason why I made this change is because:
- It uses the .options module that we have, thus using one consistent way of checking dependencies.
- Checking that both pandas and pyarrow is installed "enforces" that the user has followed the docs
- The user should have pandas anyway, if they are using the `fetch_pandas*` functions.

src/snowflake/connector/result_set.py
```diff
@@ -130,32 +127,37 @@ def _fetch_arrow_batches(
...
 
-    def _fetch_arrow_all(self):
+    def _fetch_arrow_batches(
+        self,
+    ) -> Iterator[Table]:
+        """Fetches all the results as Arrow Tables, chunked by Snowflake back-end."""
+        self.__can_create_arrow_iter()
+        return self._create_iter(iter_unit=IterUnit.TABLE_UNIT, structure="arrow")
+
+    def _fetch_arrow_all(self) -> Optional[Table]:
```
sfc-gh-mkeller, line 138: We should also expose a `_fetch_arrow_batches` similar to how `_fetch_pandas_batches` works.

Also idk if this is strictly part of this ticket, but I want to expose `fetch_arrow_batches` and `fetch_arrow_all` through the regular `cursor`. Could you please add this too?

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -14,10 +14,10 @@
 import pytest
 
 try:
-    from snowflake.connector.constants import TABLE_UNIT
+    from snowflake.connector.constants import IterUnit
 except ImportError:
     # This is because of olddriver tests
-    TABLE_UNIT = "table"
+    IterUnit.TABLE_UNIT = "table"
```
sfc-gh-mkeller, line 20: This isn't enough, do something like
```suggestion
    class IterUnit(enum.Enum):
        ROW_UNIT = "row"
        TABLE_UNIT = "table"
```
Just make sure to import `enum`

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1037,3 +1037,29 @@ def test_pandas_telemetry(
...
 
         finish(conn, table)
+
+
+@pytest.mark.parametrize("result_format", ["pandas", "arrow"])
+def test_batch_to_pandas_arrow(conn_cnx, result_format):
+    rowcount = 10
+    with conn_cnx() as cnx:
+        cursor = cnx.cursor()
+        cursor.execute(SQL_ENABLE_ARROW)
```
sfc-gh-mkeller, line 1047: I would prefer embedding these into the connection object, like you do it in your other PR, but this is more of a nitpick than a hard requirement

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1037,3 +1037,29 @@ def test_pandas_telemetry(
...
+        cursor.execute(
+            f"select seq4() as foo, seq4() as bar from table(generator(rowcount=>{rowcount})) order by foo asc"
+        )
+        batches = cursor.get_result_batches()
+        assert len(batches) == 1
+        batch = batches[0]
+
+        # check that size, columns, and FOO column data is correct
+        if result_format == "pandas":
+            df = batch.to_pandas()
```
sfc-gh-mkeller, line 1057: Can we add sanity checks for the returned types in the if and else clauses?

test/integ/test_cursor.py
```diff
@@ -1259,6 +1259,24 @@ def test_resultbatch(
                     t.message["type"] == TelemetryField.GET_PARTITIONS_USED
                     for t in telemetry_data.records
                 )
+        telemetry_data = []
+        add_log_mock = mock.Mock()
+        add_log_mock.side_effect = lambda datum: telemetry_data.append(datum)
+        con._telemetry.add_log_to_batch = add_log_mock
```
sfc-gh-mkeller, line 1265: I believe that this is a mistake here, as it does what the code above does, without the usage of `capture_sf_telemetry`

sfc-gh-mkeller (changes requested):

src/snowflake/connector/result_set.py
```diff
@@ -118,10 +118,7 @@ def _report_metrics(self) -> None:
                 metrics.get(DownloadMetrics.parse.value),
             )
 
-    def _fetch_arrow_batches(
-        self,
-    ) -> Iterator[Table]:
-        """Fetches a all the results as Arrow Tables, chunked by Snowflake back-end."""
+    def __can_create_arrow_iter(self) -> None:
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: I did this so that this function is private, because I don't expect it be used outside of this class. I referred to this: https://www.geeksforgeeks.org/private-methods-in-python/

test/integ/pandas/test_arrow_pandas.py
sfc-gh-kzaveri: Odd, I added this before re-requesting review, seems like it didn't successfully push.

src/snowflake/connector/result_batch.py
```diff
@@ -519,18 +556,57 @@ def from_data(
...
+        return self._create_iter(iter_unit=IterUnit.TABLE_UNIT)
+
+    def to_arrow(self) -> Optional[Table]:
+        """Returns this batch as a pyarrow Table"""
+        return next(self._get_arrow_iter(), None)
+
+    def to_pandas(self, **kwargs) -> "pandas.DataFrame":
+        """Returns this batch as a pandas DataFrame"""
+        self._check_can_use_pandas()
+        table = next(self._get_arrow_iter())
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: The reason why I didn't add default value here is because if table = None, this function returns a dataframe. The `test_empty` in `test.arrow_pandas.py` expects no dataframes to be yielded. Hence, a `StopIteration` is what we want here. I do not have a way of making this more explicit - I'm open to your suggestions. Also, this function still returns a DataFrame if table is None (line 589), but if a StopIteration is raised, that code will never execute.

src/snowflake/connector/result_set.py
sfc-gh-mkeller: You already have this code in the PR. I must have forgot the first half of the code review, sorry! 😅

test/integ/test_cursor.py
sfc-gh-kzaveri: You're right, sorry. Removed this old code and kept the version that uses `capture_sf_telemetry`. Both versions got mixed up with the rebase

src/snowflake/connector/result_batch.py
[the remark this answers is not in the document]
sfc-gh-kzaveri: I'm just finding that this causes a problem in Python Version > 3.6, looking in to what I can do to address this.

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1047,23 +1047,25 @@ def test_pandas_telemetry(
...
+                f"select seq4() as foo, seq4() as bar from table(generator(rowcount=>{rowcount})) order by foo asc"
+            )
+            batches = cur.get_result_batches()
+            assert len(batches) == 1
+            batch = batches[0]
+
+            # check that size, columns, and FOO column data is correct
+            if result_format == "pandas":
+                df = batch.to_pandas()
+                assert type(df) is pandas.DataFrame
```
sfc-gh-mkeller, line 1062: I think the usual way of doing this would be to use `assert isinstance(df, pandas.DataFrame)`, but I think what you have also works.
The only reason why you wouldn't want to use `isinstance` if you don't want inheritance, but then I think that you might want to use `==` instead of `is`.

sfc-gh-mkeller (approved):
Please see my one comment, but otherwise 🚢 🚢

sfc-gh-mkeller (approved):
After tests please feel free to 🚢 🚢

PR #735, opened 2021-06-02
SNOW-361357 Provide interface exposing `schema` info to `ResultBatches`

sfc-gh-kzaveri:
This PR addresses changes requested in  https://snowflakecomputing.atlassian.net/browse/SNOW-361357

It passes the cursor's description list to each of the result batches, and can be accessed by each individual batch with `.schema`.

The added test has an example of the usage for this function

### DIFF
--- a/src/snowflake/connector/arrow_iterator.pyx
+++ b/src/snowflake/connector/arrow_iterator.pyx
@@ -4,6 +4,7 @@
 
 # distutils: language = c++
 # cython: language_level=3
+from typing import Iterator
 
 from cpython.ref cimport PyObject
 from libc.stdint cimport *
--- a/src/snowflake/connector/arrow_iterator.pyx
+++ b/src/snowflake/connector/arrow_iterator.pyx
@@ -4,7 +4,6 @@
 
 # distutils: language = c++
 # cython: language_level=3
-from typing import Iterator
 
 from cpython.ref cimport PyObject
 from libc.stdint cimport *
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -121,7 +121,7 @@ class ResultMetadata(NamedTuple):
 
     @classmethod
     def from_column(cls, col: Dict[str, Any]):
-        """Initializes a ResultMetadata object from the column description in the query response"""
+        """Initializes a ResultMetadata object from the column description in the query response."""
         return cls(
             col["name"],
             FIELD_NAME_TO_ID[col["type"].upper()],
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -15,6 +15,7 @@ from threading import Lock, Timer
 from typing import (
     IO,
     TYPE_CHECKING,
+    Any,
     Dict,
     Iterator,
     List,
@@ -119,11 +120,11 @@ class ResultMetadata(NamedTuple):
     is_nullable: bool
 
     @classmethod
-    def from_column(cls, col):
-        type_value = FIELD_NAME_TO_ID[col["type"].upper()]
+    def from_column(cls, col: Dict[str, Any]):
+        """Initializes a ResultMetadata object from the column description in the query response"""
         return cls(
             col["name"],
-            type_value,
+            FIELD_NAME_TO_ID[col["type"].upper()],
             None,
             col["length"],
             col["precision"],
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -776,7 +776,6 @@ class SnowflakeCursor(object):
             for column in data["rowtype"]
         ]
 
-        # could copy description here
         result_chunks = create_batches_from_response(
             self, self._query_result_format, data, self._description
         )
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -776,10 +776,9 @@ class SnowflakeCursor(object):
             for column in data["rowtype"]
         ]
 
+        # could copy description here
         result_chunks = create_batches_from_response(
-            self,
-            self._query_result_format,
-            data,
+            self, self._query_result_format, data, self._description
         )
 
         self._result_set = ResultSet(
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -118,6 +118,19 @@ class ResultMetadata(NamedTuple):
     scale: int
     is_nullable: bool
 
+    @classmethod
+    def from_column(cls, col):
+        type_value = FIELD_NAME_TO_ID[col["type"].upper()]
+        return cls(
+            col["name"],
+            type_value,
+            None,
+            col["length"],
+            col["precision"],
+            col["scale"],
+            col["nullable"],
+        )
+
 
 def exit_handler(*_):  # pragma: no cover
     """Handler for signal. When called, it will raise SystemExit with exit code FORCE_EXIT."""
@@ -226,7 +239,7 @@ class SnowflakeCursor(object):
                 logger.info(e)
 
     @property
-    def description(self):
+    def description(self) -> List[ResultMetadata]:
         return self._description
 
     @property
@@ -763,17 +776,8 @@ class SnowflakeCursor(object):
         if self._total_rowcount == -1 and not is_dml and data.get("total") is not None:
             self._total_rowcount = data["total"]
 
-        self._description = [
-            ResultMetadata(
-                column["name"],
-                FIELD_NAME_TO_ID[column["type"].upper()],
-                None,
-                column["length"],
-                column["precision"],
-                column["scale"],
-                column["nullable"],
-            )
-            for column in data["rowtype"]
+        self._description: List[ResultMetadata] = [
+            ResultMetadata.from_column(col) for col in data["rowtype"]
         ]
 
         result_chunks = create_batches_from_response(
--- a/src/snowflake/connector/result_batch.py
+++ b/src/snowflake/connector/result_batch.py
@@ -37,7 +37,7 @@ DOWNLOAD_TIMEOUT = 7  # seconds
 
 if TYPE_CHECKING:  # pragma: no cover
     from .converter import SnowflakeConverterType
-    from .cursor import SnowflakeCursor
+    from .cursor import ResultMetadata, SnowflakeCursor
     from .vendored.requests import Response
 
 if installed_pandas:
@@ -72,7 +72,7 @@ def create_batches_from_response(
     cursor: "SnowflakeCursor",
     _format: str,
     data: Dict[str, Any],
-    schema: List[Tuple[str, int, int, int, int, bool]],
+    schema: List["ResultMetadata"],
 ) -> List["ResultBatch"]:
     column_converters: List[Tuple[str, "SnowflakeConverterType"]] = []
     arrow_context: Optional["ArrowConverterContext"] = None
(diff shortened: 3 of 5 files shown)


src/snowflake/connector/result_batch.py
```diff
@@ -208,12 +214,14 @@ def __init__(
...
         remote_chunk_info: Optional["RemoteChunkInfo"],
         column_names: Sequence[str],
+        schema: List["ResultMetadata"],
         use_dict_result: bool,
     ):
         self.rowcount = rowcount
         self._chunk_headers = chunk_headers
         self._remote_chunk_info = remote_chunk_info
         self._column_names = column_names
+        self._schema = schema
```
sfc-gh-kzaveri, line 224: schema is an attribute of the base class `ResultBatch`

sfc-gh-kzaveri:
I can add a public function `schema()` as a getter if that is preferred. Currently, it is a protected member `_schema`.

test/integ/test_cursor.py
```diff
@@ -1318,6 +1318,34 @@ def test_resultbatch_lazy_fetching(conn_cnx, result_format, patch_path):
...
+    with conn_cnx(
+        session_parameters={"python_connector_query_result_format": result_format}
+    ) as con:
+        with con.cursor() as cur:
+            cur.execute(
+                f"select seq4() as c1, randstr(1,random()) as c2 "
+                f"from table(generator(rowcount => {rowcount}));"
+            )
+            result_batches = cur.get_result_batches()
+            batch_schemas = [batch._schema for batch in result_batches]
```
sfc-gh-cshi, line 1340: _schema is the exposed interface so it should be a public attribute. A '_' prefix is not needed here. Please make sure when there is no result(row count is 0) we still have the schema.

sfc-gh-kzaveri: Sounds good. I will add a test for this and update the function accordingly.

sfc-gh-kzaveri: Updated `_schema` to `schema` and added the test

test/integ/test_cursor.py
```diff
@@ -1318,6 +1318,64 @@ def test_resultbatch_lazy_fetching(conn_cnx, result_format, patch_path):
...
+            result_batches = cur.get_result_batches()
+            batch_schemas = [batch.schema for batch in result_batches]
+            for schema in batch_schemas:
+                # 2 columns in result
+                assert len(schema) == 2
+                # all batches should have the same schema
+                assert schema[0] == ResultMetadata("C1", 0, None, None, 10, 0, False)
+                assert schema[1] == ResultMetadata(
+                    "C2", 2, None, 16777216, None, None, False
+                )
```
sfc-gh-mkeller, line 1348: Little simpler:
```suggestion
                assert schema == [
                    ResultMetadata("C1", 0, None, None, 10, 0, False),
                    ResultMetadata("C2", 2, None, 16777216, None, None, False),
                ]
```

test/integ/test_cursor.py
```diff
@@ -1318,6 +1318,64 @@ def test_resultbatch_lazy_fetching(conn_cnx, result_format, patch_path):
...
 
+@pytest.mark.skipolddriver(reason="new feature in v2.5.0")
+@pytest.mark.parametrize(
+    "result_format,patch_path",
+    (
+        ("json", "snowflake.connector.result_batch.JSONResultBatch.create_iter"),
+        ("arrow", "snowflake.connector.result_batch.ArrowResultBatch.create_iter"),
+    ),
+)
+def test_resultbatches_can_access_schema(conn_cnx, result_format, patch_path):
```
sfc-gh-mkeller, line 1329: This test seems to be mostly just a copy of `test_resultbatch_lazy_fetching`. Could we not just add the assert statement to the end of that test?

test/integ/test_cursor.py
```diff
@@ -1318,6 +1318,64 @@ def test_resultbatch_lazy_fetching(conn_cnx, result_format, patch_path):
...
+                )
+
+
+@pytest.mark.skipolddriver(reason="new feature in v2.5.0")
+@pytest.mark.parametrize(
+    "result_format,patch_path",
+    (
+        ("json", "snowflake.connector.result_batch.JSONResultBatch.create_iter"),
+        ("arrow", "snowflake.connector.result_batch.ArrowResultBatch.create_iter"),
+    ),
```
sfc-gh-mkeller, line 1357: `patch_path` is not used
```suggestion
    "result_format", ("json","arrow"),
```

sfc-gh-mkeller (changes requested):

src/snowflake/connector/arrow_iterator.pyx
```diff
@@ -4,6 +4,7 @@
 
 # distutils: language = c++
 # cython: language_level=3
+from typing import Iterator
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: Whoops, this must have slipped through while rebasing

src/snowflake/connector/cursor.py
```diff
@@ -117,6 +117,19 @@ class ResultMetadata(NamedTuple):
...
+        return cls(
+            col["name"],
+            type_value,
+            None,
+            col["length"],
+            col["precision"],
+            col["scale"],
+            col["nullable"],
+        )
+
```
sfc-gh-kzaveri, line 132: Using a class method so that its easier to see how the arguments line up with the class fields

src/snowflake/connector/result_batch.py
```diff
@@ -72,21 +72,25 @@ def create_batches_from_response(
...
-        column_converters: List[Tuple[str, "SnowflakeConverterType"]] = [
-            (
-                c["type"],
-                cursor._connection.converter.to_python_method(c["type"].upper(), c),
+
+        def col_to_converter(col):
+            type_name = col["type"].upper()
+            python_method = cursor._connection.converter.to_python_method(
+                type_name, col
             )
```
sfc-gh-kzaveri, line 89: Added a small function here so that we can name the two values in the tuple (type name and python method)

src/snowflake/connector/result_batch.py
```diff
@@ -123,6 +124,7 @@ def create_batches_from_response(
                         compressedSize=c["compressedSize"],
                     ),
                     column_names,
+                    schema,
                     column_converters,
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: Removed `column_names` as that can easily be extracted from the schema. `column_converters` still preserved since they need more than what the schema contains.

test/integ/test_cursor.py
```diff
@@ -1306,9 +1306,18 @@ def test_resultbatch_lazy_fetching(conn_cnx, result_format, patch_path):
             #  first fetchone call
             with mock.patch(patch_path, return_value=iter([(1,)])) as patched_download:
                 cur.execute(
-                    f"select seq4() from table(generator(rowcount => {rowcount}));"
+                    f"select seq4() as c1, randstr(1,random()) as c2 "
+                    f"from table(generator(rowcount => {rowcount}));"
                 )
                 result_batches = cur.get_result_batches()
+                batch_schemas = [batch.schema for batch in result_batches]
```
sfc-gh-kzaveri, line 1313: I assert that the schema that the is available before the download, so that test verifies this case

src/snowflake/connector/cursor.py
```diff
@@ -117,6 +117,19 @@ class ResultMetadata(NamedTuple):
     scale: int
     is_nullable: bool
 
+    @classmethod
+    def from_column(cls, col):
```
sfc-gh-mkeller, line 121: Type-hint for `col` please. I know that `json.loads` can give you a lot of things, but I think we can well-define at least the key part of it.

Also please add a one-liner docstring describing what this is for.

src/snowflake/connector/cursor.py
```diff
@@ -117,6 +117,19 @@ class ResultMetadata(NamedTuple):
...
+        type_value = FIELD_NAME_TO_ID[col["type"].upper()]
+        return cls(
+            col["name"],
+            type_value,
```
sfc-gh-mkeller, line 125: It's never re-used, lets inline it
```suggestion
        return cls(
            col["name"],
            FIELD_NAME_TO_ID[col["type"].upper()],
```

src/snowflake/connector/cursor.py
sfc-gh-mkeller: Maybe for the future if we find ourselves leaving a comment like this that might mean that adding a docstring saying the same thing could be beneficial

src/snowflake/connector/result_batch.py
```diff
@@ -72,21 +72,25 @@ def create_batches_from_response(
...
     total_len: int = data.get("total", 0)
     first_chunk_len = total_len
     rest_of_chunks: List["ResultBatch"] = []
     if _format == "json":
-        column_converters: List[Tuple[str, "SnowflakeConverterType"]] = [
-            (
-                c["type"],
-                cursor._connection.converter.to_python_method(c["type"].upper(), c),
+
+        def col_to_converter(col):
```
sfc-gh-mkeller, line 85: If you add a function, don't forget to type-hint it :D
```suggestion
        def col_to_converter(col -> Dict[str, Any]) -> Tuple[str, "SnowflakeConverterType"]:
```

Again, I'm not 100% on the key type-hint, but I think `str` is the only thing it can be

src/snowflake/connector/result_batch.py
```diff
@@ -112,17 +116,20 @@ def create_batches_from_response(
             chunk_headers[SSE_C_ALGORITHM] = SSE_C_AES
             chunk_headers[SSE_C_KEY] = qrmk
 
+        def remote_chunk_info(c):
```
sfc-gh-mkeller, line 119: Type-hints, please!

src/snowflake/connector/result_batch.py
```diff
@@ -207,13 +210,13 @@ def __init__(
         rowcount: int,
         chunk_headers: Optional[Dict[str, str]],
         remote_chunk_info: Optional["RemoteChunkInfo"],
-        column_names: Sequence[str],
+        schema: List["ResultMetadata"],
```
sfc-gh-mkeller, line 213: You usually want to be specific when you define a return type, or the type of a variable.
But also want to be as generic as possible when defining a function's arguments.
For this reason we should be using `Sequence` for arguments. Python also advises the same practice: https://docs.python.org/3/library/typing.html#typing.List

src/snowflake/connector/result_batch.py
```diff
@@ -311,15 +314,15 @@ def __init__(
...
```
sfc-gh-mkeller, line 317: Same as my other comment

src/snowflake/connector/result_batch.py
```diff
@@ -329,7 +332,7 @@ def from_data(
         cls,
         data: Sequence[Sequence[Any]],
         data_len: int,
-        column_names: Sequence[str],
+        schema: List["ResultMetadata"],
```
sfc-gh-mkeller, line 335: Same as my other comment

src/snowflake/connector/result_batch.py
```diff
@@ -429,11 +432,15 @@ def __init__(
         context: "ArrowConverterContext",
         use_dict_result: bool,
         numpy: bool,
-        column_names: Sequence[str],
+        schema: List["ResultMetadata"],
```
sfc-gh-mkeller, line 435: Same as my other comment

src/snowflake/connector/result_batch.py
```diff
@@ -501,7 +508,7 @@ def from_data(
...
```
sfc-gh-mkeller, line 511: Change this too

src/snowflake/connector/result_set.py
```diff
@@ -155,7 +155,8 @@ def _fetch_pandas_all(self, **kwargs):
         if table:
             return table.to_pandas(**kwargs)
         else:
-            return pandas.DataFrame(columns=self.batches[0]._column_names)
+            column_names = [col.name for col in self.batches[0].schema]
```
sfc-gh-mkeller, line 158: What do you think about introducing a `property` into `ResultBatch` that generates this dynamically?

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):
🚢 🚢

PR #739, opened 2021-06-07
SNOW-364567 SNOW-303944 Remove substitution of dash for underscore in account names

sfc-gh-mknister:
Replacing '-' with '_' has led to a problem with certain accounts. Removing this logic so as to not break existing customers.


sfc-gh-stakeda (approved):

sfc-gh-mkeller (approved):
🚢 🚢

PR #744, opened 2021-06-08
SNOW-362198 removing upper certifi pin

### LINKED ISSUE #732: SNOW-363198: setup.py defines two certifi package versions
setup.py defines two `certifi` versions

- `certifi<2021.0.0`
  located at: https://github.com/snowflakedb/snowflake-connector-python/blob/master/setup.py#L214
- `certifi>=2017.4.17`
 located at: https://github.com/snowflakedb/snowflake-connector-python/blob/master/setup.py#L230

Both of those rules resolved to same version `2020.12.5` until recently. Now two days ago `certifi` 2021.5.30 was released: https://pypi.org/project/certifi/#history

Now following is possible:
- some app installs latests certifi package, getting version 2021.5.30
- installing snowflake-connector-python fails because it wants certifi<2021.0.0

Is there a reason for disallowing certifi 2021 versions? Is it a good practice to have several times the same package in `install_requires` with differing version restrictions?

sfc-gh-mkeller:
SNOW-362198

We hit an issue when the first release of `certifi` in 2021 was released, in a very quick attempt to fix our internal pipelines I bumped the upper limit for it.
Now I believe that was a mistake and we should not have any upper limit for dependencies such as `certifi`.

Closes #732

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -211,7 +211,6 @@ setup(
         "boto3>=1.4.4,<2.0.0",
         # While requests is vendored, we use regular requests to perform OCSP checks
         "requests<3.0.0",
-        "certifi<2022.0.0>=2017.4.17",
         "pytz",
         "pycryptodomex>=3.2,!=3.5.0,<4.0.0",
         "pyOpenSSL>=16.2.0,<21.0.0",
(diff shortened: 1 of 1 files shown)


sfc-gh-kzaveri (approved):

sfc-gh-kzaveri:
LGTM

PR #749, opened 2021-06-11
SNOW-365900 Test key/pair authentication with regionless URL

sfc-gh-mknister:
SNOW-365900
This fixes the issue of JWT tokens being invalid for account aliases with dashes in them, and runs a test against one of these accounts with an org alias in QA. The current Jenkins credentials running against a dev deployment would require us to set up org/account creation with every Jenkins run, so it seems more feasible to test against an org that doesn't have to be re-created each run.

### DIFF
--- a/test/integ/test_key_pair_authentication.py
+++ b/test/integ/test_key_pair_authentication.py
@@ -7,7 +7,6 @@
 import uuid
 from datetime import datetime, timedelta
 from os import path
-from test.integ.conftest import RUNNING_AGAINST_LOCAL_SNOWFLAKE
 
 import jwt
 import pytest
@@ -65,10 +64,8 @@ def test_get_token_from_private_key(input_account, expected_account):
     )
 
 
-@pytest.mark.internal
-@pytest.mark.skipif(
-    not RUNNING_AGAINST_LOCAL_SNOWFLAKE,
-    reason="connection timeouts occur when attempting to connect to this external account with automated testing",
+@pytest.mark.skip(
+    reason="Test is for running locally only. Credentials will fail on Jenkins and Github."
 )
 def test_regionless_url_JWT_token_validity():
 
--- a/test/integ/test_key_pair_authentication.py
+++ b/test/integ/test_key_pair_authentication.py
@@ -7,6 +7,7 @@
 import uuid
 from datetime import datetime, timedelta
 from os import path
+from test.integ.conftest import RUNNING_AGAINST_LOCAL_SNOWFLAKE
 
 import jwt
 import pytest
@@ -65,6 +66,10 @@ def test_get_token_from_private_key(input_account, expected_account):
 
 
 @pytest.mark.internal
+@pytest.mark.skipif(
+    not RUNNING_AGAINST_LOCAL_SNOWFLAKE,
+    reason="connection timeouts occur when attempting to connect to this external account with automated testing",
+)
 def test_regionless_url_JWT_token_validity():
 
     test_user = "admin"
--- a/test/integ/test_key_pair_authentication.py
+++ b/test/integ/test_key_pair_authentication.py
@@ -22,7 +22,7 @@ import snowflake.connector
     "input_account,expected_account",
     [
         ("s3testaccount.global", "S3TESTACCOUNT.GLOBAL"),
-        ("acct-with-dashes", "ACCT_WITH_DASHES"),
+        ("acct-with-dashes", "ACCT-WITH-DASHES"),
         ("testaccount.extra", "TESTACCOUNT"),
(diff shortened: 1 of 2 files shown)


sfc-gh-stakeda (approved):

test/integ/test_key_pair_authentication.py
```diff
@@ -64,6 +64,49 @@ def test_get_token_from_private_key(input_account, expected_account):
...
+    cnx.cursor().execute(
+        """
+    alter user {user} set rsa_public_key='{public_key}'
+    """.format(
+            user=test_user, public_key=public_key_der_encoded
+        )
+    )
+
+    db_config["private_key"] = private_key_der
+    with snowflake.connector.connect(**db_config) as _:
```
sfc-gh-mkeller, line 106: I believe that you don't actually have to name it anything if you don't want to use it
```suggestion
    with snowflake.connector.connect(**db_config):
```

test/integ/test_key_pair_authentication.py
```diff
@@ -64,6 +64,49 @@ def test_get_token_from_private_key(input_account, expected_account):
...
+        "timezone": "UTC",
+    }
+
+    cnx = snowflake.connector.connect(**db_config_with_pw)
+    cursor = cnx.cursor()
+    cursor.execute(
+        """
+    use role accountadmin
+    """
+    )
```
sfc-gh-mkeller, line 93: You never close the connection or the cursor, which is why we prefer using the with clauses, they close things for you automatically
```suggestion
    with snowflake.connector.connect(**db_config_with_pw) as cnx:
        with cnx.cursor() as cursor:
            cursor.execute("use role accountadmin")
```

test/integ/test_key_pair_authentication.py
```diff
@@ -64,6 +64,49 @@ def test_get_token_from_private_key(input_account, expected_account):
...
+
+    private_key_der, public_key_der_encoded = generate_key_pair(2048)
+
+    cnx.cursor().execute(
+        """
+    alter user {user} set rsa_public_key='{public_key}'
+    """.format(
+            user=test_user, public_key=public_key_der_encoded
+        )
+    )
```
sfc-gh-mkeller, line 103: There's no reason for your short queries to be this long, also [f-strings](https://www.python.org/dev/peps/pep-0498/)
```suggestion
    cnx.cursor().execute(f"alter user {test_user} set rsa_public_key='{public_key_der_encoded}'")
    )
```
Also indent it if you switch to using a with statement for `cnx`

test/integ/test_key_pair_authentication.py
```diff
@@ -64,6 +64,49 @@ def test_get_token_from_private_key(input_account, expected_account):
     )
 
 
+@pytest.mark.skipolddriver
+def test_regionless_url_JWT_token_validity(request, db_parameters):
```
sfc-gh-mkeller, line 68: Seems to be unused
```suggestion
def test_regionless_url_JWT_token_validity(db_parameters):
```

sfc-gh-mkeller (changes requested):

test/integ/test_key_pair_authentication.py
```diff
@@ -64,6 +65,43 @@ def test_get_token_from_private_key(input_account, expected_account):
     )
 
 
+@pytest.mark.internal
+@pytest.mark.skipif(
+    not RUNNING_AGAINST_LOCAL_SNOWFLAKE,
+    reason="connection timeouts occur when attempting to connect to this external account with automated testing",
+)
+def test_regionless_url_JWT_token_validity(db_parameters):
```
sfc-gh-mkeller, line 73: Also unused
```suggestion
def test_regionless_url_JWT_token_validity():
```

test/integ/test_key_pair_authentication.py
```diff
@@ -64,8 +65,12 @@ def test_get_token_from_private_key(input_account, expected_account):
     )
 
 
-@pytest.mark.skipolddriver
-def test_regionless_url_JWT_token_validity(request, db_parameters):
+@pytest.mark.internal
+@pytest.mark.skipif(
+    not RUNNING_AGAINST_LOCAL_SNOWFLAKE,
+    reason="connection timeouts occur when attempting to connect to this external account with automated testing",
+)
```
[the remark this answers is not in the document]
sfc-gh-mknister: I tried running this against Jenkins but Jenkins did not seem to like the credentials either- I was hoping this would just be run locally. I don't think it can work with either GithubActions or Jenkins.

test/integ/test_key_pair_authentication.py
sfc-gh-mknister: Good point- took this out too.

PR #754, opened 2021-06-15
SNOW-365900 Remove manual test for JWT token validity

sfc-gh-mknister:
Removing test - will create confluence link instead


sfc-gh-mkeller (approved):
🚢 🚢

PR #766, opened 2021-06-21
SNOW 370785 Add Fetch Arrow Tests

sfc-gh-kzaveri:
This PR address changes requested in [SNOW-370785](https://snowflakecomputing.atlassian.net/browse/SNOW-370785)

It tests that `fetch_arrow_all` and `fetch_arrow_batches` behave correctly with empty queries.
It also cleans up some old code to be a little simpler.

### DIFF
--- a/test/integ/pandas/test_arrow_pandas.py
+++ b/test/integ/pandas/test_arrow_pandas.py
@@ -1038,3 +1038,16 @@ def test_batch_to_pandas_arrow(conn_cnx, result_format):
                 assert arrow_table.shape == (10, 2)
                 assert arrow_table.column_names == ["FOO", "BAR"]
                 assert arrow_table.to_pydict()["FOO"] == list(range(rowcount))
+
+
+def test_simple_arrow_fetchall(conn_cnx):
+    rowcount = 100000
+    with conn_cnx() as cnx:
+        with cnx.cursor() as cur:
+            cur.execute(SQL_ENABLE_ARROW)
+            cur.execute(
+                f"select seq4() as foo from table(generator(rowcount=>{rowcount})) order by foo asc"
+            )
+            arrow_table = cur.fetch_arrow_all()
+            assert arrow_table.shape == (rowcount, 1)
+            assert arrow_table.to_pydict()["FOO"] == list(range(rowcount))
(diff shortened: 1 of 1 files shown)


test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1084,3 +1038,16 @@ def test_batch_to_pandas_arrow(conn_cnx, result_format):
...
+    rowcount = 100000
+    with conn_cnx() as cnx:
+        with cnx.cursor() as cur:
+            cur.execute(SQL_ENABLE_ARROW)
+            cur.execute(
+                f"select seq4() as foo from table(generator(rowcount=>{rowcount})) order by foo asc"
+            )
+            arrow_table = cur.fetch_arrow_all()
+            assert arrow_table.shape == (rowcount, 1)
+            assert arrow_table.to_pydict()["FOO"] == list(range(rowcount))
```
sfc-gh-kzaveri, line 1053: I have added a simple test like this for now. The reason why is because `fetch_arrow_all` is a new cursor function and so I thought it would be valuable to add a basic test. In theory though the new `fetch_arrow_*` functions behave quite similarly to their `fetch_pandas_*` counterparts + go through the same code path as when you call `to_pandas/to_arrow` which is also tested, so perhaps this test is not so necessary.

Please let me know if you think there is any value to this test/more should be added since `fetch_arrow_*` are new/is trivial and not needed.

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -65,9 +65,7 @@ def test_scaled_tinyint(conn_cnx):
     cases = ["NULL", 0.11, -0.11, "NULL", 1.27, -1.28, "NULL"]
     table = "test_arrow_tiny_int"
     column = "(a number(5,2))"
-    values = (
-        "(" + "),(".join(["{}, {}".format(i, c) for i, c in enumerate(cases)]) + ")"
-    )
+    values = ",".join([f"({i}, {c})" for i, c in enumerate(cases)])
```
sfc-gh-kzaveri, line 68: This is just a small thing that caught my eye and noticed it could be a little simpler

sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-mkeller (approved):
🚢 🚢

PR #773, opened 2021-06-24
Remove offending log line.

sfc-gh-hchaturvedi:
Skip logging meta data from AWS as it might contain sensitive information. This is a temporary fix to for the problem. A more holistic approach is being explored which might take a little longer.


sfc-gh-mkeller (approved):
🚢 🚢

PR #775, opened 2021-06-26
SNOW-361211 Adds Updated Fetching Logic

sfc-gh-kzaveri:
This PR begins to address performance issues investigated in: [SNOW-361211](https://snowflakecomputing.atlassian.net/browse/SNOW-361211)

Specifically, it updates the existing fetching logic to leverage all 4 threads of the ThreadPoolExecutor at all times.

This PR also adds debug logs for improved logging/debugging purposes.

### DIFF
--- a/src/snowflake/connector/result_batch.py
+++ b/src/snowflake/connector/result_batch.py
@@ -30,8 +30,6 @@ from .options import installed_pandas, pandas
 from .time_util import DecorrelateJitterBackoff, TimerContextManager
 from .vendored import requests
 
-# from threading import Lock
-
 logger = getLogger(__name__)
 
 MAX_DOWNLOAD_RETRY = 10
@@ -51,8 +49,6 @@ else:
 SSE_C_ALGORITHM = "x-amz-server-side-encryption-customer-algorithm"
 SSE_C_KEY = "x-amz-server-side-encryption-customer-key"
 SSE_C_AES = "AES256"
-
-# load_lock = Lock()
 
 
 @unique
--- a/src/snowflake/connector/result_batch.py
+++ b/src/snowflake/connector/result_batch.py
@@ -99,7 +99,7 @@ def create_batches_from_response(
         arrow_context = ArrowConverterContext(cursor._connection._session_parameters)
     if "chunks" in data:
         chunks = data["chunks"]
-        logger.debug("chunk size=%s", len(chunks))
+        logger.debug(f"chunk size={len(chunks)}")
         # prepare the downloader for further fetch
         qrmk = data.get("qrmk")
         chunk_headers: Dict[str, Any] = {}
@@ -109,9 +109,7 @@ def create_batches_from_response(
                 chunk_headers[header_key] = header_value
                 if "encryption" not in header_key:
                     logger.debug(
-                        "added chunk header: key=%s, value=%s",
-                        header_key,
-                        header_value,
+                        f"added chunk header: key={header_key}, value={header_value}"
                     )
         elif qrmk is not None:
             logger.debug(f"qrmk={qrmk}")
@@ -271,7 +269,7 @@ class ResultBatch(abc.ABC):
             try:
                 with TimerContextManager() as download_metric:
                     logger.debug(
-                        "started downloading result batch of size %d", self.rowcount
+                        f"started downloading result batch of size {self.rowcount}"
                     )
                     response = requests.get(
                         self._remote_chunk_info.url,
@@ -281,8 +279,7 @@ class ResultBatch(abc.ABC):
                     )
                     if response.ok:
                         logger.debug(
-                            "successfully downloaded result batch of size %d",
-                            self.rowcount,
+                            f"successfully downloaded result batch of size {self.rowcount}"
                         )
                         break
             except Exception as e:
@@ -451,10 +448,10 @@ class JSONResultBatch(ResultBatch):
             return iter(self._data)
         response = self._download()
         # Load data to a intermediate form
-        logger.debug("started loading result batch of size %d", self.rowcount)
+        logger.debug(f"started loading result batch of size {self.rowcount}")
         with TimerContextManager() as load_metric:
             downloaded_data = self._load(response)
-        logger.debug("finished loading result batch of size %d", self.rowcount)
+        logger.debug(f"finished loading result batch of size {self.rowcount}")
         self._metrics[DownloadMetrics.load.value] = load_metric.get_timing_millis()
         # Process downloaded data
         with TimerContextManager() as parse_metric:
@@ -589,10 +586,10 @@ class ArrowResultBatch(ResultBatch):
         if self._local:
             return self._from_data(self._data, iter_unit)
         response = self._download()
-        logger.debug("started loading result batch of size %d", self.rowcount)
+        logger.debug(f"started loading result batch of size {self.rowcount}")
         with TimerContextManager() as load_metric:
             loaded_data = self._load(response, iter_unit)
-        logger.debug("finished loading result batch of size %d", self.rowcount)
+        logger.debug(f"finished loading result batch of size {self.rowcount}")
         self._metrics[DownloadMetrics.load.value] = load_metric.get_timing_millis()
         return loaded_data
 
--- a/src/snowflake/connector/result_batch.py
+++ b/src/snowflake/connector/result_batch.py
@@ -270,6 +270,9 @@ class ResultBatch(abc.ABC):
         for retry in range(MAX_DOWNLOAD_RETRY):
             try:
                 with TimerContextManager() as download_metric:
+                    logger.debug(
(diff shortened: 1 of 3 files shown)


test/integ/test_cursor.py
```diff
@@ -1331,11 +1337,20 @@ def test_resultbatch_lazy_fetching_and_schemas(conn_cnx, result_format, patch_pa
...
-                #  short circuits and just parses (for JSON batches) and then return an
-                #  an iterator through that data, so we expect the call count to be 5
-                assert patched_download.call_count == 5
+                first_output = cur.fetchone()
+                second_output = cur.fetchone()  # Trigger pre-fetching
+                assert first_output == (0,)
+                assert second_output == (1,)
+
+                # delay so that the thread pool downloads for all submitted chunks
+                time.sleep(1)
```
sfc-gh-kzaveri, line 1346: I added this delay because the TPE needs time to call `_create_iter` on each batch that is in the pool. 
To avoid this I could mock the ThreadPoolExecutor object and just have my mock class call `_create_iter` right away. Let me know what you think and I can implement it.

src/snowflake/connector/result_batch.py
```diff
@@ -270,13 +270,20 @@ def _download(self, **kwargs) -> "Response":
...
                     response = requests.get(
                         self._remote_chunk_info.url,
                         headers=self._chunk_headers,
                         timeout=DOWNLOAD_TIMEOUT,
                         stream=True,
                     )
                     if response.ok:
+                        logger.debug(
+                            "successfully downloaded result batch of size %d",
+                            self.rowcount,
```
sfc-gh-mkeller, line 285: This type of logging is error prone, let's use f-strings instead
```suggestion
                            f"successfully downloaded result batch of size {self.rowcount}",
```

src/snowflake/connector/result_batch.py
```diff
@@ -270,13 +270,20 @@ def _download(self, **kwargs) -> "Response":
         for retry in range(MAX_DOWNLOAD_RETRY):
             try:
                 with TimerContextManager() as download_metric:
+                    logger.debug(
+                        "started downloading result batch of size %d", self.rowcount
```
sfc-gh-mkeller, line 274:
```suggestion
                        f"started downloading result batch of size {self.rowcount}"
```

src/snowflake/connector/result_batch.py
```diff
@@ -444,8 +451,10 @@ def create_iter(
             return iter(self._data)
         response = self._download()
         # Load data to a intermediate form
+        logger.debug("started loading result batch of size %d", self.rowcount)
```
sfc-gh-mkeller, line 454:
```suggestion
        logger.debug(f"started loading result batch of size {self.rowcount}")
```

src/snowflake/connector/result_batch.py
```diff
@@ -444,8 +451,10 @@ def create_iter(
...
         with TimerContextManager() as load_metric:
             downloaded_data = self._load(response)
+        logger.debug("finished loading result batch of size %d", self.rowcount)
```
sfc-gh-mkeller, line 457:
```suggestion
        logger.debug(f"finished loading result batch of size {self.rowcount}")
```

src/snowflake/connector/result_batch.py
```diff
@@ -580,8 +589,10 @@ def _create_iter(
         if self._local:
             return self._from_data(self._data, iter_unit)
         response = self._download()
+        logger.debug("started loading result batch of size %d", self.rowcount)
```
sfc-gh-mkeller, line 592:
```suggestion
        logger.debug(f"started loading result batch of size {self.rowcount}")
```

src/snowflake/connector/result_batch.py
```diff
@@ -580,8 +589,10 @@ def _create_iter(
...
         with TimerContextManager() as load_metric:
             loaded_data = self._load(response, iter_unit)
+        logger.debug("finished loading result batch of size %d", self.rowcount)
```
sfc-gh-mkeller, line 595:
```suggestion
        logger.debug(f"finished loading result batch of size {self.rowcount}")
```

src/snowflake/connector/result_set.py
```diff
@@ -61,12 +64,40 @@ def result_set_iterator(
...
     to continue iterating through the rest of the ``ResultBatch``.
     """
-    for it in pre_iter:
-        for element in it:
-            yield element
-    for it in post_iter:
-        for element in it.create_iter(**kw):
-            yield element
+
+    yield from first_batch_iter
```
sfc-gh-mkeller, line 68: I'm sorry I don't remember anymore if this is in line with how pre-fetching worked in the old implementation. Should this not be after we submit the next 4 batches to `pool`?

src/snowflake/connector/result_set.py
```diff
@@ -61,12 +64,40 @@ def result_set_iterator(
...
+    with ThreadPoolExecutor(4) as pool:
+        # Fill up window
+        for _ in range(min(4, len(unfetched_batches))):
+            unconsumed_batches.append(
+                pool.submit(unfetched_batches.popleft().create_iter, **kw)
+            )
+
+        i = 1
+        while unconsumed_batches:
+            logger.debug("user requesting to consume result batch %d", i)
```
sfc-gh-mkeller, line 80:
```suggestion
            logger.debug(f"user requesting to consume result batch {i}")
```

src/snowflake/connector/result_set.py
```diff
@@ -61,12 +64,40 @@ def result_set_iterator(
...
+
+            # Submit the next un-fetched batch to the pool
+            if unfetched_batches:
+                logger.debug(
+                    "queuing download of result batch of size %d",
+                    unfetched_batches[0].rowcount,
```
sfc-gh-mkeller, line 86:
```suggestion
                    f"queuing download of result batch of size {unfetched_batches[0].rowcount}"
```

src/snowflake/connector/result_set.py
```diff
@@ -61,12 +64,40 @@ def result_set_iterator(
...
+                )
+                future = pool.submit(unfetched_batches.popleft().create_iter, **kw)
+                unconsumed_batches.append(future)
+
+            future = unconsumed_batches.popleft()
+
+            # this will raise an exception if one has occurred
+            batch_iterator = future.result()
+
+            logger.debug("user began consuming result batch %d", i)
```
sfc-gh-mkeller, line 96:
```suggestion
            logger.debug(f"user began consuming result batch {i}")
```

src/snowflake/connector/result_set.py
```diff
@@ -61,12 +64,40 @@ def result_set_iterator(
...
+            yield from batch_iterator
+            logger.debug("user finished consuming result batch %d", i)
```
sfc-gh-mkeller, line 98:
```suggestion
            logger.debug(f"user finished consuming result batch {i}")
```

src/snowflake/connector/result_set.py
```diff
@@ -185,18 +216,20 @@ def _create_iter(
...
-        post_download_iters = self.batches[5:]
+        first_batch_iter = self.batches[0].create_iter(**kwargs)
+
+        # Iterator[Tuple] Futures that have not been consumed by the user
+        unconsumed_batches: Deque[Future[Iterator[Tuple]]] = deque()
+
+        # batches that have not been fetched
+        unfetched_batches = deque(self.batches[1:])
+        for num, batch in enumerate(unfetched_batches):
+            logger.debug("result batch %d has size %d", num + 1, batch.rowcount)
```
sfc-gh-mkeller, line 227:
```suggestion
            logger.debug(f"result batch {num + 1} has size {batch.rowcount}")
```

test/integ/test_cursor.py
sfc-gh-mkeller: Usually the best way to test for something that might take up to `x` seconds is to check in a while loop where we allow failures, but if the condition is true then we short circuit, where `x` is what you'd deem unacceptable performance.

Now in this case this might not make much sense (for 1 second). Just keep this approach in mind in case you'd need to write a test that depends on some network activity. I'd say that mocking would take too much effort in this case.

sfc-gh-mkeller (changes requested):

src/snowflake/connector/result_set.py
sfc-gh-kzaveri: You are right, the `pre_init` in the chunk downloader pre-fetches the first 4 batches before any yielding. I'll update this.

test/integ/test_cursor.py
sfc-gh-kzaveri: Thanks for letting me know 👍 . Since I just changed the iterator to pre-fetch though, it turns out I won't need to add a delay at all (the test remains unchanged). Good to know in case I need to write a test like this again.

sfc-gh-mkeller (approved):
🚢 🚢

PR #777, opened 2021-06-26
SNOW-373625 remove five-year-old line from README

swanderz:
thought it'd make sense to drop this line given that this library has a sizable chunk of C now.


sfc-gh-mkeller (approved):
Thank you @swanderz 🚢 🚢

PR #780, opened 2021-06-28
SNOW-361210 Adds Retry Logic and Tests

sfc-gh-kzaveri:
This PR addresses changes requested in: [SNOW-361210](https://snowflakecomputing.atlassian.net/browse/SNOW-361210).

The core functionality for retries was already there, but there were a few differences between this and the old `ChunkDownloader` error handling. See [this comment](https://snowflakecomputing.atlassian.net/browse/SNOW-361210?focusedCommentId=1683481) for more details. In short, the old code didn't log/raise retry errors quite correctly, and accepted status codes other than 200 which this PR addresses.

I also added quite a few unit tests to verify that retries are working as we expect.

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -210,7 +210,7 @@ def raise_failed_request_error(
     response: Response,
 ) -> None:
     TelemetryService.get_instance().log_http_request_error(
-        "HttpError%s" % str(response.status_code),
+        f"HttpError{response.status_code}",
         url,
         method,
         SQLSTATE_CONNECTION_WAS_NOT_ESTABLISHED,
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -149,15 +149,7 @@ COMPILER = COMPILER
 
 CLIENT_NAME = CLIENT_NAME  # don't change!
 CLIENT_VERSION = CLIENT_VERSION
-PYTHON_CONNECTOR_USER_AGENT = (
-    "{name}/{version} ({platform}) {python_implementation}/{python_version}".format(
-        name=CLIENT_NAME,
-        version=SNOWFLAKE_CONNECTOR_VERSION,
-        python_implementation=IMPLEMENTATION,
-        python_version=PYTHON_VERSION,
-        platform=PLATFORM,
-    )
-)
+PYTHON_CONNECTOR_USER_AGENT = f"{CLIENT_NAME}/{SNOWFLAKE_CONNECTOR_VERSION} ({PLATFORM}) {IMPLEMENTATION}/{PYTHON_VERSION}"
 
 NO_TOKEN = "no-token"
 
@@ -621,12 +613,7 @@ class SnowflakeRestful(object):
         if "Content-Length" in headers:
             del headers["Content-Length"]
 
-        full_url = "{protocol}://{host}:{port}{url}".format(
-            protocol=self._protocol,
-            host=self._host,
-            port=self._port,
-            url=url,
-        )
+        full_url = f"{self._protocol}://{self._host}:{self._port}{url}"
         ret = self.fetch(
             "get",
             full_url,
@@ -664,12 +651,7 @@ class SnowflakeRestful(object):
         socket_timeout=DEFAULT_SOCKET_CONNECT_TIMEOUT,
         _include_retry_params=False,
     ):
-        full_url = "{protocol}://{host}:{port}{url}".format(
-            protocol=self._protocol,
-            host=self._host,
-            port=self._port,
-            url=url,
-        )
+        full_url = f"{self._protocol}://{self._host}:{self._port}{url}"
         if self._connection._probe_connection:
             from pprint import pprint
 
@@ -916,14 +898,9 @@ class SnowflakeRestful(object):
             except Exception:
                 logger.info("data is not JSON")
         logger.error(
-            "Failed to get the response. Hanging? "
-            "method: {method}, url: {url}, headers:{headers}, "
-            "data: {data}".format(
-                method=method,
-                url=full_url,
-                headers=headers,
-                data=data,
-            )
+            f"Failed to get the response. Hanging? "
+            f"method: {method}, url: {full_url}, headers:{headers}, "
+            f"data: {data}"
         )
         Error.errorhandler_wrapper(
             conn,
@@ -996,7 +973,7 @@ class SnowflakeRestful(object):
 
                 if is_retryable_http_code(raw_ret.status_code):
                     error = get_http_retryable_error(raw_ret.status_code)
-                    logger.debug("%s. Retrying...", error)
+                    logger.debug(f"{error}. Retrying...")
                     # retryable server exceptions
                     raise RetryRequest(error)
 
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -15,6 +15,7 @@ import traceback
 import uuid
 from io import BytesIO
 from threading import Lock
+from typing import Dict, Optional, Type
 
 import OpenSSL.SSL
 
@@ -86,6 +87,7 @@ from .time_util import (
 )
 from .tool.probe_connection import probe_connection
 from .vendored import requests
+from .vendored.requests import Response
 from .vendored.requests.adapters import HTTPAdapter
 from .vendored.requests.auth import AuthBase
 from .vendored.requests.exceptions import (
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/network.py
```diff
@@ -187,6 +189,54 @@ def is_retryable_http_code(code: int) -> bool:
...
+        None,
+        InterfaceError,
+        {
+            "msg": f"{response.status_code} {response.reason}: {method} {url}",
+            "errno": ER_FAILED_TO_REQUEST,
+            "sqlstate": SQLSTATE_CONNECTION_WAS_NOT_ESTABLISHED,
+        },
+    )
+
+
```
sfc-gh-kzaveri, line 239: Pulled out the error raising functions so that they can be re-used by ResultBatch's download.

src/snowflake/connector/result_batch.py
```diff
@@ -270,18 +278,33 @@ def _download(self, **kwargs) -> "Response":
...
+                        error: Error = get_http_retryable_error(response.status_code)
+                        raise RetryRequest(error)
+                    elif response.status_code == UNAUTHORIZED:
+                        # make a unauthorized error
+                        raise_okta_unauthorized_error(None, response)
+                    else:
+                        raise_failed_request_error(None, chunk_url, "get", response)
+
+            except (RetryRequest, Exception) as e:
+                e = e.args[0] if isinstance(e, RetryRequest) else e
```
sfc-gh-kzaveri, line 304: This "unwraps" the underlying error so that the error we raise is the true HTTPError that occurred.

test/unit/test_result_batch.py
```diff
@@ -0,0 +1,155 @@
...
+    mock_get.return_value = create_mock_response(200)
+    chunk_info = MockRemoteChunkInfo("http://www.chunk-url.com")
+
+    result_batch = JSONResultBatch(100, None, chunk_info, [], [], True)
+    response = result_batch._download()
+
+    # successful on first try
+    assert mock_get.call_count == 1
+    assert response.status_code == 200
+
```
sfc-gh-kzaveri, line 75: The logic for the tests below are very similar but I kept them apart because they each test a different case

src/snowflake/connector/network.py
```diff
@@ -949,56 +995,21 @@ def _request_exec(
                     return ret
 
                 if is_retryable_http_code(raw_ret.status_code):
-                    ex = STATUS_TO_EXCEPTION.get(
-                        raw_ret.status_code, OtherHTTPRetryableError
-                    )
-                    exi = ex(code=raw_ret.status_code)
-                    logger.debug("%s. Retrying...", exi)
+                    error = get_http_retryable_error(raw_ret.status_code)
+                    logger.debug("%s. Retrying...", error)
```
sfc-gh-mkeller, line 999:
```suggestion
                    logger.debug(f"{error}. Retrying...")
```

src/snowflake/connector/result_batch.py
```diff
@@ -270,18 +278,33 @@ def _download(self, **kwargs) -> "Response":
         for retry in range(MAX_DOWNLOAD_RETRY):
             try:
                 with TimerContextManager() as download_metric:
+                    chunk_url = self._remote_chunk_info.url
```
sfc-gh-mkeller, line 281: This could be in-lined into the next expression, it doesn't seem to be reused afterwards

src/snowflake/connector/result_batch.py
sfc-gh-mkeller, line 304: I'd put this into the if clause

test/unit/test_result_batch.py
```diff
@@ -0,0 +1,155 @@
...
+from snowflake.connector.sqlstate import (
+    SQLSTATE_CONNECTION_REJECTED,
+    SQLSTATE_CONNECTION_WAS_NOT_ESTABLISHED,
+)
+
+REQUEST_MODULE_PATH = (
+    "snowflake.connector.vendored.requests"
+    if importlib.util.find_spec("snowflake.connector.vendored.requests")
+    else "requests"
+)
```
sfc-gh-mkeller, line 52: We usually do something similar to:
```python
try:  # pragma: no cover
	from snowflake.connector.vendored import requests
    # Set a variable here for path
except ImportError:
	import requests
	# Set same variable to other path
```
I wonder if you see any upside to tapping into `importlib` instead. If so, do you think that we could turn this into a function and use it everywhere where we do the fall back imports for old driver tests?

test/unit/test_result_batch.py
```diff
@@ -0,0 +1,155 @@
...
+
+MockRemoteChunkInfo = namedtuple("MockRemoteChunkInfo", "url")
+
+
+def create_mock_response(status_code):
+    mock_resp = Mock()
+    mock_resp.status_code = status_code
+    mock_resp.raw = "success" if status_code == OK else "fail"
+    return mock_resp
```
sfc-gh-mkeller, line 61: I can see this being useful to other tests. Should we move it to shared place like `test/helpers.py`?

test/unit/test_result_batch.py
```diff
@@ -0,0 +1,155 @@
...
+        (555, OtherHTTPRetryableError),  # random 5xx error
+    ],
+)
+def test_retryable_response_download(errcode, error_class):
+    # retryable exceptions
+    with mock.patch(REQUEST_MODULE_PATH + ".get") as mock_get:
+        mock_get.return_value = create_mock_response(errcode)
+
+        chunk_info = MockRemoteChunkInfo("http://www.chunk-url.com")
+        result_batch = JSONResultBatch(100, None, chunk_info, [], [], True)
```
sfc-gh-mkeller, line 97: These lines seem to be repeated a few times, could we move the definition to module level?

sfc-gh-mkeller (changes requested):

test/unit/test_result_batch.py
sfc-gh-kzaveri: In this case I actually don't need to import, I only needed a way to check which dependency was available to set `REQUEST_MODULE_PATH` accordingly. I initially had what you suggested but this would result in "unused import" errors by the linting check so I decided to do this and avoid the import entirely.

This might be beneficial in cases where we don't want to do the import (like here). Some people also suggest not doing the try..except way in case the module we try importing actually does exist but one of ITS imports raises the ImportError. 
I was following the recommendation here: https://stackoverflow.com/a/14050282
I can't think of any other reasons other than readability(?) and maybe some flexibility down the line (since the importlib module seems to have some other useful functions)

sfc-gh-mkeller: If you just want to silence the unused import statement (or other) warnings you can always put `# NOQA` after a line

src/snowflake/connector/result_batch.py
sfc-gh-kzaveri: I reuse it [here](https://github.com/snowflakedb/snowflake-connector-python/pull/780/files/07fef8da97cf07c8a4cc8b9500acd2ff069896c7#diff-69943088e0213b0053377355cbbb0d7264f23bde709fd06cabc763c663d536f3R301) (so that the error message contains the url)

test/unit/test_result_batch.py
sfc-gh-kzaveri: Sure, I'll do that. In that case it might also be useful for it to accept a dictionary (in case we want to test more than just different status codes). I could add that to this PR or leave a To-do to add this when its needed. What do you suggest?

test/unit/test_result_batch.py
sfc-gh-kzaveri: Oh I see, thanks, thats good to know in the future. I think in this case though I prefer doing it with `importlib` just to avoid importing since all I need is to set this variable.

sfc-gh-mkeller (approved):
🚢 🚢

test/unit/test_result_batch.py
```diff
@@ -0,0 +1,146 @@
...
+            error = ex.value
+            assert error.errno == ER_FAILED_TO_REQUEST
+            assert error.sqlstate == SQLSTATE_CONNECTION_WAS_NOT_ESTABLISHED
+        assert mock_get.call_count == MAX_DOWNLOAD_RETRY
+
+
+def test_retries_until_success():
+    with mock.patch(REQUEST_MODULE_PATH + ".get") as mock_get:
+        error_codes = [BAD_REQUEST, UNAUTHORIZED, 201]
+        mock_responses = [create_mock_response(code) for code in error_codes + [OK]]
```
sfc-gh-mkeller, line 133: This ` + [OK]` is kind of hidden, please add an explanation

sfc-gh-mkeller (approved):
I only have small asks, once you address them feel free to merge the PR.

PR #781, opened 2021-06-29
Update requirements_37.reqs

abneetwats24:
botocore==1.18.18 breaks s3fs which is needed to load parquet data files.
botocore==1.20.49 Solve the dependency for s3fs without breaking snowflake-connector-python


sfc-gh-mkeller:
If you don't mind @abneetwats24 let me take this over with #782

abneetwats24:
Okay tx.

PR #791, opened 2021-07-06
SNOW-361263: Fix describe method when running `insert into ...`

sfc-gh-jdu:
Before this change, if we run `cursor.describe("insert into ...")` (or `delete ...`), an "out of index" error will be returned because `data['rowset']` is empty, in `_init_result_and_meta` method in `cursor.py`. This PR fixes this bug and adds a corresponding test.

### DIFF
--- a/test/integ/test_cursor.py
+++ b/test/integ/test_cursor.py
@@ -1222,9 +1222,10 @@ def test_out_of_range_year(conn_cnx, result_format, cursor_type):
 
 
 @pytest.mark.skipolddriver
-def test_describe(conn_cnx):
+def test_describe(conn_cnx, db_parameters):
     with conn_cnx() as con:
         with con.cursor() as cur:
+            # test select
             description = cur.describe(
                 "select * from VALUES(1, 3.1415926, 'snow', TO_TIMESTAMP('2021-01-01 00:00:00'))"
             )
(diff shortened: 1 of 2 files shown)


test/integ/test_cursor.py
```diff
@@ -1235,3 +1236,15 @@ def test_describe(conn_cnx):
...
+            # test insert
+            cur.execute(
+                "create table {name} (aa int)".format(name=db_parameters["name"])
+            )
+            description = cur.describe(
+                "insert into {name}(aa) values({value})".format(
+                    name=db_parameters["name"], value="1234"
+                )
+            )
+            assert description[0][0].startswith("number of rows inserted")
```
sfc-gh-okostakis, line 1249: What should be the value in (rest of) the message? 0 or -1?

sfc-gh-okostakis (approved):
The fix seems reasonable. So does adding into the tests the case which triggered this PR. Please fix the test to include the expected value.

test/integ/test_cursor.py
sfc-gh-jdu: It only has "number of rows inserted" as a part of metadata and no value after this. I  changed it to use `==`.

sfc-gh-mkeller (approved):
I pushed a small change onto your changes @sfc-gh-jdu
Randomizing table names allows us to parallelize test runs instead of reusing the same name between multiple tests during the same run.

sfc-gh-jdu:
@sfc-gh-mkeller there is a merge gate `continuous-integration/jenkins/pr-head` failing, though it's not required. I'm wondering whether the PR is ready to merge?

sfc-gh-mkeller:
It should be safe to ignore for now. The failure is caused by GS dumps.

I thinks it's safe to say that that is not caused by your change, but please check out the output of Jenkins jobs just so that you get familiar to what failure is likely a server side issue (and hopefully safe to ignore by us on the client side).

Please note that there was a bug where Jenkins runs were not collecting GS dumps, this should have been fixed yesterday!

PR #792, opened 2021-07-13
SNOW-371967 Added _s3_connection_pool_size setting to SnowflakeConnection

sfc-gh-cconner:
SNOW-371967
Adding option to set connection pool size for `boto3`.
Note: this feature is going away soon, so it'll stay undocumented until then


sfc-gh-mkeller (approved):
🚢 🚢

PR #794, opened 2021-07-13
SNOW-359246 run semgrep in temp dir and use new baseline ref

sfc-gh-spandey:
SNOW-359246

run semgrep in temp dir to prevent potential issue of .git folder ownership being changed to root
also use new baseline ref parameter to stop reporting issue in upstream branch since current branch was created


sfc-gh-mkeller (approved):
🚢 🚢

Issue #750, opened 2021-06-15
SNOW-368531: Corrupted dates with executemany


juhap:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using?

Python 3.8.5

2. What operating system and processor architecture are you using?

Linux-4.19.104-microsoft-standard-x86_64-with-glibc2.29 (Ubuntu 20.04 under WSL2)

3. What are the component versions in the environment?

asn1crypto==1.4.0
azure-common==1.1.27
azure-core==1.15.0
azure-storage-blob==12.8.1
boto3==1.17.94
botocore==1.20.94
certifi==2020.12.5
cffi==1.14.5
chardet==3.0.4
cryptography==3.4.7
idna==2.10
install==1.3.4
isodate==0.6.0
jmespath==0.10.0
msrest==0.6.21
oauthlib==3.1.1
oscrypto==1.2.1
pycparser==2.20
pycryptodomex==3.10.1
PyJWT==2.1.0
pyOpenSSL==20.0.1
python-dateutil==2.8.1
pytz==2021.1
requests==2.25.1
requests-oauthlib==1.3.0
s3transfer==0.4.2
six==1.16.0
snowflake-connector-python==2.4.4
urllib3==1.26.5

4. What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.

```
import datetime
import logging
import os

from snowflake import connector


def test():
    # This will pick the connection information from env variables like SNOWFLAKE_USER etc.
    variables = ['account', 'user', 'password', 'database', 'schema', 'role', 'warehouse']
    envprefix = 'SNOWFLAKE'
    names = dict({f:f'{envprefix}_{f.upper()}' for f in variables})
    params = dict({name: os.environ.get(env_name, None) for name, env_name in names.items()})
    connector.paramstyle='qmark'
    connection = connector.connect(**params)
    cursor = connection.cursor()

    # Before running create a test table in Snowflake
    """"
    create table date_test (
        birth_date date
    );
    """

    # Insert 1969-02-07
    cursor.executemany(
        'INSERT INTO date_test (birth_date) values (?)',
        [[datetime.date(1969, 2, 7)]]
    )
    cursor.close()

    # Insert 1969-01-01
    cursor = connection.cursor()
    cursor.executemany(
        'INSERT INTO date_test (birth_date) values (?)',
        [[datetime.date(1969, 1, 1)]]
    )
    cursor.close()

[...]

PR #796, opened 2021-07-19
SNOW-368531 Fix `executemany` insert qmarks behaviour

sfc-gh-kzaveri:
This PR addresses changes requested in [SNOW-368531](https://snowflakecomputing.atlassian.net/browse/SNOW-368531)/ [GH Issue](https://github.com/snowflakedb/snowflake-connector-python/issues/750).

Our behavior when resolving the "type" of value had a bug. The best way to illustrate this would be with an example.

When using `executemany` to insert values, you would use a command such as this:

```python
    cursor.executemany(
        'INSERT INTO date_test (birth_date) values (?)',
        [[datetime.date(1969, 2, 7)],[datetime.date(1969, 1, 1)]]
    )
```
The `executemany` function would then transform the values in to a list like this ([see this](https://github.com/snowflakedb/snowflake-connector-python/blob/master/src/snowflake/connector/cursor.py#L985)) :
  ```python 

[ datetime.date(1969, 2, 7), datetime.date(1969, 1, 1) ] 

```

We would then determine the corresponding "snowflake type" for this list. [Lists are type "TEXT"](https://github.com/snowflakedb/snowflake-connector-python/blob/master/src/snowflake/connector/converter.py#L59) and thus when inserting these datetimes, Snowflake would interpret the epochs as "TEXT" which leads to the invalid dates appearing in the DB.

To fix this, I have set the "snowflake type" to  "TEXT" by default. However, if all of the values of the list have a certain type, say "DATE" in this case, we would update the "snowflake type" to "DATE" - which leads to a valid insert. 

Furthermore, users can now use Tuples to specify exactly which type they would like.
For example, they may do this:
```python
    cursor.executemany(
        'INSERT INTO date_test (birth_date) values (?)',
        [
               [ ("DATE", datetime.date(1969, 2, 7)) ],
               [ ("DATE", datetime.date(1969, 1, 1)) ]
        ]
    )
```

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -11,8 +11,8 @@ import re
 import sys
 import uuid
 import warnings
-from collections import namedtuple
 from difflib import get_close_matches
+from functools import partial
 from io import StringIO
 from logging import getLogger
 from threading import Lock
@@ -24,6 +24,7 @@ from typing import (
     Generator,
     Iterable,
     List,
+    NamedTuple,
     Optional,
     Tuple,
     Type,
@@ -203,6 +204,13 @@ for m in [method for method in dir(errors) if callable(getattr(errors, method))]
 strptime("20150102030405", "%Y%m%d%H%M%S")
 
 logger = getLogger(__name__)
+
+
+class TypeAndBinding(NamedTuple):
+    """Stores the type name and the Snowflake binding."""
+
+    type: str
+    binding: Optional[str]
 
 
 class SnowflakeConnection(object):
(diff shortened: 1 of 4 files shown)


src/snowflake/connector/connection.py
```diff
@@ -1086,16 +1090,29 @@ def _process_params_qmarks(
...
+                )
+
+        for idx, v in enumerate(params):
+            if isinstance(v, list):
+                snowflake_type = self.converter.snowflake_type(v)
+                all_param_data = list(map(get_snowflake_type_and_binding, v))
+                first_type = all_param_data[0].type
+                # if all elements have the same snowflake type, update snowflake_type
+                if all(param_data.type == first_type for param_data in all_param_data):
+                    snowflake_type = first_type
```
sfc-gh-kzaveri, line 1105: This is to set `snowflake_type` to "LIST", then it obtains the type and binding for each parameter in the list.
If all types are the same -> update snowflake type.

src/snowflake/connector/connection.py
```diff
@@ -1053,7 +1054,11 @@ def _process_params_qmarks(
             }
             Error.errorhandler_wrapper(self, cursor, ProgrammingError, errorvalue)
 
-        for idx, v in enumerate(params):
+        TypeAndBinding = namedtuple("TypeAndBinding", ["type", "binding"])
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Move out of this function and make sure to update the return type.
Also check if we could use this and `get_snowflake_type_and_binding` in any other places, like `_process_params`

sfc-gh-mkeller (changes requested):

src/snowflake/connector/connection.py
```diff
@@ -1049,6 +1058,47 @@ def _write_params_to_byte_rows(
...
+                        "automatically mapped to Snowflake data "
+                        "type. Specify the snowflake data type "
+                        "explicitly.".format(v.__class__.__name__.lower()),
+                        "errno": ER_NOT_IMPLICITY_SNOWFLAKE_DATATYPE,
+                    },
+                )
+            return TypeAndBinding(
+                snowflake_type,
+                self.converter.to_snowflake_bindings(snowflake_type, v),
+            )
```
sfc-gh-mkeller, line 1100: I have the tiniest nitpick:
The return statement can be moved out from the if-else clauses this way:
```suggestion
            snowflake_type, v = v
        else:
            snowflake_type = self.converter.snowflake_type(v)
            if snowflake_type is None:
                Error.errorhandler_wrapper(
                    self,
                    cursor,
                    ProgrammingError,
                    {
                        "msg": "Python data type [{}] cannot be "
                        "automatically mapped to Snowflake data "
                        "type. Specify the snowflake data type "
                        "explicitly.".format(v.__class__.__name__.lower()),
                        "errno": ER_NOT_IMPLICITY_SNOWFLAKE_DATATYPE,
                    },
                )
        return TypeAndBinding(
            snowflake_type,
            self.converter.to_snowflake_bindings(snowflake_type, v),
        )
```

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller:
Love the way you used the `map` with `partial` 👍

sfc-gh-mkeller (approved):
🚢 🚢

sfc-gh-kzaveri:
I verified that this issue: https://snowflakecomputing.atlassian.net/browse/SNOW-303368 is also solved by this change.
See:
https://snowflakecomputing.atlassian.net/browse/SNOW-303368?focusedCommentId=1825333

PR #800, opened 2021-07-21
SNOW-414830 Allow Storage Client to Use Sessions

sfc-gh-kzaveri:
This PR partly address changes requested in [SNOW-414830](https://snowflakecomputing.atlassian.net/browse/SNOW-414830).

It does so by allowing the storage client to use existing HTTP Sessions to make requests. This avoids the overhead of re-establishing a connection with a new HTTP session thereby improving the performance.

The session management in `network.py` is also changed with this PR. In particular, a `_session_map` is maintained, which is a dict from hostnames to `SessionPool` objects. A `SessionPool` maintains a stack of idle connections and all active connections.
This improves the old session management by:
- using a stack instead of a queue, as we want more recently used sessions to be re-used
- segregating sessions by hostname. See [this](https://snowflakecomputing.atlassian.net/browse/SNOW-378283). In particular,
> if you’re making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase 

Some minor logging is also added to monitor how many unique connections are used.

Things to do:
- [X] Run benchmarks
- [x] Test the session creation/deletion

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -15,7 +15,7 @@ import traceback
 import uuid
 from io import BytesIO
 from threading import Lock
-from typing import Dict, List, Optional, Set
+from typing import Dict, List, Optional, Set, Type
 
 import OpenSSL.SSL
 
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -776,7 +776,7 @@ class SnowflakeRestful(object):
 
         include_retry_params = kwargs.pop("_include_retry_params", False)
 
-        with self._use_requests_session() as session:
+        with self._use_requests_session(full_url) as session:
             retry_ctx = RetryCtx(timeout, include_retry_params)
             while True:
                 ret = self._request_exec_wrapper(
@@ -1089,11 +1089,13 @@ class SnowflakeRestful(object):
             finally:
                 session.close()
         else:
-            session_pool = self._sessions_map[url]
+            parsed_url = urlparse(url)
+            hostname = parsed_url.netloc
+            session_pool = self._sessions_map[hostname]
             session = session_pool.get_session()
-            logger.debug(f"Session status for SessionPool '{url}', {session_pool}")
+            logger.debug(f"Session status for SessionPool '{hostname}', {session_pool}")
             try:
                 yield session
             finally:
                 session_pool.return_session(session)
-            logger.debug(f"Session status for SessionPool '{url}', {session_pool}")
+            logger.debug(f"Session status for SessionPool '{hostname}', {session_pool}")
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -321,6 +321,8 @@ class SessionPool(object):
                 s.close()
             except Exception as e:
                 logger.info(f"Session cleanup failed: {e}")
+        self._active_sessions.clear()
+        self._idle_sessions.clear()
 
 
 class SnowflakeRestful(object):
@@ -1087,12 +1089,11 @@ class SnowflakeRestful(object):
                 session.close()
         else:
             try:
-                parsed_url = urlparse(url)
-                hostname = parsed_url.netloc
+                hostname = urlparse(url).hostname
             except Exception:
                 hostname = None
 
-            session_pool = self._sessions_map[hostname]
+            session_pool: SessionPool = self._sessions_map[hostname]
             session = session_pool.get_session()
             logger.debug(f"Session status for SessionPool '{hostname}', {session_pool}")
             try:
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -288,7 +288,7 @@ class SnowflakeAuth(AuthBase):
         return r
 
 
-class SessionPool(object):
+class SessionPool:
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/network.py
```diff
@@ -246,6 +248,46 @@ def __call__(self, r):
         return r
 
 
+class SessionPool(object):
```
sfc-gh-mkeller, line 251: I'm pretty sure this is implicitly done for all classes since Python 3
```suggestion
class SessionPool:
```

src/snowflake/connector/network.py
```diff
@@ -246,6 +248,46 @@ def __call__(self, r):
...
+    def return_session(self, session: Session) -> None:
+        """Places an active session back into the idle session stack."""
+        try:
+            self._active_sessions.remove(session)
+        except KeyError:
+            logger.debug("session doesn't exist in the active session pool. Ignored...")
+        self._idle_sessions.append(session)
+
+    def __repr__(self):
+        return f"Active request sessions: {len(self._active_sessions)}, Idle: {len(self._idle_sessions)}"
```
sfc-gh-mkeller, line 276: I like to put the class name in the beginning of these, the rest of the change is just a different idea (you don't have to actually accept this).
```suggestion
        return f"SessionPool {len(self._active_sessions)}/{{len(self._active_sessions)} + len(self._idle_sessions)} active sessions"
```

src/snowflake/connector/network.py
```diff
@@ -1070,28 +1106,17 @@ def _use_requests_session(self):
...
-                    logger.debug(
-                        "session doesn't exist in the active session pool. "
-                        "Ignored..."
-                    )
+                session_pool.return_session(session)
                 logger.debug(
-                    "Active requests sessions: %s, idle: %s",
-                    len(self._active_sessions),
-                    len(self._idle_sessions),
+                    f"Session status for SessionPool '{hostname}', {session_pool}"
```
sfc-gh-mkeller, line 1121: I'm not sure whether this has the effect you wanted, `session_pool` should be casted to string here. You only defined `__repr__` and not `__str__`. Please verify that this logs what you expect, I might also be wrong and maybe Python tries `__str__` and then `__repr__`.

test/unit/test_session_manager.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def get_mock_connection():
+    mock_conn = mock.Mock()
+    mock_conn.disable_request_pooling = False
+    mock_conn._ocsp_mode = lambda: True
+    return mock_conn
+
+
+def close_sessions(rest: SnowflakeRestful, num_session_pools):
```
sfc-gh-mkeller, line 31:
```suggestion
def close_sessions(rest: SnowflakeRestful, num_session_pools: int) -> None:
```

test/unit/test_session_manager.py
```diff
@@ -0,0 +1,109 @@
...
+
+hostname_1 = "sfctest0.snowflakecomputing.com"
+url_1 = f"https://{hostname_1}:443/session/v1/login-request"
+
+hostname_2 = "sfc-ds2-customer-stage.s3.amazonaws.com"
+url_2 = f"https://{hostname_2}/rgm1-s-sfctest0/stages/"
+url_3 = f"https://{hostname_2}/rgm1-s-sfctst0/stages/another-url"
+
+
+def get_mock_connection():
```
sfc-gh-mkeller, line 24:
```suggestion
def get_mock_connection() -> None:
```

test/unit/test_session_manager.py
```diff
@@ -0,0 +1,109 @@
...
+):
+    """
+    Creates 'num_sessions' sessions to 'url'. This is recursive so that idle sessions
+    are not reused.
+    """
+    if num_sessions == 0:
+        return
+    with rest._use_requests_session(url):
+        create_session(rest, num_sessions - 1, url)
+        pass
```
sfc-gh-mkeller, line 49:
```suggestion
```

test/unit/test_session_manager.py
```diff
@@ -0,0 +1,109 @@
...
+def close_sessions(rest: SnowflakeRestful, num_session_pools):
+    """Helper function to call SnowflakeRestful.close(). Asserts close was called on all SessionPools."""
+    with mock.patch("snowflake.connector.network.SessionPool.close") as close_mock:
+        rest.close()
+        assert close_mock.call_count == num_session_pools
+
+
+def create_session(
+    rest: SnowflakeRestful, num_sessions: int = 1, url: Optional[str] = None
+):
```
sfc-gh-mkeller, line 40:
```suggestion
) -> None:
```

test/unit/test_session_manager.py
sfc-gh-mkeller: But I actually think that this doesn't have to be a function, but the stuff this fn returns could be a module level variable

sfc-gh-mkeller (changes requested):

src/snowflake/connector/network.py
sfc-gh-kzaveri: The logging worked but you're right, it looks like `__str__` is used when its available, I'll change this to be `__str__`

sfc-gh-mkeller (approved):
🚢 🚢

PR #807, opened 2021-07-27
SNOW-378268 Python connector S3 PUT Failure Fix

sfc-gh-hchaturvedi:
Description
 
Use s3 Regional URL information communicated from GS gets lost while executing actual PUT command. The driver instead of using the s3 regional URL uses the default S3 URL. This has the potential to go unnoticed in any environment where the S3 default URL has not been blocked. For the same reason, this problem happens consistently in Privatelink env where the default S3 URL is unavailable.

The issue happens because use_s3_regional_url information is not correctly in the upload client class object

### DIFF
--- a/src/snowflake/connector/file_transfer_agent.py
+++ b/src/snowflake/connector/file_transfer_agent.py
@@ -412,10 +412,10 @@ class SnowflakeFileTransferAgent(object):
             self._stage_location_type
         )
         client_meta = SFResourceMeta(
-            storage_util,
-            self._stage_info,
-            self._use_accelerate_endpoint,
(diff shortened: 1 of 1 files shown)


sfc-gh-mkeller (approved):
Please also manually verify the correctness of this code, as we don't have a PL automated test environment

sfc-gh-hchaturvedi:
This fix has been manually verified by myself and @sfc-gh-rturlapati in privatelink and non privatelink env.

sfc-gh-hchaturvedi:
SNOW-378268 S3 PUT Fix

PR #811, opened 2021-07-28
SNOW-414845 Fix Pandas and JSON Parallel Fetch

sfc-gh-kzaveri:
1. What JIRA issue is this PR adressing? Make sure that there is an accompanying issue to your PR.

   Fixes [SNOW-361211](https://snowflakecomputing.atlassian.net/browse/SNOW-361211) with what should be the final performance-related bug. See performance runs 7903, 7911 on the ClientPerfs Dashboard.

   - [X] I am adding a new automated test(s) to verify correctness of my new code
   - [X] I am adding new logging messages
      -  logging when `_parse` is invoked.
   - [ ] I am modyfying authorization mechanisms
   - [ ] I am modyfying OCSP code

   This changes the `_get_pandas_iter` function to return an iterator rather than have it be a generator. Since it must return 
   an iterator, when `create_iter` is submitted to the TPE, the download will happen. 
  Likewise, JSONResultBatch's `_parse` function is changed to return a list rather than be a generator (and hence parsing 
  happens when the function is called rather than when the result is iterated over).

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -17,6 +17,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
+    Generator,
     Iterator,
     List,
     NamedTuple,
@@ -1015,7 +1016,9 @@ class SnowflakeCursor(object):
             self.execute(command, param, _do_reset=False)
         return self
 
-    def _result_iterator(self):
+    def _result_iterator(
+        self,
+    ) -> Union[Generator[Dict, None, None], Generator[Tuple, None, None]]:
         """Yields the elements from _result and raises an exception when appropriate."""
         for _next in self._result:
             if isinstance(_next, Exception):
@@ -1111,7 +1114,7 @@ class SnowflakeCursor(object):
             self._inner_cursor = None
         self._prefetch_hook = None
 
-    def __iter__(self):
+    def __iter__(self) -> Union[Iterator[Dict], Iterator[Tuple]]:
         """Iteration over the result set."""
         # set _result if _result_set is not None
         if self._result is None and self._result_set is not None:
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -1015,14 +1015,9 @@ class SnowflakeCursor(object):
             self.execute(command, param, _do_reset=False)
         return self
 
-    def fetchone(self):
-        """Fetches one row."""
-        if self._prefetch_hook is not None:
-            self._prefetch_hook()
-        if self._result is None and self._result_set is not None:
-            self._result = iter(self._result_set)
-        try:
-            _next = next(self._result)
+    def _result_iterator(self):
+        """Yields the elements from _result and raises an exception when appropriate."""
+        for _next in self._result:
             if isinstance(_next, Exception):
                 Error.errorhandler_wrapper_from_ready_exception(
                     self._connection,
@@ -1030,7 +1025,16 @@ class SnowflakeCursor(object):
                     _next,
                 )
             self._rownumber += 1
-            return _next
+            yield _next
+
+    def fetchone(self) -> Optional[Union[Dict, Tuple]]:
+        """Fetches one row."""
+        if self._prefetch_hook is not None:
+            self._prefetch_hook()
+        if self._result is None and self._result_set is not None:
+            self._result = iter(self._result_set)
+        try:
+            return next(self._result_iterator())
         except StopIteration:
             return None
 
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/result_batch.py
```diff
@@ -453,11 +457,16 @@ def _parse(
...
-                        InterfaceError,
-                        {"msg": msg, "errno": ER_FAILED_TO_CONVERT_ROW_TO_PYTHON_TYPE},
+                    result_list.append(
+                        Error.errorhandler_make_exception(
+                            InterfaceError,
+                            {
+                                "msg": msg,
+                                "errno": ER_FAILED_TO_CONVERT_ROW_TO_PYTHON_TYPE,
+                            },
+                        )
```
sfc-gh-kzaveri, line 467: I need to double check the behaviour when there is an error (since it seems like an exception and the row_result would be added to the result_list).
There are no tests so I'll see what it returns and go from there.

test/unit/test_result_batch.py
```diff
@@ -39,23 +38,30 @@
     OtherHTTPRetryableError,
     ServiceUnavailableError,
 )
-from snowflake.connector.result_batch import MAX_DOWNLOAD_RETRY, JSONResultBatch
+
+try:
+    from snowflake.connector.result_batch import MAX_DOWNLOAD_RETRY, JSONResultBatch
+except ImportError:
+    MAX_DOWNLOAD_RETRY = None
+    JSONResultBatch = None
```
sfc-gh-kzaveri, line 46: Changes in this file are to fix olddriver tests

src/snowflake/connector/cursor.py
```diff
@@ -1015,22 +1015,26 @@ def executemany(self, command, seqparams):
...
+        for _next in self._result:
             if isinstance(_next, Exception):
                 Error.errorhandler_wrapper_from_ready_exception(
                     self._connection,
                     self,
                     _next,
                 )
             self._rownumber += 1
-            return _next
+            yield _next
```
sfc-gh-kzaveri, line 1028: This function is an iterator over `self._result` that raises exceptions as it iterates. This behaviour is used in both `fetchone` and `__iter__` so I pulled it out to a common function here.

src/snowflake/connector/cursor.py
```diff
@@ -1109,9 +1113,10 @@ def reset(self):
 
     def __iter__(self):
         """Iteration over the result set."""
+        # set _result if _result_set is not None
         if self._result is None and self._result_set is not None:
             self._result = iter(self._result_set)
-        return iter(self._result)
+        return self._result_iterator()
```
sfc-gh-kzaveri, line 1119: After setting `self._result` it returns the iterator `self._result_iterator()` which now supports raising exceptions.

sfc-gh-mkeller (changes requested):
Just type-hint problems and a suggestion, otherwise this is good to go by me

src/snowflake/connector/result_batch.py
```diff
@@ -438,8 +440,10 @@ def _load(self, response: "Response") -> List:
 
     def _parse(
         self, downloaded_data
-    ) -> Union[Iterator[Union[Dict, Exception]], Iterator[Union[Tuple, Exception]]]:
+    ) -> Union[List[Union[Dict, Exception]], List[Union[Tuple, Exception]]]:
         """Parses downloaded data into its final form."""
+        logger.debug(f"parsing for result batch of size {self.rowcount}")
```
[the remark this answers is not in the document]
sfc-gh-kzaveri: This is a good idea - I hadn't considered extracting data from the url but this will be useful to identify the batches in the logs. I'll update all the logs.

src/snowflake/connector/result_batch.py
```diff
@@ -258,6 +258,16 @@ def uncompressed_size(self) -> Optional[int]:
     def column_names(self) -> List[str]:
         return [col.name for col in self.schema]
 
+    @property
```
sfc-gh-mkeller, line 261: We should not make this a `property`. We shouldn't `urlsplit` for every log line we do. Let's calculate this in `__init__` and then reuse that.

src/snowflake/connector/result_batch.py
```diff
@@ -258,6 +258,16 @@ def uncompressed_size(self) -> Optional[int]:
...
+    def id(self) -> str:
+        """Returns an id for the chunk."""
+        if self._remote_chunk_info:
+            parsed_url = urlparse(self._remote_chunk_info.url)
+            path_parts = parsed_url.path.split("/")
```
sfc-gh-mkeller, line 266: Tiny bit more efficient
```suggestion
            path_parts = parsed_url.path.rsplit("/", 1)
```

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):
🚢 🚢

PR #837, opened 2021-08-21
SNOW-436322 mark HeartBeatTimer threads as daemon threads

### LINKED ISSUE #836: SNOW-436322: Setting client_session_keep_alive=True hangs Python on exit, even with atexit hooks
After creating a Snowflake connection with `client_session_keep_alive=True` set, all connections (and connection pools / SQLAlchemy Engine objects) must be closed before the program will exit.

Additionally, this can't be done automatically with an [`atexit` hook](https://docs.python.org/3/library/atexit.html), because the hook will not be invoked for the same reason the program won't exit in the first place.

The reason for this is that setting `client_session_keep_alive` will create a non-daemon background thread per connection to keep the connection alive, and the Python process won't exit until all non-daemon threads have completed. Additionally, `atexit` hooks won't be called until all non-daemon threads have completed, since this is what defines a process as ready to exit.

The simplest solution here would be to mark these threads as daemon threads, since they are not performing any system-critical work that should block the program from exiting.

I will send a patch shortly that does this.

   Python 3.8.6

   Linux-5.10.25-linuxkit-x86_64-with-glibc2.2.5

jkinkead:

   Fixes #836 

   This marks non-critical HeartBeatTimer threads as daemon threads (threads that should not keep the Python process from exiting while they are running).

### DIFF
--- a/src/snowflake/connector/time_util.py
+++ b/src/snowflake/connector/time_util.py
@@ -25,6 +25,8 @@ class HeartBeatTimer(Timer):
     def __init__(self, client_session_keep_alive_heartbeat_frequency, f):
         interval = client_session_keep_alive_heartbeat_frequency
         super(HeartBeatTimer, self).__init__(interval, f)
+        # Mark this as a daemon thread, so that it won't prevent Python from exiting.
+        self.daemon = True
 
     def run(self):
         while not self.finished.is_set():


sfc-gh-mkeller:
Tests look good, thank you @jkinkead

sfc-gh-mkeller (approved):

PR #846, opened 2021-09-01
SNOW-443844 Implement AWS signature V4

sfc-gh-mkeller:

   Fixes SNOW-443844

   - [x] I am modifying AWS authorization mechanism

   Right now with the new PUT/GET implementation these actions are inaccessible on `eufrankfurt`. This is because we were using a legacy signing procedure which is not supported on this S3 deployment.
   In this PR I implement the newest signature procedure. See documentation here for implementation details: https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html

### DIFF
--- a/src/snowflake/connector/s3_storage_client.py
+++ b/src/snowflake/connector/s3_storage_client.py
@@ -205,6 +205,7 @@ class SnowflakeS3RestClient(SnowflakeStorageClient):
         region_name: str,
         service_name: str,
         amzdate: str,
+        short_amzdate: str,
         canonical_request_hash: bytes,
     ) -> Tuple[str, str]:
         """Given all the necessary information construct a V4 string to sign.
@@ -216,7 +217,6 @@ class SnowflakeS3RestClient(SnowflakeStorageClient):
         portion of amzdate), region name and service we want to use (this is only s3
         in our case).
         """
-        short_amzdate = amzdate[:8]
         scope = f"{short_amzdate}/{region_name}/{service_name}/aws4_request"
         return (
             "\n".join(
@@ -311,6 +311,7 @@ class SnowflakeS3RestClient(SnowflakeStorageClient):
                 self.region_name,
                 "s3",
                 amzdate,
+                short_amzdate,
                 self._hash_bytes_hex(canonical_request.encode("utf-8")).lower(),
             )
             kDate = self._sign_bytes(
--- a/src/snowflake/connector/s3_storage_client.py
+++ b/src/snowflake/connector/s3_storage_client.py
@@ -4,16 +4,18 @@
 
 from __future__ import division
 
-import base64
+import binascii
+import re
 import xml.etree.cElementTree as ET
 from datetime import datetime
 from io import IOBase
 from logging import getLogger
+from operator import itemgetter
 from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Optional, Tuple, Union
 
 from cryptography.hazmat.primitives import hashes, hmac
 
-from .compat import quote
+from .compat import quote, urlparse
 from .constants import (
     HTTP_HEADER_CONTENT_TYPE,
     HTTP_HEADER_VALUE_OCTET_STREAM,
@@ -40,6 +42,8 @@ ERRORNO_WSAECONNABORTED = 10053  # network connection was aborted
 
 EXPIRED_TOKEN = "ExpiredToken"
 ADDRESSING_STYLE = "virtual"  # explicit force to use virtual addressing style
+
+RE_MULTIPLE_SPACES = re.compile(r" +")
 
 
 class S3Location(NamedTuple):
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/s3_storage_client.py
```diff
@@ -87,84 +92,134 @@ def __init__(
...
     @staticmethod
-    def _has_expired_token(response: requests.Response) -> bool:
+    def _construct_string_to_sign(
+        region_name: str,
+        service_name: str,
+        amzdate: str,
+        canonical_request_hash: bytes,
+    ) -> Tuple[str, str]:
+        short_amzdate = amzdate[:8]
+        scope = f"{short_amzdate}/{region_name}/{service_name}/aws4_request"
```
sfc-gh-hchaturvedi, line 209: What's short_amzdate for ? And can you share how it's different from amzdate?

sfc-gh-mkeller: So from https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html As the first step you generate a canonical request and hash that to generate a string to sign, which we then hash and and include as a header to our HTTP request.

I think your question is best answered by examples:
Example canonical request:
```
GET
/
Action=ListUsers&Version=2010-05-08
content-type:application/x-www-form-urlencoded; charset=utf-8
host:iam.amazonaws.com
x-amz-date:20150830T123600Z

content-type;host;x-amz-date
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

Assume which when hashed gives: `f536975d06c0309214f805bb90ccff089219ecd68b2577efef23edd43b7e1a59`

Subsequently derived string to sign:
```
AWS4-HMAC-SHA256
20150830T123600Z
20150830/us-east-1/iam/aws4_request
f536975d06c0309214f805bb90ccff089219ecd68b2577efef23edd43b7e1a59
```

Now see the blueprint for the string to sign:
```
StringToSign =
    Algorithm + \n +
    RequestDateTime + \n +
    CredentialScope + \n +
    HashedCanonicalRequest
```

Now notice that `CredentialScope` starts with a date in the example. Which is really just the first 8 characters of the `RequestDateTime`.

sfc-gh-mkeller:
The TL;DR of this PR is that we're upgrading from the AWS signature V2 to V4.
Full documentation can be found here: https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html

These are a steps necessary in the process:
1. Generate a canonical request, here's an example:
   ```
   GET
   /
   Action=ListUsers&Version=2010-05-08
   content-type:application/x-www-form-urlencoded; charset=utf-8
   host:iam.amazonaws.com
   x-amz-date:20150830T123600Z
   
   content-type;host;x-amz-date
   e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
   ```
   and then hash it, for this example assume hash: `f536975d06c0309214f805bb90ccff089219ecd68b2577efef23edd43b7e1a59`
2. Generate a key to sign. Please see the details here: https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html (I feel like if I summarized it it wouldn't be much shorter than the actual documentation)
3. With the hash from step 1 generate a string to sign, here's the example for our example:`
   ```
   AWS4-HMAC-SHA256
   20150830T123600Z
   20150830/us-east-1/iam/aws4_request
   f536975d06c0309214f805bb90ccff089219ecd68b2577efef23edd43b7e1a59
   ```
   and sign it with the key generated in step 2.
4. Finally add an `Authorization` header like so: `Authorization: AWS4-HMAC-SHA256 Credential=<access key ID>/<credential scope>, SignedHeaders=<SignedHeaders>, Signature=<signature from step 3>` Where all of these parts come from the previous steps.

If you would like more details about the signing process, please schedule a meeting with me where we can go through the documentation together, since I actually think that the AWS documentation is good once you understand what they are talking about (their docs are definitely written by an engineer who works on this a lot).

src/snowflake/connector/s3_storage_client.py
sfc-gh-hchaturvedi: Thanks for the explanation. Instead of hardcoding `[8:]` can we use an native datetime API and pass in the amzdate to it that returns the date part only? This change would then be needed at all instances where `short_amzdate` is derived in a similar manner.

sfc-gh-mkeller: I believe `short_amzdate = amzdate[:8]` is more efficient than deriving another string from `datetime`.
```pycon
>>> import timeit
>>> import datetime
>>> t = datetime.datetime.utcnow()
>>> amzdate = t.strftime("%Y%m%dT%H%M%SZ")
>>> timeit.timeit("amzdate[:8]", globals=locals())
0.08361914503620937
>>> timeit.timeit("t.strftime('%Y%m%d')", globals=locals())
1.7320209199679084
```
I do see that I do `amzdate[:8]` twice tho, I can get rid of that second one 👍

src/snowflake/connector/s3_storage_client.py
```diff
@@ -87,84 +92,145 @@ def __init__(
...
-        content_type: str = "",
-    ) -> bytes:
-        res = verb + "\n" + content_md5 + "\n" + content_type + "\n"
-        res += amzdate + "\n" + canonicalized_headers + canonicalized_element
-        return res.encode("UTF-8")
+        canonical_uri_parameter: str,
+        query_parts: Dict[str, str],
+        canonical_headers: Optional[Dict[str, Union[str, List[str]]]] = None,
+        payload_hash: str = "",
+    ) -> Tuple[str, str]:
```
sfc-gh-mkeller, line 173: Essentially this function does step 1 from https://github.com/snowflakedb/snowflake-connector-python/pull/846#issuecomment-914643727 signed headers are basically computed in here and they will be necessary for part of the string to sign generation later on.

src/snowflake/connector/s3_storage_client.py
```diff
@@ -87,84 +92,145 @@ def __init__(
...
+        )
 
     @staticmethod
-    def _has_expired_token(response: requests.Response) -> bool:
+    def _construct_string_to_sign(
+        region_name: str,
+        service_name: str,
+        amzdate: str,
+        canonical_request_hash: bytes,
+    ) -> Tuple[str, str]:
```
sfc-gh-mkeller, line 209: This does part of step 3 from my explanation https://github.com/snowflakedb/snowflake-connector-python/pull/846#issuecomment-914643727 the result of this then gets hashed.

src/snowflake/connector/s3_storage_client.py
```diff
@@ -197,48 +263,82 @@ def _send_request_with_authentication_and_retry(
...
         verb: str,
-        resources: str,
         retry_id: Union[int, str],
+        query_parts: Optional[Dict[str, str]] = None,
         x_amz_headers: Optional[Dict[str, str]] = None,
         headers: Optional[Dict[str, str]] = None,
-        content_type: str = "",
-        data: Union[bytes, bytearray, IOBase, None] = None,
+        payload: Union[bytes, bytearray, IOBase, None] = None,
+        unsigned_payload: bool = False,
```
sfc-gh-mkeller, line 271: Here I'm adding this new ability to do an unsigned payload when doing a HTTP request. This was previously not a thing with V2 signatures.

src/snowflake/connector/s3_storage_client.py
```diff
@@ -197,48 +263,82 @@ def _send_request_with_authentication_and_retry(
...
-        if not x_amz_headers:
+        if x_amz_headers is None:
             x_amz_headers = {}
-        if not headers:
+        if headers is None:
             headers = {}
+        if payload is None:
+            payload = b""
+        if query_parts is None:
+            query_parts = {}
```
sfc-gh-mkeller, line 280: Here I define some better default values, for example if `x_amz_headers={}` then that would have redefined it.

src/snowflake/connector/s3_storage_client.py
```diff
@@ -248,18 +348,16 @@ def get_file_header(self, filename: str) -> Optional[FileHeader]:
...
         url = self.endpoint + f"/{path}"
 
-        _resource = self._construct_canonicalized_element(
-            bucket_name=self.s3location.bucket_name, request_uri=path
-        )
         retry_id = "HEAD"
         self.retry_count[retry_id] = 0
         response = self._send_request_with_authentication_and_retry(
-            url, "HEAD", _resource, retry_id
+            url=url, verb="HEAD", retry_id=retry_id
```
sfc-gh-mkeller, line 360: I explicitly named each argument to this function now everywhere

src/snowflake/connector/s3_storage_client.py
```diff
@@ -197,48 +263,82 @@ def _send_request_with_authentication_and_retry(
...
+        x_amz_headers["host"] = parsed_url.hostname
+        if unsigned_payload:
+            x_amz_headers["x-amz-content-sha256"] = UNSIGNED_PAYLOAD
+        else:
+            x_amz_headers["x-amz-content-sha256"] = (
+                SnowflakeS3RestClient._hash_bytes_hex(payload).lower().decode()
+            )
 
-        def generate_authenticated_url_and_args() -> Tuple[bytes, Dict[str, bytes]]:
+        def generate_authenticated_url_and_args_v4() -> Tuple[bytes, Dict[str, bytes]]:
```
sfc-gh-mkeller, line 293: This function does the whole signature process

src/snowflake/connector/storage_client.py
```diff
@@ -186,21 +186,22 @@ def preprocess(self) -> None:
...
+                    f'file_name="{meta.dst_file_name}"'
+                )
+                meta.dst_file_size = 0
+                meta.result_status = ResultStatus.SKIPPED
+                self.preprocessed = True
+                return
+        # Uploading
+        if meta.require_compress:
+            self.compress()
+        self.get_digest()
```
sfc-gh-mkeller, line 204: This was a bug, where we tried to `HEAD` a file even if overwrite was on

src/snowflake/connector/s3_storage_client.py
```diff
@@ -441,25 +529,23 @@ def download_chunk(self, chunk_id: int) -> None:
...
             if response.status_code in (200, 206):
                 self.write_downloaded_chunk(chunk_id, response.content)
             response.raise_for_status()
 
     def transfer_accelerate_config(self) -> bool:
-        url = self.endpoint + "/?accelerate"
-        _resource = self._construct_canonicalized_element(
-            bucket_name=self.s3location.bucket_name, subresource={"accelerate": None}
-        )
+        query_parts = (("accelerate", ""),)
```
sfc-gh-mkeller, line 542: I found that we used to construct these URLs, but then I needed to extract some parts of them later on, query parameters were the ones that were the most complicated ones to parse out.

So instead I switched to having them in the form of `Tuple[Tuple[str, str], ...]`, where the strings refer to each key and value.
This way I don't need to parse them out of the URL and they are easy to turn into a string like `accelerate` in this case, but it also makes me not having to parse the URL again later on.

src/snowflake/connector/s3_storage_client.py
```diff
@@ -87,84 +92,145 @@ def __init__(
...
+
+    @staticmethod
+    def _hash_bytes_hex(_input: bytes) -> bytes:
+        """Convenience function, same as _hash_bytes, but returns result in hex form."""
+        return binascii.hexlify(SnowflakeS3RestClient._hash_bytes(_input))
+
+    @staticmethod
+    def _construct_query_string(
+        query_parts: Tuple[Tuple[str, str], ...],
     ) -> str:
```
sfc-gh-mkeller, line 131: This function simply turns an input of the form `(("key1", "valu1"), ("key2", "value2"),)` to the string `key1=value1&key2=value2`.
It also works with simple end-points like accelerate where `(("accelerate", ""),)` turns into `accelerate`

src/snowflake/connector/s3_storage_client.py
sfc-gh-hchaturvedi: Do you see us making `HTTP` requests at all?

sfc-gh-hchaturvedi: I thought everything was `HTTPS`

src/snowflake/connector/s3_storage_client.py
```diff
@@ -197,48 +263,83 @@ def _send_request_with_authentication_and_retry(
...
+                ("AWS4" + self.credentials.creds["AWS_SECRET_KEY"]).encode("utf-8"),
+                short_amzdate,
             )
-            authorization_header = (  # TODO
-                "AWS " + self.credentials.creds["AWS_KEY_ID"] + ":" + signature.decode()
+            kRegion = self._sign_bytes(kDate, self.region_name)
+            kService = self._sign_bytes(kRegion, "s3")
+            signing_key = self._sign_bytes(kService, "aws4_request")
+
+            signature = self._sign_bytes_hex(signing_key, string_to_sign).lower()
```
sfc-gh-hchaturvedi, line 325: So we're chaining the signing key and generating it at each step. I am assuming this is Amazon API standard.

sfc-gh-hchaturvedi (approved):
LGTM! Just make sure this has been tested and adheres to Amazon V4 standard in every way..

src/snowflake/connector/s3_storage_client.py
sfc-gh-mkeller: Yup https://docs.aws.amazon.com/general/latest/gr/sigv4-calculate-signature.html

sfc-gh-mkeller: > I thought everything was HTTPS

It is, this is an option where we can skip hashing the payload into the canonical request which then gets hashed into the string to sign and makes it to the authorization header.

We use this only during PUTs, so that we don't need to hash each file that we are putting. I checked and this is also what `boto3` does.

In detail what this does is it replaces the payload's hash in the canonical request with `UNSIGNED-PAYLOAD` which than cascades like normal.

PR #861, opened 2021-09-09
SNOW-454169 Bump pandas version to <1.4

sfc-gh-mkeller:
Tests #786


sfc-gh-mkeller:
Tests look good, the FIPS container has been hanging for a few days (I'll be looking into that issue next week), but otherwise this looks good!

PR #874, opened 2021-09-23
SNOW-196953: Include Standard C++ headers

### LINKED ISSUE #406: SNOW-196953: include standard C++ headers everywhere that needs them.
**NOTE:** this issue comes from closed PR #388 . See the discussion and diff there for more background.
<hr>

I ran [`cpplint`](https://github.com/cpplint/cpplint), an open-source linter that checks compliance with the Google Style Guide for C++, over the C++ code in this project.

```shell
pip install cpplint

cpplint \
    --filter=-build/c++11,-build/include_subdir,-build/header_guard,-whitespace/,-line_length \
    --recursive ./src/snowflake/connector/cpp
```

Some of the recommendations from this are only cosmetic and best left to maintainers, but I found at least one class of problem that I think is worth fixing.

There are several places in the code where the necessary headers from the C++ standard library aren't being directly included. For example, `Arrowiterator/BinaryConverter` uses `shared_ptr` but doesn't include `<memory>`. You're not getting any compiler warnings/errors because `<memory>` is probably included by some other header upstream of `Arrowiterator/BinaryConverter.hpp`.

## How would this change improve `snowflake-connector-python`?

The main benefit of this is explained in ["Why Include What You Use?"](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md#allow-refactoring)

> Suppose you refactor foo.h so it no longer uses vectors. You'd like to remove #include <vector> from foo.h, to reduce compile time -- template class files such as vector can include a lot of code. But can you? In theory yes, but in practice maybe not: some other file may be #including you and using vectors, and depending (probably unknowingly) on your #include <vector> to compile. Your refactor could break code far away from you.

> This is most compelling for a very large codebase (such as Google's). In a small codebase, it's practical to just compile everything after a refactor like this, and clean up any errors you see. When your codebase contains hundreds of thousands of source files, identifying and cleaning up the errors can be a project in itself. In practice, people are likely to just leave the #include <vector> line in there, even though it's unnecessary.

> Here, it's the actual 'include what you use' policy that saves the day. If everyone who uses vector is #including <vector> themselves, then you can remove <vector> without fear of breaking anything.

chuckthesoftwarechef:

   Fixes #406 

  Hello Snowflake team!  I have been looking for a way to contribute back to the community and I saw @sfc-gh-mkeller tagged this as a good first issue.  I have run the cpplint command and fixed all of the C++ files so that they include the standard C++ header files that they utilize instead of relying on a transitive include which may break in the future.  

  Please let me know what else I can do to help this merge!
  
  \~- Chuck -\~

### DIFF
--- a/src/snowflake/connector/cpp/ArrowIterator/BinaryConverter.cpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/BinaryConverter.cpp
@@ -3,6 +3,7 @@
 //
 
 #include "BinaryConverter.hpp"
+#include <memory>
 
 namespace sf
 {
--- a/src/snowflake/connector/cpp/ArrowIterator/BinaryConverter.hpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/BinaryConverter.hpp
@@ -7,6 +7,7 @@
 
 #include "IColumnConverter.hpp"
 #include "logging.hpp"
+#include <memory>
 
 namespace sf
 {
--- a/src/snowflake/connector/cpp/ArrowIterator/BooleanConverter.cpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/BooleanConverter.cpp
@@ -3,6 +3,7 @@
 //
 
 #include "BooleanConverter.hpp"
+#include <memory>
 
 namespace sf
 {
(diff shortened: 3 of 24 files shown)


sfc-gh-mkeller (approved):
Tests and code are looking good! 🚢 🚢

sfc-gh-mkeller:
Thank for your PR @chuckthesoftwarechef
I'd invite you to submit a PR that'd add `cpplint` as a pre-commit hook to our code base :)

PR #847, opened 2021-09-01
SNOW-157438 Print SSO URL Upon Failure

sfc-gh-jbahk:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   https://github.com/snowflakedb/snowflake-connector-python/issues/303

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [x] I am adding new logging messages
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #883, opened 2021-09-30
Wait for URL input after SSO login

### LINKED ISSUE #303: SNOW-157438: Feature request: print `sso_url` when unable to open browser when using federated auth
Hi,

When working with tools like Google Colab or other Jupyter-based tools, it often happens that data scientists or engineers are connected to a more powerful instance (generally a VM without any browser) to work with a lot of data. If a company is using federated authentication (such as G Suite) in order to allow their employees to access data, they currently cannot connect to a Snowflake database with such a setup because calling `connect` with `authenticator='externalbrowser'` will fail with `ER_UNABLE_TO_OPEN_BROWSER` error code. A very simple solution that would help this use case (and which is used by other tools, such as https://github.com/googlecolab/colabtools) is to simply print out the `sso_url` when we are unable to open a browser. The user can then click on the link, open it using their own local browser, and authenticate from there.

I tried it myself, by just replacing the lines in the if-block (https://github.com/snowflakedb/snowflake-connector-python/blob/d83d7fbeaf56d42aa433079dc9622737a061436a/auth_webbrowser.py#L108) by a simple `print(sso_url)`, and it worked fine!

It probably also works for other IdP than Google, but I only tried with Google given that this is our setup.

I was wondering what you thought about that, and if you think it would be useful to also implement it for other languages. I find it especially useful for Python as this is a common use case for data scientists/engineers, which are most of the time using Python.

Thanks in advance for considering this simple feature request that would us quite a bit!

iserko:

   Fixes #303

   PR #847 only added printing of the SSO URL however that is not enough when you depend on the response or in this case the URL you are redirected to as part of the SSO Authentication. This PR waits for the user to copy/paste the URL they are redirected to and extracts the token from it.

I can now use Snowflake from Google Colab by installing from my branch
```
pip install "https://github.com/iserko/snowflake-connector-python/archive/refs/heads/SNOW-XXXXXX_wait_for_url_input_after_sso.zip#egg=snowflake-connector-python[secure-local-storage]"
```

Caveat, prior to installing it, you may need to ensure you have the build requirements:
```
pip install "setuptools>=40.6.0" wheel cython "pyarrow>=5.0.0,<5.1.0"
```

Example from Google Colab:
Screenshot 2021-09-30 at 21 10 00

### DIFF
--- a/src/snowflake/connector/auth_webbrowser.py
+++ b/src/snowflake/connector/auth_webbrowser.py
@@ -41,11 +41,6 @@ BUF_SIZE = 16384
 # Snowflake server
 
 
-def _get_url_from_input(self):
-    # function here to facilitate tests better
-    return input("Enter the URL the SSO URL redirected you to: ")
-
-
 class AuthByWebBrowser(AuthByPlugin):
     """Authenticates user by web browser. Only used for SAML based authentication."""
 
@@ -140,7 +135,7 @@ class AuthByWebBrowser(AuthByPlugin):
                     "URL you are redirected to into the terminal."
                 )
                 print(f"URL: {sso_url}")
-                url = _get_url_from_input()
+                url = input("Enter the URL the SSO URL redirected you to: ")
                 self._process_get_url(url)
                 if not self._token:
                     # Input contained no token, either URL was incorrectly pasted,
--- a/src/snowflake/connector/auth_webbrowser.py
+++ b/src/snowflake/connector/auth_webbrowser.py
@@ -41,6 +41,11 @@ BUF_SIZE = 16384
 # Snowflake server
 
 
+def _get_url_from_input(self):
+    # function here to facilitate tests better
+    return input("Enter the URL the SSO URL redirected you to: ")
+
+
 class AuthByWebBrowser(AuthByPlugin):
     """Authenticates user by web browser. Only used for SAML based authentication."""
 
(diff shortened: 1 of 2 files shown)


feluelle:
Nice @iserko I was about to open a PR to fix this too! Thank you very much 😁 This looks great! 🚀 

cc @pecigonzalo

sfc-gh-mkeller:
This works for your use-case. However; I'm not sure that this would work for every SSO user.
I know that in the URL we try to open we embed a port number where we expect the token to be posted to and then we open an HTTP server.
I think doing something like that would be a better user experience than pasting back the url that the SSO redirects you to.

feluelle:
Good point! This was my first thought too, but I also like this approach. We can additionally make the port configurable through an env var. WDYT? If yes, how should we name it? Do you have a convention for those?

oliverholworthy:
One idea that might improve usability could be instead of redirecting to `localhost` and seeing an error message (e.g. "This site can’t be reached").

`http://localhost:49251/?token=rVdZl6LIEn73V3jsx-qSVURPW3...`

image


We could do something like this Google Auth mechanism:

image

where the redirect is a page dedicated to copying the token, which can then be pasted in the token prompt box?

`https://accounts.google.com/o/oauth2/approval/v2/approvalnativeapp?...`

image

Would require a dedicated page for this on the snowflake site I suppose?

oliverholworthy:
This PR only changes the branch where we failed to open the browser, so perhaps this can be merged as it is?

Since It doesn't change the functionality for people when the browser does open successfully. And those that don't have access to a browser will now be able to login 😃   

Improvements to the rendering of the token (e.g. in a dedicated page on the snowflake website) vs copying the URL (this PR) could be made in the future change while maintaining backward compatibility by handling both bare tokens and URLs with a token query parameter in the parse method.

src/snowflake/connector/auth_webbrowser.py
```diff
@@ -122,23 +118,17 @@ def authenticate(self, authenticator, service_name, account, user, password):
...
+                    "please open the following url manually then paste the "
+                    "URL you are redirected to into the terminal."
                 )
-                self.handle_failure(
-                    {
-                        "code": ER_UNABLE_TO_OPEN_BROWSER,
-                        "message": "Unable to open a browser in this environment.",
-                    }
-                )
-                return  # required for test case
```
sfc-gh-mkeller, line 138: I'm sure that this broke some tests

sfc-gh-mkeller (changes requested):

src/snowflake/connector/auth_webbrowser.py
iserko: @sfc-gh-mkeller thanks will take a look (and also rebase)

iserko:

iserko:

iserko:
recheck

src/snowflake/connector/auth_webbrowser.py
iserko: @sfc-gh-mkeller you were absolutely right, I've added some additional test edge cases.
I've added them to the "fail_webbrowser" test case using `pytest.mark.parametrize`. Let me know if you'd like me to use a different approach.

src/snowflake/connector/auth_webbrowser.py
```diff
@@ -258,6 +248,9 @@ def _check_post_requested(self, data):
             ":".join(origin_line.split(":")[1:]).strip(),
         )
 
+    def _process_get_url(self, url):
```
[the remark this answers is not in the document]
iserko: added

PR #906, opened 2021-10-15
SNOW-487250 remove D_GLIBCXX_USE_CXX11_ABI

sfc-gh-mkeller:

   Fixes SNOW-487250

   This build flag was added back in the day for what I think was a workaround for the really old compiler we internally used.
   This flag has been causing issues to some of our customers, who are using Anaconda.
   In this PR I want to remove it as it doesn't seem necessary anymore.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -137,7 +137,6 @@ if _ABLE_TO_COMPILE_EXTENSIONS:
                     ext.extra_compile_args.append("-isystem" + numpy.get_include())
                     if "std=" not in os.environ.get("CXXFLAGS", ""):
                         ext.extra_compile_args.append("-std=c++11")
-                        ext.extra_compile_args.append("-D_GLIBCXX_USE_CXX11_ABI=0")
 
                 ext.library_dirs.append(
                     os.path.join(current_dir, self.build_lib, "snowflake", "connector")


sfc-gh-jbahk (approved):

sfc-gh-mkeller:
@xhochy this might be of interest to you

xhochy:
Oh, this should not have worked for you. You should still need this flag to produce `manylinuxXXXX` wheels. Wheels have not yet made the migration to the newer ABI. Maybe testing doesn't cover this correctly but I had fun in the past with Tensorflow and pyarrow where the order of imports could lead to segmentation faults because of that: https://uwekorn.com/2019/09/15/how-we-build-apache-arrows-manylinux-wheels.html

Regarding Anaconda users: I've contributed the line `if "std=" not in os.environ.get("CXXFLAGS", ""):` here to actually cover this. In conda-forge the compilers set the `CXXFLAGS` to value that contains more information about the build environment. Probably Anaconda is doing the same, it could be that your users are not using the correct compilers (i.e. they use their system ones instead of the Anaconda provided ones). Or they install the snowflake wheel and otherwise use C++ packages from Anaconda or conda-forge. This is sadly doesn't work and isn't fixable without abandoning the all-wheels users.

xhochy:
Maybe this flag is set nowadays in the manylinux images by default and thus this works in CI but may cause issues for users building from source by themselves.

sfc-gh-mkeller:
Hmm, I might have made a mistake a by merging this before tagging you. Let me revert this and re-open the investigation.

ayushdg:
> Regarding Anaconda users: I've contributed the line if "std=" not in os.environ.get("CXXFLAGS", ""): here to actually cover this. In conda-forge the compilers set the CXXFLAGS to value that contains more information about the build environment. Probably Anaconda is doing the same, it could be that your users are not using the correct compilers (i.e. they use their system ones instead of the Anaconda provided ones). Or they install the snowflake wheel and otherwise use C++ packages from Anaconda or conda-forge. This is sadly doesn't work and isn't fixable without abandoning the all-wheels users.

Assuming I'm getting PyArrow and it's dependencies from `Anaconda/conda-forge` am I right in understanding that installing the snowflake connector (wheels) via pip wouldn't work? If so is the only solution to also have a conda package for the connector and/or build from source using conda compilers only?

xhochy:
Yes. An easy reproducer for this:

```
% docker run -ti condaforge/mambaforge /bin/bash
Unable to find image 'condaforge/mambaforge:latest' locally
latest: Pulling from condaforge/mambaforge
a39c84e173f0: Pull complete
c3ac7d55f8df: Pull complete
Digest: sha256:0d86182f5ef56ba9031fec2287536f7f1263d33469d9945f166ba31277062f48
Status: Downloaded newer image for condaforge/mambaforge:latest
(base) root@dfc0baf79233:/# mamba install pyarrow
… omitted …
(base) root@dfc0baf79233:/# pip install snowflake-connector-python
… omitted …
(base) root@dfc0baf79233:/# python
Python 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:20:46)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pyarrow
>>> import snowflake.connector.arrow_iterator
ImportError: /opt/conda/lib/python3.9/site-packages/snowflake/connector/arrow_iterator.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZNK5arrow16KeyValueMetadata5valueEl
```

> If so is the only solution to also have a conda package for the connector and/or build from source using conda compilers only?

Yes, I'm using Snowflake is this setup using the conda-forge package maintained at https://github.com/conda-forge/snowflake-connector-python-feedstock

xhochy:
I'm though a bit confused by the actual error message. `arrow::KeyValueMetadata::value(long) const` doesn't look like something that is related to the ABI differences. This above test uses `pyarrow=5` which should be the same the connector is built against.

ayushdg:
Here's another datapoint for reference which has the same `arrow::KeyValueMetadata::value(long) const` error though in this case I think I'm building from source using my system compiler:
```bash
mamba create --name snowflake-connector -c conda-forge pyarrow=5.0.* python=3.8 -y
conda activate snowflake-connector
pip install git+https://github.com/snowflakedb/snowflake-connector-python.git@c37c0713ddf2eb66c7cf15eb38099040c73c0a0a

python -c "import pyarrow;import snowflake.connector.arrow_iterator"

ImportError: /miniconda3/envs/snowflake-connector/lib/python3.8/site-packages/snowflake/connector/arrow_iterator.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZNK5arrow16KeyValueMetadata5valueEl
```

while this works with `CXXFLAGS`:
```bash
pip uninstall snowflake-connector-python -y
CXXFLAGS="-std=c++14" pip install git+https://github.com/snowflakedb/snowflake-connector-python.git@c37c0713ddf2eb66c7cf15eb38099040c73c0a0a
python -c "import pyarrow;import snowflake.connector.arrow_iterator" # Works
```

sfc-gh-mkeller:
@xhochy is installing 1 dependency with `pip` into a conda environment okay, or is there another way something like this should be done? Should @ayushdg maybe be doing some PATH magic to prevent using the system compiler?

PR #919, opened 2021-10-28
SNOW 372033 jwt expiration issue

sfc-gh-hchaturvedi:

   Fixes SNOW-372033

   - [x] I am adding new logging messages - the logs have been verified to be innocuous. 

  Multiple users have been running into problems with JWT based authentication where the multiple connection requests fail with `Invalid JWT` error.  We have reasons to believe that the most common cause is an expired token. There are two reasons for this issue: 
  
1. Besides the expiration imposed by the JWT, there is another expiration imposed by Snowflake backend, which is `60 seconds` from the time of creation of the JWT, ergo, if a token reaches Snowflake backend `60 second` post it's creation the backend will consider it invalid and respond with a generic error message - `Invalid JWT Token`
2. Snowflake also employs intermittent throttling if a particular backend instance is being inundated with connection requests. This is implemented in the form of `503` HTTP error code. So far the drivers have had a canopy approach to handle `503s` which is just retry the connection. This however doesn't work for JWT because there's high possibility that the JWT has passed its `60 second` lifetime mark during this retry mechanism. The driver can potentially keep retrying connection with an expired token till the backend finally can accept the request only to return an `Invalid JWT Token` error. 

The current fix addresses this problem by moving the connection timeout logic from the Network layer abstraction to the connection layer abstraction within the driver.  Every authenticator now has the option to implement their own timeout handler. The fix implements a custom handler for JWT based auth and a default handler for all other authenticators. 

The JWT handler instead of waiting to receive an error message, restricts the time spent in retrying to a configurable value that defaults to `10 seconds`. After this duration it traps the connection timeout exception to regenerate the token and initiates a retry with a newly generated token. This highly improves the chance of the backend receiving an unexpired token.  The retry attempts are also configurable but default to `10` since the timeout is quite small. 

The fix has been manually tested.


src/snowflake/connector/auth.py
```diff
@@ -517,6 +571,34 @@ def write_temporary_credentials(self, host, user, session_parameters, response):
...
+        if not self._retry_ctx.should_retry():
+            raise OperationalError(
+                msg="Could not connect to Snowflake backend after {} attempt(s)."
+                "Aborting".format(self._retry_ctx.get_current_retry_count()),
+                errno=ER_FAILED_TO_CONNECT_TO_DB,
+            )
+        else:
+            logger.debug(
+                "Hit connection timeout, attempt number {}. Will retry in a bit...".format(
+                    self._retry_ctx.get_current_retry_count()
```
sfc-gh-mkeller, line 596: I find f-strings not only easier to read, but from my testing they seem to be faster too:
```pycon
>>> import timeit
>>> timeit.timeit('"some string: {}".format(\'asd\')')
0.16692085900285747
>>> timeit.timeit('f"some string: {\'asd\'}"')
0.05588944599730894
```

src/snowflake/connector/auth_keypair.py
```diff
@@ -47,7 +54,16 @@ def __init__(self, private_key, lifetime_in_seconds: int = LIFETIME):
         self._private_key = private_key
         self._jwt_token = ""
         self._jwt_token_exp = 0
-        self._lifetime = timedelta(seconds=lifetime_in_seconds)
+        self._lifetime = timedelta(
+            seconds=os.getenv("JWT_LIFETIME_IN_SECONDS", lifetime_in_seconds)
+        )
+        self._jwt_retry_attempts = os.getenv(
+            "JWT_CNXN_RETRY_ATTEMPTS", self.DEFAULT_JWT_RETRY_ATTEMPTS
+        )
```
sfc-gh-mkeller, line 62: I believe that this should be an `int`
```suggestion
        self._jwt_retry_attempts = int(os.getenv(
            "JWT_CNXN_RETRY_ATTEMPTS", self.DEFAULT_JWT_RETRY_ATTEMPTS
        ))
```

src/snowflake/connector/auth.py
```diff
@@ -517,6 +571,34 @@ def write_temporary_credentials(self, host, user, session_parameters, response):
...
         return
 
+    # Default timeout handler. This will trigger if the authenticator
+    # hasn't implemented one. By default we retry on timeouts and use
+    # jitter to deduce the time to sleep before retrying. The sleep
+    # time ranges between 1 and 16 seconds.
+    def handle_timeout(
+        self,
+        authenticator: str,
+        service_name: any,
```
sfc-gh-mkeller, line 581: Shouldn't this be a `str`?
Also, I think [any](https://docs.python.org/3/library/functions.html#any) isn't what you meant anyway

src/snowflake/connector/auth_keypair.py
```diff
@@ -47,7 +54,16 @@ def __init__(self, private_key, lifetime_in_seconds: int = LIFETIME):
         self._private_key = private_key
         self._jwt_token = ""
         self._jwt_token_exp = 0
-        self._lifetime = timedelta(seconds=lifetime_in_seconds)
+        self._lifetime = timedelta(
+            seconds=os.getenv("JWT_LIFETIME_IN_SECONDS", lifetime_in_seconds)
+        )
+        self._jwt_retry_attempts = os.getenv(
+            "JWT_CNXN_RETRY_ATTEMPTS", self.DEFAULT_JWT_RETRY_ATTEMPTS
```
sfc-gh-mkeller, line 61: I think we should signify that this is a static variable by not using `self`
```suggestion
            "JWT_CNXN_RETRY_ATTEMPTS", AuthByKeyPair.DEFAULT_JWT_RETRY_ATTEMPTS
```

src/snowflake/connector/auth.py
sfc-gh-mkeller: Later in this PR you declare this function as
```suggestion
        service_name: Optional[str],
```

src/snowflake/connector/auth_keypair.py
```diff
@@ -131,3 +147,39 @@ def update_body(self, body):
...
+            raise OperationalError(
+                msg="Could not connect to Snowflake backend after {} attempt(s)."
+                "Aborting".format(self._current_retry_count),
+                errno=ER_FAILED_TO_CONNECT_TO_DB,
+            )
+        else:
+            logger.debug(
+                "Hit JWT timeout, attempt {}. Retrying...".format(
+                    self._current_retry_count
+                )
```
sfc-gh-mkeller, line 176:
```suggestion
                f"Hit JWT timeout, attempt {self._current_retry_count}. Retrying..."
```

src/snowflake/connector/auth_keypair.py
```diff
@@ -131,3 +147,39 @@ def update_body(self, body):
...
+        service_name: Optional[str],
+        account: str,
+        user: str,
+        password: Optional[str],
+    ) -> None:
+        if self._current_retry_count > self._jwt_retry_attempts:
+            logger.debug("Exhausted max login attempts. Aborting connection")
+            raise OperationalError(
+                msg="Could not connect to Snowflake backend after {} attempt(s)."
+                "Aborting".format(self._current_retry_count),
```
sfc-gh-mkeller, line 169:
```suggestion
                msg=f"Could not connect to Snowflake backend after {self._current_retry_count} attempt(s). Aborting",
```

src/snowflake/connector/auth_keypair.py
```diff
@@ -47,7 +54,16 @@ def __init__(self, private_key, lifetime_in_seconds: int = LIFETIME):
...
+        self._lifetime = timedelta(
+            seconds=os.getenv("JWT_LIFETIME_IN_SECONDS", lifetime_in_seconds)
+        )
+        self._jwt_retry_attempts = os.getenv(
+            "JWT_CNXN_RETRY_ATTEMPTS", self.DEFAULT_JWT_RETRY_ATTEMPTS
+        )
+        self._jwt_cnxn_wait_time = timedelta(
+            seconds=os.getenv("JWT_CNXN_WAIT_TIME", self.DEFAULT_JWT_CNXN_WAIT_TIME)
+        )
+        self._current_retry_count = 0
```
sfc-gh-mkeller, line 66: This seems to be instantiated by both of the auth classes that you touched, should we add it in a common base class between the two?

src/snowflake/connector/network.py
```diff
@@ -1042,26 +1043,38 @@ def _request_exec(
...
-                err,
-                exc_info=True,
-            )
-            raise RetryRequest(err)
+            parsed_url = parse_url(full_url)
+            if "login-request" in parsed_url.path:
+                logger.debug(
+                    "Hit a timeout error while logging in. Will be handled by "
+                    "authenticator. Ignore the following. Error stack: %s ",
+                    err,
```
sfc-gh-mkeller, line 1063:
```suggestion
                    f"authenticator. Ignore the following. Error stack: {err} ",
```

src/snowflake/connector/network.py
```diff
@@ -1042,26 +1043,38 @@ def _request_exec(
...
+                )
+                raise OperationalError(
+                    msg="ConnectionTimeout occurred. Will be handled by authenticator",
+                    errno=ER_CONNECTION_TIMEOUT,
+                )
+            else:
+                logger.debug(
+                    "Hit retryable client error. Retrying... Ignore the following "
+                    "error stack: %s",
+                    err,
```
sfc-gh-mkeller, line 1074:
```suggestion
                    f"error stack: {err}",
```

src/snowflake/connector/auth.py
```diff
@@ -237,13 +276,28 @@ def authenticate(
             {k: v for (k, v) in body["data"].items() if k != "PASSWORD"},
         )
 
+        # accommodate any authenticator specific timeout requirements here.
+        # login_timeout comes from user configuration.
+        # Between login timeout and auth specific
+        # timeout use whichever value is smaller
+        if getattr(auth_instance, "get_timeout", None) is not None:
```
sfc-gh-mkeller, line 283: Since the actual value never gets used
```suggestion
        if hasattr(auth_instance, "get_timeout"):
```

src/snowflake/connector/connection.py
```diff
@@ -1023,20 +1024,63 @@ def __authenticate(self, auth_instance):
...
+            )
+        except OperationalError as op:
+            logger.debug("Operational Error raised at authentication")
+            if hasattr(auth_instance, "handle_timeout") and (
+                auth_instance.can_handle_exception(op)
+            ):
+                timeout_handler = auth_instance
+            else:
+                # Default handler
+                timeout_handler = auth
```
sfc-gh-mkeller, line 1050: Should we not put the default handler into the base class of the authenticating classes?
Then this grab it from this one class wouldn't be necessary and all auth classes would have `handle_timeout` function by default.

src/snowflake/connector/connection.py
```diff
@@ -1023,20 +1024,63 @@ def __authenticate(self, auth_instance):
...
+                        role=self.role,
+                        passcode=self._passcode,
+                        passcode_in_password=self._passcode_in_password,
+                        mfa_callback=self._mfa_callback,
+                        password_callback=self._password_callback,
+                        session_parameters=self._session_parameters,
+                    )
+                except OperationalError as auth_op:
+                    if auth_op.errno == ER_FAILED_TO_CONNECT_TO_DB:
+                        raise auth_op
```
sfc-gh-mkeller, line 1080:
```suggestion
                        raise
```
Be careful with this, there's a subtle difference between the two:

```pycon
>>> def go() -> None:
...     raise ImportError('something')
... 
>>> try:
...     go()
... except ImportError as e:
...     raise
... 
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in go
ImportError: something
>>> try:
...     go()
... except ImportError as e:
...     raise e
... 
  File "<stdin>", line 4, in <module>
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in go
ImportError: something
```
Notice how `raise e` added an extra frame to the error's stack. Do you think that we should be adding this extra frame?

sfc-gh-mkeller (changes requested):

src/snowflake/connector/auth.py
```diff
@@ -109,13 +111,50 @@
...
 ID_TOKEN = "ID_TOKEN"
 MFA_TOKEN = "MFATOKEN"
+DEFAULT_MAX_CNXN_RETRY_ATTEMPTS = 1
+
+
+class AuthRetryCtx:
+    def __init__(self):
+        self._current_retry_count = 0
+        self._max_retry_attempts = int(
+            getenv("MAX_CNXN_RETRY_ATTEMPTS", DEFAULT_MAX_CNXN_RETRY_ATTEMPTS)
```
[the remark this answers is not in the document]
sfc-gh-hchaturvedi: Ha I assumed it was a common short hand - it stands for `connection`

src/snowflake/connector/connection.py
sfc-gh-hchaturvedi: It would have been a problem if we were still using Python 2 but given that Python 3 also includes the original source of exception, adding the extra frame helps with error tracing better.

src/snowflake/connector/auth.py
```diff
@@ -109,13 +111,50 @@
...
+        )
+        self._backoff = DecorrelateJitterBackoff(1, 16)
+        self._current_sleep_time = 1
+
+    def get_current_retry_count(self) -> int:
+        return self._current_retry_count
+
+    def increment_retry(self) -> None:
+        self._current_retry_count += 1
```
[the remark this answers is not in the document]
sfc-gh-hchaturvedi: I understand but I want to restrict what can be done on and to  `_current_retry_count`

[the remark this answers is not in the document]
sfc-gh-mkeller: But since you are prepending a `_` you're signaling that other classes should not be interacting with it.
Like other classes can still just modify `self._current_retry_count` even if we add these getter and setter.

src/snowflake/connector/auth.py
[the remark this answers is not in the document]
sfc-gh-mkeller: I never got where the `x` came from. I have been using `con` for `connection`.
However I'd prefer if we were consistent over what I think is right. Please check which one would be more consistent and make this call yourself.

src/snowflake/connector/auth.py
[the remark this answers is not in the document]
sfc-gh-hchaturvedi: Sure but having these specific functions (to be precise there's not setter there's just an increment function) communicates the intended actions on the data.

src/snowflake/connector/auth_by_plugin.py
```diff
@@ -37,3 +93,36 @@ def handle_failure(self, ret):
                 "sqlstate": SQLSTATE_CONNECTION_WAS_NOT_ESTABLISHED,
             },
         )
+
+    def handle_timeout(
+        self,
+        authenticator: str,
+        service_name: any,
```
sfc-gh-mkeller, line 100: You haven't fixed this, please see: https://github.com/snowflakedb/snowflake-connector-python/pull/919#discussion_r741351525

sfc-gh-mkeller (changes requested):

src/snowflake/connector/auth_by_plugin.py
sfc-gh-hchaturvedi: Oh yeah about this - the `any` type just signifies that we don't care for the type. It can be _any_ type. I think I should have put everything as _any_ based on that logic. Here's the reference for you https://docs.python.org/3/library/typing.html#typing.Any

sfc-gh-hchaturvedi: the fact that we don't care about the type is re-enforced by the first line in the function definition - I delete all the arguments to prevent any accidental usage.

src/snowflake/connector/auth_by_plugin.py
```diff
@@ -4,13 +4,69 @@
...
+
+    def get_current_retry_count(self) -> int:
+        return self._current_retry_count
+
+    def increment_retry(self) -> None:
+        self._current_retry_count += 1
+
+    def should_retry(self) -> bool:
+        """Decides whether retry should be performed
+        on the connection.
```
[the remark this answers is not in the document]
sfc-gh-hchaturvedi: Are we following Google style guide now everywhere? Is this a requirement for merging?

src/snowflake/connector/auth_by_plugin.py
sfc-gh-hchaturvedi: Updating to Optional

src/snowflake/connector/auth_by_plugin.py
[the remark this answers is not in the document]
sfc-gh-mkeller: We do for doc-strings specifically. @sfc-gh-kwagner and I made this  choice back when we started adding doc-strings. If you'd like we can revisit alternatives, but for now please follow it for doc-strings.

sfc-gh-mkeller (approved):

PR #925, opened 2021-10-29
Allow web authenticator to select its listening port and address

### LINKED ISSUE #303: SNOW-157438: Feature request: print `sso_url` when unable to open browser when using federated auth

pecigonzalo:

   Fixes #303

By allowing users to specify their socket port and address, we can use this from within containers running DBT with SSO authentication.

Example compose configuration:
```
version: "3.8"
services:
  dbt:
    image: some-magic-dbt-image
    environment:
      SF_AUTH_SOCKET_ADDR: 0.0.0.0
      SF_AUTH_SOCKET_PORT: 8080
      BROWSER: /bin/echo
    ports:
      - 8080:8080
    volumes:
      - ~/.dbt:/root/.dbt # Mount your DBT profile
      - ./:/usr/app # Mount your app
      - ~/.cache/snowflake:/root/.cache/snowflake # Mount cache so it does not ask again
```

This also leverages the clever trick in https://github.com/snowflakedb/snowflake-connector-python/issues/303#issuecomment-949988461 to avoid requiring https://github.com/snowflakedb/snowflake-connector-python/pull/883

---

Im not sure this is the right place to put these env vars or and its code but we need a place to set those parameters with env-vars. Let me know if you have a preference.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1435,9 +1435,10 @@ class SnowflakeConnection(object):
         """Checks whether given status is currently running."""
         return status in (
             QueryStatus.RUNNING,
-            QueryStatus.RESUMING_WAREHOUSE,
             QueryStatus.QUEUED,
+            QueryStatus.RESUMING_WAREHOUSE,
             QueryStatus.QUEUED_REPARING_WAREHOUSE,
+            QueryStatus.BLOCKED,
             QueryStatus.NO_DATA,
         )
 
--- a/test/unit/test_connection.py
+++ b/test/unit/test_connection.py
@@ -122,7 +122,7 @@ def test_is_still_running():
         (QueryStatus.RESUMING_WAREHOUSE, True),
         (QueryStatus.QUEUED_REPARING_WAREHOUSE, True),
         (QueryStatus.RESTARTED, False),
-        (QueryStatus.BLOCKED, False),
+        (QueryStatus.BLOCKED, True),
         (QueryStatus.NO_DATA, True),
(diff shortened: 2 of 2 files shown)


pecigonzalo:
Reopened as https://github.com/snowflakedb/snowflake-connector-python/pull/932

PR #927, opened 2021-11-02
SNOW-499499 Make violating NOT NULL constraint raise IntegrityError

### LINKED ISSUE #922: SNOW-499499: Violating NOT NULL constraint should raise IntegrityError instead of ProgrammingError
Trying to save a NULL in a NOT NULL column raises `ProgrammingError`:
```python
>>> import snowflake.connector
>>> conn = snowflake.connector.connect(...)
>>> conn.cursor().execute('CREATE TABLE FOO(BAR char NOT NULL)')
>>> conn.cursor().execute('INSERT INTO FOO VALUES (NULL)')
  File "/home/tim/.virtualenvs/django39/lib/python3.9/site-packages/snowflake/connector/cursor.py", line 777, in execute
    Error.errorhandler_wrapper(
  File "/home/tim/.virtualenvs/django39/lib/python3.9/site-packages/snowflake/connector/errors.py", line 255, in errorhandler_wrapper
  File "/home/tim/.virtualenvs/django39/lib/python3.9/site-packages/snowflake/connector/errors.py", line 310, in hand_to_other_handler
    cursor.errorhandler(connection, cursor, error_class, error_value)
  File "/home/tim/.virtualenvs/django39/lib/python3.9/site-packages/snowflake/connector/errors.py", line 189, in default_errorhandler
    raise error_class(
snowflake.connector.errors.ProgrammingError: NULL result in a non-nullable column
```
It would be consistent with other database drivers to instead raise `IntegrityError`. Some tests in Django's test suite fail because of this difference.

timgraham:

   Fixes #922

   Should be straightforward.

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -55,6 +55,7 @@ from .errorcode import (
 from .errors import (
     DatabaseError,
     Error,
+    IntegrityError,
     InterfaceError,
     NotSupportedError,
     ProgrammingError,
@@ -787,8 +788,10 @@ class SnowflakeCursor:
                 "sqlstate": self._sqlstate,
                 "sfqid": self._sfqid,
             }
+            is_integrity_error = code == "100072"  # NULL result in a non-nullable column
+            error_class = IntegrityError if is_integrity_error else ProgrammingError
             Error.errorhandler_wrapper(
-                self.connection, self, ProgrammingError, errvalue
+                self.connection, self, error_class, errvalue
             )
         return self
 
--- a/test/integ/test_cursor.py
+++ b/test/integ/test_cursor.py
@@ -994,6 +994,17 @@ def test_fetch_out_of_range_timestamp_value(conn, result_format):
             cur.fetchone()
 
 
+@pytest.mark.skipolddriver
+def test_null_in_non_null(conn):
+    table_name = random_string(5, "null_in_non_null")
+    error_msg = "NULL result in a non-nullable column"
+    with conn() as cnx:
+        cur = cnx.cursor()
+        cur.execute(f"create temp table {table_name}(bar char not null)")
+        with pytest.raises(errors.IntegrityError, match=error_msg):
+            cur.execute(f'insert into {table_name} values (null)')
+
+
 @pytest.mark.parametrize("sql", (None, ""), ids=["None", "empty"])
 def test_empty_execution(conn, sql):
     """Checks whether executing an empty string, or nothing behaves as expected."""


src/snowflake/connector/cursor.py
```diff
@@ -778,8 +779,10 @@ def execute(
                 "sqlstate": self._sqlstate,
                 "sfqid": self._sfqid,
             }
+            null_in_non_nullable = err == "NULL result in a non-nullable column"
```
timgraham, line 782: Perhaps it's better to check `code` (which appears to be `'100072'` for this error) if that's how this works.

timgraham:
@sfc-gh-mkeller Should I add a connection parameter to toggle the behavior change?

timgraham:
@sfc-gh-mkeller Is backward compatibility blocking this? Should I add a connection parameter to toggle the behavior change?

sfc-gh-mkeller:
> @sfc-gh-mkeller Is backward compatibility blocking this? Should I add a connection parameter to toggle the behavior change?

Yes, we'd need to pre-announce the Exception class change before we can change it.
But my real problem is that your PR covers only one case, while I think `IntegrityError` is actually a pretty big category. I'd rather do one big breaking change if we could collect all the cases that should raise this type of Exception.

sfc-gh-mkeller:
Maybe @sfc-gh-jbahk could help by collecting more exception cases where database integrity was violated

timgraham:
I haven't encountered any other cases in Django's test suite. As the documentation notes, "Snowflake supports defining and maintaining constraints, but does not enforce them, except for NOT NULL constraints, which are always enforced."

sfc-gh-mkeller:
> I haven't encountered any other cases in Django's test suite. As the documentation notes, "Snowflake supports defining and maintaining constraints, but does not enforce them, except for NOT NULL constraints, which are always enforced."

Nevermind then, in that case the PR is complete.
For future here's the referenced documentation: https://docs.snowflake.com/en/sql-reference/constraints.html

Sorry, for not accepting this earlier @timgraham I totally forgot to post my comment earlier and thanks to @sfc-gh-jbahk for bringing this up to me again. Let me review it in detail then

test/integ/test_cursor.py
```diff
@@ -975,6 +975,15 @@ def test_fetch_out_of_range_timestamp_value(conn, result_format):
             cur.fetchone()
 
 
+@pytest.mark.skipolddriver
+def test_null_in_non_null(conn):
+    with conn() as cnx:
+        cur = cnx.cursor()
+        cur.execute("create table null_in_non_null(bar char not null)")
```
sfc-gh-mkeller, line 982: Please use our table name randomizer instead of using a hard-coded table name.

Also make this a temporary table, that way you won't need drop it in a finally clause.

test/integ/test_cursor.py
```diff
@@ -975,6 +975,15 @@ def test_fetch_out_of_range_timestamp_value(conn, result_format):
...
+        with pytest.raises(errors.IntegrityError):
```
sfc-gh-mkeller, line 983: Please add a match regex here, that way we also verify that the error message hasn't been screwed up in the future.

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):

PR #908, opened 2021-10-15
SNOW-487268 make dbapi.Binary bytes not str

timgraham:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #907

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   I think it's self-explanatory.

PR #972, opened 2021-11-16
Test 908

sfc-gh-mkeller:
Tests #908


timgraham:
Have I understood correctly that I should add  `pytest.mark.skipolddriver` to all of the tests for my bug fixes since they're not going to pass with `snowflake-connector-python==1.9.1`?

sfc-gh-mkeller:
> Have I understood correctly that I should add `pytest.mark.skipolddriver` to all of the tests for my bug fixes since they're not going to pass with `snowflake-connector-python==1.9.1`?

That's right!

PR #1060, opened 2022-02-08
SNOW-514682 fix hanging FIPS tests

sfc-gh-mkeller:

   Fixes SNOW-514682

   FIPS tests have been hanging for a while, this PR tries to fix this


.github/workflows/build_test.yml
```diff
@@ -180,7 +180,7 @@ jobs:
            download_name: linux
          - image_name: macos-latest
            download_name: macos
-         - image_name: windows-latest
+         - image_name: windows-2019
```
sfc-gh-mkeller, line 183: `windows-latest` was recently updated to `windows-2022`
This seems to be causing `test_connection.test_invalid_proxy` to hang on the new Windows machines, I believe that these machines let `localhost:3333` hang instead of refusing connection.
I've filed SNOW-541832 to follow up on this.

test/unit/test_session_manager.py
```diff
@@ -19,7 +28,7 @@
 
 mock_conn = mock.Mock()
 mock_conn.disable_request_pooling = False
-mock_conn._ocsp_mode = lambda: True
+mock_conn._ocsp_mode = lambda: DEFAULT_OCSP_MODE
```
sfc-gh-mkeller, line 31: This was the source of the hanging, during tear-down we tried to do `._ocsp_mode().name` but `bool` doesn't have a `name` method in it. This was then infinitely retried.
We should ramp up the efforts to introduce `mypy` in our code-base to catch errors like this in the future.

sfc-gh-jbahk (approved):

PR #1030, opened 2022-01-13
SNOW-499493 Add Interval Precision to Timestamp

sfc-gh-jbahk:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-499493

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

PR #1104, opened 2022-04-27
Clip nano precision int64_t arrow values

ihor-mozil:
This prevents data from overflows.

   Fixes case 00343142

   This code avoid overflows in arrow table iterator.

### DIFF
--- a/src/snowflake/connector/cpp/ArrowIterator/CArrowTableIterator.cpp
+++ b/src/snowflake/connector/cpp/ArrowIterator/CArrowTableIterator.cpp
@@ -789,7 +789,15 @@ void CArrowTableIterator::convertTimestampColumn(
                 structArray->GetFieldByName(sf::internal::FIELD_NAME_EPOCH))->Value(rowIdx);
               int32_t fraction = std::static_pointer_cast<arrow::Int32Array>(
                 structArray->GetFieldByName(sf::internal::FIELD_NAME_FRACTION))->Value(rowIdx);
-              val = epoch * sf::internal::powTenSB4[9] + fraction;
+              // Clip the epoch to max/min values avoiding overflows
+              int powTenSB4 = sf::internal::powTenSB4[9];
+              if (epoch > (INT64_MAX / powTenSB4)) {
+                val = INT64_MAX;
+              } else if (epoch < (INT64_MIN / powTenSB4)) {
+                val = INT64_MIN;
+              } else {
+                val = epoch * powTenSB4 + fraction;
+              }
             }
             break;
           default:


ihor-mozil:

ihor-mozil:
recheck

ihor-mozil:
recheck

sfc-gh-mkeller:
When I check out this code the test in #1108 is still giving me incorrect values:
```
LOW_TS: timestamp[ns] not null
HIGH_TS: timestamp[ns] not null
----
LOW_TS: [[1677-09-21 00:12:43.145224192]]
HIGH_TS: [[2262-04-11 23:47:16.854775807]]
[(datetime.datetime(1400, 1, 1, 1, 2, 3), datetime.datetime(9999, 1, 1, 1, 2, 3))]
```
These timstamps are less wrong, but they don't fully seem to solve the issue

ihor-mozil:
@sfc-gh-mkeller How do you propose to fix this?

sfc-gh-mkeller:
In https://github.com/snowflakedb/snowflake-connector-python/pull/1030 where we actually switched from constructing a `datetime.DateTime` with 2 integers instead of 1 float for all other code-paths. 
I think the same should be done to `CArrowIterator.cpp`

ihor-mozil:
Ok, how to do plan to overcome int64 parquet file nano precision limitation?
https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp
The documentation says:
```Due to the range of the int64 type, timestamps using the NANOS unit can only represent values between 1677-09-21 00:12:43 and 2262-04-11 23:47:16. Values outside of this range can not be represented with the NANOS unit. (Other precisions have similar limits but those are outside of the domain for practical everyday usage.)```

**Comment by sfc-gh-mkeller** [comment]
We don't use Parquet for fetching results, Snowflake has it's own representation of timestamps in Arrow format. Which is why the code you changed has 2 integers that it combines into a single float.
Snowflake gives us 1-3 ints for each timestamp. The whole number part of the timestamp, the decimal part of the timestamp and the timezone offset.

You can already our implementation work well for timestamps outside of the range you mention. In #1108 when you `.fetchall()` we correctly handle those timestamps.

**Comment by ihor-mozil** [comment]
If I correctly understand the documentation r.fetch_arrow_all() https://docs.snowflake.com/en/user-guide/python-connector-api.html#fetch_arrow_all returns PyArrow table which is apache/parquet-format.
This format has limitations described in the link I sent earlier.
I don't understand how the nano precision column in arrow table can hold values beyond the documented range.
I'm not sure if it's possible to squeeze two integer values into single pyarrow column. Can you please elaborate on that?

**Comment by sfc-gh-mkeller** [comment]
> If I correctly understand the documentation r.fetch_arrow_all() [docs.snowflake.com/en/user-guide/python-connector-api.html#fetch_arrow_all](https://docs.snowflake.com/en/user-guide/python-connector-api.html#fetch_arrow_all) returns PyArrow table which is apache/parquet-format.

`fetch_arrow_all()` does return a `pyarrow.Table`, but parquet is different from arrow.
I'm no expert of Parquet, so instead let me point you to the original author of Arrow's explanation: https://stackoverflow.com/a/56481636


> This format has limitations described in the link I sent earlier.
> I don't understand how the nano precision column in arrow table can hold values beyond the documented range.

I'm sure this is an issue with Parquet, but we don't use it

> I'm not sure if it's possible to squeeze two integer values into single pyarrow column. Can you please elaborate on that?

You're right about this, however note that our C-extension does more than read the results' Arrow representation and serve that.
The code you modified does actually implement some conversion on the data before converting it to `pyarrow.Table` and returning the result to the user.

**Comment by ihor-mozil** [comment]
Here is a simple python test:
```
import pyarrow as pa
ns_table = pa.Table.from_arrays(
        [pa.array(['9999-12-31', '1400-01-01']
).cast('timestamp[ns]')], names = ['ns_timestamps'])
print(ns_table)
```
Results:
```
pyarrow.Table
ns_timestamps: timestamp[ns]
----
ns_timestamps: [[1816-03-29 05:56:08.066277376,1984-07-21 23:34:33.709551616]]
```

**Comment by ihor-mozil** [comment]
> I'm no expert of Parquet, so instead let me point you to the original author of Arrow's explanation: https://stackoverflow.com/a/56481636

Both use int64 to store data. Sorry for mixing up the terms.

**Comment by sfc-gh-mkeller** [comment]
It seems like I misunderstood your issue, in #1108 you mention that the timestamps: `1400-01-01 01:02:03` and `9999-01-01 01:02:03` are fetched incorrectly. Those timestamps have no sub-second parts.

I apologize, some of things I talked about are not related to the actual issue.

It does look like [arrow supports nano second timestamps](https://arrow.apache.org/docs/python/timestamps.html), but it doesn't require it.
In the case you posted I believe that since there is no nanosecond precision needed your example timestamp should fit into a `pyarrow.Table`. Please see:
```pycon
>>> import datetime
>>> import pyarrow as pa
>>> pa.array([datetime.datetime(1400, 1, 1, 1, 2, 3), datetime.datetime(9999, 1, 1, 1, 2, 3)], type=pa.timestamp('s'))
<pyarrow.lib.TimestampArray object at 0x10f15fac0>
[
  1400-01-01 01:02:03,
  9999-01-01 01:02:03
]
>>> pa.array([datetime.datetime(1400, 1, 1, 1, 2, 3), datetime.datetime(9999, 1, 1, 1, 2, 3)], type=pa.timestamp('ms'))
<pyarrow.lib.TimestampArray object at 0x10f15fb20>
[
  1400-01-01 01:02:03.000,
  9999-01-01 01:02:03.000
]
>>> pa.array([datetime.datetime(1400, 1, 1, 1, 2, 3), datetime.datetime(9999, 1, 1, 1, 2, 3)], type=pa.timestamp('ns'))
  File "pyarrow/array.pxi", line 315, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 39, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 143, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 99, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Could not convert datetime.datetime(1400, 1, 1, 1, 2, 3) with type datetime.datetime: out of bounds for nanosecond resolution
```

In an ideal world we should handle these cases instead of just preventing overflows.

> In my opinion gracefully handle overflows is better than having problems with dates comparison when '9999-12-31' becomes less than a '1400-01-01' after converting to pyarrow.

We're in total agreement here. But your change is a breaking change already, so we'd need to pre-announce it and allow some time for all users to learn of this change before we actually release it. I would like to fully flush out our solution before we change anything.

Since `pyarrow` supports more than microsecond precision what was done in #1030 is not a full solution to this issue, but what was done in it should get us most of the way there still.

ihor-mozil:
I'm sorry I still don't understand what is the proposal to fix this.

> It does look like [arrow supports nano second timestamps](https://arrow.apache.org/docs/python/timestamps.html), but it doesn't require it.

I'm not sure who/what is **it** that does not require nano precision timestamps.

[Snowflake timestamp](https://docs.snowflake.com/en/sql-reference/data-types-datetime.html#timestamp) the default precision is 9 (nanoseconds). `cursor.fetch_arrow_all()` should fetch the timestamp into nano precision pyarrow column. the types conversion is here: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/result_batch.py#L665

So back to the problem: `1400-01-01` snowflake timestamp can not be fit into int64 - pyarrow.timestamp('ns') column which is used by pyarrow for nano precision timestamps.

> In an ideal world we should handle these cases instead of just preventing overflows.

> I would like to fully flush out our solution before we change anything.

So the question is what to expect:
- Option 1: pyarrow.timestamp('ns') handles dates like `1400-01-01` or `9999-21-31` but how???
- Option 2: Clip values to the max/min int64 values `1400-01-01` becomes `1677-09-21`
- Option 3: leave overflows in the code `1400-01-01` becomes `1984-07-22`
- Option 4: Change the pyarrow timestamp column to timestamp('ms') default and document that fetch_arrow_all() does not support nanoseconds, precision nanoseconds will be converted to milliseconds (but this is even bigger breaking change...)
- Other options??

sfc-gh-mkeller:
> I'm not sure who/what is it that does not require nano precision timestamps.

Arrow doesn't. In my last comment I posted some code that shows that your timestamps do actually fit into a `pyarrow.Table` if you don't use nanosecond precision.
I personally think that Snowflake's back-end should be smart enough to to realize that your timestamps don't need nanosecond precision and it should serve them at second precision and therefore we should be able to fit them into a `pyarrow.Table`. To verify this we'll need to talk to our back-end team.

> Option 1: pyarrow.timestamp('ns') handles dates like 1400-01-01 or 9999-21-31 but how???

This isn't an option, you were right. I just didn't understand the issue previously.

> Option 2: Clip values to the max/min int64 values 1400-01-01 becomes 1677-09-21

I think we should be raising errors in the cases that you have identified (just like arrow does in my last comment).

If I was a Snowflake customer and I was fetching `1400-01-01 01:02:03.999999999` I'd expect it to show up as exactly as that, I'd say that any other result would be an error from my perspective. Now we can absolutely make this configurable, it sounds like you would rather have it snap to the smallest/largest timestamps instead and we should allow you to do that.

However; by default I say we should be raising an error for each of these timestamps.

> Option 3: leave overflows in the code 1400-01-01 becomes 1984-07-22

We both agree that this is not an option

> Option 4: Change the pyarrow timestamp column to timestamp('ms') default and document that fetch_arrow_all() does not support nanoseconds, precision nanoseconds will be converted to milliseconds (but this is even bigger breaking change...)

I don't think this would solve anything. It'd just make the default case less likely to error.

sfc-gh-yixie:
Created PR #1204 to fix this problem.

ihor-mozil:
Thanks @sfc-gh-mkeller, @sfc-gh-yixie. I'm closing this one since the new PR looks like a better fix to me. I hope it will not take much time to merge the https://github.com/snowflakedb/snowflake-connector-python/pull/1204

sfc-gh-yixie:
@ihor-mozil we'll merge it this week. It will be a part of the release of Aug 2022.

PR #1117, opened 2022-05-04
DATA-1339 Set application name based on streamlit being in sys.modules

sfc-gh-zblackwood:

   https://snowflakecomputing.atlassian.net/browse/DATA-1339

   In order to correctly attribute connector usage to streamlit, check sys.modules for the presence of "streamlit" and if found, and not set otherwise, use that to set the application appropriately.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -268,8 +268,11 @@ class SnowflakeConnection:
 
         self.heartbeat_thread = None
 
-        if "application" not in kwargs and ENV_VAR_PARTNER in os.environ.keys():
-            kwargs["application"] = os.environ[ENV_VAR_PARTNER]
+        if "application" not in kwargs:
+            if ENV_VAR_PARTNER in os.environ.keys():
+                kwargs["application"] = os.environ[ENV_VAR_PARTNER]
+            elif "streamlit" in sys.modules:
+                kwargs["application"] = "streamlit"
 
         self.converter = None
         self.__set_error_attributes()
--- a/test/unit/test_connection.py
+++ b/test/unit/test_connection.py
@@ -20,7 +20,7 @@ except ImportError:
     QueryStatus = None
 
 
-def fake_connector() -> snowflake.connector:
+def fake_connector() -> snowflake.connector.SnowflakeConnection:
     return snowflake.connector.connect(
         user="user",
         account="account",
--- a/test/unit/test_connection.py
+++ b/test/unit/test_connection.py
@@ -137,16 +137,13 @@ def test_is_still_running():
         )
 
 
-@pytest.mark.skipolddriver
-@patch("snowflake.connector.network.SnowflakeRestful._post_request")
-def test_partner_env_var(mockSnowflakeRestfulPostRequest):
-    PARTNER_NAME = "Amanda"
+@pytest.fixture
+def mock_post_requests(monkeypatch):
+    request_body = []
 
-    request_body = {}
-
-    def mock_post_request(url, headers, json_body, **kwargs):
+    def mock_post_request(request, url, headers, json_body, **kwargs):
         nonlocal request_body
-        request_body = json.loads(json_body)
+        request_body.append(json.loads(json_body))
         return {
             "success": True,
             "message": None,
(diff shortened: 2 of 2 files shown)


src/snowflake/connector/connection.py
```diff
@@ -268,8 +268,11 @@ def __init__(self, **kwargs):
 
         self.heartbeat_thread = None
 
-        if "application" not in kwargs and ENV_VAR_PARTNER in os.environ.keys():
-            kwargs["application"] = os.environ[ENV_VAR_PARTNER]
+        if "application" not in kwargs:
+            if ENV_VAR_PARTNER in os.environ.keys():
+                kwargs["application"] = os.environ[ENV_VAR_PARTNER]
+            elif "streamlit" in sys.modules:
```
sfc-gh-yixie, line 274: This might me inaccurate for two reasons?
1. "streamlit" won't count in some cases even though a connection is created from a streamlit.
2. "streamlit" is used as long as it's in sys.modules. If a connection is created but it is unrelated to streamlit, it will still be counted.

Is it possible to pass in something when streamlit creates a connection? This is more accurate I think.

sfc-gh-mkeller: @sfc-gh-yixie it looks like you were not invited to the meeting yesterday.
Streamlit does not interact with any connectors, users need to manually pipe the data into it and open their own connections. Due to this we couldn't find any other way to do track usage other than what's in this PR.

We all understand that your first comment is a real issue and there's no solution to it unfortunately, but we saw no other way to track Streamlit usage.

test/unit/test_connection.py
```diff
@@ -172,3 +173,39 @@ def mock_post_request(url, headers, json_body, **kwargs):
...
+    with patch.dict(sys.modules, {"streamlit": "foo"}):
+        # connection
+        con = snowflake.connector.connect(
+            user="user",
+            account="account",
+            password="testpassword",
+            database="TESTDB",
+            warehouse="TESTWH",
+        )
+        assert con.application == "streamlit"
```
sfc-gh-mkeller, line 209: This same change could be done to the original test too
```suggestion
        assert snowflake.connector.connect(
            user="user",
            account="account",
            password="testpassword",
            database="TESTDB",
            warehouse="TESTWH",
        ).application == "streamlit"
```

test/unit/test_connection.py
```diff
@@ -172,3 +173,39 @@ def mock_post_request(url, headers, json_body, **kwargs):
         assert con.application == PARTNER_NAME
 
     assert request_body["data"]["CLIENT_ENVIRONMENT"]["APPLICATION"] == PARTNER_NAME
+
+
+@pytest.mark.skipolddriver
+@patch("snowflake.connector.network.SnowflakeRestful._post_request")
+def test_imported_module(mockSnowflakeRestfulPostRequest):
```
sfc-gh-mkeller, line 180: Please parametrize the already existing test

sfc-gh-mkeller (changes requested):

sfc-gh-kwagner (approved):
LGTM as long as the test comments are addressed

sfc-gh-zblackwood:
@sfc-gh-kwagner @sfc-gh-mkeller I created a fixture to be reused by both tests. Let me know if that looks good.

sfc-gh-mkeller:
@sfc-gh-zblackwood  I'm not convinced that a fixture is the right choice the way things are. The current implementation (maintaining a list of requests) suggests that you'd like this to be reused by other tests, but there's no way to customize the responses it returns. Therefore I don't think that it'd be useful for any other tests.

I do think that having a fixture like this would be useful for other unit tests though. Could we make the responses customizable? It'd be like our own [VCR.py](https://github.com/kevin1024/vcrpy)

src/snowflake/connector/connection.py
sfc-gh-yixie: Thanks for the context.

test/unit/test_connection.py
```diff
@@ -19,45 +20,43 @@
...
-                    "token": "TOKEN",
-                    "masterToken": "MASTER_TOKEN",
-                    "idToken": None,
-                    "parameters": [
-                        {"name": "SERVICE_NAME", "value": "FAKE_SERVICE_NAME"}
-                    ],
-                },
-            }
-        return ret
+def fake_connector() -> snowflake.connector:
```
sfc-gh-mkeller, line 23:
```suggestion
def fake_connector() -> snowflake.connector.SnowflakeConnection:
```

sfc-gh-mkeller (approved):

Issue #1115, opened 2022-05-03
SNOW-584971: fetch_pandas_all returns JSON columns as string not dict


kbroughton:
I believe this is similar to several other issues and PRs:

* https://github.com/snowflakedb/snowflake-connector-python/issues/1069
* https://github.com/snowflakedb/snowflake-connector-python/issues/828
* https://github.com/snowflakedb/snowflake-connector-python/pull/1059

1. What version of Python are you using?

python 3.9.7

2. What operating system and processor architecture are you using?

Linux-5.10.47-linuxkit-x86_64-with-glibc2.31

3. What are the component versions in the environment (`pip freeze`)?
Built wheel using:
ci/build_docker.sh
which generates
snowflake_connector_python-2.7.6-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl

4. What did you do?

Create a DB with json data.
I check that it is being stored in a JSON friendly format because i can query the data using `:` dotted access

```python
cursor = ctx.cursor().execute(f"""
    SELECT 
      *
    FROM MYDATASET.PUBLIC.MYTABLE_VARIANT
    WHERE DATA:jsonPayload:bytes_sent > 3500
    LIMIT 2
    """)
result = cursor.fetch_pandas_all()
print(type(result['DATA'].iloc[0]), result['DATA'].iloc[0])
```

The output shows that it is <class 'str'>.
```
<class 'str'> {
  "insertId": "sb2haif1xnn2k",
  "jsonPayload": {
    "bytes_sent": "0",
    "connection": {
      "dest_ip": "2xx.85.147.95",
      "dest_port": "443",
      "protocol": "6",
      "src_ip": "10.128.0.xx",
      "src_port": "48516"
    },
```

5. What did you expect to see?

I expected that the dataframe column for DATA would automatically convert to JSON, in this case: <class 'dict'>.

If I use the Snowflake UI, I can copy the DATA column and it pastes as JSON. It also displays as JSON.

6. Can you set logging to DEBUG and collect the logs?

I don't think this is necessary since this is less a bug than a feature request. However, i marked it as a bug because I think it is a bug with respected to everybody's expectation of how it would work, and related to the many issues raised over fetch_pandas_all() not respecting schema.

Issue #1120, opened 2022-05-05
SNOW-586325: Please add instructions on how to run tests


kbroughton:
## What is the current behavior?

Pull requests template requires users to add tests for new code. However, there is no explanation of how to run tests.
In my case, I found ci/tests_darwin.sh but there are a ton of variables that need to be defined and no sample.env.sh to explain what variables are required/optional.  For example, you have config files in 
```
$ ls .github/workflows/parameters/public
parameters_aws.py.gpg   parameters_azure.py.gpg parameters_gcp.py.gpg
```
which are decrypted, but since we can't decrypt them, we have to reverse engineer what the file is like to set the variables to run the tests.
```
gpg --quiet --batch --yes --decrypt --passphrase="${PARAMETERS_SECRET}" ${PARAMS_FILE} > test/parameters.py
```

## What is the desired behavior?

There should be clear instructions on how to run the tests. Ideally, support an all-local test run for someone who just wants to make a one-time-fix PR.

## How would this improve `snowflake-connector-python`?
PRs requires tests, but nobody can run tests to check that their new tests work.

sfc-gh-mkeller:
I think what you're looking for is here https://github.com/snowflakedb/snowflake-connector-python/tree/main/test#running-tests
We should have clear instructions on how to create your own `parameters.py` file and how to run the tests too there

sfc-gh-aling:
To clean up and re-prioritize bugs and feature requests we are closing all issues older than 6 months as of March 1, 2023. If there are any issues or feature requests that you would like us to address, please re-create them. For urgent issues, opening a support case with this link [Snowflake Community](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge) is the fastest way to get a response.

Issue #1069, opened 2022-03-02
SNOW-554970: fetch_pandas_all does not respect date format set by DATE_OUTPUT_FORMAT session parameter


sfc-gh-jhunt:
Please answer these questions before submitting your issue. Thanks!

1. What version of Python are you using?

Python 3.9.7 (running under Anaconda 4.11.0)

2. What operating system and processor architecture are you using?

macOS-10.16-x86_64-i386-64bit

3. What are the component versions in the environment (`pip freeze`)?

Output of conda list (I presume this is OK, as pip freeze in conda gives odd output)

**Name                    Version                   Build  Channel**
appnope                   0.1.2           py39hecd8cb5_1001  
argon2-cffi               20.1.0           py39h9ed2024_1  
asn1crypto                1.4.0                    pypi_0    pypi
async_generator           1.10               pyhd3eb1b0_0  
attrs                     21.2.0             pyhd3eb1b0_0  
backcall                  0.2.0              pyhd3eb1b0_0  
blas                      1.0                         mkl  
bleach                    4.0.0              pyhd3eb1b0_0  
bottleneck                1.3.2            py39he3068b8_1  
brotli                    1.0.9                hb1e8313_2  
ca-certificates           2021.9.30            hecd8cb5_1  
certifi                   2021.10.8        py39hecd8cb5_0  
cffi                      1.14.6           py39h2125817_0  
charset-normalizer        2.0.12                   pypi_0    pypi
cryptography              36.0.1                   pypi_0    pypi
cycler                    0.10.0           py39hecd8cb5_0  
debugpy                   1.4.1            py39h23ab428_0  
decorator                 5.1.0              pyhd3eb1b0_0  
defusedxml                0.7.1              pyhd3eb1b0_0  
entrypoints               0.3              py39hecd8cb5_0  
fonttools                 4.25.0             pyhd3eb1b0_0  
freetype                  2.10.4               ha233b18_0  
icu                       58.2                 h0a44026_3  
idna                      3.3                      pypi_0    pypi
importlib-metadata        4.8.1            py39hecd8cb5_0  
importlib_metadata        4.8.1                hd3eb1b0_0  
intel-openmp              2021.3.0          hecd8cb5_3375  
ipykernel                 6.4.1            py39hecd8cb5_1  
ipython                   7.27.0           py39h01d92e1_0  
ipython_genutils          0.2.0              pyhd3eb1b0_1  
[...]

Issue #828, opened 2021-08-16
SNOW-427759: fetch_pandas_all not using int64 for NUMBER(38, 0)


jonashaag:
Connector version 2.5.1.

`X` is a `NUMBER(38, 0)` column with values that fit into `int8`. It is fetched as decimal number rather than int64, unless you explicitly cast to `NUMBER(18,  0)`:

```
# eng is a SQLAlchemy engine object
In [23]: res = eng.execute("select X from table(result_scan('...'))")
In [24]: res.cursor._result._fetch_arrow_all()
Out[24]:
pyarrow.Table
X: decimal(38, 0)

In [27]: cur = eng.execute("select X::number(18,0) from table(result_scan('...'))")
In [28]: cur.cursor._result._fetch_arrow_all()
Out[28]:
pyarrow.Table
LiteralRedacted0: int64
```

jonashaag:
Aha, this only seems to happen when doing a `result_scan`.

jonashaag:
Here's a minimal example.

```pycon
# eng is SQLAlchemy engine object
>>> eng.execute("create table xxx.snow427759 (x number(38,0))")
>>> eng.execute("insert into xxx.snow427759 (x) values (1)")
>>> res = eng.connect().execute("select * from xxx.snow427759")
>>> res.cursor._result._fetch_arrow_all()
pyarrow.Table
X: int8
>>> res2 = eng.connect().execute(f"select * from table(result_scan('{res.cursor.sfqid}'))")
>>> res2.cursor._result._fetch_arrow_all()
pyarrow.Table
X: decimal(38, 0)
```

sfc-gh-mkeller:
@jonashaag I tried reproducing your issue , I found that the following worked as expected.
```python
import snowflake.connector
import pyarrow
config = {...}

with snowflake.connector.connect(
    **config,
) as cnx:
    with cnx.cursor() as cur:
        cur.execute("select 1::number(38,0) as x")
        res = cur.fetch_arrow_all()
        assert pyarrow.types.is_int8(res["X"].type)
        sfqid = cur.sfqid
        cur.execute(f"select * from table(result_scan('{sfqid}'))")
        res = cur.fetch_arrow_all()
        assert pyarrow.types.is_int8(res["X"].type)
```

I tried using a table and also running the `result_scan` from a brand new connection and I am unable to reproduce your issue.
Could you see if you could reproduce the issue with our Python connector only?

jonashaag:
[...]

PR #1121, opened 2022-05-05
Make fetch all pandas json friendly

kbroughton:

Fixes https://github.com/snowflakedb/snowflake-connector-python/issues/1115

Related to other issues where schema/column types are not being respected

* https://github.com/snowflakedb/snowflake-connector-python/issues/1069
* https://github.com/snowflakedb/snowflake-connector-python/issues/828

 [x] I am adding a new automated test(s) to verify correctness of my new code
      However, due to https://github.com/snowflakedb/snowflake-connector-python/issues/1120 I haven't been able to run the added tests.
 [x] I am adding new logging messages Debug logging for trying to convert column types in _fetch_pandas_all
 [] I am modifying authorization mechanisms
 [] I am adding new credentials
 [] I am modifying OCSP code
 [] I am adding a new dependency

* result_set.py
Add the following function
```python
    def _convert_string_to_json(self, dataframe: pandas.DataFrame) -> pandas.DataFrame:
        """Converts string columns to JSON columns.
        This is necessary for the JSON columns in the dataframe to be
        serialized properly."""
```
and use it in _fetch_pandas_all()

* test_pandas_tools.py
```python
def test_json_string_in_db_to_dict_in_datafram_query
    """Tests whether fetch_pandas_all can return a VARIANT field as an object not a string."""
```


src/snowflake/connector/pandas_tools.py
```diff
@@ -225,6 +228,12 @@ def write_pandas(
         parquet_columns = "$1:" + ",$1:".join(f'"{c}"' for c in df.columns)
     else:
         parquet_columns = "$1:" + ",$1:".join(df.columns)
+    parquet_columns = ",".join(
+        [
+            "$1:"+x if x not in column_formats else column_formats[x].substitute(col=x)
+            for x in df.columns
+        ]
```
sfc-gh-turbaszek, line 235:
```suggestion
        "$1:" + x if x not in column_formats else column_formats[x].substitute(col=x)
        for x in df.columns
```
I think list comprehension can be skipped and you can use generator.

src/snowflake/connector/result_set.py
```diff
@@ -180,9 +181,24 @@ def _fetch_pandas_batches(self, **kwargs) -> Iterator[pandas.DataFrame]:
...
+
+        This is necessary for the JSON columns in the dataframe to be
+        serialized properly.
+        """
+        for col in dataframe:
+            if dataframe[col].dtype == "object":
+                try:
+                    dataframe[col] = dataframe[col].apply(loads)
+                    logger.debug(f"Converted column {col} to json")
+                except Exception as e:
```
sfc-gh-turbaszek, line 195: Do we expect some specific error? Using `Exception` is generally discouraged if possible

kbroughton: I see one of these two:

the JSON object must be str, bytes or bytearray, not dict
TypeError

or
Expecting value: line 1 column 1 (char 0)
JSONDecodeError

The first one occurs if the data has already been converted to json somehow. I think I only see it because I have the modified connector installed.

The JSONDecodeError is any type that gets converted to PyObject like timestamps etc that aren't json.

So I could `execpt JSONDecodeError as e`
and that would probably do it. If there were a lib that inspects if a string is convertible to JSON, i could do that instead of the try/except.

src/snowflake/connector/pandas_tools.py
kbroughton: fixed

src/snowflake/connector/result_set.py
kbroughton: fixed

kbroughton:

PR #966, opened 2021-11-12
SNOW-499804 Add Geography Types

sfc-gh-kwagner:
Description

Added support for geoJson to show up as GEOGRAPHY when enabled

Testing

Added testing for when it is disabled and enabled

Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-499804

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Recognizes geoJson as a GEOGRAPHY type in the connector

PR #1158, opened 2022-05-27
SNOW-592647 consolidate definitions and resolve circular dependency issues

sfc-gh-mkeller:

   Fixes SNOW-592647

   In this PR I fix an issue that was introduced in #966 where a new type was added to `snowflake.connector.constants.FIELD_TYPES`, but this type was not added to `FIELD_TYPE_TO_PA_TYPE` defined in `snowflake.connector.result_batch.ArrowResultBatch._create_empty_table`.
   In this PR I move the later definition into the former to prevent them drifting apart in the future.
   I had to then move around definitions and import statements to prevent circular import issues and I also rewrote a couple of tests.


src/snowflake/connector/errors.py
```diff
@@ -120,6 +113,8 @@ def telemetry_msg(self) -> str | None:
 
     def generate_telemetry_exception_data(self) -> dict[str, str]:
         """Generate the data to send through telemetry."""
+        from .description import CLIENT_NAME, SNOWFLAKE_CONNECTOR_VERSION
```
sfc-gh-aling, line 116: may I ask why we move the import into methods?

sfc-gh-mkeller: To prevent circular imports. I tried moving things around to not have to do this, but I cannot find a way to cleanly separate the modules from each other.

sfc-gh-aling: the major concern I have here is perf, we're doing import per call which could add overhead to some extent.

I see `CLIENT_NAME, SNOWFLAKE_CONNECTOR_VERSION` are just string constants, I guess they are not the one s causing the circular import. which two modules are circular importing each other?

one technique I used before is to cut off circular import by writing helper methods and passing classes in the helper method.

sfc-gh-mkeller: This isn't the case, when import statements are re-executed they become a lookup of the already imported modules, they don't actually re-import anything. See [docs](https://docs.python.org/3/reference/import.html#the-module-cache)

sfc-gh-sfan (commented):
The change itself looks good to me. I'm wondering if there is a test or linter to make sure we don't accidentally introduce circular dependencies in the future?

src/snowflake/connector/constants.py
```diff
@@ -17,25 +26,58 @@
...
-    FieldType(name="TIMESTAMP_TZ", dbapi_type=[DBAPI_TYPE_TIMESTAMP]),
-    FieldType(name="TIMESTAMP_NTZ", dbapi_type=[DBAPI_TYPE_TIMESTAMP]),
-    FieldType(name="OBJECT", dbapi_type=[DBAPI_TYPE_BINARY]),
-    FieldType(name="ARRAY", dbapi_type=[DBAPI_TYPE_BINARY]),
-    FieldType(name="BINARY", dbapi_type=[DBAPI_TYPE_BINARY]),
-    FieldType(name="TIME", dbapi_type=[DBAPI_TYPE_TIMESTAMP]),
-    FieldType(name="BOOLEAN", dbapi_type=[]),
-    FieldType(name="GEOGRAPHY", dbapi_type=[DBAPI_TYPE_STRING]),
-]
+    pa_type: Callable[[], DataType]
```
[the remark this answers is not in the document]
sfc-gh-yixie: Same question. Why do we need that lambda with no parameter?

DESCRIPTION.md
```diff
@@ -10,13 +10,18 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 
 
+- v2.7.9()
+
+   - Fixed a bug where empty results containing GEOOGRAPHY type raised IndexError
```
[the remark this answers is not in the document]
sfc-gh-mkeller:
```suggestion
   - Fixed a bug where empty results containing GEOGRAPHY type raised IndexError
```

src/snowflake/connector/constants.py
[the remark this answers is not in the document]
sfc-gh-mkeller: Since `pyarrow` is an optional dependency we cannot resolve `pyarrow.int64` for sure.
If we didn't have `pyarrow` installed then the lookup invoked by `pyarrow.int64` would trigger an error, so I wrap these definitions into lambdas. This way I still get [lookup magic from snowflake.connector.options](https://github.com/snowflakedb/snowflake-connector-python/blob/b132ab1eb8a1b193191c1e7f0897cb7df618dedc/src/snowflake/connector/options.py#L38) with nice errors if they are needed, but I don't resolve them until I necessarily have to

src/snowflake/connector/errors.py
sfc-gh-aling: @sfc-gh-mkeller , thanks for the link.

yeah, I see how it works, but I feel my concern still stands -- though it's not doing re-import, but it still needs to look up import in the cache per method call, especially when there're plenty of module loaded (big cache map) and if the code is on the critical path (frequently called), then this could be a perf issue.

sorry for being paranoid on performance (because previously I suffered a lot from improving perf..). I didn't mean to block the PR, will check out and try by myself.

sfc-gh-mkeller: Please run some benchmarks too @sfc-gh-aling

sfc-gh-mkeller:
> The change itself looks good to me. I'm wondering if there is a test or linter to make sure we don't accidentally introduce circular dependencies in the future?

Python itself will do this for you, just importing `snowflake.connector` caused Python to stop before I moved the import statements into the functions that needed them.

sfc-gh-jdu (approved):

sfc-gh-sfan (approved):

src/snowflake/connector/errors.py
sfc-gh-aling: @sfc-gh-mkeller , I found that the circular import only takes place in [options.py](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/options.py#L15)

```python
from .errors import MissingDependencyError
```

as an workaround to fix circular import, you could also do:
```python
import snowflake.connector.errors as errors
# when you need to use MissingDependencyError, do:
errors.MissingDependencyError(...)
```

in such a way you can avoid per call import in errors.py, the cost is that in this way, we're loading the whole errors module instead of a single error class which might cost extra memory (not too much I believe).

sfc-gh-aling: also I did a naive perf running against the two approaches:

I have two module a and b importing each other, and I have a dumb script testing calling. My program shows that there's a perf diff, please check my scripts below:

```bash
import per call cost time: 0.462610958
import module and call cost time: 0.13138283300000003
```

---- perf test scripts -----

```python
# module_a.py
import module_b


def func_a_import_module_and_call():
    module_b.func_b_circular_import_callee()


def func_a_import_per_call():
    from module_b import func_b_circular_import_callee
    func_b_circular_import_callee()
``` 

```python
# module_b.py
from module_a import func_a_import_module_and_call, func_a_import_per_call


def func_b_import_module_and_call():
    func_a_import_module_and_call()


def func_b_import_per_call():
    func_a_import_per_call()
    pass


def func_b_circular_import_callee():
    pass
```

```python
# comparing perf
import timeit

setup = """
from module_b import func_b_import_module_and_call, func_b_import_per_call
"""

st_import_per_call = """
func_b_import_per_call()
"""

print(
    'import per call cost time:',
    timeit.timeit(
        stmt=st_import_per_call,
        setup=setup,
        number=1000000
    )
)


st_import_module_and_call = """
func_b_import_module_and_call()
"""

print(
    'import module and call cost time:',
    timeit.timeit(
        stmt=st_import_module_and_call,
        setup=setup,
        number=1000000
    )
)
```

sfc-gh-yixie (approved):

sfc-gh-mkeller:
Thanks to https://github.com/snowflakedb/snowflake-connector-python/pull/1158#discussion_r890685816  I found that in the case of `from . import subpackage` the imported module acts lazy, so importing modules this way allows us to import everything at module level only and resolve them only when we need them only.
Good find @sfc-gh-aling  , thank you!

sfc-gh-aling (approved):

PR #1164, opened 2022-06-13
update to test_util.py

### LINKED ISSUE #1163: SNOW-610826: update test_util.py os.getenv() behavior
Hey, 

I'm not using the template because honestly the problem i'm trying to solve upstream of us isn't really a bug per-se more just a conflict in behaviors between libraries. 

As part of a pipeline for validating the importability of code for an orchestration system, we often mock `os.environ.get()` to return a string value. This allows the user of the library to test the validity of their code without either setting a default or setting up environment variables for their workflow. 

Unfortunately, when we do this it interacts with `snowflake.connector.test_util` in some unintended ways. Specifically , `RUNNING_ON_JENKINS = os.getenv("JENKINS_HOME") is not None` , starts being set to a value that causes it to evaluate to True even when not explicitly set. This PR simply sets a value for the default return on that `os.environ.get()` and tests for its presence instead of assuming None. 

I realize this edge case is a "my problem" down stream and we have some ways of dealing with it as an edge case, but because all of this is executed on import instead of as a function call. Usually I'd suggest isolating less code to be running on import - but I think this is an easy solution and doesn't impact your larger code base behaviour in a meaningful way. 

TIA for your time.

dylanbstorey:

   Fixes #1163

  Updating the behaviour to provide an incredibly unlikely default value as an argument to `os.getenv` and then using the presence of that value to determine if `JENKINS_HOME` is set in the environment.


dylanbstorey:
recheck

dylanbstorey:
recheck

src/snowflake/connector/test_util.py
```diff
@@ -10,7 +10,9 @@
 
 from .compat import IS_LINUX
 
-RUNNING_ON_JENKINS = os.getenv("JENKINS_HOME") is not None
+
+JENKINS_HOME_DEFAULT_RETURN_VALUE= '$!$@46217d2d-1000-40b2-b4e5-9fdbab719c7c'
+RUNNING_ON_JENKINS = os.getenv("JENKINS_HOME", JENKINS_HOME_DEFAULT_RETURN_VALUE) is not JENKINS_HOME_DEFAULT_RETURN_VALUE
```
sfc-gh-mkeller, line 15: You usually want to test equality of strings with `==`, or `!=` not with `is`.
```suggestion
RUNNING_ON_JENKINS = os.getenv("JENKINS_HOME", JENKINS_HOME_DEFAULT_RETURN_VALUE) != JENKINS_HOME_DEFAULT_RETURN_VALUE
```

src/snowflake/connector/test_util.py
```diff
@@ -10,7 +10,9 @@
 
 from .compat import IS_LINUX
 
-RUNNING_ON_JENKINS = os.getenv("JENKINS_HOME") is not None
+
+JENKINS_HOME_DEFAULT_RETURN_VALUE= '$!$@46217d2d-1000-40b2-b4e5-9fdbab719c7c'
```
sfc-gh-mkeller, line 14: Formatting
```suggestion
JENKINS_HOME_DEFAULT_RETURN_VALUE = '$!$@46217d2d-1000-40b2-b4e5-9fdbab719c7c'
```

sfc-gh-mkeller (changes requested):
I'm sorry, could you please explain how this solves your issue?
To me:
`os.getenv("JENKINS_HOME") is not None`
and
`os.getenv("JENKINS_HOME", JENKINS_HOME_DEFAULT_RETURN_VALUE) is not JENKINS_HOME_DEFAULT_RETURN_VALUE`
does the same thing.

If you overwrite `os.getenv` then its going to trigger both. Maybe something like this would be better?
```
RUNNING_ON_JENKINS = "JENKINS_HOME" in os.environ
```

PR #1195, opened 2022-07-15
SNOW-574657 in-memory cache impementation

sfc-gh-mkeller:

   Fixes SNOW-574657

   - [x] I am modifying OCSP code: I'm adding the new cache to the OCSP checks
   - [x] I am adding a new dependency: `typing_extensions` for `typing` feature backports

### DIFF
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -75,7 +75,7 @@ class SFDictCache(Generic[K, V]):
             raise
         if is_expired(t):
             self._expiration(k)
-            del self._cache[k]
+            self.__delitem(k)
             raise KeyError
         if should_record_hits:
             self._hit(k)
@@ -95,6 +95,7 @@ class SFDictCache(Generic[K, V]):
             expiry=now() + self._entry_lifetime,
             entry=v,
         )
+        self.telemetry["size"] = len(self._cache)
 
     def __getitem__(
         self,
@@ -150,12 +151,24 @@ class SFDictCache(Generic[K, V]):
             self._cache.clear()
             self._reset_telemetry()
 
+    def __delitem(
+        self,
+        key: K,
+    ) -> None:
+        """Non-locking version of __delitem__.
+
+        This should only be used by internal functions when already
+        holding self._lock.
+        """
+        del self._cache[key]
+        self.telemetry["size"] = len(self._cache)
+
     def __delitem__(
         self,
         key: K,
     ) -> None:
         with self._lock:
-            del self._cache[key]
+            self.__delitem(key)
 
     def __contains__(
         self,
@@ -189,10 +202,12 @@ class SFDictCache(Generic[K, V]):
             }
             with self._lock:
                 self._cache.update(to_insert)
+                self.telemetry["size"] = len(self._cache)
         elif isinstance(other, SFDictCache):
             to_insert: dict[K, CacheEntry[V]] = {k: v for k, v in other._cache.items()}
             with self._lock:
                 self._cache.update(to_insert)
+                self.telemetry["size"] = len(self._cache)
         else:
             raise TypeError
 
@@ -203,6 +218,7 @@ class SFDictCache(Generic[K, V]):
                     self.__getitem(k, should_record_hits=False)
                 except KeyError:
                     continue
+            self.telemetry["size"] = len(self._cache)
 
     # Telemetry related functions, these can be plugged by child classes
     def _reset_telemetry(self) -> None:
@@ -215,6 +231,7 @@ class SFDictCache(Generic[K, V]):
             "hit": 0,
             "miss": 0,
             "expiration": 0,
+            "size": 0,
         }
 
     def _hit(self, k: K) -> None:
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -44,7 +44,7 @@ class SFDictCache(Generic[K, V]):
         self._entry_lifetime = datetime.timedelta(seconds=entry_lifetime)
         self._cache: dict[K, CacheEntry[V]] = {}
         self._lock = Lock()
-        self._setup_telemetry()
+        self._reset_telemetry()
 
     @classmethod
     def from_dict(
@@ -71,11 +71,9 @@ class SFDictCache(Generic[K, V]):
         try:
             t, v = self._cache[k]
         except KeyError:
-            # Call telemetry function
             self._miss(k)
             raise
         if is_expired(t):
-            # Call telemetry function
             self._expiration(k)
             del self._cache[k]
             raise KeyError
@@ -119,45 +117,23 @@ class SFDictCache(Generic[K, V]):
         return iter(self.keys())
 
     def keys(self) -> list[K]:
-        keys: list[K] = []
-        with self._lock:
-            for k in list(self._cache.keys()):
-                try:
-                    _ = self.__getitem(k, should_record_hits=False)
-                    keys.append(k)
-                except KeyError:
-                    continue
-        return keys
-
-    def __items(self) -> list[tuple[K, V]]:
-        """Non-locking version of items.
-
-        This should only be used by internal functions when already
-        holding self._lock.
-        """
-        values: list[tuple[K, V]] = []
-        for k in list(self._cache.keys()):
-            try:
-                v = self.__getitem(k, should_record_hits=False)
-                values.append((k, v))
-            except KeyError:
-                continue
-        return values
+        return [k for k, _ in self.items()]
 
     def items(self) -> list[tuple[K, V]]:
         with self._lock:
-            return self.__items()
-
-    def values(self) -> list[V]:
-        values: list[V] = []
-        with self._lock:
+            values: list[tuple[K, V]] = []
             for k in list(self._cache.keys()):
                 try:
+                    # TODO: this function could be further optimized by removing
+                    #  the need here to call __getitem here and call se;f._cache.keys()
                     v = self.__getitem(k, should_record_hits=False)
-                    values.append(v)
+                    values.append((k, v))
                 except KeyError:
                     continue
         return values
+
+    def values(self) -> list[V]:
+        return [v for _, v in self.items()]
 
     def get(
         self,
@@ -172,6 +148,7 @@ class SFDictCache(Generic[K, V]):
     def clear(self) -> None:
         with self._lock:
             self._cache.clear()
+            self._reset_telemetry()
 
     def __delitem__(
         self,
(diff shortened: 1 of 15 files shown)


sfc-gh-kwagner:
@sfc-gh-mkeller do we need a security review for this since you're modifying the OCSP code?

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def _hit(self, k: K) -> None:
+        """This function gets called when a hit occurs.
+
+        Functions that hit every entry (like values) is not going to count.
+
+        Note that while this function does not interact with lock, but it's only
+        called from contexts where the lock is already held.
+        """
+        del k
```
sfc-gh-aling, line 245: may I ask why `del K` is called here?

sfc-gh-mkeller:
> @sfc-gh-mkeller do we need a security review for this since you're modifying the OCSP code?

I don't think so, I didn't actually change any of the OCSP code. I just save what a function returns into the new cache and return it from cache if it's present. Obviously I can do a review if you think I should do one.

sfc-gh-kwagner:
> > @sfc-gh-mkeller do we need a security review for this since you're modifying the OCSP code?
> 
> I don't think so, I didn't actually change any of the OCSP code. I just save what a function returns into the new cache and return it from cache if it's present. Obviously I can do a review if you think I should do one.

I think we should do one at least for the epic associated with this JIRA. Can you make Harsh the reviewer since this applies to all the ecosystem drivers?

sfc-gh-mkeller:
> > > @sfc-gh-mkeller do we need a security review for this since you're modifying the OCSP code?
> > 
> > 
> > I don't think so, I didn't actually change any of the OCSP code. I just save what a function returns into the new cache and return it from cache if it's present. Obviously I can do a review if you think I should do one.
> 
> I think we should do one at least for the epic associated with this JIRA. Can you make Harsh the reviewer since this applies to all the ecosystem drivers?

Oh, we have done a review for the caching epic. We forgot to record it on Jira though. I'll go and fix this!

sfc-gh-kwagner:
> > > > @sfc-gh-mkeller do we need a security review for this since you're modifying the OCSP code?
> > > 
> > > 
> > > I don't think so, I didn't actually change any of the OCSP code. I just save what a function returns into the new cache and return it from cache if it's present. Obviously I can do a review if you think I should do one.
> > 
> > 
> > I think we should do one at least for the epic associated with this JIRA. Can you make Harsh the reviewer since this applies to all the ecosystem drivers?
> 
> Oh, we have done a review for the caching epic. We forgot to record it on Jira though. I'll go and fix this!

Awesome!

sfc-gh-kwagner (approved):

sfc-gh-mkeller:
Found it: https://github.com/snowflakedb/security_reviews/pull/1956

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+                self.telemetry["size"] = len(self._cache)
+        elif isinstance(other, SFDictCache):
+            to_insert: dict[K, CacheEntry[V]] = {k: v for k, v in other._cache.items()}
+            with self._lock:
+                self._cache.update(to_insert)
+                self.telemetry["size"] = len(self._cache)
+        else:
+            raise TypeError
+
+    def _clear_expired_entries(self) -> None:
```
sfc-gh-aling, line 214: I don't see this method is used?

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+    ) -> bool:
+        with self._lock:
+            if key in self._cache.keys():
+                try:
+                    self.__getitem(key, should_record_hits=True)
+                    return True
+                except KeyError:
+                    # Fall through
+                    pass
+        return False
```
sfc-gh-aling, line 185: I think this could be simplified as:
```suggestion
        with self._lock:
            try:
                self.__getitem(key, should_record_hits=True)
                return True
            except KeyError:
                # Fall through
                return False
```

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+
+def is_expired(d: datetime.datetime) -> bool:
+    return now() >= d
+
+
+class SFDictCache(Generic[K, V]):
+    """A generic in-memory cache that acts somewhat like a dictionary.
+
+    Notes:
+    - Unlike normal dictionaries keys, values and items return list
```
sfc-gh-sfan, line 36: Maybe it's just me but I find this note not very fluent. How about:

`keys(), values() and items() return list materialized at call time, unlike normal dictionaries that return a view object`

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+      materialized at call time, instead of returning a view object.
+    """
+
+    def __init__(
+        self,
+        entry_lifetime: int = 60 * 60 * 24,  # 1 day in seconds
```
sfc-gh-sfan, line 42: Maybe make this value a macro `DAY_IN_SECONDS`?

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+        from the other cache.
+        """
+        if isinstance(other, dict):
+            t = now() + self._entry_lifetime
+            to_insert: dict[K, CacheEntry[V]] = {
+                k: CacheEntry(expiry=t, entry=v) for k, v in other.items()
+            }
+            with self._lock:
+                self._cache.update(to_insert)
+                self.telemetry["size"] = len(self._cache)
```
sfc-gh-sfan, line 205: This is shared between two if branches and could be extracted:

```
if ...:
  ...
elif ...:
 ...
else:
  ...

with self._lock:
    self._cache.update(to_insert)
    self.telemetry["size"] = len(self._cache)
```

src/snowflake/connector/network.py
```diff
@@ -907,14 +920,14 @@ def _request_exec_wrapper(
...
-            conn: SnowflakeConnection,
-            timed_out: bool = True
+        self,
+        e: Exception,
+        full_url: str,
+        method: str,
+        retry_timeout: int,
+        retry_count: int,
+        conn: SnowflakeConnection,
+        timed_out: bool = True,
```
sfc-gh-sfan, line 930: Do we need a merge gate checking for format? It is a bit weird that this PR contains unrelated format changes.

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -64,9 +67,30 @@
 from snowflake.connector.telemetry_oob import TelemetryService
 from snowflake.connector.time_util import DecorrelateJitterBackoff
 
+from .cache import SFDictCache
+
+OCSP_CACHE: SFDictCache[
+    tuple[bytes, bytes, int, str],
+    tuple[Exception | None, Certificate, Certificate, CertId, bytes],
+] = SFDictCache(
+    entry_lifetime=60 * 60 * 24,  # A day in seconds
```
sfc-gh-sfan, line 76: Similarly, use a macro. Maybe the macro could be put into `constants` so it can be shared.

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -1322,13 +1373,23 @@ def _validate_certificates_sequential(
...
 
         for issuer, subject in cert_data:
-            r = self.validate_by_direct_connection(
-                issuer, subject, telemetry_data, hostname, do_retry=do_retry
-            )
-            results.append(r)
+            cert_id, _ = self.create_ocsp_request(issuer=issuer, subject=subject)
+            cache_key = generate_cache_key(cert_id, subject)
+            cache_hit = OCSP_CACHE.get(cache_key)
+            if cache_hit is None:
```
sfc-gh-sfan, line 1379: I'm curious if we plan to support a `loader` concept (similar to Guava's loading cache) so if a key is not present in the cache, the cache triggers loader and returns it, and also caches the loaded result for next call.

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @classmethod
+    def from_dict(
+        cls,
+        _dict: dict[K, V],
+        **kw,
+    ) -> Self:
+        cache = cls(**kw)
+        for k, v in _dict.items():
+            cache[k] = v
```
sfc-gh-aling, line 57: I feel we may want to `deepcopy` k and v here to keep an internal copy for the SFDictCache.

otherwise we have to be aware that the SFDictCache is keeping the same reference as the dict derived from

src/snowflake/connector/cache.py
```diff
@@ -0,0 +1,264 @@
...
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+def is_expired(d: datetime.datetime) -> bool:
+    return now() >= d
+
+
+class SFDictCache(Generic[K, V]):
```
sfc-gh-aling, line 32: is the cache expected to be internally used only?

my concern is that improper usage pattern leads to memory monotonically increment (as we're only doing lazy deletion here) and after long running, users encounter out of memory error.

or this is not a valid concern?

src/snowflake/connector/cache.py
sfc-gh-mkeller: To get around linters complaining about this argument being unused

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-mkeller: This PR only introduces an in-memory cache, so not yet. I'm still flushing out the exact save/load strategies.
I'll keep this is mind for the final strategies though!

src/snowflake/connector/cache.py
sfc-gh-mkeller: Internal only, yes. This exact idea was discussed in the design doc for the project, please take a look at that

sfc-gh-sfan (approved):
LGTM

src/snowflake/connector/cache.py
sfc-gh-aling: sounds reasonable to me, I don't have strong opinions on deepcopy the objects. As long as we're aware of the "same references" here and know how to deal with it then we are good.

src/snowflake/connector/cache.py
sfc-gh-aling: I see, keeping it should be fine

src/snowflake/connector/cache.py
sfc-gh-aling: I see, I think there should be a better way to ignore the lint error, maybe try adding `# noqa` on the line where the error is reported? https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

I'm not sure what side effect that del K could give us (e.g. reduce ref cnt xx)

src/snowflake/connector/cache.py
sfc-gh-aling: yes, I remember that. we know there's a fixed amount of cache and small size.

src/snowflake/connector/cache.py
sfc-gh-mkeller: It does reduce ref count, but it should only delete the local scope's ref and not the outer scope, but I hear you.
I'll change it for something else!

sfc-gh-jdu (approved):

sfc-gh-aling (approved):
LGTM

PR #1200, opened 2022-07-21
SNOW-612517 disable gzip compression

sfc-gh-mkeller:

   Fixes SNOW-612517

   It seems like there was a bug in our old gzip compression code, because `GzipFile` was never closed. One user hit this issue, I believe that the issue is very dependent on their gzip library's code as we were unable to  reproduce the issue. This user has been testing this issue for days without encountering the issue since. I'd like to merge this fix pre-maturely, because I see no harm in it even if didn't fully fix the issue.

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import collections
 import contextlib
+import gzip
 import itertools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/network.py
```diff
@@ -994,11 +1006,8 @@ def _request_exec(
         logger.debug("socket timeout: %s", socket_timeout)
         try:
             if not catch_okta_unauthorized_error and data and len(data) > 0:
-                gzdata = BytesIO()
-                gzip.GzipFile(fileobj=gzdata, mode="wb").write(data.encode("utf-8"))
-                gzdata.seek(0, 0)
                 headers["Content-Encoding"] = "gzip"
-                input_data = gzdata
+                input_data = gzip.compress(data.encode("utf-8"))
```
sfc-gh-aling, line 1010: code change itself looks good to me.

have you sanity checked that the compress mechanisms under two approaches are the same?
you know, from the issue description looks like closing the gzip file is also an option. but if the effect is the same, then less code wins.

sfc-gh-mkeller: Yes, I have manually compressed a file with both methods and then uncompressed the created GZIP bytes and got back the same things.
The effect should be the same.

sfc-gh-aling (approved):
LGTM

PR #1201, opened 2022-07-21
SNOW-629080: investigate and implement lastrowid

sfc-gh-aling:

   Fixes SNOW-629080

    Snowflake doesn't support lastrowid, according to [PEP249](https://peps.python.org/pep-0249/#lastrowid), we should return None for this attribute.

    ```definition in pep249
    Cursor.lastrowid
    This read-only attribute provides the rowid of the last modified row 
    (most databases return a rowid only when a single INSERT operation is performed).
    If the operation does not set a rowid or if the database does not support rowids,
    this attribute should be set to None.
    The semantics of .lastrowid are undefined in case the last executed
    statement modified more than one row, e.g. when using INSERT with .executemany().
    ```

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -368,8 +368,8 @@ class SnowflakeCursor:
         return hasattr(self, "_is_file_transfer") and self._is_file_transfer
 
     @property
-    def lastrowid(self):
-        # snowflake does not support lastrowid in which case None should be returned as per PEP249
+    def lastrowid(self) -> None:
+        """Snowflake does not support lastrowid in which case None should be returned as per PEP249."""
         return None
 
     def callproc(self, procname: str, args: T = ()) -> T:
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -367,6 +367,11 @@ class SnowflakeCursor:
         """Whether the command is PUT or GET."""
         return hasattr(self, "_is_file_transfer") and self._is_file_transfer
 
+    @property
+    def lastrowid(self):
+        # snowflake does not support lastrowid in which case None should be returned as per PEP249
+        return None
+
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -367,6 +367,11 @@ def is_file_transfer(self):
         """Whether the command is PUT or GET."""
         return hasattr(self, "_is_file_transfer") and self._is_file_transfer
 
+    @property
+    def lastrowid(self):
+        # snowflake does not support lastrowid in which case None should be returned as per PEP249
```
sfc-gh-mkeller, line 372:
```suggestion
        """Snowflake does not support lastrowid in which case None should be returned as per PEP249."""
```

src/snowflake/connector/cursor.py
```diff
@@ -367,6 +367,11 @@ def is_file_transfer(self):
         """Whether the command is PUT or GET."""
         return hasattr(self, "_is_file_transfer") and self._is_file_transfer
 
+    @property
+    def lastrowid(self):
```
sfc-gh-mkeller, line 371:
```suggestion
    def lastrowid(self) -> None:
```

test/unit/test_cursor.py
```diff
@@ -27,3 +34,9 @@ class FileTransferType(Enum):
 )
 def test_get_filetransfer_type(sql, _type):
     assert SnowflakeCursor.get_file_transfer_type(sql) == _type
+
+
+def test_cursor_attribute():
+    fake_conn = FakeConnection()
+    cursor = SnowflakeCursor(fake_conn)
+    assert not cursor.lastrowid
```
sfc-gh-mkeller, line 42:
```suggestion
    assert cursor.lastrowid is None
```

sfc-gh-mkeller (changes requested):

sfc-gh-mkeller (approved):

PR #1209, opened 2022-07-26
SNOW-611961 in-file cache implementation

sfc-gh-mkeller:

   Fixes SNOW-611961

   In this PR I add an extension of #1195 that adds file caching on top of in-memory caches.
   
   ## On disk save strategy:
   On every write to the in-memory cache (`__setitem__`, `update`) we generate a random number 0-9 and if it's 1 then we save to disk. This is done so if multi-processing starts multiple connectors with no cache available we distribute their writes randomly. The only other time we try to save is during tear-down.
   One caveat of save is that we try to acquire the [file-lock](https://py-filelock.readthedocs.io/en/latest/api.html#filelock.FileLock) (which is required to be able to write to the cache file) in a non-blocking manner and if we don't succeed we log this and give up. So if a rogue process held the file lock forever then no file caching would be done.
   However; after taking a look at how `filelock` is implemented this is unlikely to happen. It uses OS APIs to lock file parts to file descriptors, so if a connector held the file lock and it was SIGABORTED out of existence the OS would still clean up the open file descriptors and release those file parts. So someone would really have to hold the filelock maliciously for file caching to not work at all.
   I think this is acceptable for caches.
   
   ## On disk load strategy:
   On every miss and expiration we check whether the cache file has been modified since we last loaded it. If yes, then we load and try to turn that frown upside down!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,6 +57,7 @@ install_requires =
     urllib3>=1.21.1,<1.27
     certifi>=2017.4.17
     typing_extensions>=4.3,<5
+    filelock>=3.5,<4
 include_package_data = True
 package_dir =
     =src
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -241,9 +241,7 @@ class SFDictCache(Generic[K, V]):
         given another SFDictCache then the timestamps will be taken
         from the other cache.
 
-        Returns a tuple of 2 booleans. The first one describes whether self learnt
-        from other and the second describes whether other could learn from self
-        (it's important to notice that other has not learnt anythin from self).
+        Returns a boolean. It describes whether self learnt anything from other.
 
         Note that clear_expired_entries will be called on both caches. To
         prevent deadlocks this is done without acquiring other._lock. The
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -5,10 +5,17 @@
 from __future__ import annotations
 
 import datetime
+import logging
+import os
+import pickle
+import platform
+import random
+import tempfile
 from collections.abc import Iterator
 from threading import Lock
 from typing import Generic, TypeVar
 
+from filelock import FileLock, Timeout
 from typing_extensions import NamedTuple, Self
 
 from . import constants
@@ -16,6 +23,8 @@ from . import constants
 now = datetime.datetime.now
 
 T = TypeVar("T")
+
+logger = logging.getLogger(__name__)
 
 
 class CacheEntry(NamedTuple, Generic[T]):
@@ -42,6 +51,7 @@ class SFDictCache(Generic[K, V]):
         self,
         entry_lifetime: int = constants.DAY_IN_SECONDS,
     ) -> None:
+        """Inits a SFDictCache with lifetime."""
         self._entry_lifetime = datetime.timedelta(seconds=entry_lifetime)
         self._cache: dict[K, CacheEntry[V]] = {}
         self._lock = Lock()
@@ -129,14 +139,12 @@ class SFDictCache(Generic[K, V]):
     def items(self) -> list[tuple[K, V]]:
         with self._lock:
             values: list[tuple[K, V]] = []
-            for k in list(self._cache.keys()):
-                try:
-                    # TODO: this function could be further optimized by removing
-                    #  the need here to call _getitem here and call self._cache.keys()
-                    v = self._getitem(k, should_record_hits=False)
-                    values.append((k, v))
-                except KeyError:
-                    continue
+            for k, v in list(self._cache.items()):
+                if is_expired(v.expiry):
+                    self._expiration(k)
+                    self._delitem(k)
+                else:
+                    values.append((k, v.entry))
         return values
 
     def values(self) -> list[V]:
@@ -188,38 +196,76 @@ class SFDictCache(Generic[K, V]):
                 # Fall through
                 return False
 
+    def _update(
+        self,
+        other: dict[K, V] | list[tuple[K, V]] | SFDictCache[K, V],
+    ) -> bool:
+        to_insert: dict[K, CacheEntry[V]]
+        self._clear_expired_entries()
+        if isinstance(other, (list, dict)):
+            expiry = now() + self._entry_lifetime
+            if isinstance(other, list):
+                g = iter(other)
+            elif isinstance(other, dict):
+                g = iter(other.items())
+            to_insert = {k: CacheEntry(expiry=expiry, entry=v) for k, v in g}
+        elif isinstance(other, SFDictCache):
+            other._clear_expired_entries()
+            others_items = list(other._cache.items())
+            # Only accept values from another cache if their key is not in self,
+            #  or if expiry is later the self known one
+            to_insert = {
+                k: v
+                for k, v in others_items
+                if (
+                    # self doesn't have this key
+                    k not in self._cache
+                    # other has newer expiry time
+                    or self._cache[k].expiry < v.expiry
+                )
+            }
+        else:
+            raise TypeError
+        self._cache.update(to_insert)
+        self.telemetry["size"] = len(self._cache)
+        return len(to_insert) > 0
+
     def update(
         self,
-        other: dict[K, V] | SFDictCache[K, V],
-    ) -> None:
-        """Insert multiple values at the same time.
+        other: dict[K, V] | list[tuple[K, V]] | SFDictCache[K, V],
+    ) -> bool:
+        """Insert multiple values at the same time, if dicts could learn from the other.
 
-        If this function is given a dictionary, expiration timestamps
+        If this function is given a dictionary, or list expiration timestamps
         will be all the same a self._entry_lifetime form now. If it's
         given another SFDictCache then the timestamps will be taken
         from the other cache.
+
+        Returns a tuple of 2 booleans. The first one describes whether self learnt
+        from other and the second describes whether other could learn from self
+        (it's important to notice that other has not learnt anythin from self).
+
+        Note that clear_expired_entries will be called on both caches. To
+        prevent deadlocks this is done without acquiring other._lock. The
+        intended behavior is to use this function with an unpickled/unused cache.
+        If live caches are are being merged then use .items() on them first and
+        merge those into the other caches.
         """
-        if isinstance(other, dict):
-            t = now() + self._entry_lifetime
-            to_insert: dict[K, CacheEntry[V]] = {
-                k: CacheEntry(expiry=t, entry=v) for k, v in other.items()
-            }
-        elif isinstance(other, SFDictCache):
-            to_insert: dict[K, CacheEntry[V]] = {k: v for k, v in other._cache.items()}
-        else:
-            raise TypeError
         with self._lock:
-            self._cache.update(to_insert)
-            self.telemetry["size"] = len(self._cache)
+            return self._update(other)
 
     def _clear_expired_entries(self) -> None:
+        for k in list(self._cache.keys()):
+            try:
+                self._getitem(k, should_record_hits=False)
+            except KeyError:
+                continue
+        self.telemetry["size"] = len(self._cache)
+
+    def clear_expired_entries(self) -> None:
+        """Remove expired entries from the cache."""
         with self._lock:
-            for k in self._cache.keys():
-                try:
-                    self._getitem(k, should_record_hits=False)
-                except KeyError:
-                    continue
-            self.telemetry["size"] = len(self._cache)
+            self._clear_expired_entries()
 
     # Telemetry related functions, these can be plugged by child classes
     def _reset_telemetry(self) -> None:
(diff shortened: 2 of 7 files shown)


src/snowflake/connector/cache.py
```diff
@@ -260,3 +306,185 @@ def _expiration(self, k: K) -> None:
...
+        #  of whether the lock is locked, but unless this function gets extended I
+        #  feel like it's an overkill. Make sure to change the bool right after
+        #  self._lock.acquire() and self._lock.release().
+        currently_holding = True
+        try:
+            try:
+                t, v = self._cache[k]
+            except KeyError:
+                # Load if we should and try to see if we can turn the miss into a hit
+                self._lock.release()
```
sfc-gh-aling, line 355: is the re-release/re-acquire necessary here? I don't see `_load_if_should` use `self._lock`

src/snowflake/connector/cache.py
```diff
@@ -260,3 +306,185 @@ def _expiration(self, k: K) -> None:
...
+                currently_holding = True
+                expire_item = True
+                if loaded:
+                    t, v = self._cache[k]
+                    if not is_expired(t):
+                        expire_item = False
+                if expire_item:
+                    self._expiration(k)
+                    self._delitem(k)
+                    raise KeyError
```
sfc-gh-aling, line 383: I think we could reuse `self._getitem` here.

setup.cfg
```diff
@@ -57,6 +57,7 @@ install_requires =
     urllib3>=1.21.1,<1.27
     certifi>=2017.4.17
     typing_extensions>=4.3,<5
+    filelock>=3.5,<4
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Yup, got the go ahead!

src/snowflake/connector/cache.py
```diff
@@ -174,52 +182,90 @@ def __delitem__(
...
+            if isinstance(other, list):
+                g = iter(other)
+            elif isinstance(other, dict):
+                g = iter(other.items())
+            to_insert = {k: CacheEntry(expiry=expiry, entry=v) for k, v in g}
+        elif isinstance(other, SFDictCache):
+            other._clear_expired_entries()
+            others_items = list(other._cache.items())
+            # Only accept values from another cache if their key is not in self,
+            #  or if expiry is later the self known one
```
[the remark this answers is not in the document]
sfc-gh-mkeller: When I load a cache from disk and it has value that was added later then mine was then I don't want to throw away the fresher cached value for an older one

src/snowflake/connector/cache.py
```diff
@@ -174,52 +182,90 @@ def __delitem__(
...
 
-        If this function is given a dictionary, expiration timestamps
+        If this function is given a dictionary, or list expiration timestamps
         will be all the same a self._entry_lifetime form now. If it's
         given another SFDictCache then the timestamps will be taken
         from the other cache.
+
+        Returns a tuple of 2 booleans. The first one describes whether self learnt
+        from other and the second describes whether other could learn from self
+        (it's important to notice that other has not learnt anythin from self).
```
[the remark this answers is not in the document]
sfc-gh-mkeller: That's right I'll update this, thanks for catching it!

src/snowflake/connector/cache.py
```diff
@@ -129,14 +139,12 @@ def keys(self) -> list[K]:
...
-                    v = self.__getitem(k, should_record_hits=False)
-                    values.append((k, v))
-                except KeyError:
-                    continue
+            for k, v in list(self._cache.items()):
+                if is_expired(v.expiry):
+                    self._expiration(k)
+                    self._delitem(k)
+                else:
+                    values.append((k, v.entry))
```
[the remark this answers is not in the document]
sfc-gh-mkeller: No, unfortunately I cannot reuse `_getitem` as it may remove elements without us knowing whether it has hit an expiration, or not. In which case the file cache would want to call `load_if_should()` but not over and over again, only once.

I also couldn't overwrite that function in this child class because that would cause other functions (like `clear_expired_entries` to call the file version, which is not what we want.

src/snowflake/connector/cache.py
```diff
@@ -260,3 +306,185 @@ def _expiration(self, k: K) -> None:
...
+            self._load()
+
+    def __getitem__(self, k: K) -> V:
+        """Returns an element if it hasn't expired yet in a thread-safe way."""
+        self._lock.acquire()
+        # TODO: This variable could be replaced by a wrapper class that keeps track
+        #  of whether the lock is locked, but unless this function gets extended I
+        #  feel like it's an overkill. Make sure to change the bool right after
+        #  self._lock.acquire() and self._lock.release().
+        currently_holding = True
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Nope, it only tells you whether the lock is locked and not whether you own the lock currently.
This in-line with anyone being able to unlock `threading.Lock`m they don't actually need to own it.

src/snowflake/connector/cache.py
```diff
@@ -260,3 +306,185 @@ def _expiration(self, k: K) -> None:
...
+            self.file_path = os.path.expanduser(file_path)
+        else:
+            current_platform = platform.system().lower()
+            if current_platform is None or current_platform not in file_path:
+                self.file_path = next(iter(file_path.values()))
+            else:
+                self.file_path = os.path.expanduser(file_path[current_platform])
+        self.file_timeout = file_timeout
+        self._file_lock_path = f"{self.file_path}.lock"
+        self._file_lock = FileLock(self._file_lock_path, timeout=self.file_timeout)
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Oh I forgot about this, I need to write a manager class on top of them

src/snowflake/connector/cache.py
sfc-gh-mkeller: `self._lock` is for thread safety, it should not be held while doing file operations.
Holding it while trying to acquire `self._file_lock` could lead to a deadlock.

src/snowflake/connector/cache.py
sfc-gh-mkeller: I disagree, how would we patch in the `_load_if_should`s and the single time retries for misses and expirations? `self._getitem` raises a `KeyError` for both of those events

src/snowflake/connector/cache.py
```diff
@@ -260,3 +306,185 @@ def _expiration(self, k: K) -> None:
...
+                self._lock.release()
+
+    def _setitem(self, k: K, v: V) -> None:
+        super()._setitem(k, v)
+        self._save_if_should()
+
+    def _load(self) -> bool:
+        """Load cache from disk if possible, returns whether it was able to load."""
+        try:
+            with open(self.file_path, "rb") as r_file:
```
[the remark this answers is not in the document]
sfc-gh-mkeller: No, I made load work as:
1. I acquire file lock
2. I write a temporary file to disk next to the actual cache
3. I move the temporary file to replace the permanent file

This way we don't need a lock to load, replace actions are atomic on all OSes

src/snowflake/connector/cache.py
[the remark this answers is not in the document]
sfc-gh-mkeller: Maybe I can write an `update` function that works like how it does for regular `dict`s, but I'm going to keep this around with a different name. Would that work for you?

sfc-gh-sfan (commented):
Do we have a doc or could we add some comment on how the locking is expected to work in different scenarios? I have a hard time reasoning through those probably due to my lack of knowledge here :(

src/snowflake/connector/cache.py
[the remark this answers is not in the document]
sfc-gh-aling: thanks for the explanation, and yes I agree this is a valid scenario.
code looks good, I love the idea of introducing a new parameter to control the behavior.

src/snowflake/connector/cache.py
```diff
@@ -101,15 +112,15 @@ def __setitem(
             expiry=now() + self._entry_lifetime,
             entry=v,
         )
-        self.telemetry["size"] = len(self._cache)
+        self._add_or_remove()
```
sfc-gh-sfan, line 115: I'm wondering if it makes sense to have a subclass that handles telemetry? For example, `self._telemetry.add_or_remove()`. This could reduce ambiguity as `_add_or_remove` also sounds like an action.

src/snowflake/connector/cache.py
```diff
@@ -260,3 +319,218 @@ def _expiration(self, k: K) -> None:
...
+        # Once we decided on where to put the file cache make sure that this
+        #  place is readable/writable by us
+        random_string = "".join(random.choice(string.ascii_letters) for _ in range(5))
+        tmp_file, tmp_file_path = tempfile.mkstemp(
+            dir=os.path.dirname(self.file_path),
+        )
+        try:
+            with open(tmp_file, "w") as w_file:
+                # Might raise an OSError
+                w_file.write(random_string)
```
sfc-gh-sfan, line 369: Do we need to wrap the error and give a clearer error message like "Specified file path is not readable/writable"?

src/snowflake/connector/cache.py
```diff
@@ -260,3 +319,218 @@ def _expiration(self, k: K) -> None:
...
+        self._lock.acquire()
+        # TODO: This variable could be replaced by a wrapper class that keeps track
+        #  of whether the lock is locked, but unless this function gets extended I
+        #  feel like it's an overkill. Make sure to change the bool right after
+        #  self._lock.acquire() and self._lock.release().
+        currently_holding = True
+        try:
+            try:
+                t, v = self._cache[k]
+            except KeyError:
```
sfc-gh-sfan, line 395: Curious: did we choose `try-except` instead of key lookup first in `if` because of performance reasons? Just feel a bit awkward with the nested `try-except`.

src/snowflake/connector/cache.py
```diff
@@ -260,3 +319,218 @@ def _expiration(self, k: K) -> None:
...
+                        raise
+                else:
+                    self._miss(k)
+                    raise
+            if is_expired(t):
+                self._lock.release()
+                currently_holding = False
+                loaded = self._load_if_should()
+                self._lock.acquire()
+                currently_holding = True
```
sfc-gh-sfan, line 416: Can this be extracted as well given it is used twice? Maybe `_load_if_should_without_lock` (a terrible name...)

src/snowflake/connector/cache.py
```diff
@@ -260,3 +319,218 @@ def _expiration(self, k: K) -> None:
...
+            return self._load()
+        return False
+
+    def _should_save(self) -> bool:
+        """Decide whether we should save.
+
+        This is a simple random number generator to randomize writes across processes
+        that are possibly saving the same values in this cache.
+        """
+        return random.randint(0, 9) == 1
```
sfc-gh-sfan, line 500: Do we want to make this `9` a static variable? I guess we don't need to change this a lot but in case it is needed it's easier to adjust.

src/snowflake/connector/cache.py
```diff
@@ -260,3 +319,218 @@ def _expiration(self, k: K) -> None:
...
+    def _setitem(self, k: K, v: V) -> None:
+        super()._setitem(k, v)
+        self._save_if_should()
+
+    def _load(self) -> bool:
+        """Load cache from disk if possible, returns whether it was able to load."""
+        try:
+            with open(self.file_path, "rb") as r_file:
+                other = pickle.load(r_file)
+            _ = self._update(
```
sfc-gh-aling, line 441: nit
```suggestion
            self._update(
```

src/snowflake/connector/cache.py
sfc-gh-aling: I don't see `_load_if_should` tries to acquire `self._file_lock` (as you explained load doesn't require file lock) so deadlock shouldn't be an issue?

also I think actually locking the whole scope helps avoid reloading the same version of file multiple times.

let's say if we have 2 threads reached here at the same time trying to acquire the lock here, in the current implementation, the file might be loaded twice as _should_load is not protected by lock. As contrast, if we don't release, then the second thread will block (which could be a bad thing I admit), and when it acquire the lock it will  find the file cache is already reloaded to the latest saving one time of read.

src/snowflake/connector/cache.py
sfc-gh-aling: gotcha, I just feel seeing this pattern a couple of times in the code, maybe we could make this a util func?

```
                if expire_item:
                    self._expiration(k)
                    self._delitem(k)
                    raise KeyError
```

src/snowflake/connector/cache.py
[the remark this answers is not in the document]
sfc-gh-aling: please see my comment about the re-release/acquire first, if we choose not to do re-release/re-acquire, then `currently_holding` may not be needed.

src/snowflake/connector/cache.py
sfc-gh-aling: gotcha, I just feel seeing this pattern a couple of times in the code, maybe we could make this a util func?

```
                if expire_item:
                    self._expiration(k)
                    self._delitem(k)
                    raise KeyError
```

src/snowflake/connector/cache.py
sfc-gh-aling: I don't see `_load_if_should` tries to acquire `self._file_lock` (as you explained load doesn't require file lock) so deadlock shouldn't be an issue?

also I think actually locking the whole scope helps avoid reloading the same version of file multiple times.

let's say if we have 2 threads reached here at the same time trying to acquire the lock here, in the current implementation, the file might be loaded twice as _should_load is not protected by lock. As contrast, if we don't release, then the second thread will block (which could be a bad thing I admit), and when it acquire the lock it will  find the file cache is already reloaded to the latest saving one time of read.

src/snowflake/connector/cache.py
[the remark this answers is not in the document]
sfc-gh-aling: I'm contemplating over the following implementation. because just looking at the code, I think the effect is the same as what you write here:
```python
for k in self._cache.keys():
    try:
        values.append((k, self._get_item(k, should_record_hits=False))
    except KeyError:
        pass
```

src/snowflake/connector/cache.py
sfc-gh-mkeller: I think the current design is good enough for what I'd want out of cache telemetry. It's important to note that the consumer of this class and its telemetry is us and not our users and this API is not exposed to them, so it can be changed anytime.

src/snowflake/connector/cache.py
sfc-gh-mkeller: Sure, I can add this

src/snowflake/connector/cache.py
sfc-gh-mkeller: Sure, this is probably a good idea

src/snowflake/connector/cache.py
```diff
@@ -260,3 +319,218 @@ def _expiration(self, k: K) -> None:
...
+            return False
+
+    def _save(self) -> bool:
+        """Save cache to disk if possible, returns whether it was able to save."""
+        self._clear_expired_entries()
+        try:
+            with self._file_lock:
+                self._load_if_should()
+                _dir, fname = os.path.split(self.file_path)
+                # TODO add exception handling for folder not being writeable
```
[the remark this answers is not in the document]
sfc-gh-mkeller: My bad, I forgot to resolve this

src/snowflake/connector/cache.py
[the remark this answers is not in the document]
sfc-gh-mkeller: My bad, I thought this was in the `SFDictFileCache`, yes what you posted almost works.
```python
for k in list(self._cache.keys()):
    try:
        values.append((k, self._get_item(k, should_record_hits=False))
    except KeyError:
        pass
```
will do the trick. However; you cannot change a regular dictionary while iterating over it, so the `list` becomes necessary.

src/snowflake/connector/cache.py
sfc-gh-mkeller: You're right, nothing tries to acquire both locks in the current implementation. This was also done to help future developers working with this code.
My thinking was also to minimize the time we hold `self._lock` for, which means that we cannot perform file operations if `self._lock` is held as you point out exactly.

I agree, in the case where multiple threads (in one process) are missing some keys that are all now saved on disk, but be aware that this will affect performance in other cases, like a thread performing a lookup that results in a hit. That is if memory has: `cache_in_memory={"a": 1, "b": 2}, cache_on_disk={"a": 1, "b": 2, "c": 3}` when thread 1 tries to lookup `d` thread 2 won't be able to resolve `cache["a"]` until thread 1 is done loading.

src/snowflake/connector/cache.py
sfc-gh-mkeller: Sure, I've made this refactor

src/snowflake/connector/cache.py
sfc-gh-mkeller: I don't think that this is possible without some pretty complicated code as it needs to return a boolean and also possibly throw a value. However; this would be fixed by https://github.com/snowflakedb/snowflake-connector-python/blob/b119328abf999ef3729efa6a488e8e0967139201/src/snowflake/connector/cache.py#L408
Maybe we should just prioritize this to actually be done?

sfc-gh-sfan: Sure. This is just a nit :)

sfc-gh-sfan (approved):

sfc-gh-aling (approved):
LGTM

PR #1216, opened 2022-08-02
SNOW-630797 Support table_type in pandas_tools

sfc-gh-sfan:
Description

table_type could support transient table as well, which is more
flexible comparing to create_temp_table

Testing

Added integ test

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -8,6 +8,7 @@ import collections.abc
 import os
 import random
 import string
+import warnings
 from functools import partial
 from logging import getLogger
 from tempfile import TemporaryDirectory
@@ -49,6 +50,7 @@ def write_pandas(
     auto_create_table: bool = False,
     create_temp_table: bool = False,
     overwrite: bool = False,
+    table_type: str = "",
 ) -> tuple[
     bool,
     int,
@@ -102,10 +104,12 @@ def write_pandas(
             I.e. identifiers will be coerced to uppercase by Snowflake.  (Default value = True)
         auto_create_table: When true, will automatically create a table with corresponding columns for each column in
             the passed in DataFrame. The table will not be created if it already exists
-        create_temp_table: Will make the auto-created table as a temporary table
+        create_temp_table: (Deprecated) Will make the auto-created table as a temporary table
         overwrite: When true, and if auto_create_table is true, then it drops the table. Otherwise, it
         truncates the table. In both cases it will replace the existing contents of the table with that of the passed in
             Pandas DataFrame.
+        table_type: The table type of to-be-created table. The supported table types include ``temp``/``temporary``
+            and ``transient``. Empty means permanent table as per SQL convention.
 
     Returns:
         Returns the COPY INTO command's results to verify ingestion in the form of a tuple of whether all chunks were
@@ -124,6 +128,22 @@ def write_pandas(
                 compression, compression_map.keys()
             )
         )
+
+    if create_temp_table:
+        warnings.warn(
+            "create_temp_table is deprecated, we still respect this parameter when it is True but "
+            'please consider using `table_type="temp"` instead',
+            DeprecationWarning,
+            # warnings.warn -> write_pandas
+            stacklevel=2,
+        )
+        table_type = "temp"
+
+    if table_type and table_type.lower() not in ["temp", "temporary", "transient"]:
+        raise ValueError(
+            "Unsupported table type. Expected table types: temp/temporary, transient"
+        )
+
     if quote_identifiers:
         location = (
             (('"' + database + '".') if database else "")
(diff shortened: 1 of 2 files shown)


sfc-gh-sfan:
Will add a test for invalid table type. Thanks CodeCov report :)

sfc-gh-jdu (approved):

src/snowflake/connector/pandas_tools.py
```diff
@@ -49,6 +50,7 @@ def write_pandas(
     auto_create_table: bool = False,
     create_temp_table: bool = False,
     overwrite: bool = False,
+    table_type: str = "",
```
sfc-gh-mkeller, line 53: Could we introduce a unique [enum](https://docs.python.org/3/library/enum.html#creating-an-enum). Otherwise this should have more precise type-hints like `typing.Literal["", "temp", "temporary", "transient"]`

test/integ/pandas/test_pandas_tools.py
```diff
@@ -139,9 +139,10 @@ def test_write_pandas_with_overwrite(
         "gzip",
     ],
 )
-@pytest.mark.parametrize("quote_identifiers", [True, False])
-@pytest.mark.parametrize("auto_create_table", [True, False])
-@pytest.mark.parametrize("create_temp_table", [True, False])
+@pytest.mark.parametrize("quote_identifiers", [True])
+@pytest.mark.parametrize("auto_create_table", [True])
+@pytest.mark.parametrize("create_temp_table", [True])
```
sfc-gh-mkeller, line 144: Did you mean to turn the 8 tests into 1?

sfc-gh-mkeller: We should remove `create_temp_table` and replace `if create_temp_table` with `if table_type.lower() in ("temp", "temporary")`

sfc-gh-mkeller (changes requested):

test/integ/pandas/test_pandas_tools.py
sfc-gh-sfan: Crap this is not intended. I changed this to quickly verify some changes 😂 Will revert

src/snowflake/connector/pandas_tools.py
sfc-gh-sfan: I considered using `enum` in the snowpark but @sfc-gh-yixie suggested that using a string is more user friendly when discussing about similar changes in snowpark. I'll add a more precise type-hints (and make same change in snowpark as well). Thanks for catching this.

test/integ/pandas/test_pandas_tools.py
sfc-gh-sfan: BTW I still want to test `create_temp_table` because I have a deprecation warning check. Both `table_type` and `create_temp_table` will be passed to `write_pandas` and we test their expected behavior respectively.

src/snowflake/connector/pandas_tools.py
sfc-gh-sfan: Hmmmm `Literal` is only available from 3.8 and above (https://docs.python.org/3/library/typing.html#typing.Literal). So I'm not sure if we could do this given we still support 3.7 :(

sfc-gh-sfan: Discussed offline with @sfc-gh-mkeller , this is indeed caused by unable to import `Literal`. We import from `typing_extensions` instead.

test/integ/pandas/test_pandas_tools.py
```diff
@@ -142,6 +142,7 @@ def test_write_pandas_with_overwrite(
 @pytest.mark.parametrize("quote_identifiers", [True, False])
 @pytest.mark.parametrize("auto_create_table", [True, False])
 @pytest.mark.parametrize("create_temp_table", [True, False])
+@pytest.mark.parametrize("table_type", ["", "temp", "temporary", "transient"])
```
sfc-gh-mkeller, line 145: Please move this test into it's own test. This test is very thorough and it's one of the slowest one. Let's try to not add more checks to it.
On the bright side I think for your test you can skip a bunch of checks that this one does.

test/integ/pandas/test_pandas_tools.py
```diff
@@ -219,6 +219,79 @@ def test_write_pandas(
             cnx.execute_string(drop_sql)
 
 
+@pytest.mark.parametrize("table_type", ["", "temp", "temporary", "transient"])
+def test_write_pandas_table_type(
+    conn_cnx: Callable[..., Generator[SnowflakeConnection, None, None]],
+    table_type: str,
+):
+    with conn_cnx() as cnx:
+        table_name = f"driver_versions_{table_type}"
```
sfc-gh-mkeller, line 228: You should use the name randomizer for table names: https://github.com/snowflakedb/snowflake-connector-python/blob/0aec3eff1e4d2fe0f3845814645b7a66246a88f4/test/integ/pandas/test_pandas_tools.py#L54

```suggestion
        table_name = random_string(5, "write_pandas_table_type_")
```

test/integ/pandas/test_pandas_tools.py
```diff
@@ -219,6 +219,79 @@ def test_write_pandas(
...
+            assert table_info[0]["kind"] == expected_table_kind
+        finally:
+            cnx.execute_string(drop_sql)
+
+
+def test_write_pandas_create_temp_table_deprecation_warning(
+    conn_cnx: Callable[..., Generator[SnowflakeConnection, None, None]],
+):
+    with conn_cnx() as cnx:
+        table_name = "driver_versions"
```
sfc-gh-mkeller, line 259:
```suggestion
        table_name = random_string(5, "driver_versions_")
```

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1223, opened 2022-08-12
SNOW-532972 fix upload size behavior

sfc-gh-mkeller:

   Fixes SNOW-532972

  	As part of the [SDKless rewrite](https://github.com/snowflakedb/snowflake-connector-python/pull/725) it seems like we uninstentionally changed behavior of uploading files. Unless you create an external unencrypted stage we always encrypt a file that the client uploads and before rewrite [it used to be that upload size was set to the file before encryption](https://github.com/snowflakedb/snowflake-connector-python/blob/498e84c974a74072f899e6e5416a0317fee9d2f7/src/snowflake/connector/file_transfer_agent.py#L629), but in the rewrite we changed it so that the upload size was the true upload size. Encrypted files are 16B padded, [source](https://github.com/snowflakedb/snowflake-connector-python/blob/498e84c974a74072f899e6e5416a0317fee9d2f7/src/snowflake/connector/encryption_util.py#L108).
  	This is why in this ticket the empty file's size became 16B instead of 0.

### DIFF
--- a/src/snowflake/connector/storage_client.py
+++ b/src/snowflake/connector/storage_client.py
@@ -167,6 +167,8 @@ class SnowflakeStorageClient(ABC):
             self.encryption_metadata = SnowflakeEncryptionUtil.encrypt_stream(
                 meta.encryption_material, src_stream, encrypted_stream
             )
+            # Reset streams
+            encrypted_stream.seek(0)
             src_stream.seek(0)
             meta.upload_size = src_stream.seek(0, os.SEEK_END)
             src_stream.seek(0)
--- a/src/snowflake/connector/storage_client.py
+++ b/src/snowflake/connector/storage_client.py
@@ -159,7 +159,7 @@ class SnowflakeStorageClient(ABC):
                 meta.real_src_file_name,
                 tmp_dir=self.tmp_dir,
             )
-            meta.upload_size = os.path.getsize(self.data_file)
+            meta.upload_size = os.path.getsize(meta.real_src_file_name)
         else:
             encrypted_stream = BytesIO()
             src_stream = meta.src_stream or meta.intermediate_stream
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/storage_client.py
```diff
@@ -159,17 +159,19 @@ def encrypt(self) -> None:
...
             src_stream.seek(0)
             self.encryption_metadata = SnowflakeEncryptionUtil.encrypt_stream(
                 meta.encryption_material, src_stream, encrypted_stream
             )
-            src_stream.seek(0)
-            meta.upload_size = encrypted_stream.seek(0, os.SEEK_END)
+            # Reset streams
             encrypted_stream.seek(0)
+            src_stream.seek(0)
+            meta.upload_size = src_stream.seek(0, os.SEEK_END)
```
sfc-gh-sfan, line 173: Is this correct that we are not using `encrypted_stream` for upload size?

sfc-gh-jdu (approved):
@sfc-gh-sfan probably we want to import this change it stored proc connector too?

sfc-gh-sfan:
This does not really apply because for stored proc encryption is handled by the server, so sp connector does not do encryption.

src/snowflake/connector/storage_client.py
sfc-gh-mkeller: Yup, see my PR description. I should have a detailed explanation in there. I verified that old connector versions did this

sfc-gh-sfan: But is that the desired behavior? We explicitly add this padding for stored proc so that the returned result is the padded size. Encrypted size is consistent with the `ls @stage` result for the file 🤔

sfc-gh-mkeller:
After some offline discussion the team believes that this is the better behavior, we'll go ahead and keep things the way they are.
However; this PR should serve as a record of our decision.

PR #1226, opened 2022-08-18
SNOW-641540 Fix dtypes of fetch_pandas_all for empty result

sfc-gh-sfan:

   Fixes SNOW-641540

### DIFF
--- a/src/snowflake/connector/result_batch.py
+++ b/src/snowflake/connector/result_batch.py
@@ -545,6 +545,7 @@ class ArrowResultBatch(ResultBatch):
         self._context = context
         self._numpy = numpy
         self._number_to_decimal = number_to_decimal
+        self._dataframe = None
 
     def __repr__(self) -> str:
         return f"ArrowResultChunk({self.id})"
@@ -646,7 +647,7 @@ class ArrowResultBatch(ResultBatch):
         return self._create_iter(iter_unit=IterUnit.TABLE_UNIT, connection=connection)
 
     def _create_empty_table(self) -> Table:
-        """Returns emtpy Arrow table based on schema"""
+        """Returns empty Arrow table based on schema"""
         if installed_pandas:
             # initialize pyarrow type array corresponding to FIELD_TYPES
             FIELD_TYPE_TO_PA_TYPE = [e.pa_type() for e in FIELD_TYPES]
@@ -666,9 +667,13 @@ class ArrowResultBatch(ResultBatch):
         self, connection: SnowflakeConnection | None = None, **kwargs
     ) -> pandas.DataFrame:
         """Returns this batch as a pandas DataFrame"""
+        if self._dataframe is not None:
+            return self._dataframe
+
         self._check_can_use_pandas()
         table = self.to_arrow(connection=connection)
-        return table.to_pandas(**kwargs)
+        self._dataframe = table.to_pandas(**kwargs)
+        return self._dataframe
 
     def _get_pandas_iter(
         self, connection: SnowflakeConnection | None = None, **kwargs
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/result_batch.py
```diff
@@ -666,9 +667,13 @@ def to_pandas(
         self, connection: SnowflakeConnection | None = None, **kwargs
     ) -> pandas.DataFrame:
         """Returns this batch as a pandas DataFrame"""
+        if self._dataframe is not None:
```
sfc-gh-jdu, line 670: I'm a little bit confused that why do we need to cache here? Doesn't the change in result_set.py already fix the issue?

sfc-gh-sfan: I'm not sure how expensive `to_pandas()` call is. The change in `result_set.py` leads to double calling `to_pandas()` when the dataframe is empty. The first call happens in https://github.com/snowflakedb/snowflake-connector-python/blob/0e9af4be39c63d728ecf89ac4003fb50dd098135/src/snowflake/connector/result_batch.py#L683.

src/snowflake/connector/result_batch.py
```diff
@@ -545,6 +545,7 @@ def __init__(
         self._context = context
         self._numpy = numpy
         self._number_to_decimal = number_to_decimal
+        self._dataframe = None
```
sfc-gh-mkeller, line 548: missing type-hint

src/snowflake/connector/result_batch.py
sfc-gh-mkeller: We can't cache results. This was considered when we did the distributed fetch redesign, but was rejected when we couldn't cache results without blowing up our memory footprint.

Consider the following:
1. I run a query with 37 gb of results.
2. I run `cursor.fetch_pandas_batches` to fetch results in batches so that I don't need to hold all results in memory at once.
3. This caching will force the user to hold all previously seen results in memory since [result sets hold references to all of their result batches](https://github.com/snowflakedb/snowflake-connector-python/blob/8108f63c053fdd74a739a1a0e8418856288f9753/src/snowflake/connector/result_set.py#L115) and now those would hold a reference to their previously seen results.

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1169,6 +1169,16 @@ def assert_dtype_equal(a, b):
     )
 
 
+def assert_pandas_batch_types(batch, expected_types):
```
sfc-gh-mkeller, line 1172: missing type-hints

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1169,6 +1169,16 @@ def assert_dtype_equal(a, b):
...
+    assert batch.dtypes is not None
+
+    pandas_dtypes = batch.dtypes
+    # pd.string is represented as an np.object
+    # np.dtype string is not the same as pd.string (python)
+    for i, typ in enumerate(expected_types):
+        assert_dtype_equal(pandas_dtypes[i].type, numpy.dtype(typ).type)
```
sfc-gh-mkeller, line 1179: I think `zip` would work better here:
```suggestion
    for a, b in zip(pandas_dtypes, expected_types):
        assert_dtype_equal(a.type, numpy.dtype(b).type))
```

sfc-gh-mkeller (changes requested):

src/snowflake/connector/result_batch.py
sfc-gh-sfan: I see. Then let's just do two `to_pandas()`. I think the overhead should be quite minimal given it is empty anyways.

sfc-gh-jdu (approved):

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1229, opened 2022-08-18
SNOW-647539 Closing the cursor erases the last rowcount

sfc-gh-hachouraria:

   Fixes [snowflake-sqlalchemy/#244](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/244)

SQLAlchemy's batch delete and update operations (in ORM) return
rowcount of those operations after it closes the cursor. This
fails to work with Snowflake engine and returns None because we
are erasing the last known rowcount when the cursor is closed.

More information on the issue can be read on the original issue:
https://github.com/snowflakedb/snowflake-sqlalchemy/issues/244

In comparison the SQLite DB-API implementation does not remove
the rowcount information after the cursor is closed. Since the
cursor is not expected to be reusable after closure anyway this
appears safe to do.

This change preserves the existing behavior of the internal
cursor.reset() method that is responsible for removing the
rowcount information during close, but specializes the behavior
when the reset is performed in context of the cursor closing.

This approach preserves the existing use of cursor.reset() over
open cursors while also fixing the behavior for SQLAlchemy's use.

Testing:

  - Unit tests added for the reset and close methods of Cursor
  - Ran existing tests via `tox -e 'py37'` and verified no failures
     occurred due to rowcount related values
  - (Manual) Re-tested SQLAlchemy with the modified connector
    and verified it returns the valid rowcount for delete and
    update calls.

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -399,7 +399,7 @@ class SnowflakeCursor:
                 return False
 
             with self._lock_canceling:
-                self.reset()
+                self.reset(closing=True)
                 self._connection = None
                 del self.messages[:]
                 return True
@@ -1175,15 +1175,18 @@ class SnowflakeCursor:
             },
         )
 
-    def reset(self):
+    def reset(self, closing=False):
         """Resets the result set."""
-        self._total_rowcount = -1  # reset the rowcount
+        # SNOW-647539: Do not erase the rowcount
+        # information when closing the cursor
+        if not closing:
+            self._total_rowcount = -1
         if self._result_state != ResultState.DEFAULT:
             self._result_state = ResultState.RESET
         if self._result is not None:
             self._result = None
         if self._inner_cursor is not None:
-            self._inner_cursor.reset()
+            self._inner_cursor.reset(closing=closing)
             self._result = None
             self._inner_cursor = None
         self._prefetch_hook = None
--- a/test/integ/test_cursor.py
+++ b/test/integ/test_cursor.py
@@ -747,6 +747,9 @@ def test_closed_cursor(conn, db_parameters):
         with pytest.raises(InterfaceError, match="Cursor is closed in execute") as err:
             c.execute(f"select aa from {table_name}")
         assert err.value.errno == errorcode.ER_CURSOR_IS_CLOSED
+        assert (
+            c.rowcount == 5
+        ), "SNOW-647539: rowcount should remain available after cursor is closed"
 
 
 def test_fetchmany(conn, db_parameters):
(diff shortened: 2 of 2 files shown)


sfc-gh-hachouraria:

src/snowflake/connector/cursor.py
```diff
@@ -1175,15 +1175,18 @@ def scroll(self, value, mode="relative"):
             },
         )
 
-    def reset(self):
+    def reset(self, closing=False):
```
sfc-gh-mkeller, line 1178: I also fins this name misleading. If I want to not close the cursor why would I want the `rowcount` to be reset?
Also if we want `rowcount` to stay after a `reset` call do we want `_result_set` to stick around too?

test/integ/test_cursor.py
```diff
@@ -1039,11 +1042,16 @@ def test_empty_execution(conn, sql):
...
 def test_reset_fetch(conn, reuse_results):
-    """Tests behavior after resetting the cursor."""
+    """Tests behavior after resetting an open cursor."""
     with conn(reuse_results=reuse_results) as cnx:
         with cnx.cursor() as cur:
             cur.execute("select 1")
+            assert cur.rowcount == 1
             cur.reset()
+            assert (
+                cur.rowcount is None
```
sfc-gh-mkeller, line 1052: `rowcount` should never be None, it should be `-1` when it is reset according to https://peps.python.org/pep-0249/#rowcount

src/snowflake/connector/cursor.py
sfc-gh-hachouraria: Thank you for reviewing this Mark,

> I also find this name misleading.

To clarify, is this note regarding the use of the name `closing` for the optional argument?

> If I want to not close the cursor why would I want the rowcount to be reset?

The existing behavior of `cursor.reset()` is mainly around the correctness of multiple executions performed over a single cursor. The reset call ensures that upon a failure of the new execution the past attributes are not returned back.

In the case of `cursor.close()` calling `cursor.reset()` internally, no further executions will be permitted so it appears OK to retain the current `rowcount`.

> Also if we want `rowcount` to stay after a `reset` call do we want `_result_set` to stick around too?

No, the spec does strictly say no more operations should be permitted after a `cursor.close()` and I do not see any part in SQLAlchemy accessing data or requesting executions after closes. It appears to only be the cursor's attributes (`rowcount` and `description`) that appear accessed after `cursor.close()`.

test/integ/test_cursor.py
sfc-gh-hachouraria: The internal `_total_rowcount` behind this attribute method is `-1` but the property method wraps and returns `None` in that case.

This test was just asserting a behavior that has been the same since 2017 after this commit (I could not find a reason for this change): https://github.com/snowflakedb/snowflake-connector-python/commit/534674ad46d121a2249790814b06cac8876ce217#diff-6ef89a5da3fcd97c38c8bf752b68deb48e573ea5557334e5a5dd9d89ddcad03e

Perhaps the change above was made following the note in the PEP-249?

> Future versions of the DB API specification could redefine the latter case to have the object return None instead of -1.

If the behavior of the property method `cursor.rowcount` needs to be changed back to returning `-1` we can perhaps do it over a separate JIRA?

sfc-gh-mkeller (approved):
:ship: :ship:

sfc-gh-mkeller:
recheck

PR #1231, opened 2022-08-19
SNOW-645247: Python connector telemetry is partially broken

sfc-gh-aling:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-645247

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [x] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

Screen Shot 2022-08-18 at 5 43 07 PM

PR #1236, opened 2022-08-24
SNOW-644849: Add telemetry about imported pacakages at runtime

sfc-gh-jdu:

   SNOW-644849

   - [x] I am adding a new telemetry message

   We would like to know what packages are imported at Python runtime along with connector, which can help us understand the user scenarios of python connector. The message will looks like this:
```
{
  "source": "PythonSnowpark",
  "type": "client_imported_packages",
  "value": "{'pdb', 'bz2', 'abc', 'ssl', 'faulthandler', 'fnmatch', 'pytest', 'shlex', 'fractions', 'enum', 'queue', 'certifi', 'traceback', 'pprint', 'pwd', 'pickle', 'ast', 'uu', 'pkg_resources', 'codecs', 'gc', 'hmac', 'copy', 'errno', 'copyreg', 'subprocess', 'snowflake', 'weakref', 'decimal', 'logging', 'dataclasses', 'typing', 'numpy', 'pytest_forked', 'ipaddress', 'operator', 'signal', 'cmd', 'email', 'gettext', 'datetime', 'importlib', 'functools', 'base64', 'calendar', 'configparser', 'codeop', 'tempfile', 'cryptography', 'distutils', 'dis', 'contextvars', 'linecache', 'uuid', 'contextlib', 'posixpath', 'unittest', 'Cryptodome', 'selectors', 'iniconfig', 'binascii', 'locale', 'grp', 'select', 'encodings', 'posix', 'plistlib', 'sre_compile', 'platform', 'pycparser', 'ntpath', 'time', 'itertools', 'textwrap', 'csv', 'zlib', 'unicodedata', 'token', 'sys', 'struct', 'reprlib', 'keyword', 'charset_normalizer', 'sysconfig', 'cffi', 'attr', 'socket', 'pyexpat', 'numbers', 'jwt', 'teamcity', 'pathlib', 'random', 'opcode', 'py', 'json', 'threading', 'zipfile', 'ctypes', 'math', 'code', 'gzip', 'marshal', 'collections', 'genericpath', 'os', 'mimetypes', 'pytz', 'xml', 'concurrent', 'bdb', 'string', 'difflib', 'argparse', 'quopri', 'http', 'inspect', 'warnings', 'pluggy', 'hashlib', 'urllib', 'zipimport', 'OpenSSL', 'bisect', 'site', 'glob', 'stat', 'html', 'multiprocessing', 'atexit', 'io', 'secrets', 'sre_constants', 'lzma', 'heapq', 'tokenize', 'pkgutil', 'sre_parse', 'shutil', 'tomli', 'webbrowser', 'asyncio', 'asn1crypto', 'stringprep', 'test', 're', 'builtins', 'cython_runtime', 'types', 'array'}"
}
```

### DIFF
--- a/src/snowflake/connector/__init__.py
+++ b/src/snowflake/connector/__init__.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 apilevel = "2.0"
 threadsafety = 2
 paramstyle = "pyformat"
-log_imported_packages_in_telemetry = True
 
 import logging
 from logging import NullHandler
@@ -91,5 +90,4 @@ __all__ = [
     "ROWID",
     # Extended data type (experimental)
     "Json",
-    "log_imported_packages_in_telemetry",
 ]
--- a/src/snowflake/connector/__init__.py
+++ b/src/snowflake/connector/__init__.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 apilevel = "2.0"
 threadsafety = 2
 paramstyle = "pyformat"
+log_imported_packages_in_telemetry = True
 
 import logging
 from logging import NullHandler
@@ -90,4 +91,5 @@ __all__ = [
     "ROWID",
     # Extended data type (experimental)
     "Json",
+    "log_imported_packages_in_telemetry",
 ]
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -83,7 +83,12 @@ from .network import (
     SnowflakeRestful,
 )
 from .sqlstate import SQLSTATE_CONNECTION_NOT_EXISTS, SQLSTATE_FEATURE_NOT_SUPPORTED
-from .telemetry import TelemetryClient, TelemetryData, TelemetryField
+from .telemetry import (
+    TelemetryClient,
+    TelemetryData,
+    TelemetryField,
+    generate_telemetry_data,
+)
 from .telemetry_oob import TelemetryService
 from .time_util import HeartBeatTimer, get_time_millis
 from .util_text import construct_hostname, parse_account, split_statements
@@ -1565,12 +1570,18 @@ class SnowflakeConnection:
                 for k in sys.modules.keys()
                 if not k.startswith("_")
             }
-            obj = {
-                TelemetryField.KEY_TYPE.value: TelemetryField.IMPORTED_PACKAGES.value,
-                TelemetryField.KEY_SOURCE.value: self.application
-                if self.application
-                else CLIENT_NAME,
-                "value": str(imported_modules),
-            }
             ts = get_time_millis()
-            self._log_telemetry(TelemetryData(obj, ts))
+            self._log_telemetry(
+                TelemetryData(
+                    generate_telemetry_data(
+                        from_dict={
+                            TelemetryField.KEY_TYPE.value: TelemetryField.IMPORTED_PACKAGES.value,
+                            TelemetryField.KEY_SOURCE.value: self.application
+                            if self.application
+                            else CLIENT_NAME,
+                            TelemetryField.KEY_VALUE.value: str(imported_modules),
+                        }
+                    ),
+                    ts,
+                )
+            )
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -965,11 +965,6 @@ class SnowflakeConnection:
             )
             self._use_openssl_only = os.environ["SF_USE_OPENSSL_ONLY"] == "True"
 
-        if "log_imported_packages_in_telemetry" in kwargs:
-            self._log_imported_packages_in_telemetry = kwargs[
-                "log_imported_packages_in_telemetry"
-            ]
-
     def cmd_query(
         self,
         sql: str,
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1566,8 +1566,8 @@ class SnowflakeConnection:
                 if not k.startswith("_")
             }
             obj = {
-                TelemetryField.KEY_TYPE: TelemetryField.IMPORTED_PACKAGES.value,
-                TelemetryField.KEY_SOURCE: self.application
+                TelemetryField.KEY_TYPE.value: TelemetryField.IMPORTED_PACKAGES.value,
+                TelemetryField.KEY_SOURCE.value: self.application
                 if self.application
                 else CLIENT_NAME,
                 "value": str(imported_modules),
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -198,8 +198,8 @@ DEFAULT_CONFIGURATION: dict[str, tuple[Any, type | tuple[type, ...]]] = {
         (type(None), str),
     ),  # Path to connection diag whitelist json
     "log_imported_packages_in_telemetry": (
-        None,
-        (type(None), bool),
+        True,
+        bool,
     ),  # Whether to log imported packages in telemetry
 }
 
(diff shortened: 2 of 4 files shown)


test/integ/test_connection.py
```diff
@@ -1107,3 +1108,54 @@ def test_ocsp_cache_working(conn_cnx):
...
+        # these imports are not used but for testing
+        import html.parser  # noqa: F401
+        import json  # noqa: F401
+        import multiprocessing as mp  # noqa: F401
+        from datetime import date  # noqa: F401
+        from math import sqrt  # noqa: F401
+
+        with snowflake.connector.connect(**config) as conn:
+            with capture_sf_telemetry.patch_connection(conn, False) as telemetry_test:
+                conn._log_telemetry_imported_packages(conn._imported_vars)
```
sfc-gh-jdu, line 1137: that's the best way I can think of to test telemetry based on the current test infra (I'm not able to get the telemetry sent during connection creation). Please comment if you have a better way.

sfc-gh-kwagner: Yea I think based on the complexity of this, it may be easier to just get the packages in the python environment and send that as telemetry. What do you think @sfc-gh-achandrasekaran

sfc-gh-sfan:
I kinda agree this is a little bit too hacky, and I'm not particularly sure if `getouterframes` works in stored proc or not :(

sfc-gh-jdu:
> I kinda agree this is a little bit too hacky...(scratched my comment about sp as I thought this is snowpark by mistake).

It will not work in SP (similar to https://github.com/snowflakedb/snowpark-python/pull/274), and I don't expect it works in SP because per our discussion, we can just collect pakcages information of a SP from snowhouse. Though it's pretty hacky, it still works (you can find in the test) and I'm just looking for a better way to do that...

sfc-gh-jdu:
I changed to use `sys.modules` that we can capture the usages of packages other than pandas. Note that as long as pandas exists in the environment, it will be pulled into sys.modules. But our goal is actually to get some information about what packages are used along with connector (or snowpark) and pandas. Note that we are not able to get pandas (also pyarrow) usage in this telemetry records.

src/snowflake/connector/connection.py
```diff
@@ -1541,3 +1555,22 @@ def _all_async_queries_finished(self) -> bool:
             not self.is_still_running(self.get_query_status(q)) for q in queries
         )
         return all(finished_async_queries)
+
+    def _log_telemetry_imported_packages(self) -> None:
+        if self._log_imported_packages_in_telemetry:
+            # filter out duplicates caused by submodules
+            # and internal modules with names starting with an underscore
+            imported_modules = {
+                k.split(".", maxsplit=1)[0]
```
sfc-gh-sfan, line 1564: It's possible that several packages would have the same top level module, e.g. `snowflake`. But this can be considered as corner cases. There probably isn't a one-size-fit-all solution.

sfc-gh-sfan (approved):

src/snowflake/connector/connection.py
sfc-gh-jdu: yea, that's possible but I think these packages are rare. Otherwise the telemetry will be blown up by submodules (simply import numpy you will get ~30 submodules)

sfc-gh-kwagner (approved):
LGTM

src/snowflake/connector/connection.py
```diff
@@ -1541,3 +1560,28 @@ def _all_async_queries_finished(self) -> bool:
...
+            }
+            ts = get_time_millis()
+            self._log_telemetry(
+                TelemetryData(
+                    generate_telemetry_data(
+                        from_dict={
+                            TelemetryField.KEY_TYPE.value: TelemetryField.IMPORTED_PACKAGES.value,
+                            TelemetryField.KEY_SOURCE.value: self.application
+                            if self.application
+                            else CLIENT_NAME,
```
sfc-gh-mkeller, line 1581: Do we need to manually feed the `source` key-value pair into each of the telemetry calls, @sfc-gh-aling ?
If so then should this not be part of `generate_telemetry_data`?

test/integ/test_connection.py
```diff
@@ -1107,3 +1108,73 @@ def test_ocsp_cache_working(conn_cnx):
...
+            assert len(telemetry_test.records) > 0
+            for t in telemetry_test.records:
+                if (
+                    t.message[TelemetryField.KEY_TYPE.value]
+                    == TelemetryField.IMPORTED_PACKAGES.value
+                ):
+                    assert (
+                        new_application_name
+                        == t.message[TelemetryField.KEY_SOURCE.value]
+                    )
```
sfc-gh-mkeller, line 1170: I'm a fan of using a generator of booleans and `any` in cases of this. If the telemetry changes then the test in its current form would silently not test anything. With `any` we make sure that if occurs once.
```suggestion
            any(
                [
                    t.message[TelemetryField.KEY_TYPE.value] == TelemetryField.IMPORTED_PACKAGES.value
                    and new_application_name == t.message[TelemetryField.KEY_SOURCE.value]
                    for t in telemetry_test.records:
                ]
            )
```

src/snowflake/connector/__init__.py
```diff
@@ -9,6 +9,7 @@
 apilevel = "2.0"
 threadsafety = 2
 paramstyle = "pyformat"
+log_imported_packages_in_telemetry = True
```
sfc-gh-mkeller, line 12: Let's please use a connection parameter instead of this global variable

src/snowflake/connector/connection.py
```diff
@@ -192,6 +197,10 @@ def DefaultConverterClass():
         None,
         (type(None), str),
     ),  # Path to connection diag whitelist json
+    "log_imported_packages_in_telemetry": (
+        None,
+        (type(None), bool),
```
sfc-gh-mkeller, line 202: If we are going to switch to only having a connection parameter then this should default to `True`

sfc-gh-mkeller (changes requested):

src/snowflake/connector/connection.py
sfc-gh-jdu: We should have `source`. `generate_telemetry_data` doesn't set it https://github.com/snowflakedb/snowflake-connector-python/blob/2c3f6d59cb824c568de15d5c646cddf35ebe8a3e/src/snowflake/connector/telemetry.py#L196-L215, and when we log job data, we do set `source`. https://github.com/snowflakedb/snowflake-connector-python/blob/2c3f6d59cb824c568de15d5c646cddf35ebe8a3e/src/snowflake/connector/cursor.py#L1221

test/integ/test_connection.py
sfc-gh-jdu: done

src/snowflake/connector/__init__.py
sfc-gh-jdu: done

src/snowflake/connector/connection.py
sfc-gh-jdu: done

sfc-gh-mkeller (approved):
:ship: :ship:

src/snowflake/connector/connection.py
sfc-gh-aling: in my previous [fix PR](https://github.com/snowflakedb/snowflake-connector-python/pull/1231/files), I didn't realize that `source` could be part of the basic data for telemetry. I thought it is an operation based item as the exception telemetry does not include the `source` information.

but looks like `source` could/should be part of the basic information for those non-oob telemetry data.

let's add it?

sfc-gh-aling: we could have a separate jira for the change.

sfc-gh-jdu: Yes, I think we should add it! Adam, can I merge this PR first, then you can modify my commit?

sfc-gh-mkeller: Yes, let's have Adam's PR update this function call!

PR #1252, opened 2022-09-15
SNOW-364424: Exposed to_parquet kwargs parameter in write_pandas

sfc-gh-heshah:

   [SNOW-364424: Support for providing kwargs to to_parquet when using write_pandas](https://snowflakecomputing.atlassian.net/browse/SNOW-364424)

        - It's a small 2 line change. The use of `**kwargs` in the `df.to_parquet` function seems very niche so if we could get some idea of how the customer uses the functionality, I could write a test for it. Just as a sanity check, does the user just want access to the non-exposed parameters of `to_parquet` (including `engine`, `index`, `partition_cols`, `storage_options`), or actually `**kwargs`?

4. Please describe how your code solves the related issue.

   Added `**kwargs` to `write_pandas` arguments and passed the parameter to the `to_parquet` call in the function.

### DIFF
--- a/test/integ/pandas/test_pandas_tools.py
+++ b/test/integ/pandas/test_pandas_tools.py
@@ -45,14 +45,12 @@ sf_connector_version_df = LazyVar(
 
 @pytest.mark.parametrize("quote_identifiers", [True, False])
 @pytest.mark.parametrize("auto_create_table", [True, False])
-@pytest.mark.parametrize("index", [True, False])
-@pytest.mark.parametrize("engine", ["auto", "pyarrow"])
+@pytest.mark.parametrize("index", [False])
 def test_write_pandas_with_overwrite(
     conn_cnx: Callable[..., Generator[SnowflakeConnection, None, None]],
     quote_identifiers: bool,
     auto_create_table: bool,
     index: bool,
-    engine: str,
 ):
     """Tests whether overwriting table using a Pandas DataFrame works as expected."""
     random_table_name = random_string(5, "userspoints_")
@@ -93,7 +91,6 @@ def test_write_pandas_with_overwrite(
                 quote_identifiers=quote_identifiers,
                 auto_create_table=auto_create_table,
                 overwrite=True,
-                engine=engine,
                 index=index,
             )
             # Write dataframe with 1 row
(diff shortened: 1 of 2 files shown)


sfc-gh-aalam (approved):
LGTM. I'll also wait for @sfc-gh-mkeller to view this before merging

test/integ/pandas/test_pandas_tools.py
```diff
@@ -45,10 +45,14 @@
 
 @pytest.mark.parametrize("quote_identifiers", [True, False])
 @pytest.mark.parametrize("auto_create_table", [True, False])
+@pytest.mark.parametrize("index", [True, False])
+@pytest.mark.parametrize("engine", ["auto", "pyarrow"])
```
sfc-gh-mkeller, line 49: These tests are really slow relatively to the others. I'd like to not make it run 16 times instead of 4.
Do you think that it'd be enough to test `**kwargs` with sending in constant values in the already existing 4 tests, instead of the parametrized ones?

test/integ/pandas/test_pandas_tools.py
```diff
@@ -142,6 +152,8 @@ def test_write_pandas_with_overwrite(
 @pytest.mark.parametrize("quote_identifiers", [True, False])
 @pytest.mark.parametrize("auto_create_table", [True, False])
 @pytest.mark.parametrize("create_temp_table", [True, False])
+@pytest.mark.parametrize("index", [True, False])
+@pytest.mark.parametrize("engine", ["auto", "pyarrow"])
 def test_write_pandas(
```
sfc-gh-mkeller, line 157: Same as the other comment. However; this one runs 8 times already and your change would make it run 32 times.

sfc-gh-mkeller (changes requested):

test/integ/pandas/test_pandas_tools.py
sfc-gh-heshah: I'm going to remove the parametrization on the `engine` arg as `auto` uses `pyarrow` by default. It looks like for the `index` arg there's no incentive to set it to True as the index is saved by default in the metadata in a faster way when left as None, but I think setting it to False could be heavily used to save space for highly transformed DataFrames, so I would like to leave the test for it there.

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1258, opened 2022-09-22
SNOW-344567: Exposing write_pandas keyword arguments in pd_writer

### LINKED ISSUE #679: SNOW-344567: Expose `write_pandas` keyword arguments in `pd_writer`
## What is the current behavior?

Today there are two ways to write pandas DataFrames to Snowflake:

1. Using pandas' `DataFrame.to_sql` along with [`pd_writer`](https://github.com/snowflakedb/snowflake-connector-python/blob/15f7b2239f4db5a38662dba18c1ae9321263dc9a/src/snowflake/connector/pandas_tools.py#L207)

2. Using [`write_pandas`](https://github.com/snowflakedb/snowflake-connector-python/blob/15f7b2239f4db5a38662dba18c1ae9321263dc9a/src/snowflake/connector/pandas_tools.py#L43).

Looking at the implementation of `pd_writer` it appears to be a relatively lightweight wrapper function around `write_pandas`. However `pd_writer` doesn't expose a way for users to specify some of the keyword arguments used when calling `write_pandas` (e.g. `parallel=`) 

## What is the desired behavior?

It would be great if `pd_writer` could allow users to pass keyword arguments through to `write_pandas`. This already happens today with `quote_identifiers=`, but not with other parameters like `compression=`, `chunk_size=`, `parallel=`, etc. 

## How would this improve `snowflake-connector-python`?

It would provide better feature parity between `pd_writer` and `write_pandas`. One concrete improvement is if a user needs to specify `parallel=1` to avoid multi-threading issues like https://github.com/snowflakedb/snowflake-connector-python/issues/156 then `pd_writer` can't be used today. The feature request proposed here would allow `pd_writer` to be used in this case. 

## References, Other Background

N/A

sfc-gh-heshah:

   Fixes #679 

       - `pd_writer` functionality has to be used with `snowflake-sqlalchemy` as the `DataFrame.to_sql` function requires a sqlalchemy engine, so the tests are in a separate PR in the `snowflake-sqlalchemy` repository.

    Adding `write_pandas` non-deprecated keyword arguments to both `pd_writer` and `make_pd_writer` so users can actually use them in conjunction with `DataFrame.to_sql` and `snowflake-sqlalchemy`.

### DIFF
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -292,13 +292,14 @@ def write_pandas(
 
 
 def make_pd_writer(
-    quote_identifiers: bool = True,
+    **kwargs,
 ) -> Callable[
     [
         pandas.io.sql.SQLTable,
         sqlalchemy.engine.Engine | sqlalchemy.engine.Connection,
         Iterable,
         Iterable,
+        Any,
     ],
     None,
 ]:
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/pandas_tools.py
```diff
@@ -291,13 +291,26 @@ def write_pandas(
...
         sqlalchemy.engine.Engine | sqlalchemy.engine.Connection,
         Iterable,
         Iterable,
+        int,
+        str,
+        int,
+        bool,
+        bool,
+        bool,
+        Literal,
```
sfc-gh-mkeller, line 313: I don't think this is correct

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1261, opened 2022-09-27
SNOW-664933 Fix write_pandas with empty dataframe

sfc-gh-stan:

   Fixes #SNOW-664933

  `write_pandas` used to raise `ValueError` when we attempt to write a dataframe with no rows, this is caused by `chunk_helper` attempting to generate chunks of 0 rows (`for i in range(0, len(lst), n):` when `n=0`). With this code change, we will just generate one chunk when the dataframe has fewer rows than the chunk size, which includes the special case of 0 rows.

### DIFF
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -35,7 +35,7 @@ logger = getLogger(__name__)
 
 def chunk_helper(lst: T, n: int) -> Iterator[tuple[int, T]]:
     """Helper generator to chunk a sequence efficiently with current index like if enumerate was called on sequence."""
-    if n >= len(lst):
+    if len(lst) == 0:
         yield 0, lst
         return
     for i in range(0, len(lst), n):
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -35,6 +35,9 @@ logger = getLogger(__name__)
 
 def chunk_helper(lst: T, n: int) -> Iterator[tuple[int, T]]:
     """Helper generator to chunk a sequence efficiently with current index like if enumerate was called on sequence."""
+    if n >= len(lst):
+        yield 0, lst
+        return
     for i in range(0, len(lst), n):
         yield int(i / n), lst[i : i + n]
 
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -12,7 +12,7 @@ import warnings
 from functools import partial
 from logging import getLogger
 from tempfile import TemporaryDirectory
-from typing import TYPE_CHECKING, Callable, Iterable, Iterator, Sequence, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Iterable, Iterator, Sequence, TypeVar
 
 from typing_extensions import Literal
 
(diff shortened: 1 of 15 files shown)


test/integ/pandas/test_pandas_tools.py
```diff
@@ -300,6 +300,21 @@ def test_invalid_table_type_write_pandas(
             )
 
 
+def test_empty_dataframe_auto_create_table_write_pandas(
```
sfc-gh-stan, line 303: The test name seems redundant, the issue we fix is irrelevant to `auto_create_table`.

sfc-gh-sfan (approved):
LGTM. Can you also port this to sp connector?

src/snowflake/connector/pandas_tools.py
```diff
@@ -35,6 +35,9 @@
 
 def chunk_helper(lst: T, n: int) -> Iterator[tuple[int, T]]:
     """Helper generator to chunk a sequence efficiently with current index like if enumerate was called on sequence."""
+    if n >= len(lst):
+        yield 0, lst
+        return
```
sfc-gh-mkeller, line 40: So it worked for shorter lists already:
```
>>> list(chunk_helper([1, 2], 3))
[(0, [1, 2])]
```
There's only one case the original function didn't work for and so this is more explicit:
```suggestion
    if n == len(lst):
        yield 0, lst
        return
```

sfc-gh-mkeller (changes requested):

test/integ/pandas/test_pandas_tools.py
```diff
@@ -300,6 +300,21 @@ def test_invalid_table_type_write_pandas(
...
+    table_name = random_string(5, "empty_dataframe_")
+    df = pandas.DataFrame([], columns=["name", "balance"])
+    with conn_cnx() as cnx:
+        try:
+            success, num_chunks, num_rows, _ = write_pandas(
+                cnx, df, table_name, auto_create_table=True
+            )
+            assert success and num_chunks == 1 and num_rows == 0
+        finally:
+            cnx.execute_string(f"DROP TABLE IF EXISTS {table_name}")
```
sfc-gh-mkeller, line 315: No need to call that helper fn (execute is enough), also if we used a temp table then this wouldn't be necessary at all.
```suggestion
            cnx.execute(f"DROP TABLE IF EXISTS {table_name}")
```

test/integ/pandas/test_pandas_tools.py
```diff
@@ -300,6 +300,21 @@ def test_invalid_table_type_write_pandas(
...
 
+def test_empty_dataframe_write_pandas(
+    conn_cnx: Callable[..., Generator[SnowflakeConnection, None, None]],
+):
+    table_name = random_string(5, "empty_dataframe_")
+    df = pandas.DataFrame([], columns=["name", "balance"])
+    with conn_cnx() as cnx:
+        try:
+            success, num_chunks, num_rows, _ = write_pandas(
+                cnx, df, table_name, auto_create_table=True
```
sfc-gh-mkeller, line 311: We could use a temp table for this:
```suggestion
                cnx, df, table_name, auto_create_table=True, table_type="temp",
```

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1264, opened 2022-09-28
replace union types on cursor results by concrete types

### LINKED ISSUE #1263: SNOW-669760: Union types returned by cursor.fetchone (and similar) require extensive casting on user side

   Python 3.6.8

   Linux-5.16.13-generic-aarch64-with-centos-7.9.2009-AltArch

   This has no impact on the issue.

4. What did you do?

   `cursor.execute(sql_query).fetchone()`

  Regarding methods fetchone, fetchmany, fetchall, __iter__ and _result_iterator:
  I expect to get a `list` type from the methods on a SnowflakeCursor and a `dict` from the same methods on the DictCursor. Instead, I get a `dict | list` union type either way.

   Not relevant

**Problem**

Because the SnowflakeCursor/DictCursor returns a Union type as result for method like fetchone, users of the connector have to cast the result to make it work with mypy: e.g., `cast(list, cursor.fetchone())[0]`.

This is unnecessary, DictCursor should return dicts and SnowflakeCursor (or a ListCursor) should return lists.

sfc-gh-nsemmler:

   Fixes #1263

None of the above

   Add a new subclass ListCursor to SnowflakeCursor and make the latter abstract, so that DictCursor returns dicts and ListCursor returns a list instead of a union type

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -15,6 +15,7 @@ import uuid
 from enum import Enum
 from logging import getLogger
 from threading import Lock, Timer
+from abc import ABC
 from typing import (
     IO,
     TYPE_CHECKING,
@@ -26,6 +27,7 @@ from typing import (
     NoReturn,
     Sequence,
     TypeVar,
+    Generic
 )
 
 from snowflake.connector.result_batch import create_batches_from_response
@@ -155,7 +157,10 @@ class ResultState(Enum):
     RESET = 3
 
 
-class SnowflakeCursor:
+T2 = TypeVar('T2', list, dict)
+
+
+class SnowflakeCursor(ABC, Generic[T2]):
     """Implementation of Cursor object that is returned from Connection.cursor() method.
 
     Attributes:
@@ -1085,9 +1090,7 @@ class SnowflakeCursor:
             self.execute(command, params=param, _do_reset=False, **kwargs)
         return self
 
-    def _result_iterator(
-        self,
-    ) -> Generator[dict, None, None] | Generator[tuple, None, None]:
+    def _result_iterator(self) -> Generator[T2, None, None]:
         """Yields the elements from _result and raises an exception when appropriate."""
         try:
             for _next in self._result:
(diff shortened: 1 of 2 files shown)


sfc-gh-nsemmler:

sfc-gh-nsemmler:
recheck

sfc-gh-dschumann (approved):
LGTM, great contribution!

src/snowflake/connector/cursor.py
```diff
@@ -155,7 +157,10 @@ class ResultState(Enum):
     RESET = 3
 
 
-class SnowflakeCursor:
+T2 = TypeVar("T2", list, dict)
```
sfc-gh-stan, line 160: `T2 = TypeVar("T2", tuple, dict)`?

sfc-gh-mkeller (changes requested):

src/snowflake/connector/cursor.py
sfc-gh-nsemmler: thx fixed

src/snowflake/connector/cursor.py
```diff
@@ -1105,7 +1108,7 @@ def _result_iterator(
             else:
                 yield None
 
-    def fetchone(self) -> dict | tuple | None:
+    def fetchone(self) -> T2 | None:
```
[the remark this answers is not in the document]
sfc-gh-nsemmler: @sfc-gh-mkeller Sorry for the long delay. I made the SnowflakeCursor abstract, so you cannot do `SnowflakeCursor(use_dict_result=True)`. This should work as is. 

I rather have something now that removes the Union types and have a cleaner solution later. Our current code base is littered with statements like the following because mypy produces an erro.
`my_result = cast(list, cursor.fetchone())[0]`

sfc-gh-nsemmler:
After more thought, I have to realize that my use of the generics is wrong. As the type of the generic is decided by the subclass or instantiation, we cannot resolve the type in the SnowflakeCursor class. 

Making this work would require one of the following:
1. SnowflakeCursor forwards the generic to the underlying _result object. This requires a change to the implementation of a different object.
2. SnowflakeCursor becomes abstract and DictCursor/TupleCursor implement the _result object as generating dict/tuple respectively. This would break current instantiation of the SnowflakeCursor class.

PR #1273, opened 2022-10-03
SNOW-657710 Parallelize async query status check when closing connection

sfc-gh-stan:

   Fixes #SNOW-657710

   Parallelize the status checks by running all the GET requests in a ThreadPoolExecutor.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -5,7 +5,6 @@
 
 from __future__ import annotations
 
-import copy
 import logging
 import os
 import re
@@ -13,6 +12,8 @@ import sys
 import uuid
 import warnings
 import weakref
+from collections import OrderedDict
+from concurrent.futures.thread import ThreadPoolExecutor
 from difflib import get_close_matches
 from functools import partial
 from io import StringIO
@@ -267,8 +268,8 @@ class SnowflakeConnection:
         self._errorhandler = Error.default_errorhandler
         self._lock_converter = Lock()
         self.messages = []
-        self._async_sfqids = set()
-        self._done_async_sfqids = set()
+        self._async_sfqids = OrderedDict()
+        self._done_async_sfqids = OrderedDict()
         self.telemetry_enabled = False
         self._session_parameters: dict[str, str | int | bool] = {}
         logger.info(
@@ -1459,8 +1460,10 @@ class SnowflakeConnection:
         status_ret = QueryStatus[status]
         # If query was started by us and it has finished let's cache this info
         if sf_qid in self._async_sfqids and not self.is_still_running(status_ret):
-            self._async_sfqids.remove(sf_qid)
-            self._done_async_sfqids.add(sf_qid)
+            self._async_sfqids.pop(
+                sf_qid, None
+            )  # Prevent KeyError when multiple threads try to remove the same query id
+            self._done_async_sfqids[sf_qid] = sf_qid
         return status_ret, status_resp
 
     def get_query_status(self, sf_qid: str) -> QueryStatus:
@@ -1492,7 +1495,7 @@ class SnowflakeConnection:
         queries = status_resp["data"]["queries"]
         if self.is_an_error(status):
             if sf_qid in self._async_sfqids:
-                self._async_sfqids.remove(sf_qid)
+                self._async_sfqids.pop(sf_qid, None)
             message = status_resp.get("message")
             if message is None:
                 message = ""
@@ -1541,13 +1544,27 @@ class SnowflakeConnection:
 
     def _all_async_queries_finished(self) -> bool:
         """Checks whether all async queries started by this Connection have finished executing."""
-        queries = copy.copy(
-            self._async_sfqids
-        )  # get_query_status might update _async_sfqids, let's copy the list
-        finished_async_queries = (
-            not self.is_still_running(self.get_query_status(q)) for q in queries
-        )
-        return all(finished_async_queries)
+
+        queries = list(self._async_sfqids.keys())
+        if not queries:
+            return True
+        num_workers = min(self.client_prefetch_threads, len(queries))
+        found_unfinished_query = False
+
+        def check_status_helper(
+            sfq_id: str,
+        ):  # We should upgrade to using cancel_futures=True once supporting 3.9+
+            nonlocal found_unfinished_query
+            if not found_unfinished_query and self.is_still_running(
+                self.get_query_status(sfq_id)
+            ):
+                found_unfinished_query = True
+
+        with ThreadPoolExecutor(max_workers=num_workers) as tpe:
+            for sfq_id in queries:
+                tpe.submit(check_status_helper, sfq_id)
+
+        return not found_unfinished_query
 
     def _log_telemetry_imported_packages(self) -> None:
         if self._log_imported_packages_in_telemetry:
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -267,8 +267,8 @@ class SnowflakeConnection:
         self._errorhandler = Error.default_errorhandler
         self._lock_converter = Lock()
         self.messages = []
-        self._async_sfqids = {}
-        self._done_async_sfqids = {}
+        self._async_sfqids: dict[str, None] = {}
+        self._done_async_sfqids: dict[str, None] = {}
         self.telemetry_enabled = False
         self._session_parameters: dict[str, str | int | bool] = {}
         logger.info(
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/connection.py
```diff
@@ -1541,13 +1543,27 @@ def is_an_error(status: QueryStatus) -> bool:
...
-        )
-        return all(finished_async_queries)
+
+        queries = list(reversed(self._async_sfqids.keys()))
+        if not queries:
+            return True
+        num_workers = min(self.client_prefetch_threads, len(queries))
+        found_unfinished_query = False
+
+        def check_status_helper(
```
sfc-gh-stan, line 1553: If we want to avoid submitting all checks at the same time, we could also use a `callback` like:
```
with ThreadPoolExecutor(num_workers) as tpe:
    def check_status_callback(future: Future):
        nonlocal found_unfinished_query, tpe
        status = future.result()
        if self.is_still_running(status):
            found_unfinished_query = True
        if not found_unfinished_query and queries:
            try:
                sfq_id = queries.pop()
                tpe.submit(self.get_query_status, sfq_id).add_done_callback(check_status_callback)
            except IndexError:
                pass  # handle racing condition
```

src/snowflake/connector/connection.py
```diff
@@ -267,8 +267,8 @@ def __init__(self, **kwargs):
         self._errorhandler = Error.default_errorhandler
         self._lock_converter = Lock()
         self.messages = []
-        self._async_sfqids = set()
-        self._done_async_sfqids = set()
+        self._async_sfqids = {}
```
sfc-gh-stan, line 270: `dict` starts to preserve insertion order since 3.7, so using `reverse(dict.keys())` allows us to start checking the last submitted queries first.

src/snowflake/connector/connection.py
```diff
@@ -1541,13 +1543,27 @@ def is_an_error(status: QueryStatus) -> bool:
...
         """Checks whether all async queries started by this Connection have finished executing."""
-        queries = copy.copy(
-            self._async_sfqids
-        )  # get_query_status might update _async_sfqids, let's copy the list
-        finished_async_queries = (
-            not self.is_still_running(self.get_query_status(q)) for q in queries
-        )
-        return all(finished_async_queries)
+
+        queries = list(reversed(list(self._async_sfqids.keys())))
```
sfc-gh-stan, line 1547: This could be simplified to `list(reversed(self._async_sfqids.keys()))` once we support 3.8+

src/snowflake/connector/connection.py
```diff
@@ -1541,13 +1543,27 @@ def is_an_error(status: QueryStatus) -> bool:
...
+        def check_status_helper(
+            sfq_id: str,
+        ):  # We should upgrade to using cancel_futures=True once supporting 3.9+
+            nonlocal found_unfinished_query
+            if not found_unfinished_query and self.is_still_running(
+                self.get_query_status(sfq_id)
+            ):
+                found_unfinished_query = True
+
+        with ThreadPoolExecutor(max_workers=num_workers) as tpe:
```
sfc-gh-hachouraria, line 1562: Thank you for the improvements here!

Minor ask: For easier troubleshooting through logging, could an appropriately named `thread_name_prefix` be passed to this constructor to differentiate it from other thread-pools at runtime (such as those that download result batches, customer's own thread-pools, etc.)

src/snowflake/connector/connection.py
sfc-gh-mkeller: Missing type-hints on these definitions

src/snowflake/connector/connection.py
sfc-gh-mkeller: If so, how about?
```suggestion
        if sys.version_info >= (3, 8):
            queries = list(reversed(self._async_sfqids.keys()))
        else:
            queries = list(reversed(list(self._async_sfqids.keys())))
```
This makes refactoring this bit a child's play in the future.

src/snowflake/connector/connection.py
sfc-gh-mkeller: I think [deque](https://docs.python.org/3/library/collections.html#deque-objects)s would be better here, there's no up-side to storing the keys as the values too, but you double the memory we use to keep the order things are inserted in.

src/snowflake/connector/connection.py
```diff
@@ -1541,13 +1543,27 @@ def is_an_error(status: QueryStatus) -> bool:
...
+            for sfq_id in queries:
+                tpe.submit(check_status_helper, sfq_id)
```
sfc-gh-mkeller, line 1564: I think you forgot to wait on the futures that [submit](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit) returns.
Could I instead suggest using [map](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map)? The trick with that helper is to cast it to a list to wait on all of the results.
Otherwise you'll need to construct a list of these futures and wait on each one of them.

sfc-gh-mkeller (changes requested):

src/snowflake/connector/connection.py
sfc-gh-stan: I think using `with` means it does an implicit `shutdown(wait=True)` when exiting. Since all the work it does is in the `callback`, I don't need to wait on the futures explicitly in the main thread. But maybe this is not good for code readability?

sfc-gh-stan: We could get rid of `callback` and replace it with `map` and `as_completed`, and do the post-processing in the main thread 🤔

test/integ/test_async.py
```diff
@@ -235,3 +235,13 @@ def test_close_connection_with_running_async_queries(conn_cnx):
...
         assert not con._all_async_queries_finished()
     assert len(con._done_async_sfqids) < 2 and con.rest is None
+
+
+def test_close_connection_with_completed_async_queries(conn_cnx):
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            cur.execute_async("select 1")
+            cur.execute_async("select 2")
+        time.sleep(5)
```
sfc-gh-mkeller, line 245: Instead of doing these hard sleeps, I think doing a while checks like this would be faster most of the time:
https://github.com/snowflakedb/snowflake-connector-python/blob/2d78cb0f397af24cc118c8942c83a852441be6cd/test/integ/test_async.py#L88
Unless you don't want the result caching to kick in, in that case we should maybe split up https://github.com/snowflakedb/snowflake-connector-python/blob/2d78cb0f397af24cc118c8942c83a852441be6cd/src/snowflake/connector/connection.py#L1374 into a non-caching implementation and a caching wrapper on it. Or maybe `mock.patch.dict` the async caching dict could also disable modifying the cache 🤣

src/snowflake/connector/connection.py
sfc-gh-mkeller: You're right! :)

src/snowflake/connector/connection.py
sfc-gh-mkeller: > I think using `with` means it does an implicit `shutdown(wait=True)` when exiting. Since there is no post-processing, technically I don't need to wait on the futures explicitly in the main thread. But maybe that is not good for code readability?

I forgot about this, good catch!

src/snowflake/connector/connection.py
```diff
@@ -1541,13 +1548,36 @@ def is_an_error(status: QueryStatus) -> bool:
...
+                self.get_query_status(sfq_id)
+            )
+
+        with ThreadPoolExecutor(
+            max_workers=num_workers, thread_name_prefix="async_query_check_"
+        ) as tpe:
+
+            futures = (tpe.submit(async_query_check_helper, sfqid) for sfqid in queries)
+            for future in as_completed(futures):
+                if not found_unfinished_query and future.result():
```
sfc-gh-stan, line 1577: When we support 3.9+ and switch to using [`cancel_futures`](https://docs.python.org/3.9/library/concurrent.futures.html#concurrent.futures.Executor.shutdown), we should handle [`CancelledError`](https://docs.python.org/3.9/library/concurrent.futures.html#concurrent.futures.CancelledError) here.

sfc-gh-mkeller: Put this as a comment in the code, otherwise we'll not find it!

sfc-gh-mkeller (approved):
:ship: :ship:

src/snowflake/connector/connection.py
sfc-gh-stan: Sorry, this is an outdated comment.
```
def _all_async_queries_finished():
    tpe = ThreadPoolExecutor()
    futures = (tpe.submit(check_query_status, q) for q in queries)
    for f in as_completed(futures):
        if f.result():
            found_running_query = true
            break
    tpe.shutdown(cancel_futures=true)
    return not found_running_query
```
The code should look something like this, we shouldn't need to access cancelled `future.result()`, after the `break`, so there is no need to handle the exception.

PR #1274, opened 2022-10-04
SNOW-534004: Added database and schema to the queries related to temporary stage

### LINKED ISSUE #1034: SNOW-534004: write_pandas does not use schema provided

```
Python 3.10.1 (main, Jan 10 2022, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)]
```

```
Linux-5.15.14-200.fc35.x86_64-x86_64-with-glibc2.34
```

```
asn1crypto==1.4.0
certifi==2021.10.8
cffi==1.15.0
charset-normalizer==2.0.10
cryptography==36.0.1
idna==3.3
numpy==1.22.1
oscrypto==1.2.1
pandas==1.3.5
pycparser==2.21
pycryptodomex==3.12.0
PyJWT==2.3.0
pyOpenSSL==21.0.0
python-dateutil==2.8.2
pytz==2021.3
requests==2.27.1
six==1.16.0
snowflake-connector-python==2.7.3
urllib3==1.26.8
```

4. What did you do?

Use `write_pandas` for data insertion:
```
from snowflake import connector
from snowflake.connector.pandas_tools import write_pandas
import pandas

schema_name = "my_schema"
table_name = "my_table"
data = pandas.DataFrame(data=[
        [1],
    ],
    columns=["my_id"])


with connector.connect(
        user="USERNAME",
        password="PASSWORD",
        account="ACCOUNT",
        warehouse="WAREHOUSE",
        database="DATABASE"
    ) as conn:
    write_pandas(conn=conn, df=data, schema=schema_name, table_name=table_name)
```

 Error message received:
```
090106 (22000): Cannot perform CREATE TEMPSTAGE. This session does not have a current schema. Call 'USE SCHEMA', or use a qualified name.
```

   Data inserted into table

unrelated. 

`write_pandas` uses cursor without schema specified. Either it has to be specified during cursor creation or schema name should be specified in every call (found 3 within that function)

jekaterinakletnaja:

   Fixes #1034 

   Database and schema parameters are not used in the `write_pandas` code that is related to temporary stage - it refers only to the stage’s name. Therefore, in case the connection provided to `write_pandas` has no database and/or schema, execution fails, even if database and/or schema are provided as arguments to the `write_pandas` itself. Thus, indication of database and schema in the queries related to the temporary stage solves the issue.


jekaterinakletnaja:

sfc-gh-stan (changes requested):
Would you please add an integration test in `test/integ/pandas/test_pandas_tool.py` to verify your change 🙂 ?
In order to use the pytest fixtures we have (e.g. `conn_cnx`, `db_parameters`), you need to add a file named `parameters.py` under `test/`, which contains the parameters you pass to `snowflake.connector.Connect` like:
```
#!/usr/bin/env python
CONNECTION_PARAMETERS = {
    'account': 'accountname',
    'user': 'username',
    'password': 'password',
    'schema': 'schemaname',
    'database': 'databasename',
    'protocol': 'https',
    'host': 'hostname',
    'port': 'portnumber',
}
```

sfc-gh-stan:
Hi @jekaterinakletnaja , thank you for looking into the issue and submitting a fix, we really appreciate your contribution 🎉 !  Please address my comments, I will keep monitoring this PR, let's try to merge this change soon.

src/snowflake/connector/pandas_tools.py
```diff
@@ -172,10 +172,22 @@ def write_pandas(
             stage_name = "".join(
                 random.choice(string.ascii_lowercase) for _ in range(5)
             )
+            if quote_identifiers:
```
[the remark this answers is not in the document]
jekaterinakletnaja: That makes a lot of sense, it was refactored as suggested [here](https://github.com/snowflakedb/snowflake-connector-python/pull/1274/commits/f4e5bfa9e543d04cd8cbcd5ea543c3687854dd1c).

src/snowflake/connector/pandas_tools.py
```diff
@@ -172,10 +172,22 @@ def write_pandas(
...
+                    (database + "." if database else "")
+                    + (schema + "." if schema else "")
+                    + stage_name
+                )
             create_stage_sql = (
                 "create temporary stage /* Python:snowflake.connector.pandas_tools.write_pandas() */ "
-                '"{stage_name}"'
-            ).format(stage_name=stage_name)
+                '{stage_location}'
+            ).format(stage_location=stage_location)
```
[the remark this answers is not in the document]
jekaterinakletnaja: All usages of `format` except one were replaced [here](https://github.com/snowflakedb/snowflake-connector-python/pull/1274/commits/f4e5bfa9e543d04cd8cbcd5ea543c3687854dd1c).
The one that was left as is uses backslashes in it which are not allowed in f-strings.

src/snowflake/connector/pandas_tools.py
```diff
@@ -238,7 +250,7 @@ def write_pandas(
                 if pe.msg.endswith("already exists."):
                     continue
                 raise
-        infer_schema_sql = f"SELECT COLUMN_NAME, TYPE FROM table(infer_schema(location=>'@\"{stage_name}\"', file_format=>'{file_format_name}'))"
+        infer_schema_sql = f"SELECT COLUMN_NAME, TYPE FROM table(infer_schema(location=>'@{stage_location}', file_format=>'{file_format_name}'))"
```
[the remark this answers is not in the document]
jekaterinakletnaja: That was reflected [here](https://github.com/snowflakedb/snowflake-connector-python/pull/1274/commits/f4e5bfa9e543d04cd8cbcd5ea543c3687854dd1c) as well.

jekaterinakletnaja:
> Hi @jekaterinakletnaja , thank you for looking into the issue and submitting a fix, we really appreciate your contribution 🎉 ! Please address my comments, I will keep monitoring this PR, let's try to merge this change soon.

Hi @sfc-gh-stan, appreciate your comments and suggestions! I reflected those in the latest commit and added tests.

At the same time, while adding tests and running previous ones related to `write_pandas`, I’ve run into a few issues that potentially are not related to the changes in this PR, but might be different bugs (however, due to the lack of deep knowledge about the code I might be wrong). Could you please take a look at my comments below and suggest how to approach them the best?

1. The test `test_empty_dataframe_write_pandas` in `snowflake-connector-python/test/integ/pandas/test_pandas_tools.py` is failing with
```
...
    def chunk_helper(lst: T, n: int) -> Iterator[tuple[int, T]]:
        """Helper generator to chunk a sequence efficiently with current index like if enumerate was called on sequence."""
>       for i in range(0, len(lst), n):
E       ValueError: range() arg 3 must not be zero
```
Indeed, in case a `df` is empty, `chunk_size` is 0 according to it’s [current definition](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L167), but 0 is not accepted as `step` for `range`.

2. There are new tests (added by me in order to test creation and usage of `stage_location` and `file_format_location`, for example `test_stage_location_building_db_schema`). They are failing at [this line](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L250) with message
```
.0 = <map object at 0x7f9d855177d0>

>       [f"{quote}{c}{quote} {column_type_mapping[c]}" for c in df.columns]
    )
E   KeyError: 'name'
```
even in case when all assertion related code (for ex., lines 428-443 for `test_stage_location_building_db_schema`) is commented.
It is because `column_type_mapping` (which is the result of execution of [`infer_schema_sql`](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L241)) is an empty list, but we are [searching for columns names in it](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L250). Do you have a feeling what might be wrong here? Is it expected that `column_type_mapping` is an empty list?

test/integ/pandas/test_pandas_tools.py
```diff
@@ -315,7 +315,7 @@ def test_empty_dataframe_write_pandas(
 
 
 @pytest.mark.parametrize("quote_identifiers", [True, False])
-def test_location_building_db_schema(conn_cnx, quote_identifiers: bool):
+def test_table_location_building_db_schema(conn_cnx, quote_identifiers: bool):
```
sfc-gh-stan, line 318: Can we combine the three cases (`db_schema`, `schema`, neither specified) into one using `parametrize`? Maybe something like this:
```
@pytest.mark.parametrize("database,schema,quote_identifiers,expected_location", [
    ("database", "schema", True, "\"database\".\"schema\".\"table\""),
    ("database", "schema", False, "database.schema.table"),
    (None, "schema", True, "\"schema\".\"table\""),
    (None, "schema", False, "schema.table"),
    (None, None, True, "\"table\""),
    (None, None, False, "table")
])
def test_table_location_building(conn_cnx, database:str|None, schema:str|None, quote_identifiers: bool, expected_location:str):
    """This tests that write_pandas constructs location correctly with database, schema and table name."""
    from snowflake.connector.cursor import SnowflakeCursor

    with conn_cnx() as cnx:
        def mocked_execute(*args, **kwargs):
            if len(args) >= 1 and args[0].startswith("COPY INTO"):
                location = args[0].split(" ")[2]
                assert location == expected_location
            cur = SnowflakeCursor(cnx)
            cur._result = iter([])
            return cur

        with mock.patch(
            "snowflake.connector.cursor.SnowflakeCursor.execute",
            side_effect=mocked_execute,
        ) as m_execute:
            success, nchunks, nrows, _ = write_pandas(
                cnx,
                sf_connector_version_df.get(),
                "table",
                database=database,
                schema=schema,
                quote_identifiers=quote_identifiers,
            )
            assert m_execute.called and any(
                map(lambda e: "COPY INTO" in str(e[0]), m_execute.call_args_list))
``` 
What do you think?

sfc-gh-stan: Same applies for `stage_location` and `file_format_location` tests.

test/integ/pandas/test_pandas_tools.py
```diff
@@ -416,6 +416,297 @@ def mocked_execute(*args, **kwargs):
...
+def test_stage_location_building_db_schema(conn_cnx, quote_identifiers: bool, auto_create_table: bool):
+    """This tests that write_pandas constructs stage location correctly with database and schema."""
+    from snowflake.connector.cursor import SnowflakeCursor
+
+    with conn_cnx() as cnx:
+
+        def mocked_execute(*args, **kwargs):
+            if len(args) >= 1 and args[0].lower().startswith(('create temporary stage', 'put', 'copy into', 'select')):
+                stage_location = None
+                if args[0].lower().startswith('create temporary stage'):
```
sfc-gh-stan, line 430: Thank you for the detailed test 👍 It's also acceptable to only test the case for `COPY INTO`, which is what the `table_location` test does.

sfc-gh-stan (changes requested):
Thank you for addressing my suggestions, the code change looks good to go! 
I made a few more comments about the tests, mainly concerned with code duplication/simplicity. Please let me know what you think 🙂 .

sfc-gh-stan:
> > Hi @jekaterinakletnaja , thank you for looking into the issue and submitting a fix, we really appreciate your contribution 🎉 ! Please address my comments, I will keep monitoring this PR, let's try to merge this change soon.
> 
> Hi @sfc-gh-stan, appreciate your comments and suggestions! I reflected those in the latest commit and added tests.
> 
> At the same time, while adding tests and running previous ones related to `write_pandas`, I’ve run into a few issues that potentially are not related to the changes in this PR, but might be different bugs (however, due to the lack of deep knowledge about the code I might be wrong). Could you please take a look at my comments below and suggest how to approach them the best?
> 
> 1. The existing test `test_empty_dataframe_write_pandas` in `snowflake-connector-python/test/integ/pandas/test_pandas_tools.py` is failing with
> 
> ```
> ...
>     def chunk_helper(lst: T, n: int) -> Iterator[tuple[int, T]]:
>         """Helper generator to chunk a sequence efficiently with current index like if enumerate was called on sequence."""
> >       for i in range(0, len(lst), n):
> E       ValueError: range() arg 3 must not be zero
> ```
> 
> Indeed, in case a `df` is empty, `chunk_size` is 0 according to it’s [current definition](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L167), but 0 is not accepted as `step` for `range`.
> 
> 2. ~There are new tests (added by me in order to test creation and usage of `stage_location` and `file_format_location`, for example `test_stage_location_building_db_schema`). They are failing at [this line](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L250) with message~
> 
> ```
> .0 = <map object at 0x7f9d855177d0>
> 
> >       [f"{quote}{c}{quote} {column_type_mapping[c]}" for c in df.columns]
>     )
> E   KeyError: 'name'
> ```
> 
> ~even in case when all assertion related code (for ex., lines 428-443 for `test_stage_location_building_db_schema`) is commented. It is because `column_type_mapping` (which is the result of execution of [`infer_schema_sql`](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L241)) is an empty list, but we are [searching for columns names in it](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/pandas_tools.py#L250). Do you have a feeling what might be wrong here? Is it expected that `column_type_mapping` is an empty list?~
> 
> UPD: I’ve got what was the problem in p.2, my bad. It is solved now by [the latest commit](https://github.com/snowflakedb/snowflake-connector-python/pull/1274/commits/8ed7378ec1798258ce357133eabd1e6bcfc1fba7). Question in p.1 is still actual though.

1. This was actually a bug we recently fixed with #1261. You should be able to resolve the test failure by pulling the latest code. It shouldn't be affected by your changes.

test/integ/pandas/test_pandas_tools.py
sfc-gh-stan: I think it might actually be better if we only test one query, both for simplicity and in the case if we need to tweak some of the queries, we can minimize the change to these tests.

test/integ/pandas/test_pandas_tools.py
jekaterinakletnaja: Like this idea, makes it much more clear and easier to understand. Did it [here](https://github.com/snowflakedb/snowflake-connector-python/pull/1274/commits/331b999014708427e25e8d11a3ce0aea5642cb58) for 3 of them.

test/integ/pandas/test_pandas_tools.py
jekaterinakletnaja: It was refactored to test only creation query for all 3 cases (table, stage, file format) [here](https://github.com/snowflakedb/snowflake-connector-python/pull/1274/commits/331b999014708427e25e8d11a3ce0aea5642cb58).

sfc-gh-stan (approved):

jekaterinakletnaja:
> Hi @jekaterinakletnaja , your code LGTM and has passed all the merge gates 🎉 ! Normally this would be merged right away, but since the bug fix is technically a behavior change, it can only be merged into to `main` in the Python connector's next release in January 2023. I will keep monitoring and make sure that happens. Again, thank you so much for this excellent contribution! We would like to give you credit in the next release note if you don't mind 😉 ?

Hi @sfc-gh-stan, thanks for the wonderful news! 🎉  
It’s an honour to be in the release notes 😄

Looking forward to January 2023!

PR #1275, opened 2022-10-10
Return empty Arrow table instead of None

JakobGM:
I have created no issue because the PR itself should be descriptive enough, the discussion can perhaps be taken here instead?

In the same way that `SnowflakeCursor.fetch_pandas_all()` now returns an empty DataFrame with the correct schema when the query produces 0 rows, `SnowflakeCursor.fetch_arrow_all()` should do the same.


JakobGM:

PR #1281, opened 2022-10-14
SNOW-199087: Implement multi-statement support for the Python connector.

sfc-gh-heshah:
SNOW-676491: Executing multiple queries with no complications.

SNOW-676495: Execute multiple queries with parameter bindings.

   Fixes [SNOW-199087](https://snowflakecomputing.atlassian.net/browse/SNOW-199087)

   When submitting a multi-statement query, GS returns a payload in which `resultIds` corresponding to each of the individual queries are provided. Even if the parameter (account, session, or statement level) `MULTI_STATEMENT_COUNT` is not 1 (other options allow multi-statement results), GS will still return the regular payload without `resultIds`, allowing the connector to determine whether the query submitted was multi-statement or not after inspecting the payload. I added several data members to the `cursor` class allowing the storage and retrieval of submitted query IDs. The `exec_async` functionality will also work the same way as it does for single statements (users are expected to call `get_results_by_sfqid` when they want to retrieve results of an asynchronous multi-statement query. Using the stored query IDs and the `query_result` method, the connector can set the cursor to achieve the same ready state for calling `fetch*()` after calling `nextset()` to iterate through the result sets of each query in the submitted multi-statement.


test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            cur.execute_async(
+                "select 1; select 2; select count(*) from table(generator(timeLimit => 5)); select 'b';",
+                _num_statements=4,
+            )
+            q_id = cur.sfqid
+            status = con.get_query_status(q_id)
+            assert con.is_still_running(status)
+    time.sleep(5)
```
sfc-gh-mkeller, line 136: Please don't add such a long sleep into automated tests, make it shorter and do a while loop to wait

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            for _ in range(25):
```
sfc-gh-mkeller, line 139: 25 is too long
```suggestion
            for _ in range(3):
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+                if status == QueryStatus.SUCCESS:
+                    break
+                time.sleep(1)
+            else:
+                pytest.fail(
+                    f"We should have broke out of this loop, final query status: {status}"
+                )
+
+            status = con.get_query_status_throw_if_error(q_id)
+            assert status == QueryStatus.SUCCESS
```
sfc-gh-mkeller, line 150:
```suggestion
            assert con.get_query_status_throw_if_error(q_id) == QueryStatus.SUCCESS
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+                assert not cur.fetchall()
+
+
+@pytest.mark.skipolddriver
+def test_async_exec_multi(conn_cnx):
+    """Tests whether async execution query works within a multi-statement"""
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            cur.execute_async(
+                "select 1; select 2; select count(*) from table(generator(timeLimit => 5)); select 'b';",
```
sfc-gh-mkeller, line 130:
```suggestion
                "select 1; select 2; select count(*) from table(generator(timeLimit => 2)); select 'b';",
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+def test_async_error_multi(conn_cnx):
+    """
+    Runs a query that will fail to execute and then tests that if we tried to get results for the query
+    then that would raise an exception. It also tests QueryStatus related functionality too.
+    """
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            sql = "select 1; select * from nonexistentTable"
+            cur.execute_async(sql)
+            q_id = cur.sfqid
```
sfc-gh-mkeller, line 174:
```suggestion
            q_id = cur.execute_async(sql).sfqid
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+            with pytest.raises(ProgrammingError) as sync_error:
+                cur.execute(sql)
+            while con.is_still_running(con.get_query_status(q_id)):
+                time.sleep(1)
+            status = con.get_query_status(q_id)
+            assert status == QueryStatus.FAILED_WITH_ERROR
```
sfc-gh-mkeller, line 180:
```suggestion
            assert con.get_query_status(q_id) == QueryStatus.FAILED_WITH_ERROR
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+            assert e1.value.errno == e2.value.errno == sync_error.value.errno
+
+
+@pytest.mark.skipolddriver
+def test_mix_sync_async_multi(conn_cnx):
+    """Tests sending multiple multi-statement async queries at the same time."""
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            # Setup
+            cur.execute("alter session set MULTI_STATEMENT_COUNT = 0")
```
sfc-gh-mkeller, line 196: Do the params trick I mentioned before to save 2 round trips to Snowflake

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+                assert cur.fetchall() == [(2,)]
+                assert cur.nextset() == cur
+                assert cur.fetchall() == [("b",)]
+                assert cur.nextset() == cur
+                assert cur.fetchall() == [("row1", 1), ("row2", 2), ("row3", 3)]
+                assert cur.nextset() is None
+                assert not cur.fetchall()
+            finally:
+                for table in ["smallTable", "uselessTable"]:
+                    cur.execute(f"drop table if exists {table}")
```
sfc-gh-mkeller, line 238: You can get rid of this by making these tables temporary https://docs.snowflake.com/en/user-guide/tables-temp-transient.html#temporary-tables

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+@pytest.mark.skipolddriver
+def test_done_caching_multi(conn_cnx):
+    """Tests whether get status caching is working as expected."""
+    with conn_cnx() as con:
+        with con.cursor() as cur:
+            cur.execute("alter session set MULTI_STATEMENT_COUNT=0;")
```
sfc-gh-mkeller, line 246: Do dictionary trick here too

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,268 @@
...
+            cur.execute_async(
+                "select 1; select 'a'; select count(*) from table(generator(timeLimit => 5));"
+            )
+            qid1 = cur.sfqid
+            cur.execute_async(
+                "select 2; select 'b'; select count(*) from table(generator(timeLimit => 10));"
```
sfc-gh-mkeller, line 252: too long of a sleep

src/snowflake/connector/cursor.py
```diff
@@ -1273,6 +1311,18 @@ def wait_until_ready():
                     )
                 )
             self._inner_cursor.execute(f"select * from table(result_scan('{sfqid}'))")
+
+            if (
+                self._inner_cursor._total_rowcount == 1
+                and self._inner_cursor.fetchall()
```
sfc-gh-mkeller, line 1317: `fetchall` does affect its cursor's results, so if `[("Multiple statements executed successfully.",)]` isn't what `fetchall` returns then you've affected what the cursor will return

sfc-gh-mkeller (changes requested):

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,385 @@
...
+            cur.execute_async(
+                "select 1; select 'a'; select count(*) from table(generator(timeLimit => 2));"
+            )
+            qid1 = cur.sfqid
+            cur.execute_async(
+                "select 2; select 'b'; select count(*) from table(generator(timeLimit => 2));"
+            )
+            qid2 = cur.sfqid
+            assert len(con._async_sfqids) == 2
+            time.sleep(5)
```
sfc-gh-mkeller, line 261: Way too long of a sleep
```suggestion
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,385 @@
...
+            while con.is_still_running(con.get_query_status(qid1)):
+                time.sleep(1)
+            assert con.get_query_status(qid1) == QueryStatus.SUCCESS
+            assert len(con._async_sfqids) == 1
+            assert len(con._done_async_sfqids) == 1
+            time.sleep(5)
```
sfc-gh-mkeller, line 267: This sleep is way too long
```suggestion
```

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,385 @@
...
+            assert cur.fetchall() == [(2,)]
+            assert cur.nextset() == cur
+            assert cur.fetchall() == [("b",)]
+            assert cur.nextset() == cur
+            assert cur.fetchall() == [("row1", 1), ("row2", 2), ("row3", 3)]
+            assert cur.nextset() is None
+            assert not cur.fetchall()
+
+
+def test_done_caching_multi(conn_cnx):
```
sfc-gh-mkeller, line 248: I challenge you on this test. I think this is the same as https://github.com/snowflakedb/snowflake-connector-python/blob/88843f88a5b7de84bbca3587da86c702155fd96c/test/integ/test_async.py#L166 but you send in multi-statements which I think is not testing anything new.

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,385 @@
...
+    """Tests whether multiple describe queries are rewritten properly and that this is logged."""
+    with conn_cnx(session_parameters={PARAMETER_MULTI_STATEMENT_COUNT: 0}) as con:
+        with con.cursor() as cur:
+            table1 = random_string(5, "test_desc_rewrite_multi_")
+            table2 = random_string(5, "test_desc_rewrite_multi_")
+            cur.execute(
+                f"create temp table {table1} (a int); create temp table {table2} (b int);"
+            )
+            caplog.set_level(logging.DEBUG, "snowflake.connector")
+            cur.execute(f"desc {table1}; select 3; desc {table2}; select 'b';")
```
sfc-gh-mkeller, line 286: Once we update the regex I'd like to see `[ TYPE =  { COLUMNS | STAGE } ]` being tested too from https://docs.snowflake.com/en/sql-reference/sql/desc-table.html

sfc-gh-mkeller (changes requested):

src/snowflake/connector/cursor.py
```diff
@@ -464,9 +468,7 @@ def _execute_helper(
             self._is_file_transfer = _is_put_get
         else:
             # or detect it.
-            self._is_file_transfer = self.PUT_SQL_RE.match(
-                query
-            ) or self.GET_SQL_RE.match(query)
+            self._is_file_transfer = bool(self.get_file_transfer_type(query))
```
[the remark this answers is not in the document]
sfc-gh-heshah: The server side response is different for a PUT/GET command, but I don't see much value in modifying these regexes to detect PUT/GET commands in multi-statement since they are not supported for multi-statements anyways. I think we should make very clear in the docs that PUT/GET commands are not supported for multi-statements, and leave the existing code as is. This would exactly match the implementation in JDBC.

src/snowflake/connector/cursor.py
```diff
@@ -733,13 +751,17 @@ def execute(
...
             )
-        logger.debug("sfqid: %s", self.sfqid)
 
-        logger.info("query execution done")
         if ret["success"]:
             logger.debug("SUCCESS")
             data = ret["data"]
 
+            if "resultIds" in data:
+                self._init_multi_statement_results(data)
```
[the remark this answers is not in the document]
sfc-gh-heshah: I think this is going to be more involved than moving the code up unfortunately. We are going to need to make sure the regex works for multiple alter commands in a multi-statement query, then parse through all of the groups. I also noticed we probably want to run the following code for async multi-statements as well.
```
if _exec_async:
    self.connection._async_sfqids[self._sfqid] = None
if _no_results:
    self._total_rowcount = (
        ret["data"]["total"]
        if "data" in ret and "total" in ret["data"]
        else -1
    )
    return data
```

Maybe it would be best to detect PUT/GET commands in the returned payload after all. Then we don't need to rely on `self.is_file_transfer` just before these two code blocks.

test/integ/test_multi_statement.py
```diff
@@ -0,0 +1,385 @@
...
+            for _ in range(3):
+                status = con.get_query_status(q_id)
+                if status == QueryStatus.SUCCESS:
+                    break
+                time.sleep(1)
+            else:
+                pytest.fail(
+                    f"We should have broke out of this loop, final query status: {status}"
+                )
+
```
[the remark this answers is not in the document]
sfc-gh-heshah: I think there's some delay between a query no longer running and getting a `SUCCESS` status. The test doesn't seem to pass without this additional wait.

test/integ/test_multi_statement.py
sfc-gh-heshah: It's quite similar, I used it as inspiration because I wanted to be sure the retrieval of the individual query IDs in the multi-statement wasn't messing with the async QID caching in the connector. However, I realize I'm not even pulling in the multi-statement query IDs here since it's execute_async, so I'll add that.

src/snowflake/connector/cursor.py
```diff
@@ -1154,8 +1229,20 @@ def fetchall(self) -> list[tuple] | list[dict]:
...
+        Fetches the next set of results if the previously executed query was multi-statement so that subsequent calls
+        to any of the fetch*() methods will return rows from the next query's set of results. Returns None if no more
+        query results are available.
+        """
+        self.reset()
+        if self._multi_statement_resultIds:
+            self.query_result(self._multi_statement_resultIds[0])
+            logger.info(
+                "Retrieved results for query ID: %s",
+                self._multi_statement_resultIds.popleft(),
```
sfc-gh-mkeller, line 1242:
```suggestion
                f"Retrieved results for query ID: {self._multi_statement_resultIds.popleft()}",
```

src/snowflake/connector/test_util.py
```diff
@@ -29,3 +35,37 @@
...
+    sleep_per_check: int,
+):
+    for _ in range(num_checks):
+        status = con.get_query_status(sfqid)
+        if status == QueryStatus.SUCCESS:
+            break
+        time.sleep(sleep_per_check)
+    else:
+        pytest.fail(
+            f"We should have broke out of wait loop for query success."
```
sfc-gh-mkeller, line 68:
```suggestion
            "We should have broke out of wait loop for query success."
```

src/snowflake/connector/test_util.py
```diff
@@ -29,3 +35,37 @@
...
     )
     rt_plain_logger.addHandler(ch)
+
+
+def _wait_while_query_running(
+    con: snowflake.connector.connection,
+    sfqid: str,
+    sleep_time: int,
+    dont_cache: bool = False,
+):
```
sfc-gh-mkeller, line 45:
```suggestion
) -> None:
```

src/snowflake/connector/test_util.py
```diff
@@ -29,3 +35,37 @@
...
+    while con.is_still_running(query_status(sfqid)):
+        time.sleep(sleep_time)
+
+
+def _wait_until_query_success(
+    con: snowflake.connector.connection,
+    sfqid: str,
+    num_checks: int,
+    sleep_per_check: int,
+):
```
sfc-gh-mkeller, line 60:
```suggestion
) -> None:
```

src/snowflake/connector/test_util.py
```diff
@@ -29,3 +35,37 @@
...
+    Checks if the provided still returns that it is still running, and if so,
+    sleeps for the specified time in a while loop.
+    """
+    query_status = con._get_query_status if dont_cache else con.get_query_status
+    while con.is_still_running(query_status(sfqid)):
+        time.sleep(sleep_time)
+
+
+def _wait_until_query_success(
+    con: snowflake.connector.connection,
```
sfc-gh-mkeller, line 56:
```suggestion
    con: snowflake.connector.connection.SnowflakeConnection,
```

src/snowflake/connector/test_util.py
```diff
@@ -29,3 +35,37 @@
         )
     )
     rt_plain_logger.addHandler(ch)
+
+
+def _wait_while_query_running(
+    con: snowflake.connector.connection,
```
sfc-gh-mkeller, line 41:
```suggestion
    con: snowflake.connector.connection.SnowflakeConnection,
```

test/integ/test_cursor.py
```diff
@@ -1116,6 +1119,7 @@ def test_execute_helper_params_error(conn_testaccount):
             cur._execute_helper("select %()s", statement_params="1")
 
 
+@pytest.mark.skipolddriver
```
sfc-gh-mkeller, line 1122: Are you sure this is necessary?

sfc-gh-heshah: Shouldn't be necessary anymore actually

sfc-gh-mkeller (approved):
:ship: :ship:

DESCRIPTION.md
```diff
@@ -8,11 +8,22 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
...
+    - The Snowflake parameter `MULTI_STATEMENT_COUNT` can be altered at the account, session, or statement level. An additional argument, `num_statements`, can be provided to `execute` to use this parameter at the statement level. It *must* be provided to `executemany` to submit a multi-statement query through the method. Note that bulk insert optimizations available through `executemany` are not available when submitting multi-statement queries.
+      - By default the parameter is 1, meaning only a single query can be submitted at a time
+      - Set to 0 to submit any number of statements in a multi-statement query
+      - Set to >1 to submit the specified exact number of statements in a multi-statement query
+    - Bindings are accepted in the same way for multi-statements as they are for single statement queries
+    - Asynchronous multi-statement query execution is supported. Users should still use `get_results_from_sfqid` to retrieve results
+    - To access the results of each query, users can call `cursor.nextset()` as specified in the DB 2.0 API (PEP-249), to iterate through each statements results
```
sfc-gh-mkeller, line 22:
```suggestion
    - To access the results of each query, users can call `SnowflakeCursor.nextset()` as specified in the DB 2.0 API (PEP-249), to iterate through each statements results
```

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1291, opened 2022-10-21
SNOW-669650: Fix `write_pandas` atomicity when `overwrite=True`

### LINKED ISSUE #1262: SNOW-669650: write_pandas is not atomic

   Python 3.9.2

    macOS-10.15.4-x86_64-i386-64bit

   snowflake-connector-python==2.8.0
   snowflake-sqlalchemy==1.4.1
   pandas==1.4.4
   pyarrow==8.0.0

4. What did you do?
```python
import pandas as pd
import snowflake.connector.pandas_tools as snowflake_pandas
import snowflake

d = {"col12": [1, 2], "col2": [3, 4]}
df = pd.DataFrame(data=d)

    user="XXX",
    password="XXX",
    account="XXX",
    database="XXX",
    schema="XXX",
    warehouse="XXX",
)

conn.cursor().execute("USE ROLE ACCOUNTADMIN;")

snowflake_pandas.write_pandas(
    conn,
    df,
    table_name="testtable",
    database="XXX",
    schema="XXX",
    overwrite=True,
    auto_create_table=True,
    quote_identifiers=False,
)

# Add bad character to column name to make the function fail
d = {"col12%": [1, 2], "col2": [3, 4]} 
df = pd.DataFrame(data=d)

snowflake_pandas.write_pandas(
    conn,
    df,
    table_name="testtable",
    database="XXX",
    schema="XXX",
    overwrite=True,
    auto_create_table=True,
    quote_identifiers=False,
)
```
   
I expected that write_pandas is atomic. When my write_pandas call fails, I should still have the previous version of my table in place. I should only see it change if everything else is successful.

sfc-gh-jdu:

   Fixes SNOW-669650 #1262 

   The original approach dropped/truncated the original table directly even if the underlying copy into query fails, which is not atomic. Now in this PR we create a table with random name first if `overwrite=True`. Then if the query succeeds, we drop the original table and rename this table back to the original table. 
   
   I refactored some code in `write_pandas` to make my change simpler, which are necessary in my opinion. I listed them in the comment


src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
+        copy_results = cursor.execute(copy_into_sql, _is_internal=True).fetchall()
+
+        if overwrite:
+            original_table_name = f"{location}{quote}{table_name}{quote}"
+            drop_object(original_table_name, "table")
+            rename_table_sql = f"ALTER TABLE IF EXISTS {target_table_name} RENAME TO {original_table_name} /* Python:snowflake.connector.pandas_tools.write_pandas() */"
+            logger.debug(f"rename table with '{rename_table_sql}'")
+            cursor.execute(rename_table_sql, _is_internal=True)
+    finally:
+        drop_object(stage_name, "stage")
```
sfc-gh-jdu, line 268: previously this stage is not dropped.

src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
+        )
+        logger.debug(f"copying into with '{copy_into_sql}'")
+        copy_results = cursor.execute(copy_into_sql, _is_internal=True).fetchall()
+
+        if overwrite:
+            original_table_name = f"{location}{quote}{table_name}{quote}"
+            drop_object(original_table_name, "table")
+            rename_table_sql = f"ALTER TABLE IF EXISTS {target_table_name} RENAME TO {original_table_name} /* Python:snowflake.connector.pandas_tools.write_pandas() */"
+            logger.debug(f"rename table with '{rename_table_sql}'")
+            cursor.execute(rename_table_sql, _is_internal=True)
```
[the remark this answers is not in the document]
sfc-gh-sfan: +1. I wonder if we could do that as well. It's possible that we cannot because ddl probably will commit the transaction.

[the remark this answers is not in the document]
sfc-gh-jdu: Yea it's a good idea. But as Shixuan said, it seems ddl cannot be rolled back. https://docs.snowflake.com/en/sql-reference/transactions.html#ddl

src/snowflake/connector/pandas_tools.py
```diff
@@ -152,37 +152,27 @@ def write_pandas(
...
-            (database + "." if database else "")
-            + (schema + "." if schema else "")
-            + (table_name)
+        location = (database + "." if database else "") + (
+            schema + "." if schema else ""
         )
     if chunk_size is None:
         chunk_size = len(df)
+
+    def random_name(length: int = 10) -> str:
```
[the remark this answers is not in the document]
sfc-gh-jdu: It will be hard with my change. Because I need to create a table with random name, and when I use previous, if the generated random name is the same as the target table name, because our SQL to create such a table is `CREATE TABLE IF EXISTS`, the intermediate table with random name will not be created if the table already exists. So I feel this change is needed and it's actually safe since I just remove the while loop and make the name more random.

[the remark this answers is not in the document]
sfc-gh-sfan: I see. It's probably an overkill but we could remove `IF EXISTS` and use a while-loop as well, but I don't think it is strictly necessary if we are confident that there won't be conflict.

sfc-gh-sfan (approved):

src/snowflake/connector/pandas_tools.py
```diff
@@ -152,37 +152,27 @@ def write_pandas(
...
-            if pe.msg.endswith("already exists."):
-                continue
-            raise
+    stage_name = random_name()
+    create_stage_sql = (
+        "create temporary stage /* Python:snowflake.connector.pandas_tools.write_pandas() */ "
+        '"{stage_name}"'
+    ).format(stage_name=stage_name)
+    logger.debug(f"creating stage with '{create_stage_sql}'")
+    cursor.execute(create_stage_sql, _is_internal=True).fetchall()
```
sfc-gh-mkeller, line 175: If the random name exists then this will just raise an exception now

src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
+
+    if auto_create_table or overwrite:
+        file_format_name = random_name()
+        file_format_sql = (
+            f"CREATE TEMP FILE FORMAT {file_format_name} "
+            f"/* Python:snowflake.connector.pandas_tools.write_pandas() */ "
+            f"TYPE=PARQUET COMPRESSION={compression_map[compression]}"
+        )
+        logger.debug(f"creating file format with '{file_format_sql}'")
+        cursor.execute(file_format_sql, _is_internal=True)
```
sfc-gh-mkeller, line 220: Again, if name is in use this will just fail now

src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
+                cursor.execute(infer_schema_sql, _is_internal=True).fetchall()
+            )
+            # Infer schema can return the columns out of order depending on the chunking we do when uploading
+            # so we have to iterate through the dataframe columns to make sure we create the table with its
+            # columns in order
+            create_table_columns = ", ".join(
+                [f"{quote}{c}{quote} {column_type_mapping[c]}" for c in df.columns]
+            )
+        finally:
+            drop_object(file_format_name, "file format")
```
sfc-gh-mkeller, line 235: It's a temp file format, I'd argue that we can save the round-trip time and have it be garbage collected by the server-side

src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
+            f"FILE_FORMAT=(TYPE=PARQUET COMPRESSION={compression_map[compression]}) "
+            f"PURGE=TRUE ON_ERROR={on_error}"
+        )
+        logger.debug(f"copying into with '{copy_into_sql}'")
+        copy_results = cursor.execute(copy_into_sql, _is_internal=True).fetchall()
+
+        if overwrite:
+            original_table_name = f"{location}{quote}{table_name}{quote}"
+            drop_object(original_table_name, "table")
+            rename_table_sql = f"ALTER TABLE IF EXISTS {target_table_name} RENAME TO {original_table_name} /* Python:snowflake.connector.pandas_tools.write_pandas() */"
```
sfc-gh-mkeller, line 264: This table needs to exists, let's remove the error catching (we should know if this ever happened)
```suggestion
            rename_table_sql = f"ALTER TABLE {target_table_name} RENAME TO {original_table_name} /* Python:snowflake.connector.pandas_tools.write_pandas() */"
```

src/snowflake/connector/pandas_tools.py
sfc-gh-mkeller: It's a temporary stage, so it only shows up in the current session and it will be deleted when the session ends.
I think we don't need to drop it!

src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
+            logger.debug(f"rename table with '{rename_table_sql}'")
+            cursor.execute(rename_table_sql, _is_internal=True)
+    finally:
+        drop_object(stage_name, "stage")
+        if overwrite:
+            drop_object(target_table_name, "table")
```
sfc-gh-mkeller, line 270: This should only be executed if an error was encountered

sfc-gh-mkeller (changes requested):

src/snowflake/connector/pandas_tools.py
```diff
@@ -202,90 +192,85 @@ def write_pandas(
...
-        # Infer schema can return the columns out of order depending on the chunking we do when uploading
-        # so we have to iterate through the dataframe columns to make sure we create the table with its
-        # columns in order
-        quote = '"' if quote_identifiers else ""
-        create_table_columns = ", ".join(
-            [f"{quote}{c}{quote} {column_type_mapping[c]}" for c in df.columns]
+        try:
+            column_type_mapping = dict(
+                cursor.execute(infer_schema_sql, _is_internal=True).fetchall()
+            )
```
sfc-gh-mkeller, line 227: This is a very neat trick! 😄

src/snowflake/connector/pandas_tools.py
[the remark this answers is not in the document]
sfc-gh-jdu: done, I moved this method to util_text.py

src/snowflake/connector/pandas_tools.py
sfc-gh-jdu: the random name should never conflict because the probability of conflict is (1/26)^10 ~1e-14

src/snowflake/connector/pandas_tools.py
sfc-gh-jdu: done

src/snowflake/connector/pandas_tools.py
sfc-gh-jdu: good catch, fixed

sfc-gh-mkeller (approved):
I think we should keep the while loop retries for the astronomically unlikely case that they occur. However; I'm not against adding them in a separate Jira and PR!

Great job on this @sfc-gh-jdu 

:ship: :ship:

PR #1327, opened 2022-11-14
SNOW-693548 fixing redos attack

sfc-gh-mkeller:

   Fixes SNOW-693548

   This fixes the CVE https://nvd.nist.gov/vuln/detail/CVE-2022-42965
   The issue comes from the greediness of the original regex. With this reworked function we have roughly linear performance. This is achieved by starting from the beginning to lazily strip away comments one-by-one instead of matching the whole SQL command with one regex. The new SQL is also better by matching more than multi-line comments.

### DIFF
--- a/src/snowflake/connector/_sql_util.py
+++ b/src/snowflake/connector/_sql_util.py
@@ -1,0 +1,52 @@
+#
+# Copyright (c) 2012-2021 Snowflake Computing Inc. All rights reserved.
+#
+
+from __future__ import annotations
+
+import re
+
+from .constants import FileTransferType
+
+COMMENT_START_SQL_RE = re.compile(
+    r"""
+                                  ^\s*(?:
+                                      /\*[\w\W]*?\*/
+                                  )""",
+    re.VERBOSE,
+)
+
+PUT_SQL_RE = re.compile(r"^\s*put", flags=re.IGNORECASE)
+GET_SQL_RE = re.compile(r"^\s*get", flags=re.IGNORECASE)
+
+
+def remove_starting_comments(sql: str) -> str:
+    """Remove all comments from the start of a SQL statement."""
+    commentless_sql = sql
+    while True:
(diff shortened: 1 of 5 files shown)


sfc-gh-mkeller:
So as it turns out the new Regex is now able to find more comments and this has broken [this old test](https://github.com/snowflakedb/snowflake-connector-python/blob/aa6c88f5a0f72feb3a2fa0329b6b253bb52e2394/test/integ/test_put_get_medium.py#L761).
Which makes this definitely a behavior change 😭 I've modified the old test in [2faa971](https://github.com/snowflakedb/snowflake-connector-python/pull/1327/commits/2faa9711272c4ba2a2834663405202e93ff907d3).

src/snowflake/connector/_sql_util.py
```diff
@@ -0,0 +1,52 @@
...
+
+import re
+
+from .constants import FileTransferType
+
+COMMENT_START_SQL_RE = re.compile(
+    r"""
+                                  ^\s*(?:
+                                      /\*[\w\W]*?\*/
+                                  )""",
```
sfc-gh-sfan, line 15: Why do we need these new lines and spaces in this regex?

sfc-gh-mkeller: It's easier to read. Here's the doc on [re.VERBOSE](https://docs.python.org/3/library/re.html#re.VERBOSE)

sfc-gh-jdu (approved):

sfc-gh-sfan (approved):

sfc-gh-ssarangpur (commented):
LGTM

sfc-gh-hpathak:
@sfc-gh-mkeller  Looks good overall for the CVE. However, I wonder if there is a potential to exploit this again via some other means. We may want to think about making this robust at a design level.

sfc-gh-aling (approved):

PR #1341, opened 2022-11-22
SNOW-679732: loosen cryptography restriction

sfc-gh-aling:

   Fixes #SNOW-679732

	cryptography has its own [API stability policy](https://cryptography.io/en/stable/api-stability/#deprecation) which is different from the semantic versioning, however, we have customers who like to catch up with the latest cryptography release for security fix, avoid dependency conflict etc.
	
	As a compromise, we decide to loosen the upper bound of cryptography pin to +3 version as per their API stability policy.
	
	The major reasoning for pinning the version is that customer will pin the connector version in the deployment and we want the connector to work with compatible cryptography lib if the environment is going to recover from the pinned version.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -8,6 +8,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 
+- v2.8.3(Unreleased)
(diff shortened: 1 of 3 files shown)


sfc-gh-mkeller (approved):
Please add changelog before merging, but otherwise looks good!

:ship: :ship:

PR #1379, opened 2022-12-16
Receive ARRAY, VARIANT, OBJECT datatypes in Arrow format

sfc-gh-heshah:

     [SNOW-703039](https://snowflakecomputing.atlassian.net/browse/SNOW-703039)

    Extending the ArrowIterator cpp files with a new VariantConverter object that relies on the previously created SnowflakeJSONDecoder class to handle undefined values.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,6 +121,7 @@ if _ABLE_TO_COMPILE_EXTENSIONS:
                     os.path.join(ARROW_ITERATOR_SRC_DIR, "StringConverter.cpp"),
                     os.path.join(ARROW_ITERATOR_SRC_DIR, "TimeConverter.cpp"),
                     os.path.join(ARROW_ITERATOR_SRC_DIR, "TimeStampConverter.cpp"),
+                    os.path.join(ARROW_ITERATOR_SRC_DIR, "VariantConverter.cpp"),
                     os.path.join(ARROW_ITERATOR_SRC_DIR, "Python", "Common.cpp"),
                     os.path.join(ARROW_ITERATOR_SRC_DIR, "Python", "Helpers.cpp"),
                     os.path.join(ARROW_ITERATOR_SRC_DIR, "Util", "time.cpp"),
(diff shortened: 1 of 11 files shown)


cpcloud:
@sfc-gh-heshah Any chance this PR is still viable? It would be really great to properly support arrays (including `undefined` values).

cpcloud:
@sfc-gh-heshah Any idea if support for nested data with arrow will ever be added?

malthe:
VARIANT data can also be for example Avro – not just JSON.

The code presented here seems to targetting VARIANT data which is JSON-encoded; or is Snowflake implicitly converting other semi-structured formats to JSON? I wouldn't think so.

cpcloud:
> The code presented here seems to targetting VARIANT data which is JSON-encoded; or is Snowflake implicitly converting other semi-structured formats to JSON? I wouldn't think so.

I'm not sure I follow. Snowflake encodes things in its internal format and doesn't preserve the source format.

malthe:
@cpcloud in this changeset, I read code such as:
```python
def _VARIANT_to_python(self, _) -> Callable[[str], Any]:
    """Use a custom json decoder (accepts undefined) to load row value"""
    return partial(json.loads, cls=SnowflakeJSONDecoder)
```
Like, we see a VARIANT value, let's decode it using JSON. But is that the only "wire format" we can get? I'm thinking both in terms of performance but also fidelity – JSON doesn't have the same data type richness as the other supported semi-structured formats.

cpcloud:
The fact that the arrow interop code [here](https://github.com/snowflakedb/snowflake-connector-python/pull/1379/files#diff-83292202803defc0126ad4e03e1710d98226bd106de5e2a40d6ad9f73cb6ea45R52) is pulling out a string and then immediately decoding it as JSON suggests that variants are always coming back as JSON.

I'd love to be wrong about this, other wire formats would be most welcome from the ibis perspective.

cpcloud:
I guess there's this option too https://docs.snowflake.com/en/user-guide/python-connector-example#label-python-connector-bypass-data-conversion which means we'd have to implement the entire conversion ourselves.

cpcloud:
Actually that method is broken and just gives back data in Python format.

PR #1405, opened 2023-01-20
SNOW-677788 query context caching for python connector

sfc-gh-aalam:

   Fixes #SNOW-677788

   This PR adds support to store and manage query context using a `QueryContextCache`. This cache stores `queryContext` received in the server response, holds it in a sorted set sorted by `priority`, and sends the cache back to the server during requests. Implementation is based on jdbc implementation: https://snowflakecomputing.atlassian.net/browse/SNOW-677787


setup.cfg
```diff
@@ -58,6 +58,7 @@ install_requires =
     certifi>=2017.4.17
     typing_extensions>=4.3,<5
     filelock>=3.5,<4
+    sortedcontainers
```
sfc-gh-aalam, line 61: Adding dependency for using `SortedSets` for `QueryContextCache`. I have added this requirement in `setup.cfg` and `requirements_3[7,8,9,10].reqs`. Is there any other place I need to update this?

src/snowflake/connector/query_context_cache.py
```diff
@@ -0,0 +1,271 @@
...
+        return self._priority < other.priority
+
+    def __hash__(self) -> int:
+        _hash = 31
+
+        _hash = _hash * 31 + self._id
+        _hash += (_hash * 31) + self._read_timestamp
+        _hash += (_hash * 31) + self._priority
+        if self._context:
+            _hash += (_hash * 31) + int.from_bytes(md5(self._context).digest(), "big")
```
sfc-gh-aalam, line 78: Is there a better way to hash bytearray?

src/snowflake/connector/query_context_cache.py
```diff
@@ -0,0 +1,271 @@
...
+        else:
+            _hash += _hash * 31
+        return _hash
+
+    def __str__(self) -> str:
+        return f"({self._id}, {self._read_timestamp}, {self._priority})"
+
+
+class QueryContextCache:
+    QUERY_CONTEXT_SCHEMA = pa.schema(
```
sfc-gh-sfan, line 88: Today we install `pyarrow` as part of `.pandas`, which means `pyarrow` is not required dependency. 

We probably should have a check on the availability of `pyarrow`, and only enable `query_context_cache` when it is available. Otherwise if `pyarrow` is not installed, query will fail here.

sfc-gh-sfan (changes requested):
I wonder if we could build on top of `cachetools.TTLCache`?

src/snowflake/connector/query_context_cache.py
```diff
@@ -0,0 +1,271 @@
...
+        ]
+    )
+
+    def __init__(self, capacity: int):
+        self._capacity = capacity
+        self._id_map: dict[int, QueryContextElement] = {}
+        self._priority_map: dict[int, QueryContextElement] = {}
+        # stores elements sorted by priority. Element with
+        # least priority value has the highest priority
+        self._treeset: set[QueryContextElement] = SortedSet()
```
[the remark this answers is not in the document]
sfc-gh-aalam: I'm happy to update the code using by using `key` for `SortedSet` but I already have an implementation of `__lt__` and `__eq__` for `QueryContextElement` based on @sfc-gh-igarish's doc: https://docs.google.com/document/d/1OOJE0lYFDkjkGoQFrnH28dyp7Eve9k8faa69Qt1OLtc/edit#heading=h.uj8lyr9wa82r. I believe `__lt__` offers more flexible comparison in this case because it can be modified to have more complex comparisons vs `key`. Do you agree?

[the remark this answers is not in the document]
sfc-gh-sfan: Ah I see. I'm not aware of jdbc's existing implementation. I still think a custom comparator is better (e.g. using `cmp_to_key` to convert a comparator to key), but given JDBC's choice, I don't have a strong opinion.

src/snowflake/connector/query_context_cache.py
sfc-gh-aalam: thanks. this was causing multiple test failures earlier

sfc-gh-sfan (approved):
LGTM

src/snowflake/connector/query_context_cache.py
```diff
@@ -0,0 +1,284 @@
...
+
+from sortedcontainers import SortedSet
+
+from .options import installed_pandas
+from .options import pyarrow as pa
+
+logger = getLogger(__name__)
+
+
+if installed_pandas:
```
sfc-gh-igarish, line 22: why dependency on Pandas?

sfc-gh-igarish (approved):
LGTM. We added disable QCC capability in JDBC. Please add the similar in Python.

src/snowflake/connector/query_context_cache.py
sfc-gh-aalam: `pyarrow` is installed with snowflake connector along with pandas therefore we use this check

PR #1418, opened 2023-01-30
mypy pre-commit hook and lot of type-hints

sfc-gh-mkeller:

   Fixes SNOW-420416

   In this PR I did a few things:
   1. Ran [autotyping](https://github.com/JelleZijlstra/autotyping) to generate some type-hints
   2. Ran a modified version of [PyAnnotate](https://github.com/dropbox/pyannotate) to generate type-hints
   3. Ran `pre-commit autoupdate` and updated some settings that I now know were useless (e.g.: excluding files that were never part of VC)
   4. I removed `snowflake.connector.mixin` as it seems to have never been used. I also checked the [DBAPI2 spec](https://peps.python.org/pep-0249/) and this seems to NOT be part of the spec. However; I'm not against reverting this change.
   5. I added `mypy` to our pre-commit config to be run on some of our files. While I fixed a lot of other typing issues, doing more than this subset will take a while and as we have agreed it'd be better to get started with some verification rather than delay this even longer.


src/snowflake/connector/auth/_auth.py
```diff
@@ -663,7 +668,7 @@ def unlock_temporary_credential_file() -> bool:
         return False
 
 
-def delete_temporary_credential(host, user, cred_type):
+def delete_temporary_credential(host, user, cred_type) -> None:
```
sfc-gh-sfan, line 671: I'm confused, why this has `-> None` but in `read_temporary_credential_file` the `-> None` is removed?

sfc-gh-sfan (commented):
Mostly looks good, just one question

src/snowflake/connector/auth/_auth.py
sfc-gh-mkeller: That function actually returns something while this one doesn't. Instead of fixing the return type of `read_temporary_credential_file` I just removed `-> None` for now.

sfc-gh-sfan: I see. Thanks for the clarification

sfc-gh-sfan (approved):

sfc-gh-jdu (approved):

.pre-commit-config.yaml
```diff
@@ -67,8 +70,39 @@ repos:
...
+                | gzip_decoder
+                | proxy
+                | secret_detector
+                | sfbinaryformat
+                | snow_logging
+                | sqlstate
+                | ssd_internal_keys
+                | test_util
+                | util_text
+                | version
```
sfc-gh-aling, line 96: what's the list here.. is that we only apply mypy check to the listed file? I didn't see cache/ocsp_snowflake etc. being listed here.

src/snowflake/connector/arrow_context.py
```diff
@@ -119,22 +121,26 @@ def TIMESTAMP_LTZ_to_python_windows(
...
+    def FIXED_to_numpy_float64(self, py_long: int, scale: int) -> float64:
         return numpy.float64(decimal.Decimal(py_long).scaleb(-scale))
 
-    def DATE_to_numpy_datetime64(self, py_days):
+    def DATE_to_numpy_datetime64(self, py_days: int) -> datetime64:
         return numpy.datetime64(py_days, "D")
 
-    def TIMESTAMP_NTZ_ONE_FIELD_to_numpy_datetime64(self, value, scale):
+    def TIMESTAMP_NTZ_ONE_FIELD_to_numpy_datetime64(
+        self, value: int, scale: int
```
sfc-gh-aling, line 137: dumb question: should it be float?

src/snowflake/connector/arrow_context.py
```diff
@@ -119,22 +121,26 @@ def TIMESTAMP_LTZ_to_python_windows(
...
+    ) -> datetime64:
         nanoseconds = int(decimal.Decimal(value).scaleb(9 - scale))
         return numpy.datetime64(nanoseconds, "ns")
 
-    def TIMESTAMP_NTZ_TWO_FIELD_to_numpy_datetime64(self, epoch, fraction):
+    def TIMESTAMP_NTZ_TWO_FIELD_to_numpy_datetime64(
+        self, epoch: int, fraction: int
```
sfc-gh-aling, line 143: should it be float?

src/snowflake/connector/auth/_auth.py
```diff
@@ -460,7 +465,7 @@ def _read_temporary_credential(self, host, user, cred_type) -> str | None:
                     "this please install keyring module using the following command : pip install "
                     "snowflake-connector-python[secure-local-storage]"
                 )
-                return
+                return None
```
sfc-gh-aling, line 468: is it mypy requiring to explicitly return None? a single `return` would just return None..

src/snowflake/connector/compat.py
```diff
@@ -23,9 +24,11 @@
 IS_MACOS = platform.system() == "Darwin"
 
 NUM_DATA_TYPES: tuple[type, ...] = ()
+numpy: ModuleType | None
 try:
-    import numpy
+    import numpy  # type: ignore
 
+    assert isinstance(numpy, ModuleType)
```
sfc-gh-aling, line 31: may I ask why we add the assertion here? I feel it will always be true

src/snowflake/connector/connection.py
```diff
@@ -329,183 +331,184 @@ def _ocsp_mode(self) -> OCSPMode:
...
         return self._user
 
     @property
-    def host(self):
+    def host(self) -> str:
         return self._host
 
     @property
-    def port(self):
+    def port(self) -> int | str:  # TODO: shouldn't be a string
```
sfc-gh-aling, line 346: if this shouldn't be a string, then our type probably should not hint it could be a string?
otherwise users could be dependent on the string behavior, and when one day we go with int only, users think we're breaking them..

src/snowflake/connector/connection.py
```diff
@@ -329,183 +331,184 @@ def _ocsp_mode(self) -> OCSPMode:
...
         return int(self._network_timeout) if self._network_timeout is not None else None
 
     @property
-    def client_session_keep_alive(self):
+    def client_session_keep_alive(self) -> bool | None:
         return self._client_session_keep_alive
 
     @client_session_keep_alive.setter
-    def client_session_keep_alive(self, value):
+    def client_session_keep_alive(self, value) -> None:
```
sfc-gh-aling, line 406: I saw you gave type hint to the `value` of the `errorhandler` setter.
are we missing the type hint for `value` of other setters?

src/snowflake/connector/cursor.py
```diff
@@ -226,12 +231,11 @@ def __init__(
         ] = []
         self._timebomb: Timer | None = None  # must be here for abort_exit method
         self._description: list[ResultMetadata] | None = None
-        self._column_idx_to_name = None
```
sfc-gh-aling, line 229: this attribute is removed?

src/snowflake/connector/cursor.py
```diff
@@ -394,16 +406,15 @@ def close(self) -> bool | None:
...
                 return False
-
             with self._lock_canceling:
                 self.reset(closing=True)
                 self._connection = None
                 del self.messages[:]
                 return True
         except Exception:
-            pass
+            return False
```
sfc-gh-aling, line 415: technically this is changing behavior from returning None to return False?

src/snowflake/connector/cursor.py
```diff
@@ -957,7 +968,7 @@ def check_can_use_pandas(self):
                 },
             )
 
-    def query_result(self, qid):
+    def query_result(self, qid: str) -> SnowflakeCursor:
```
sfc-gh-aling, line 971: can use `Self`?
```suggestion
    def query_result(self, qid: str) -> Self:
```

sfc-gh-mkeller: We should be able to. `mypy` is supposed to support this through `typing_extensions` for older versions. https://peps.python.org/pep-0673/

src/snowflake/connector/compat.py
sfc-gh-mkeller: This is done to please `mypy` More info here: https://mypy.readthedocs.io/en/stable/type_narrowing.html

src/snowflake/connector/auth/_auth.py
sfc-gh-mkeller: I think so, otherwise I made a mistake

src/snowflake/connector/arrow_context.py
sfc-gh-mkeller: This was type-hinted by observing what values functions get during our integration tests, so maybe it should be a Union? It's hard to tell with these functions.

src/snowflake/connector/arrow_context.py
sfc-gh-mkeller: https://github.com/snowflakedb/snowflake-connector-python/pull/1418#discussion_r1092324142

src/snowflake/connector/proxy.py
```diff
@@ -6,9 +6,15 @@
...
 import os
+from typing import Any
 
 
-def set_proxies(proxy_host, proxy_port, proxy_user=None, proxy_password=None):
+def set_proxies(
+    proxy_host: str | None,
+    proxy_port: str | None,
+    proxy_user: Any | None = None,
+    proxy_password: Any | None = None,
```
sfc-gh-aling, line 16:
```suggestion
    proxy_user: str | None = None,
    proxy_password: str | None = None,
```

I tend to just type hint `str` for those two to be conservative about the types we support

src/snowflake/connector/result_batch.py
```diff
@@ -36,18 +36,14 @@
 DOWNLOAD_TIMEOUT = 7  # seconds
 
 if TYPE_CHECKING:  # pragma: no cover
+    from pandas import DataFrame
+    from pyarrow import DataType, Table
```
sfc-gh-aling, line 40: wouldn't this raise import error if pandas/pyarrow not installed? though it's under TYPE_CHECKING so run time might be fine, but during static checking, will this prompt error information if the two packages are not installed

src/snowflake/connector/result_batch.py
```diff
@@ -96,9 +92,7 @@ def col_to_converter(col: dict[str, Any]) -> tuple[str, SnowflakeConverterType]:
             )
             return type_name, python_method
 
-        column_converters: list[tuple[str, SnowflakeConverterType]] = [
-            col_to_converter(c) for c in rowtypes
-        ]
+        column_converters = [col_to_converter(c) for c in rowtypes]
```
sfc-gh-aling, line 95: emmm, we're removing type hint here?

src/snowflake/connector/result_set.py
```diff
@@ -22,12 +23,10 @@
 from .time_util import get_time_millis
 
 if TYPE_CHECKING:  # pragma: no cover
-    from snowflake.connector.cursor import SnowflakeCursor
+    from pandas import DataFrame
+    from pyarrow import Table
```
sfc-gh-aling, line 27: same question as above if pandas/pyarrow is not instslled

src/snowflake/connector/result_batch.py
sfc-gh-mkeller: Yes, believe it or not this fixed an error. TBH I cannot remember if the type-hint was incorrect here, or if `mypy` can just infer the actual type and didn't want to be told what this was.

src/snowflake/connector/result_batch.py
sfc-gh-mkeller: It does not cause issues. If you look at `.pre-commit-config.yaml` you'll see that we don't install either. I've disabled these modules completely in `mypy.ini`.
So these type-hints are only ever useful in devs'  actively working on the code-base. I've given this some more thought and I believe that devs should have every dependency installed instead of duplicating. Setting them to `None` failback might give a false sense to devs in their editors.

src/snowflake/connector/arrow_context.py
sfc-gh-aling: let's just keep it integer for now

src/snowflake/connector/compat.py
sfc-gh-aling: why we need numpy declaration at line 27?
I feel we can do `# type: ignore` at line 47 to get rid of line 27 and the assertion?

src/snowflake/connector/cursor.py
sfc-gh-mkeller: Ahh, my bad!
```suggestion
return None
```

src/snowflake/connector/dbapi.py
```diff
@@ -57,16 +55,3 @@ def TimestampFromTicks(ticks):
 NUMBER = _DBAPITypeObject(get_number_types())
 DATETIME = _DBAPITypeObject(get_timestamp_types())
 ROWID = _DBAPITypeObject()
-
-
-class Json(UnicodeMixin):
```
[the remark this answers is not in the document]
sfc-gh-mkeller: I think so, I can't find it in our docs, or in the DBAPI2 spec. It was added probably because of Python 2

src/snowflake/connector/connection.py
sfc-gh-mkeller: This was just my thought. When I saw this auto generated type-hint. If it's a unit test that calls this with a string and it isn't documented then we can update the type-hint.

sfc-gh-aling (approved):

PR #1444, opened 2023-02-16
SNOW-733835 fix AuthByIdToken expired token auth

sfc-gh-mkeller:

   Fixes SNOW-733835

   We made a bug when we introduced custom auth to the connector. When an ID token expires in storage (i.e. keychain in MacOS) then when we fail to connect we remove the expired token from storage and then propagate the Exception instead of initiating a new externalbrowser auth. In this PR I fix this bug

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -804,7 +804,13 @@ class SnowflakeConnection:
                     port=self.port,
                 )
             else:
-                self.auth_class = AuthByIdToken(id_token=self._rest.id_token)
+                self.auth_class = AuthByIdToken(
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/connection.py
```diff
@@ -879,7 +887,8 @@ def __config(self, **kwargs):
                             if isinstance(accepted_types, tuple)
                             else accepted_types.__name__,
                             type(value).__name__,
-                        )
+                        ),
+                        stacklevel=4,
```
sfc-gh-sfan, line 891: ditto

sfc-gh-sfan (approved):

PR #1470, opened 2023-03-09
SNOW-749141: Fix uploading a file > 200M when the same filename already exists in a stage

sfc-gh-yixie:

   Fixes #SNOW-749141. 

Call get_digest even when the file already exists on a stage. The current code skips calling get_digest so the digest is None.

### DIFF
--- a/src/snowflake/connector/storage_client.py
+++ b/src/snowflake/connector/storage_client.py
@@ -185,7 +185,7 @@ class SnowflakeStorageClient(ABC):
     def preprocess(self) -> None:
         meta = self.meta
         logger.debug(f"Preprocessing {meta.src_file_name}")
-
+        self.get_digest()
         if not meta.overwrite:
             self.get_file_header(meta.dst_file_name)  # Check if file exists on remote
             if meta.result_status == ResultStatus.UPLOADED:
@@ -201,7 +201,6 @@ class SnowflakeStorageClient(ABC):
         # Uploading
         if meta.require_compress:
             self.compress()
-        self.get_digest()
 
         self.preprocessed = True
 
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/storage_client.py
```diff
@@ -185,7 +185,7 @@ def get_file_header(self, filename: str) -> FileHeader | None:
     def preprocess(self) -> None:
         meta = self.meta
         logger.debug(f"Preprocessing {meta.src_file_name}")
-
+        self.get_digest()
```
sfc-gh-aling, line 188: can we test it? I remember we discussed offline that we can change the chunk size to make it easier to test

src/snowflake/connector/storage_client.py
```diff
@@ -185,7 +185,7 @@ def get_file_header(self, filename: str) -> FileHeader | None:
...
         if not meta.overwrite:
             self.get_file_header(meta.dst_file_name)  # Check if file exists on remote
```
sfc-gh-aling, line 190: while this `self.get_file_header` raise error? if it raises error, then is it still correct to call `self.get_digest` ahead

sfc-gh-yixie: As discussed offline, update the PR to add a new call to `get_digest` under `if not meta.overwrite:` and keep the previous `get_digest`.

src/snowflake/connector/storage_client.py
sfc-gh-yixie: Updated `test_large_put.py` to cover the this fix. 
I manually tested by using the updated test and removing the fix. The problem was reproduced. After adding back the fix with this PR, the problem disappeared.
I also manually tested with a file 250M on aws, azure and gcp.

sfc-gh-aling (approved):

sfc-gh-yixie:
@sfc-gh-mkeller Not sure why code coverage is down by -0.05%. The two impacted indirect files, one has code coverage decrease, shouldn't be related to this PR.
https://app.codecov.io/gh/snowflakedb/snowflake-connector-python/pull/1470/indirect-changes

PR #1477, opened 2023-03-15
Add support for specifying 'RSAPublicKey' instance instead of raw bytes

### LINKED ISSUE #1276: SNOW-676645: Add connector support for keypair authentication using hardware security modules or cloud key managers
## What is the current behavior?

Current Snowflake keypair authentication requires that the user has the private key in their possession as a file. However, Hardware Security Modules (e.g. Smartcards, Yubikeys, …) and Cloud Key Managers do not expose the private key to the user. Instead, they provide cryptographic operations as API functionality. I can, for example, ask the HSM to sign a certain string using the private key. 

## What is the desired behavior?

I would like to be able to use any cryptographic device or service capable of RSA-signatures for keypair authentication. More specifically, I want an interface against which I can implement my custom signing flow.

## How would this improve `snowflake-connector-python`?

It would provide more options for secure logins. In our case, we would use Azure Key Vault to store keys and allow services with Managed Identity to read those keys. This lets our service accounts log in without exposing passwords or private keys at any phase.

## References, Other Background

As a proof of concept, I have implemented a version of keypair authentication that abstracts signing and public key operations into a separate class. An instance of this class can be provided to the Snowflake connector to sign the token.

Here's an example. I will provide a full pull request soon.

```
class AzureKeyVaultManager(KeyManager):
    def __init__(self, kvclient, keyname):
        self._kvclient = kvclient
        self._keyname = keyname

    def public_key(self):
        pubkey = self._kvclient.get_key(self._keyname)
        e = int.from_bytes(pubkey.key.e, 'big')
        n = int.from_bytes(pubkey.key.n, 'big')
        rsakey = RSAPublicNumbers(e, n)
        return rsakey.public_key()

    def sign(self, message):
        cc = self._kvclient.get_cryptography_client(self._keyname)
        hash = sha256(message).digest()
        result = cc.sign(SignatureAlgorithm.rs256, hash)
        return result.signature


    ctx = snowflake.connector.connect(
        user='antti.rasinen@foo.com',
        key_manager=AzureKeyVaultManager(kc, 'testkey'),
        authenticator='KMS',
        account='foo.west-europe.azure',
        warehouse='compute_wh',
        database='misc',
    )
```

Please note that this approach would not add new dependencies into the connector. The user of the connector library is responsible for providing the class, customized to their particular usecase.

malthe:
This can be used to externalize the JWT encoding process (see linked issue for motivation and details).

For further motivation, see https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/ – managed keys (software- or hardware protected) are the answer to this problem.

   Fixes #1276.

   This change makes use of the fact that `RSAPrivateKey` is already an interface (ABC) and the [JWT library](https://pypi.org/project/jwt/) is fully compatible with a custom implementation, for example one that externalizes the signing process to a cloud-based service such as Azure Key Vault.

   Note that we do not provide any implementations of this interface, nor any new dependencies.

### DIFF
--- a/src/snowflake/connector/auth/keypair.py
+++ b/src/snowflake/connector/auth/keypair.py
@@ -127,7 +127,9 @@ class AuthByKeyPair(AuthByPlugin):
         elif isinstance(self._private_key, RSAPrivateKey):
             private_key = self._private_key
         else:
-            raise TypeError(f"Expected bytes or RSAPrivateKey, got {type(self._private_key)}")
+            raise TypeError(
+                f"Expected bytes or RSAPrivateKey, got {type(self._private_key)}"
+            )
 
         public_key_fp = self.calculate_public_key_fingerprint(private_key)
 
--- a/src/snowflake/connector/auth/keypair.py
+++ b/src/snowflake/connector/auth/keypair.py
@@ -129,8 +129,7 @@ class AuthByKeyPair(AuthByPlugin):
         else:
             raise TypeError(self._private_key)
 
-        public_key = private_key.public_key()
-        public_key_fp = self.calculate_public_key_fingerprint(public_key)
+        public_key_fp = self.calculate_public_key_fingerprint(private_key)
 
         self._jwt_token_exp = now + self._lifetime
         payload = {
@@ -155,9 +154,9 @@ class AuthByKeyPair(AuthByPlugin):
         return {"success": False}
 
     @staticmethod
-    def calculate_public_key_fingerprint(public_key):
+    def calculate_public_key_fingerprint(private_key):
         # get public key bytes
-        public_key_der = public_key.public_bytes(
+        public_key_der = private_key.public_key().public_bytes(
             Encoding.DER, PublicFormat.SubjectPublicKeyInfo
         )
 
--- a/src/snowflake/connector/auth/keypair.py
+++ b/src/snowflake/connector/auth/keypair.py
@@ -127,7 +127,7 @@ class AuthByKeyPair(AuthByPlugin):
         elif isinstance(self._private_key, RSAPrivateKey):
             private_key = self._private_key
         else:
-            raise TypeError(self._private_key)
+            raise TypeError(f"Expected bytes or RSAPrivateKey, got {type(self._private_key)}")
 
         public_key_fp = self.calculate_public_key_fingerprint(private_key)
 
--- a/src/snowflake/connector/auth/keypair.py
+++ b/src/snowflake/connector/auth/keypair.py
@@ -47,17 +47,18 @@ class AuthByKeyPair(AuthByPlugin):
 
     def __init__(
         self,
-        private_key: bytes,
+        private_key: bytes | RSAPrivateKey,
         lifetime_in_seconds: int = LIFETIME,
     ) -> None:
         """Inits AuthByKeyPair class with private key.
 
         Args:
-            private_key: a byte array of der formats of private key
+            private_key: a byte array of der formats of private key, or an
+                object that implements the `RSAPrivateKey` interface.
             lifetime_in_seconds: number of seconds the JWT token will be valid
         """
         super().__init__()
-        self._private_key: bytes | None = private_key
+        self._private_key: bytes | RSAPrivateKey | None = private_key
         self._jwt_token = ""
         self._jwt_token_exp = 0
         self._lifetime = timedelta(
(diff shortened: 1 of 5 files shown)


malthe:
The following is an example of how to use this functionality to authenticate using a key stored in an Azure Key Vault – the `sign` and `public_key` methods were implemented by @arsatiki:
```python
from cryptography.hazmat.primitives import hashes
from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicNumbers
from cryptography.hazmat.primitives.asymmetric.rsa import AsymmetricPadding
from cryptography.hazmat.primitives.asymmetric.padding import PKCS1v15
from cryptography.hazmat.primitives.asymmetric import utils as asym_utils

class AzureManagedKey(RSAPrivateKey):
    _algorithm_mapping = {
        hashes.SHA256: SignatureAlgorithm.rs256,
        hashes.SHA384: SignatureAlgorithm.rs384,
        hashes.SHA512: SignatureAlgorithm.rs512,
    }
    _key_name = "service-sandbox-user4"
    _kvclient = client

    def decrypt(self, *args) -> bytes:
        raise NotImplementedError()

    @property
    def key_size(self) -> int:
        raise NotImplementedError()

    def sign(
        self,
        data: bytes,
        padding: AsymmetricPadding,
        algorithm: Union[asym_utils.Prehashed, hashes.HashAlgorithm],
    ) -> bytes:
        if not isinstance(padding, PKCS1v15):
            raise ValueError("Unsupported padding: %s" % padding.name)
        sig_algorithm = self._algorithm_mapping.get(type(algorithm))
        if sig_algorithm is None:
            raise ValueError("Unsupported algorithm: %s" % algorithm.name)
        cc = self._kvclient.get_cryptography_client(self._key_name)
        digest = sha256(data).digest()
        result = cc.sign(sig_algorithm, digest)
        return result.signature

    def private_numbers(self) -> "RSAPrivateNumbers":
        raise NotImplementedError()

    def private_bytes(self, *args) -> bytes:
        raise NotImplementedError()

    def public_key(self) -> RSAPublicKey:
        pubkey = self._kvclient.get_key(self._key_name)
        e = int.from_bytes(pubkey.key.e, "big")
        n = int.from_bytes(pubkey.key.n, "big")
        rsakey = RSAPublicNumbers(e, n)
        return rsakey.public_key()
```

malthe:
I have read the CLA Document and I hereby sign the CLA.

malthe:
recheck

malthe:
For further motivation, see https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/.

Using managed keys (software- or hardware protected) is the answer to this problem.

attekei:
Would be wonderful to get this merged!

malthe:
I have updated my previous comment, adding an example of a custom plugin implementation `AuthByAzureKeyVault` that works without this pull request.

The main points being addressed here are:
- Any plugin needs to inherit from `AuthByKeyPair` which leads to brittle code.
- Some duplicated code in the case of a managed key in Azure Key Vault.

src/snowflake/connector/auth/keypair.py
```diff
@@ -102,28 +103,32 @@ def prepare(
...
-            )
+            if not isinstance(private_key, RSAPrivateKey):
+                raise ProgrammingError(
+                    msg=f"Private key type ({private_key.__class__.__name__}) not supported."
+                    "\nPlease provide a valid rsa private key in DER format as bytes "
+                    "object",
+                    errno=ER_INVALID_PRIVATE_KEY,
+                )
+        else:
+            private_key = self._private_key
```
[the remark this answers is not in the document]
malthe: Fixed in 225f846d5110e8dee93008b22af34e4ab08b8531 – I just raise a `TypeError(self._private_key)` if it's not either `bytes` or an `RSAPrivateKey`.

src/snowflake/connector/auth/keypair.py
```diff
@@ -148,9 +153,9 @@ def reauthenticate(self, **kwargs: Any) -> dict[str, bool]:
         return {"success": False}
 
     @staticmethod
-    def calculate_public_key_fingerprint(private_key):
+    def calculate_public_key_fingerprint(public_key):
         # get public key bytes
-        public_key_der = private_key.public_key().public_bytes(
+        public_key_der = public_key.public_bytes(
```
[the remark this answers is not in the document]
malthe: Agreed. I have reverted this change in ac6c298 now.

src/snowflake/connector/auth/keypair.py
```diff
@@ -102,26 +103,31 @@ def prepare(
...
+                raise ProgrammingError(
+                    msg=f"Private key type ({private_key.__class__.__name__}) not supported."
+                    "\nPlease provide a valid rsa private key in DER format as bytes "
+                    "object",
+                    errno=ER_INVALID_PRIVATE_KEY,
+                )
+        elif isinstance(self._private_key, RSAPrivateKey):
+            private_key = self._private_key
+        else:
+            raise TypeError(self._private_key)
```
sfc-gh-sfan, line 130: nit: let's be more explicit on the error message: `f"Expected bytes or RSAPrivateKey, got {type(self._private_key)}"`

sfc-gh-sfan (approved):

src/snowflake/connector/auth/keypair.py
malthe: Fixed in f779da2a027a5e9892a87597dc64bfce89fc6401.

sfc-gh-sfan: Let's add an unhappy path test to cover this :) Otherwise LGTM!

malthe: Fixed in 48e5aca5070353b3594513081e4c6656fe953bd4.

sfc-gh-sfan:
Test PR in #1681

sfc-gh-sfan:
Hey can you fix the lint failures? Also, it would be great if you could add this commit for changelog: https://github.com/snowflakedb/snowflake-connector-python/commit/6f459f90763fc57d9112fc21f783abbe25c508dc (will need a rebase to main).

malthe:
@sfc-gh-sfan hope I did the right thing with that change log entry.

sfc-gh-sfan:
Can you take a look at the test failure?

test/integ/test_key_pair_authentication.py
```diff
@@ -248,8 +248,6 @@ def test_bad_private_key(db_parameters):
     )
 
     bad_private_key_test_cases = [
-        "abcd",
-        1234,
```
sfc-gh-sfan, line 252: I think removing the test is wrong. We should adapt the test to cover these cases, right?

malthe: I don't think so – it's not an integration test concern I would say, and we've got the unit test to cover it. One has to keep in mind that these _are_ regression tests, not a complete certification. We're trying to avoid regressions.

sfc-gh-sfan: How about we move these two cases to the unit test as well, in addition to `class Bad`?

malthe: Seems reasonable – good suggestion, I have pushed a change to the previous commit now in 818c3e8312777038681497e835f49541d14fef18, _moving_ those tests instead. Should help with continuity.

sfc-gh-sfan:
Merged. Thanks for your contribution :)

PR #1488, opened 2023-03-23
SNOW-761256 Fix finish_download in gcs_storage_client

sfc-gh-sfan:
Description

After super().finish_download(), self.intermediate_dst_path has either been deleted or removed. We should use the dst file name instead.

Testing

Existing test. I'm curious to learn how to test as well

   Fixes SNOW-761256

   After super().finish_download(), self.intermediate_dst_path has either been deleted or removed. We should use the dst file name instead.


sfc-gh-sfan:
What's really interesting is that the file is still downloaded and we could return the correct GET result. Given this is running in multithreading and only on a child thread, this stacktrace is actually not brought up. I'm planning to use caplog to make check if there are exceptions, but let me know if there is a better way otherwise.

sfc-gh-stan (approved):
🚢 🚢

PR #1489, opened 2023-03-24
SNOW-706838 add option to skip upload file if content matches

sfc-gh-aalam:

   Fixes #SNOW-706838

   Adding a parameter that allows to skip file uploads to stage in case contents of the remote file match the content of the local file.

### DIFF
--- a/src/snowflake/connector/storage_client.py
+++ b/src/snowflake/connector/storage_client.py
@@ -185,9 +185,12 @@ class SnowflakeStorageClient(ABC):
     def preprocess(self) -> None:
         meta = self.meta
         logger.debug(f"Preprocessing {meta.src_file_name}")
+
+        file_header = self.get_file_header(
+            meta.dst_file_name
+        )  # check if file exists on remote
         if not meta.overwrite:
             self.get_digest()  # self.get_file_header needs digest for multiparts upload when aws is used.
-            self.get_file_header(meta.dst_file_name)  # Check if file exists on remote
             if meta.result_status == ResultStatus.UPLOADED:
                 # Skipped
                 logger.debug(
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/storage_client.py
```diff
@@ -202,6 +205,15 @@ def preprocess(self) -> None:
...
             self.compress()
         self.get_digest()
+
+        if (
+            meta.skip_upload_on_content_match
+            and file_header
+            and meta.sha256_digest == file_header.digest
+        ):
+            logger.debug(f"same file contents for {meta.name}, skipping upload")
+            meta.result_status = ResultStatus.SKIPPED
```
sfc-gh-aalam, line 215: I think codecov bot think this line is not covered is because we skip test if `CONNECTION_PARAMETERS_ADMIN` is not found

sfc-gh-sfan (approved):
Do we plan to add this to snowpark? If so we probably would need to add to sp connector as well.

sfc-gh-jdu (approved):
Actually in snowpark we have a similar mechanism https://github.com/snowflakedb/snowpark-python/blob/13a9ed9e0a74505b60caffebd3bbb41835354445/src/snowflake/snowpark/session.py#L610-L613. Maybe you can replace it with this change in python connector

sfc-gh-aalam:
> Actually in snowpark we have a similar mechanism https://github.com/snowflakedb/snowpark-python/blob/13a9ed9e0a74505b60caffebd3bbb41835354445/src/snowflake/snowpark/session.py#L610-L613. Maybe you can replace it with this change in python connector

In the current behavior, what happens if users want to upload a different version of package with same name. It seems like `_resolve_imports` skips upload if an older version exists?

PR #1493, opened 2023-03-29
Fix `__iter__()` to call `_prefetch_hook()`

Kache:
Usage of `execute_async()` requires running `wait_until_ready()` (via
`_prefetch_hook()`, set in `get_results_from_sfqid()`) before fetching
results.

`fetchone()` does the above, but `__iter__()` was missed in #398.

Fix by correcting & consolidating iteration within `__iter__()` and
having `fetchone()` depend on it, instead of having two usages of
`_result_iterator()`.

Fix: #1472
Since: 8735e31d33866bdc2a5272682d8389692c7e0262
Jira: SNOW-757920

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -1177,37 +1177,9 @@ class SnowflakeCursor:
 
         return self
 
-    def _result_iterator(
-        self,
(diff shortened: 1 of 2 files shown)


sfc-gh-mkeller:
Merged through #1515 .
Thank you, @Kache ❤️

PR #1494, opened 2023-03-31
[SNOW-773047] Fix chunk-helper, issue explicit warning for non-standard Pandas index

sfc-gh-lspiegelberg:

   Fixes SNOW-773047

   - [x] I am adding a new warning

   In pandas_tools.py write_pandas tools silently assumes a RangeIndex with step size 1 as this is the default index Pandas generates for a DataFrame. Yet, the index may contain arbitrary values. For non-integer based indices the chunk_helper utility used to write data in chunks to Snowflake uses the sequence protocol, but should use iloc instead. This PR issues a warning for any index that is not a RangeIndex with step=1, but allows to still ingest the data by dropping the index via a fixed chunk_helper.

### DIFF
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -39,7 +39,7 @@ def chunk_helper(lst: pd.DataFrame, n: int) -> Iterator[tuple[int, pd.DataFrame]
         yield 0, lst
         return
     for i in range(0, len(lst), n):
-        yield int(i / n), lst.iloc[i: i + n]
+        yield int(i / n), lst.iloc[i : i + n]
 
 
 def build_location_helper(
@@ -172,11 +172,13 @@ def write_pandas(
         chunk_size = len(df)
 
     if not isinstance(df.index, pd.RangeIndex) or df.index.step != 1:
-        warnings.warn(f"Pandas Dataframe has index of type {str(type(df.index))} which will not be written."
-                      f" Consider changing the index to pd.RangeIndex(...,step=1) or "
-                      f"calling reset_index() to keep index as column(s)",
-                      UserWarning,
-                      stacklevel=2)
+        warnings.warn(
+            f"Pandas Dataframe has index of type {str(type(df.index))} which will not be written."
+            f" Consider changing the index to pd.RangeIndex(...,step=1) or "
+            f"calling reset_index() to keep index as column(s)",
+            UserWarning,
+            stacklevel=2,
+        )
 
     cursor = conn.cursor()
     stage_location = build_location_helper(
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -12,7 +12,6 @@ from logging import getLogger
 from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING, Any, Callable, Iterable, Iterator, Sequence, TypeVar
 
-import pandas as pd
 from typing_extensions import Literal
 
 from snowflake.connector import ProgrammingError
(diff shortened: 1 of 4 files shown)


src/snowflake/connector/pandas_tools.py
```diff
@@ -32,13 +33,13 @@
 logger = getLogger(__name__)
 
 
-def chunk_helper(lst: T, n: int) -> Iterator[tuple[int, T]]:
+def chunk_helper(lst: pd.DataFrame, n: int) -> Iterator[tuple[int, pd.DataFrame]]:
```
sfc-gh-jdu, line 36: We can't import and use pandas directly, because python connector doesn't install pandas by default. I guess you have to
```suggestion
def chunk_helper(lst: "pandas.DataFrame", n: int) -> Iterator[tuple[int, "pandas.DataFrame"]]:
```

src/snowflake/connector/pandas_tools.py
```diff
@@ -170,6 +171,15 @@ def write_pandas(
     if chunk_size is None:
         chunk_size = len(df)
 
+    if not isinstance(df.index, pd.RangeIndex) or df.index.step != 1:
```
sfc-gh-jdu, line 174: same here

test/integ/pandas/test_pandas_tools.py
```diff
@@ -255,6 +256,97 @@ def test_write_pandas(
...
+        )
+    )
+
+    # convert to list of tuples to compare to received output
+    pandas_df_data = [tuple(row) for row in list(pandas_df.values)]
+
+    num_of_chunks = math.ceil(len(pandas_df_data) / chunk_size)
+
+    with conn_cnx(
+        user=db_parameters["user"],
```
sfc-gh-jdu, line 286: I don't think you need to specify these parameters.

src/snowflake/connector/pandas_tools.py
sfc-gh-lspiegelberg: Good catch! I see, just saw in the same file `pandas.DataFrame` is used. Going to try that out.

test/integ/pandas/test_pandas_tools.py
sfc-gh-lspiegelberg: I copied the test above, how can I get rid off the params?

sfc-gh-jdu: `with conn_cnx() as cnx:` should be enough

sfc-gh-jdu (approved):

src/snowflake/connector/pandas_tools.py
```diff
@@ -170,6 +172,15 @@ def write_pandas(
     if chunk_size is None:
         chunk_size = len(df)
 
+    if not isinstance(df.index, pandas.RangeIndex) or df.index.step != 1:
```
sfc-gh-jdu, line 175: hmmm if df.index.start is not 0, it will be written?

sfc-gh-lspiegelberg: are you suggesting adding a check that df.index.start must be 0 as well?

sfc-gh-jdu: Yes, if range index is 2, 3, 4, ... 10, it will also become 1, 2, ..., 9 after pulling data out as a pandas dataframe, right?

sfc-gh-lspiegelberg: we can add that and be even more restrictive, i.e. only allowing the standard range-index (which is step==1 and start==0). However, I opted to issue a warning at this point to allow code to still run. Basically the semantics are:
1.) issue warning if not standard pandas index
2.) load/ingest without index

Does that sound reasonable?

sfc-gh-jdu: Yes, sounds good. We should do that to avoid behavior change

sfc-gh-lspiegelberg: fixed.

sfc-gh-sfan (approved):
LGTM. We probably should also port this change to stored proc connector?

PR #1504, opened 2023-04-07
SNOW-730092: Fix float precision lost issue in SnowSQL

sfc-gh-aling:

   Fixes #NNNN

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -12,6 +12,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
   - Fixed a bug that prints error in logs for GET command on GCS
   - Added a parameter that allows users to skip file uploads to stage if file exists on stage and contents of the file match.
+  - Fixed a bug in `converter_snowsql.py` that converting TIMESTAMP_NTZ loses precision when the timestamp value is large.
 
 - v3.0.2(March 23, 2023)
 
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -14,7 +14,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added a parameter that allows users to skip file uploads to stage if file exists on stage and contents of the file match.
   - Fixed a bug when writing a Pandas DataFrame with non-default index in `snowflake.connector.pandas_tool.write_pandas`.
   - Fixed a bug when writing a Pandas DataFrame with column names containing double quotes in `snowflake.connector.pandas_tool.write_pandas`.
-  - Fixed a bug in `converter_snowsql.py` that converting TIMESTAMP_NTZ loses precision when the timestamp value is large.
+  - Fixed a bug in SnowSQL that parsing `TIMESTAMP_NTZ` result loses precision when the timestamp value is large.
 
 - v3.0.2(March 23, 2023)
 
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/converter_snowsql.py
```diff
@@ -198,6 +198,13 @@ def _TIMESTAMP_NTZ_to_python(self, ctx: dict[str, Any]) -> Callable:
         def conv(value: str) -> str:
             microseconds, fraction_of_nanoseconds = _extract_timestamp(value, ctx)
             try:
+                new_integer_part = str(microseconds).split(".")[0]
+                origin_integer_part = value.split(".")[0]
+                if new_integer_part != origin_integer_part and float(
+                    origin_integer_part
+                ) + 1 == float(new_integer_part):
```
sfc-gh-sfan, line 205: Maybe the correct behavior for "off by more than 1" should be failing, or is there legit use case? Not sure if it is going to be a behavior change though. Ideally we should not hit this issue or we are already silently corrupting the data I think?

sfc-gh-sfan (approved):

DESCRIPTION.md
```diff
@@ -14,6 +14,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added a parameter that allows users to skip file uploads to stage if file exists on stage and contents of the file match.
   - Fixed a bug when writing a Pandas DataFrame with non-default index in `snowflake.connector.pandas_tool.write_pandas`.
   - Fixed a bug when writing a Pandas DataFrame with column names containing double quotes in `snowflake.connector.pandas_tool.write_pandas`.
+  - Fixed a bug in `converter_snowsql.py` that converting TIMESTAMP_NTZ loses precision when the timestamp value is large.
```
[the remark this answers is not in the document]
sfc-gh-aling: sure thing, I will mention it in the snowsql release notes as well

src/snowflake/connector/converter_snowsql.py
sfc-gh-aling: unfortunately we're already silently corrupting customer's data.

this is a legit use case and a limitation of the Python float number -- from what I learn  Python float 64-bit double-precision values, and once the integer part is great than a certain value, it will sacrifice and round the decimal part.

e.g. if you type `123456789123456.99999` in the REPL, python returns `123456789123457.0`

> We observed this issue from the customer who uses the max possible timestamp value 9999-12-31 23.59.59.99999999 in their table, the value queried on SnowSQL ouputs 10000-01-01 00:00:00.999999.

DESCRIPTION.md
sfc-gh-aling, line 17:
```suggestion
  - Fixed a bug in SnowSQL that parsing `TIMESTAMP_NTZ` result loses precision when the timestamp value is large.
```

sfc-gh-mkeller:
There has been a lot of work done by quite a few people to address timestamp issues similar to this in https://github.com/snowflakedb/snowflake-connector-python/commit/44862bffb1fd20286cd86f7d77ca30d525b27d47 but since this is for SnowSQL I think this *isn't worth fixing* and breaking change users for.
Users can work around this limitation by casting the columns into string and defer the math involved to our server-side instead.

PR #1507, opened 2023-04-11
SNOW-754627 python connector query context serialization/deserialization format from Apache Arrow to JSON

sfc-gh-lqin:

   Fixes #SNOW-754627

   The QCC is currently transmitted in Arrow format which has been found to be not as performant as when the cache is n JSON format. Hence we need to update to support the  cache being transmitted in JSON instead. The JDBC side reference implementation is in https://github.com/snowflakedb/snowflake-jdbc/pull/1290


test/integ/test_connection.py
```diff
@@ -52,7 +52,6 @@
 except ImportError:  # Keep olddrivertest from breaking
     ER_FAILED_PROCESSING_QMARK = 252012
 
-
```
sfc-gh-aalam, line 55: was this removed by linter? if not please add this back

test/unit/test_query_context_cache.py
```diff
@@ -11,19 +11,20 @@
     QueryContextElement,
 )
 
+installed_pandas = True
+
 try:
-    from snowflake.connector.options import installed_pandas, pandas, pyarrow
+    from snowflake.connector.options import pandas
```
sfc-gh-aalam, line 17: same comment on using `.options`

src/snowflake/connector/query_context_cache.py
```diff
@@ -10,35 +10,30 @@
...
-            pa.field("context", pa.binary(), nullable=True),
-        ]
-    )
-else:
-    QUERY_CONTEXT_SCHEMA = None
+try:
+    from snowflake.connector.options import pandas
+except ImportError:
+    installed_pandas = False
+    pandas = None
```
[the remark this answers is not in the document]
sfc-gh-aalam: Also, I think since we are not using pyarrow at all, and have switched to json ser-deser, we don't need to check if pandas is installed or not

test/unit/test_query_context_cache.py
```diff
@@ -227,64 +228,64 @@ def test_same_id_with_stale_timestamp(
     assert_cache_with_data(qcc_with_data, expected_data)
 
 
-@pytest.mark.skipif(not installed_pandas, reason="pyarrow not installed")
+@pytest.mark.skipif(not installed_pandas, reason="pandas not installed")
```
sfc-gh-aalam, line 231: all `pytest.mark.skipif` can be removed that are checking for `installed_pandas` since we are not using arrow/pyarrow ser-deser.

sfc-gh-lqin: I agree. These changes aimed to bypass the pyarrow installation check. I can remove all related checks.

test/integ/test_connection.py
sfc-gh-lqin: reverted

test/unit/test_query_context_cache.py
sfc-gh-lqin: remove this installed pandas directly

src/snowflake/connector/query_context_cache.py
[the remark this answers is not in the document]
sfc-gh-lqin: I agree. I will remove it

src/snowflake/connector/query_context_cache.py
```diff
@@ -180,91 +169,115 @@ def _get_elements(self) -> set[QueryContextElement]:
...
-                        schema=QUERY_CONTEXT_SCHEMA,
-                    )
-                    writer.write_batch(record_batch)
-
-                stream.seek(0)
-                # use same encoding use on jdbc driver
-                data = b64encode(stream.read()).decode("iso-8859-1")
-                stream.close()
+                data = {}
+                if(size >= 1):
```
sfc-gh-aalam, line 189:
```suggestion
                if size > 0:
```
a more pythonic way or writing

src/snowflake/connector/query_context_cache.py
```diff
@@ -180,91 +169,115 @@ def _get_elements(self) -> set[QueryContextElement]:
...
+                #       "context": "base64 encoded context"
+                #     },
+                #     {
+                #       "id": 2,
+                #       "read_timestamp": 123456789,
+                #       "priority": 2,
+                #       "context": "base64 encoded context"
+                #     }
+                #   ]
+                if("entries" in data):
```
sfc-gh-aalam, line 251:
```suggestion
                #   ]
                # }
                if("entries" in data):
```
nit: closing the left curly braces

src/snowflake/connector/query_context_cache.py
```diff
@@ -180,91 +169,115 @@ def _get_elements(self) -> set[QueryContextElement]:
...
+                    # Deserialize entries
+                    entries = data["entries"]
+                    for entry in entries:
+                        if not isinstance(entry.get("id"), int):
+                            raise Exception(f"Invalid type for 'id' field: Expected int, got {type(entry['id'])}")
+                        if not isinstance(entry.get("timestamp"), int):
+                            raise Exception(f"Invalid type for 'timestamp' field: Expected int, got {type(entry['timestamp'])}")
+                        if not isinstance(entry.get("priority"), int):
+                            raise Exception(f"Invalid type for 'priority' field: Expected int, got {type(entry['priority'])}")
```
sfc-gh-aalam, line 260: raising `TypeError` seems more appropriate

src/snowflake/connector/query_context_cache.py
```diff
@@ -180,91 +169,115 @@ def _get_elements(self) -> set[QueryContextElement]:
...
+                        
+                        context = entry.get("context", None) #OpaqueContext field currently is empty from GS side.
+                    
+                        if context is not None and not isinstance(context, str):
+                            raise Exception(f"Invalid type for 'context' field: Expected str, got {type(entry['context'])}")
```
sfc-gh-aalam, line 265: please raise a `TypeError` here also

src/snowflake/connector/query_context_cache.py
sfc-gh-aalam: have these new exceptions been tested? If not, could you please add negative tests for these

src/snowflake/connector/connection.py
```diff
@@ -1100,7 +1100,9 @@ def cmd_query(
             if "finalRoleName" in data:
                 self._role = data["finalRoleName"]
             if "queryContext" in data and not _no_results:
-                self.set_query_context(data["queryContext"])
+                # here the data["queryContext"] field has been automatically converted into a dict type
+                self.set_query_context(json.dumps(data["queryContext"]))
+
```
sfc-gh-aalam, line 1106:
```suggestion

```

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: updated

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: I will add some unit testcases for this exception

src/snowflake/connector/connection.py
sfc-gh-lqin: reverted

src/snowflake/connector/query_context_cache.py
```diff
@@ -89,9 +80,7 @@ def __hash__(self) -> int:
         _hash += (_hash * 31) + self._read_timestamp
         _hash += (_hash * 31) + self._priority
         if self._context:
-            _hash += (_hash * 31) + int.from_bytes(md5(self._context).digest(), "big")
-        else:
-            _hash += _hash * 31
+            _hash += (_hash * 31) + int.from_bytes(md5(self._context.encode('utf-8')).digest(), "big")
```
sfc-gh-sfan, line 83: Dumb question: How is this hash change related to the SerDe change?

src/snowflake/connector/connection.py
```diff
@@ -1100,7 +1100,9 @@ def cmd_query(
             if "finalRoleName" in data:
                 self._role = data["finalRoleName"]
             if "queryContext" in data and not _no_results:
-                self.set_query_context(data["queryContext"])
+                # here the data["queryContext"] field has been automatically converted into a dict type
+                self.set_query_context(json.dumps(data["queryContext"]))
```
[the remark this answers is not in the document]
sfc-gh-lqin: We can directly work on the original dict. I can update the testcase accordingly. Previously I did not want to modify the testcase side

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: To be consistent with the GS and JDBC changes, here `context` is `str` field, previous implementation is `bytearray`.

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: updated

src/snowflake/connector/query_context_cache.py
```diff
@@ -180,91 +169,119 @@ def _get_elements(self) -> set[QueryContextElement]:
...
                 return None
 
             try:
-                stream = BytesIO()
                 size = self.get_size()
                 id_vals = [None] * size
                 timestamp_vals = [None] * size
                 priority_vals = [None] * size
                 context_vals = [None] * size
                 self.get_elements(id_vals, timestamp_vals, priority_vals, context_vals)
```
sfc-gh-sfan, line 187: Do we still need to use 4 elements arrays? I think We probably can just use the `qce` object itself.

src/snowflake/connector/query_context_cache.py
```diff
@@ -180,91 +169,119 @@ def _get_elements(self) -> set[QueryContextElement]:
...
-                        schema=QUERY_CONTEXT_SCHEMA,
-                    )
-                    writer.write_batch(record_batch)
-
-                stream.seek(0)
-                # use same encoding use on jdbc driver
-                data = b64encode(stream.read()).decode("iso-8859-1")
-                stream.close()
+                data = {}
+                if(size > 0):
```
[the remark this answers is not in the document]
sfc-gh-lqin: You are right. I should remove this if check.

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: I think it is OK to stay with the original implementation here.

sfc-gh-sfan: I think this introduced unnecessary complications to the code so I'd like to see it optimized. Or at the very least we should create a JIRA and fix it in a follow-up PR.

src/snowflake/connector/query_context_cache.py
```diff
@@ -80,7 +66,12 @@ def __lt__(self, other: object) -> bool:
             raise NotImplementedError(
                 f"cannot compare QueryContextElement with object of type {type(other)}"
             )
-        return self._priority < other.priority
+        if self._priority != other._priority:
+            return self._priority < other._priority
+        if self._id != other._id:
+            return self._id < other._id
+        return self._read_timestamp < other._read_timestamp
```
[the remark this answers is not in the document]
sfc-gh-sfan: > this aims to fix a bug.

This is the exact reason why I'd like to see it in a separate PR because we should not have one PR do a hundred of things. If this PR's title is SerDe, we should only have SerDe related changes. Having multiple stuff in one PR just increase the possibility of it getting reverted and make debug harder.

Given this does not go into main, I'm fine either way, but just pointing out that fixing a bug is not a good reason to stitch changes.

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: I see. Let me try to optimize it in this PR then.

src/snowflake/connector/query_context_cache.py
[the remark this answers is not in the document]
sfc-gh-lqin: I see. I can revert this part change and make another PR. I just realize that this part change can not fix the multi-database query query context `merge` missing error.

sfc-gh-lqin:
Hi @sfc-gh-sfan and @sfc-gh-aalam , could you take a final round on the current changes? I will submit another PR to fix the `merge` issue. When dealing with multi-database query, and the query context priority of each database might change, and the current implementation could fail to `merge` entries due to the priority switch.

For example, Entry A priority 1->2, B 2->1, when replace old A with new A, the `priority_map` add(2, new A) will fail due to duplicate priority key `2`

sfc-gh-lqin:
I think a simple fix is to use a nest dict for priority_map, when there is priority key duplication, we use `id` as key. I will push the PR later.

PR #1515, opened 2023-04-18
Test 1493

### LINKED ISSUE #1472: SNOW-757920: SnowflakeCursor doesn't call `_prefetch_hook()` before yielding results of an `execute_async()`

Python 3.7.6

Darwin-20.6.0-x86_64-i386-64bit

   Replace with the output of `python -m pip freeze`
   Not particularly relevant, should be able to repro in a blank project.

4. What did you do?

Following from docs: https://docs.snowflake.com/en/user-guide/python-connector-example.html

```py
def exec_and_log(conn, sql_str):
    cur = conn.cursor()
    if not cur.execute_async(sql_str):
        raise Exception('Some unknown error in Snowflake SDK')

    sfqid = str(cur.sfqid or '')
    logging.info(sfqid)
    cur.get_results_from_sfqid(sfqid) 
    yield from cur
```

`_result_iterator()` will fail because `cur._result` is not set, with `TypeError: 'NoneType' object is not iterable` at:

https://github.com/snowflakedb/snowflake-connector-python/blob/8bcd1fbfc4f01ff4d48226a7a15691c9e406b67c/src/snowflake/connector/cursor.py#L1089

(I may be mistaken, but it seems clear that...)
Cursor should first wait for results, then yield them in the same manner as `fetchone()` does by calling `_prefetch_hook()`: https://github.com/snowflakedb/snowflake-connector-python/blob/8bcd1fbfc4f01ff4d48226a7a15691c9e406b67c/src/snowflake/connector/cursor.py#L1104-L1114

For example, my current workaround:

```py
def exec_and_log(conn, sql_str):
    cur = conn.cursor()
    if not cur.execute_async(sql_str):
        raise Exception('Some unknown error in Snowflake SDK')

    sfqid = str(cur.sfqid or '')
    logging.info(sfqid)

    # these private accesses to prepare internal state shouldn't be necessary
    cur._connection = conn  # have to do this if cursor was closed, despite connection still being open
    cur.get_results_from_sfqid(sfqid) 
    if cur._prefetch_hook:
        cur._prefetch_hook()  # calls wait_until_ready(), which'll set self._result_set

    yield from cur
```

But it should instead be called internally, perhaps at the beginning of `__iter__()` (note section with `self._result_set` is identical to `fetchone()`): https://github.com/snowflakedb/snowflake-connector-python/blob/8bcd1fbfc4f01ff4d48226a7a15691c9e406b67c/src/snowflake/connector/cursor.py#L1193-L1199

Afterwards, the body of `fetchone()` can be just: `return next(iter(self))`

Note: this issue is recreated from #1234 

   ```
   import logging
   import os

   for logger_name in ('snowflake.connector',):
       ch = logging.StreamHandler()
       ch.setLevel(logging.DEBUG)
       logger.addHandler(ch)
   ```

sfc-gh-aalam:
Test PR for https://github.com/snowflakedb/snowflake-connector-python/pull/1493

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -1243,7 +1243,28 @@ class SnowflakeCursor:
 
     def fetchone(self) -> dict | tuple | None:
         """Fetches one row."""
-        return next(iter(self), None)
+        if self._prefetch_hook is not None:
+            self._prefetch_hook()
+        if self._result is None and self._result_set is not None:
+            self._result = iter(self._result_set)
+            self._result_state = ResultState.VALID
+
+        try:
+            _next = next(self._result, None)
+            if isinstance(_next, Exception):
+                Error.errorhandler_wrapper_from_ready_exception(
+                    self._connection,
+                    self,
+                    _next,
+                )
+            if _next is not None:
+                self._rownumber += 1
+            return _next
+        except TypeError as err:
+            if self._result_state == ResultState.DEFAULT:
+                raise err
+            else:
+                return None
 
     def fetchmany(self, size: int | None = None) -> list[tuple] | list[dict]:
         """Fetches the number of specified rows."""
@@ -1338,30 +1359,13 @@ class SnowflakeCursor:
         if not self.connection._reuse_results:
             self._result_set = None
 
-    def __iter__(self) -> Iterator[dict] | Iterator[tuple]:
+    def __iter__(self) -> Iterator[dict | None] | Iterator[tuple | None]:
         """Iteration over the result set."""
-        if self._prefetch_hook is not None:
-            self._prefetch_hook()
-        # set _result if _result_set is not None
-        if self._result is None and self._result_set is not None:
-            self._result = iter(self._result_set)
-            self._result_state = ResultState.VALID
-
-        try:
-            for _next in self._result:
-                if isinstance(_next, Exception):
-                    Error.errorhandler_wrapper_from_ready_exception(
-                        self._connection,
-                        self,
-                        _next,
-                    )
-                self._rownumber += 1
-                yield _next
-        except TypeError as err:
-            if self._result_state == ResultState.DEFAULT:
-                raise err
-            else:
-                yield None
+        while True:
+            _next = self.fetchone()
+            if _next is None:
+                break
+            yield _next
 
     def __cancel_query(self, query) -> None:
         if self._sequence_counter >= 0 and not self.is_closed():
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -21,7 +21,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
-    Generator,
     Iterator,
     NamedTuple,
     NoReturn,
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -1243,7 +1243,7 @@ class SnowflakeCursor:
 
     def fetchone(self) -> dict | tuple | None:
         """Fetches one row."""
-        next(iter(self), None)
+        return next(iter(self), None)
 
     def fetchmany(self, size: int | None = None) -> list[tuple] | list[dict]:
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -1242,37 +1241,9 @@ def executemany(
...
-        if self._prefetch_hook is not None:
-            self._prefetch_hook()
-        if self._result is None and self._result_set is not None:
-            self._result = iter(self._result_set)
-            self._result_state = ResultState.VALID
-        try:
-            return next(self._result_iterator())
-        except StopIteration:
-            return None
+        return next(iter(self), None)
```
sfc-gh-sfan, line 1246: Instead of creating `iter` every time, we should consider using the existing `iter`.

src/snowflake/connector/cursor.py
```diff
@@ -1242,37 +1241,30 @@ def executemany(
...
             self._prefetch_hook()
         if self._result is None and self._result_set is not None:
             self._result = iter(self._result_set)
             self._result_state = ResultState.VALID
+
         try:
-            return next(self._result_iterator())
-        except StopIteration:
-            return None
+            _next = next(self._result, None)
```
sfc-gh-sfan, line 1253: This means `None` will never be a legit result as part of `next(self._result)`. Does this assumption hold? We could also omit the default value of `next` and use `StopIteration` to convey the "end of iteration" message.

src/snowflake/connector/cursor.py
```diff
@@ -1242,37 +1241,30 @@ def executemany(
...
+                    _next,
+                )
+            if _next is not None:
+                self._rownumber += 1
+            return _next
+        except TypeError as err:
+            if self._result_state == ResultState.DEFAULT:
+                raise err
+            else:
+                return None
```
sfc-gh-sfan, line 1267: I'm not entirely sure I understand what the other two states mean. Is returning `None` legit?

src/snowflake/connector/cursor.py
sfc-gh-mkeller: Right, [None by itself was never a valid value being returned](https://github.com/snowflakedb/snowflake-connector-python/blob/72c4f15dc01299ad319979f5b3294e35876d984f/src/snowflake/connector/result_set.py#L42).

src/snowflake/connector/cursor.py
sfc-gh-mkeller: > Fetch the next row of a query result set, returning a single sequence, or `None` when no more data is available.

[source](https://peps.python.org/pep-0249/#fetchone)

sfc-gh-sfan (approved):

src/snowflake/connector/cursor.py
```diff
@@ -1367,13 +1359,13 @@ def reset(self, closing: bool = False) -> None:
...
+    def __iter__(self) -> Iterator[dict | None] | Iterator[tuple | None]:
         """Iteration over the result set."""
-        # set _result if _result_set is not None
-        if self._result is None and self._result_set is not None:
-            self._result = iter(self._result_set)
-            self._result_state = ResultState.VALID
-        return self._result_iterator()
+        while True:
+            _next = self.fetchone()
+            if _next is None:
```
sfc-gh-aling, line 1366: is it possible that `None` is a valid value in self._result, in such case we exit the loop sooner that some of data are not returned to users

src/snowflake/connector/cursor.py
```diff
@@ -1367,13 +1359,13 @@ def reset(self, closing: bool = False) -> None:
         if not self.connection._reuse_results:
             self._result_set = None
 
-    def __iter__(self) -> Iterator[dict] | Iterator[tuple]:
+    def __iter__(self) -> Iterator[dict | None] | Iterator[tuple | None]:
```
sfc-gh-aling, line 1362: this might fall out of the scope of this PR but do we consider making the cursor an iterator object (\_\_next__ + __iter__) instead of just an iterable object (\_\_iter__)

sfc-gh-aling (approved):

PR #1524, opened 2023-04-21
SNOW-799180 Do not modify statement_params from user

sfc-gh-sfan:
Description

Today we are modifying statement_params provided by user if they are also a multistatement. This behavior seems awkward and could break user is they are sharing the same statement_params object between multiple queries

Testing

integ

   Fixes SNOW-799180

   Today we are modifying statement_params provided by user if they are also a multistatement. This behavior seems awkward and could break user is they are sharing the same statement_params object between multiple queries

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -8,8 +8,10 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 
-- v3.0.3(April 20, 2023)
+- v3.0.4(TBD)
+  - Fixed a bug that cursor.execute() could change user's input statement_params object if multistatement query is used.
 
+- v3.0.3(April 20, 2023)
   - Fixed a bug that prints error in logs for GET command on GCS.
   - Added a parameter that allows users to skip file uploads to stage if file exists on stage and contents of the file match.
   - Fixed a bug that occurred when writing a Pandas DataFrame with non-default index in `snowflake.connector.pandas_tool.write_pandas`.
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -9,7 +9,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 # Release Notes
 
 - v3.0.4(TBD)
-  - Fixed a bug that cursor.execute() could change user's input statement_params object if multistatement query is used.
+  - Fixed a bug in which `cursor.execute()` could modify the argument statement_params dictionary object when executing a multistatement query.
 
 - v3.0.3(April 20, 2023)
   - Fixed a bug that prints error in logs for GET command on GCS.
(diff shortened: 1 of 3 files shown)


sfc-gh-sfan:
I'm wondering if this should be considered as a behavior change. We certainly did not intend to change the statement_params from user, but do we need to worry that there are users already depending on this wrong behavior?

src/snowflake/connector/cursor.py
```diff
@@ -748,11 +749,18 @@ def execute(
             logger.warning("execute: no query is given to execute")
             return None
 
-        if _statement_params is None:
-            _statement_params = dict()
-
+        extra_statement_params = dict()
```
sfc-gh-stan, line 752: NIT:
How about we put
```
_statement_params = (_statement_params or dict()).copy()
```
before the if case?

DESCRIPTION.md
```diff
@@ -8,8 +8,10 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 
-- v3.0.3(April 20, 2023)
+- v3.0.4(TBD)
+  - Fixed a bug that cursor.execute() could change user's input statement_params object if multistatement query is used.
```
sfc-gh-stan, line 12:
```suggestion
  - Fixed a bug in which `cursor.execute()` could modify the argument statement_params dictionary object when executing a multistatement query.
```

src/snowflake/connector/cursor.py
sfc-gh-sfan: I think that will make us always copy the statement parameters, which I'm not sure if it is a good idea. Though what I did might also well be an overkill 🤔

sfc-gh-stan: I see, what about:
```
_statement_params = _statement_params or dict()
if num_statements:
    _statements_params = {**_statements_params, 'MULTI_STATEMENT_COUNT':num_statements}
```
?

sfc-gh-sfan: Hmmm but `extra_statement_params` is probably more extensible in case we want to add more parameters in the future. 

Maybe we could change the extra param check to:

```
        extra_statement_params = dict()
        if num_statements:
            extra_statement_params["MULTI_STATEMENT_COUNT"] = num_statements
        if extra_statement_params:
            _statement_params = {**(_statement_params or dict()), **extra_statement_params)}
```

sfc-gh-sfan: As discussed offline with @sfc-gh-stan, we will add a comment about `extra_statement_params` as a TODO when we need to add another statement. For now, we will add fewer lines of code.

sfc-gh-stan (approved):
LGTM. Thank you Shixuan 🙂

PR #1532, opened 2023-04-26
SNOW-754627 fix priority switch corner case error

sfc-gh-lqin:

   This PR is a follow-up fix for SNOW-754627.

   As stated in the JDBC PR side https://github.com/snowflakedb/snowflake-jdbc/pull/1366
   
   Consider a simple situation :

```
current state:
<id, priority, timestamp, context>
<1, 1, 1, None>
<2,2,2, None>
incoming changes:

merge <1, 2,1, None> 
   call remove QCE(<1,1,1, None>) -> remove priorityMap(1, <1,1,1, None>)
   call add QCE(<1,2,1, None>) -> add priorityMap(2, <1,2,1, None>) this map insertion will fail due to collision of priority key(there are two priority keys of 2)
merge <2,1,1, None>
```

Although this bug does not affect the contents in treeSet in this simple example (this is also why I thought I fixed the issue before, I only checked the final QCC entries embedded into the HTTP request, and found the previous multi-db missing query context issue was gone. However, the priorityMap contents are not correctly maintained), it can trigger inconsistent QCC treeSet and priorityMap contents in following rounds.

I also find another bug in python implementation:

Basically, in my end-to-end test, when I want to
```
def _remove_qce(self, qce: QueryContextElement) -> None:
        self._id_map.pop(qce.id)
        self._priority_map.pop(qce.priority)
        self._tree_set.remove(qce)
```
even if there is an element in tree_set with all the same field values(id, ts, priority, context), the remove API can not find the entry to remove. I checked the _hash_  values are the same. If i do
```
for ele in self._tree_set:
   if(ele.__eq__(qce)):
       self._tree_set.remove(ele)
```
This can work. 

This reason could be that the python QueryContextElement object in 
```
        self._id_map: dict[int, QueryContextElement] = {}
        self._priority_map: dict[int, QueryContextElement] = {}
        self._tree_set: set[QueryContextElement] = SortedSet()
```
are referencing the same underlying object. I use `copy.deepcopy` when adding a QCE to each map/set to avoid this tricky bug. In this way, we also can not do in-place update in `merge` function. The overhead should be very small since the QCC capacity is usually a small number. If you have any better solution, please let me know.

### DIFF
--- a/src/snowflake/connector/query_context_cache.py
+++ b/src/snowflake/connector/query_context_cache.py
@@ -11,6 +11,7 @@ from io import BytesIO
 from logging import DEBUG, getLogger
 from threading import Lock
 import json
+import copy
 
 from sortedcontainers import SortedSet
 
@@ -89,6 +90,7 @@ class QueryContextCache:
         self._capacity = capacity
         self._id_map: dict[int, QueryContextElement] = {}
         self._priority_map: dict[int, QueryContextElement] = {}
+        self._intermediate_priority_map: dict[int, QueryContextElement] = {}
 
         # stores elements sorted by priority. Element with
         # least priority value has the highest priority
@@ -101,9 +103,9 @@ class QueryContextCache:
         return self._capacity
 
     def _add_qce(self, qce: QueryContextElement) -> None:
-        self._id_map[qce.id] = qce
-        self._priority_map[qce.priority] = qce
         self._tree_set.add(qce)
+        self._id_map[qce.id] = copy.deepcopy(qce)
+        self._intermediate_priority_map[qce.priority] = copy.deepcopy(qce)
 
     def _remove_qce(self, qce: QueryContextElement) -> None:
         self._id_map.pop(qce.id)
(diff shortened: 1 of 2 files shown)


sfc-gh-sfan (commented):
Can we add a test that could cover this fix?

sfc-gh-lqin:
Sure. I will add some unit tests and push them later.

sfc-gh-lqin:
I have added a randomized unit testcase for priority switch case.

src/snowflake/connector/query_context_cache.py
```diff
@@ -115,23 +117,26 @@ def _replace_qce(
...
         self._add_qce(new_qce)
+        
+    def sync_priority_map(self):
+        """
+        Sync the _intermediate_priority_map with the _priority_map at the end of the current round of merges.
+        """
+        logger.debug(f"syncPriorityMap called priority_map size = {len(self._priority_map)}, new_priority_map size = {len(self._intermediate_priority_map)}")
+        
+        for key, value in self._intermediate_priority_map.items():
+            self._priority_map[key] = value
```
sfc-gh-sfan, line 128: nit: You could potentially do `self._priority_map.update(self._intermediate_priority_map)`

src/snowflake/connector/query_context_cache.py
```diff
@@ -115,23 +117,26 @@ def _replace_qce(
...
         self, id: int, read_timestamp: int, priority: int, context: str
     ) -> None:
         if id in self._id_map:
             qce = self._id_map[id]
-            # when id if found in cache and we are operating on a more recent timestamp
-            if read_timestamp > qce.read_timestamp:
-                if qce.priority == priority:
-                    # same priority updates the current context object
-                    qce.read_timestamp = read_timestamp
-                    qce.context = context
```
sfc-gh-sfan, line 129: We essentially removed this logic. Is this expected? Just want to make sure this is expected

src/snowflake/connector/query_context_cache.py
```diff
@@ -115,23 +117,26 @@ def _replace_qce(
...
-                else:
-                    # change in priority caused replacement of query context
-                    new_qce = QueryContextElement(id, read_timestamp, priority, context)
-                    self._replace_qce(qce, new_qce)
-            elif read_timestamp == qce.read_timestamp and priority != qce.priority:
+            if((read_timestamp > qce.read_timestamp) or (read_timestamp == qce.read_timestamp and priority != qce.priority)):
```
sfc-gh-sfan, line 138: nit: space after if, and remove extra outermost parenthesis

sfc-gh-sfan (approved):

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: Like I said in the description, because now id_map, priority_map and tree_set saves a `copy.deepcopy` of qce, we can not simply do in-place update like the old logic. We use `replace_qce` instead

src/snowflake/connector/query_context_cache.py
sfc-gh-lqin: updated it according to your suggestion

PR #1549, opened 2023-05-04
SNOW-802807 ConfigManager and config file support

sfc-gh-mkeller:

   Fixes SNOW-802807

   In this PR I add the ability for reading values out of our centralized configuration file.


src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,188 @@
...
+            type: A function that can turn str to the desired type, useful for reading value from environmental variable
+        """
+        self.name = name
+        self.type = _type
+        self.choices = choices
+        self._nest_path = _nest_path + [name]
+        self._root_parser: ConfigParser = _root_parser
+        self.env_name = env_name
+
+    def get(self):
```
sfc-gh-aling, line 49: I am mostly a fan of `@property` instead of `get()` method approach:

```python
@property
def value(self):
    ....

@value.setter
def value(self):
    ....
````

File: src/snowflake/connector/config_parser/parser.py [hunk 2]
```diff
@@ -0,0 +1,188 @@
...
+                f"{source} is not part of {self.choices}"
+            )
+        return value
+
+    def generate_name(self) -> str:
+        return ".".join(self._nest_path[1:])
+
+    def generate_env_name(self) -> str:
+        pieces = map(lambda e: e.upper(), self._nest_path[1:])
+        return f"SF{'_' + '_'.join(pieces)}"
```
sfc-gh-aling, line 68: same I think they can be property instead of methods

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class ConfigOption:
+    def __init__(
+        self,
+        name: str,
+        _root_parser: ConfigParser,
+        _nest_path: list[str],
+        _type: Callable[[str], _T] | None = None,
+        choices: Iterable[_T] | None = None,
```
sfc-gh-aling, line 32: what is `choices` here? the `Iterable[_T]` type hint makes this a little confusing to me. Do we only expect a fixed/constant set of choices -- I'm thinking of `Enum` for the `choices`

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,188 @@
...
+            # If we got a TOML table we probably want it in dictionary form
+            return e.value
+        return e
+
+
+class ConfigParser:
+    def __init__(
+        self,
+        *,
+        name: str,
```
sfc-gh-aling, line 104: I just want to call it out that `name` here is a required keyword argument, we are not making it a positional argument?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,188 @@
...
+            raise ConfigSourceError(
+                f"An unknown error happened while loading '{str(self.file_path)}"
+                f"', please see the error: {e}"
+            )
+
+    @wraps(ConfigOption.__init__)
+    def add_option(
+        self,
+        *args,
+        **kwargs,
```
sfc-gh-aling, line 140: I think we need to document what this `add_option` takes as valid inputs

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,188 @@
...
+            for sub_parser in node._sub_parsers.values():
+                _root_setter_helper(sub_parser)
+            # Deal with ConfigOptions
+            for option in node._options.values():
+                option._root_parser = self._root_parser
+                option._nest_path = self._nest_path + option._nest_path
+
+        _root_setter_helper(other)
+
+    def __getitem__(self, item: str) -> ConfigOption | ConfigParser:
```
sfc-gh-aling, line 174: since we have `__getitem__`, shall we have the counter part `__setitem__`?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,188 @@
...
+        self._check_child_conflict(new_option.name)
+        self._options[new_option.name] = new_option
+
+    def _check_child_conflict(self, name: str) -> None:
+        if name in (self._options.keys() | self._sub_parsers.keys()):
+            raise ConfigParserError(
+                f"'{name}' subparser, or option conflicts with a child element of '{self.name}'"
+            )
+
+    def add_subparser(self, other: ConfigParser) -> None:
```
sfc-gh-aling, line 157: what's the usage of subparser and do we expect users to add subparser in runtime?

from the tests I saw this function is always called along with the constructor. I'm thinking whether we just make `subparser` part of the `__init__` method and remove this `add` method.

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: I think these can be properties! 👍

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: It'd be not an enum, but a list (or tuple) of possible values.

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: I don't like property here, because if you're debugging this code in PyCharm (for example) it'll continuously evaluate it.
What's worse is that might trigger reading your config file in and cache it. I want this action in specific to be done on purpose. Therefore; I went with the `get` function to signify that this has some cost and I hope you saved your config file previously.

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: For now `ConfigParser` is only for reading values. In the future we can add this, but just not for now

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: `wraps` does that as `ConfigOption.__init__` has all of the arguments documented. PyCharm (and probably VS code) just don't render these properly

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: Since this is a public API I just thought that it'd be a better idea to require all usages to specify every argument as a named argument. It makes things less confusing.

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: Yes, these need to updated at run-time. Imagine that SnowPark wants to add its own option to the Snowflake config file. That can only be done at run-time and only by SnowPark when it gets imported.
So the idea was that SnowPark can create its own root parser and then nest that under the Connector's root parser with probably the name `snowpark`.
So something along the lines of

```python
snowpark_cp = ConfigParser(name="snowpark")
snowpark_cp.add_option(name="use_async_queries", _type=lambda s: s.lower() == "true", choices=(True, False))
from snowflake.connector.config_parser import CONFIG_PARSER
CONFIG_PARSER.add_subparser(snowpark_cp)
```

So hopefully every library using the Python connector can just maintain their own parser and once they added all option just register it to the connector (they can possibly nest more of them to create separation of options).

setup.cfg
```diff
@@ -60,6 +60,8 @@ install_requires =
     certifi>=2017.4.17
     typing_extensions>=4.3,<5
     filelock>=3.5,<4
+    platformdirs
+    tomlkit
```
sfc-gh-yixie, line 64: Do we need to get legal to review when we add new dependencies?

sfc-gh-mkeller: Sure, do you have description, or example of what to do?

src/snowflake/connector/config_parser/parser.py
sfc-gh-jdu: Can we add docstrings for all methods here?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+        _type: Callable[[str], _T] | None = None,
+        choices: Iterable[_T] | None = None,
+        env_name: str | None | Literal[False] = None,
+    ) -> None:
+        """Create a config option that can read values from different locations.
+
+        Args:
+            name: The name of the ConfigOption
+            env_name: Environmental variable value should be read from, if not supplied, we'll construct this
+            type: A function that can turn str to the desired type, useful for reading value from environmental variable
```
sfc-gh-jdu, line 40: It seems a little bit complicated to me. Do we want to support in our first version? Because for mandatory fields in connection parameters, they are all strings?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+        env_var = os.environ.get(env_name, None)
+        loaded_var: _T | None | str = env_var
+        if env_var and self.type is not None:
+            loaded_var = self.type(env_var)
+        if isinstance(loaded_var, (Table, TOMLDocument)):
+            # If we got a TOML table we probably want it in dictionary form
+            return True, loaded_var.value
+        return True, loaded_var
+
+    def _get_config(self):
```
sfc-gh-jdu, line 92: type hint

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+class ConfigParser:
+    def __init__(
+        self,
+        *,
+        name: str,
+        file_path: Path | None = None,
+    ):
+        self.name = name
+        self.file_path = file_path
+        # Objects holding subparsers and options
```
sfc-gh-jdu, line 111: Could you explain what's subparser? parser for different sections in a file?

src/snowflake/connector/connection.py
```diff
@@ -276,7 +277,11 @@ class SnowflakeConnection:
 
     OCSP_ENV_LOCK = Lock()
 
-    def __init__(self, **kwargs) -> None:
+    def __init__(
```
sfc-gh-jdu, line 280: As we previously discussed, we should also provide an argument for users that they can use a non default path? Maybe not in the first version, but just to make sure you're aware of it.

test/integ/test_connection.py
```diff
@@ -1192,3 +1192,33 @@ def check_packages(message: str, expected_packages: list[str]) -> bool:
     ) as conn, capture_sf_telemetry.patch_connection(conn, False) as telemetry_test:
         conn._log_telemetry_imported_packages()
         assert len(telemetry_test.records) == 0
+
+
+@pytest.mark.skipolddriver
```
sfc-gh-jdu, line 1197: Should we have an e2e test? (create a temp config file, read from it and connect)

test/integ/test_connection.py
```diff
@@ -1192,3 +1192,33 @@ def check_packages(message: str, expected_packages: list[str]) -> bool:
...
+                with conn.cursor() as cur:
+                    assert cur.execute("select 1;").fetchall() == [
+                        (1,),
+                    ]
+    except Exception:
+        pytest.fail("something failed", pytrace=False)
+
+
+@pytest.mark.skipolddriver
+def test_not_found_connection_name():
```
sfc-gh-jdu, line 1220: maybe more negative tests? e.g., config file is not present, wrong format, etc.

src/snowflake/connector/config_parser/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    copyfile(default_config, config_file)
+
+CONFIG_PARSER = ConfigParser(
+    name="CONFIG_PARSER",
+    file_path=config_file,
+)
+CONFIG_PARSER.add_option(
+    name="connections",
+    _type=parse,
+)
```
sfc-gh-turbaszek, line 27: What do you think about encapsulating this in factory method like `_build_config_parser`?

src/snowflake/connector/config_parser/parser.py
sfc-gh-turbaszek: +1 to what @sfc-gh-jdu . Personally I'm in favour of using methods like `getboolean` or `getfloat` when reading the values from config instead of converting them automatically to particular type.

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+        loaded_var: _T | None | str = env_var
+        if env_var and self.type is not None:
+            loaded_var = self.type(env_var)
+        if isinstance(loaded_var, (Table, TOMLDocument)):
+            # If we got a TOML table we probably want it in dictionary form
+            return True, loaded_var.value
+        return True, loaded_var
+
+    def _get_config(self):
+        e = self._root_parser._conf
```
sfc-gh-turbaszek, line 93: Can you please use more meaningful name?

test/unit/test_configparser.py
```diff
@@ -0,0 +1,208 @@
...
+from pathlib import Path
+from test.randomize import random_string
+from textwrap import dedent
+from typing import Callable, Dict, Union
+
+import pytest
+from pytest import raises
+
+try:
+    from snowflake.connector.config_parser import ConfigParser
```
sfc-gh-turbaszek, line 16: Should we name it `SnowflakeConfigParser` to avoid confusion with `configparser.ConfigParser`?

setup.cfg
sfc-gh-mkeller: The Python docs suggested it https://docs.python.org/3/library/tomllib.html
Since we support 3.7-3.9. We can't assume that the standard lib one is available

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+        return value
+
+    @property
+    def option_name(self) -> str:
+        return ".".join(self._nest_path[1:])
+
+    @property
+    def option_env_name(self) -> str:
+        pieces = map(lambda e: e.upper(), self._nest_path[1:])
+        return f"SF{'_' + '_'.join(pieces)}"
```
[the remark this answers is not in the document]
sfc-gh-mkeller: I propose that we use the shorthand `SF`

src/snowflake/connector/constants.py
```diff
@@ -14,6 +14,14 @@
...
     from pyarrow import DataType
 
+from platformdirs import PlatformDirs
+
+dirs = PlatformDirs(
+    appname="snowflake",
+    appauthor=False,
+    ensure_exists=True,
+)
+config_file = dirs.user_config_path / "config.toml"
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Yes, I still need to do this!

test/integ/test_connection.py
sfc-gh-mkeller: I have unit tests for all those. I made sure we have 100% coverage for `snowflake.connector.config_parser`

test/unit/test_configparser.py
sfc-gh-mkeller: I'm not a fan of prepending Snowflake to anything anymore. I'm not against it, but if others don't feel like we need this I'd rather not

src/snowflake/connector/connection.py
sfc-gh-mkeller: I forgot about this, let me add the capability and add it to `snowflake.connector.connect`

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: Yes exactly. One example:
```toml
[connections.preprod]
user = "user"
account = "acccount"
password = "password"

[snowcli.output]
format = "json"
verbose = true
```

Refers to:
```
Root parser
|
snowcli
|
output
|               \
(format)   (verbose)
```
So being able to nest `ConfigParser`s in each other allows for this

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: Yes, I need to finish these still

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Args:
+            name: The name of the ConfigOption
+            env_name: Environmental variable value should be read from, if not supplied, we'll construct this
+            type: A function that can turn str to the desired type, useful for reading value from environmental variable
+        """
+        self.name = name
+        self.type = _type
+        self.choices = choices
+        self._nest_path = _nest_path + [name]
```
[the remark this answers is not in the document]
sfc-gh-mkeller: `_nest_path` is necessary to be able to efficiently figure out where things are nested. So if I'm a `ConfigOption` where do I retrieve the value from the cached toml file we read? This list of strings tells you that exactly.

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,190 @@
...
+
+class ConfigOption:
+    def __init__(
+        self,
+        name: str,
+        _root_parser: ConfigParser,
+        _nest_path: list[str],
+        _type: Callable[[str], _T] | None = None,
+        choices: Iterable[_T] | None = None,
+        env_name: str | None | Literal[False] = None,
```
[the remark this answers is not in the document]
sfc-gh-mkeller: False disables reading these arguments from environmental variables. Imagine some token, we might say that's okay to store in the config file, but not in an environmental variable.
I'm still missing doc-strings. I'll cover this in that.

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: I think we don't need to do that yet, I'm a fan of only complicating the code when there's a need.
Something like what you're suggesting is usually useful for testing.

I'm not against it though. If others feel like we should do this from the get-go we can!

src/snowflake/connector/config_parser/parser.py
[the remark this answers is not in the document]
sfc-gh-jdu: but in our design doc, we intentionally decided to use full name to avoid confusion of SF. Snowservice and SnowML have already used SNOWFLAKE prefix. We should align with the original decision. https://docs.google.com/document/d/1mi3mciW7AQkWiGozhC8edzX0S_PNonf5fty4YIIOgfA/edit#heading=h.uy91oh2ef1m1

src/snowflake/connector/config_parser/parser.py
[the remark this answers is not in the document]
sfc-gh-jdu: got it, please add some comments to explain this

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: I'd prefer this over that because the `_type` function only needs to be setup once and then we'd always be able to do `parser["option_name"]` to retrieve the values over-and-over again.
With the API you're proposing every usage would need to make sure that they're casting values correctly, so `parser["option_name"]` would become `parser["option_name"].getboolean()` possibly multiple times.

I do want to acknowledge that the name `_type` is not good now weeks after I wrote that piece of code :) It did make a lot of sense that day...

src/snowflake/connector/config_parser/parser.py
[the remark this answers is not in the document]
sfc-gh-mkeller: After some more discussion, I updated the PR to use `SNOWFLAKE_` as its default prefix

test/unit/test_configparser.py
sfc-gh-mkeller: After some discussion with my team, we've decided not to do this.
If it's necessary for some user we believe that `import-as` statements should solve their issues.

src/snowflake/connector/constants.py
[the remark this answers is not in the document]
sfc-gh-mkeller: CC @sfc-gh-jcarroll and @sfc-gh-sdas explicitly for comments on this

src/snowflake/connector/config_parser/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+from pathlib import Path
+from shutil import copyfile
+
+from tomlkit import parse
+
+from ..constants import config_file
+from .parser import ConfigParser
+
+# Create default configuration file, if it doesn't exist
+if not config_file.exists():
```
sfc-gh-bwarsaw, line 16: Two questions.

* Should we be doing this check in the module globals?  It means this gets executed when the module is imported which may or may not be the right time.  Also, I'm always wary of module import side effects.  It's sometimes difficult to predict when a module will be imported.  Would it be better to have an explicit call that guarantees the file exists?

* Isn't there a potential race condition here?  What if two processes end up here at the same time?  One find the file doesn't exist but before it can copy, another process enters the same conditional.  [copyfile()](https://docs.python.org/3/library/shutil.html#shutil.copyfile) says the destination will get ovewritten, but doesn't say that the write is atomic, so what's the possibility of file corruption?  Do you think it would be better to expand this and use EAFP principles to ensure atomicity?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def _get_env(self) -> tuple[bool, str | _T | None]:
+        """Get value from environment variable if possible.
+
+        Returns whether it was able to load the data and the loaded value
+        itself.
+        """
+        if self.env_name is False:
+            return False, None
+        if self.env_name is not None:
```
sfc-gh-lspiegelberg, line 114: Nit: if/else could be shortened to `self._env_name = self._env_name if self.env_name is not None else self.default_env_name`.

src/snowflake/connector/config_parser/parser.py
sfc-gh-bwarsaw: I don't in general think that all arguments of public APIs need to be keyword-only arguments.  It can make for more verbose call sites.  That said, there are some good use cases for them (e.g. boolean arguments).  Sumit and I were just discussing this particular style recommendation.

src/snowflake/connector/config_parser/parser.py
sfc-gh-bwarsaw: I didn't realize you needed to do this, or is it just a convenience?  I wonder if using a true [TOML parser](https://docs.python.org/3/library/tomllib.html) would make things simpler?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+        self._check_child_conflict(new_option.name)
+        self._options[new_option.name] = new_option
+
+    def _check_child_conflict(self, name: str) -> None:
+        """Check if a sub-parser, or ConfigOption conflicts with given name.
+
+        Args:
+            name: Name to check against children.
+        """
+        if name in (self._options.keys() | self._sub_parsers.keys()):
```
sfc-gh-bwarsaw, line 244: Double check, but I don't think you need `.keys()` here as iteration over keys is usually the default.

src/snowflake/connector/constants.py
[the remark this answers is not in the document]
sfc-gh-bwarsaw: Wouldn't using `$XDG_CONFIG_DIR` give users more control about where they want their config files, with standardized defaults?

test/integ/test_connection.py
```diff
@@ -1192,3 +1193,47 @@ def check_packages(message: str, expected_packages: list[str]) -> bool:
...
+                tmp_config_file.write_text(tomlkit.dumps(doc))
+            with snowflake.connector.connect(
+                connection_name="default",
+                config_file_path=tmp_config_file,
+            ) as conn:
+                with conn.cursor() as cur:
+                    assert cur.execute("select 1;").fetchall() == [
+                        (1,),
+                    ]
+    except Exception:
```
sfc-gh-bwarsaw, line 1230: Why not just let exceptions percolate up?

sfc-gh-bwarsaw (changes requested):

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+            # If we got a TOML table we probably want it in dictionary form
+            return e.value
+        return e
+
+
+class ConfigParser:
+    """Read TOML configuration file with managed multi-source precedence.
+
+     This class is updatable at run-time, allowing other libraries to add their
+     options own configuration options and sub-parsers. Sub-parsers allow
```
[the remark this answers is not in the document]
sfc-gh-mkeller:
```suggestion
    own configuration options and sub-parsers. Sub-parsers allow
```

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: This isn't because of TOML, this paradigm arose from the fact that I want to nest things not only in TOML, but also in environmental variables and possibly even more.
I use this hierachy to automatically generate environmental variables like: `SNOWFLAKE_SNOWCLI_OUTPUT_VERBOSE` from my example and I forsee using this very same thing to eventually parse arguments from SnowSQL(or other CLI) CLI arguments. So my example could be `snowsql --output_format json`.

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: * Could you propose how you think this should work? When should such a function be called?
* Yes, I forgot about this... We do something similar in [cache.py](https://github.com/snowflakedb/snowflake-connector-python/blob/d6a47498eefee8c125ad37cecc2361f6ef66da45/src/snowflake/connector/cache.py#L523). Essentially make temp file next to target file, write all content and then overwrite target file with temporary one. Would this be enough?

src/snowflake/connector/config_parser/__init__.py
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
+#
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
sfc-gh-mkeller: It's enforced by [our pre-commit config](https://github.com/snowflakedb/snowflake-connector-python/blob/d6a47498eefee8c125ad37cecc2361f6ef66da45/.pre-commit-config.yaml#L55).
I added this because I wanted us to be ready for what seems to be the future when it comes.

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @property
+    def option_name(self) -> str:
+        """User-friendly name of the config option. Includes self._nest_path."""
+        return ".".join(self._nest_path[1:])
+
+    @property
+    def default_env_name(self) -> str:
+        """The default environmental variable name for this option."""
+        pieces = map(lambda e: e.upper(), self._nest_path[1:])
```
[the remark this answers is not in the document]
sfc-gh-mkeller: Sure, that's an excellent alternative!

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: The only difference is that my comment won't really fit anywhere nicely. Should we sacrifice it for being a one-liner?
Could you maybe add a suggestion edit so that we could all visualize it first?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+        except Exception as e:
+            raise ConfigSourceError(
+                f"An unknown error happened while loading '{str(self.file_path)}"
+                f"', please see the error: {e}"
+            )
+
+    def add_option(
+        self,
+        *,
+        option_cls: type[ConfigOption] = ConfigOption,
```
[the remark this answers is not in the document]
sfc-gh-mkeller: My reasoning is that since the class is pluggable with `option_cls` I think using keyword only arguments will save developers from a ton of headache.
In other parts of this file I also found that I couldn't decide on an intuitive way to order arguments.
I definitely welcome your feedback and wouldn't be opposed to another discussion about this. The only thing I ask is that we have this meeting this week!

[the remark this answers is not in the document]
sfc-gh-mkeller: > all arguments for public APIs must be keyword-only
I don't agree with this. I think it's appropriate this time because the target audience is Snowflake developers. I'd definitely not do something like this with Public APIs that are targeted at end-users.

test/integ/test_connection.py
sfc-gh-mkeller: Because I'm actually operating on `db_parameters` which has our connection options in it (including password and other sensitive information). pytest has a very useful feature where if an Exception is raised all the way up it'll print local variables from the stack. This is my way of guaranteeing that we'll never expose sensitive information.
This is my first time Catching and Swallowing all exceptions in a test like this though, so I'm open to suggestions on doing this more elegantly.

src/snowflake/connector/connection.py
```diff
@@ -307,6 +314,18 @@ def __init__(self, **kwargs) -> None:
...
 
         self.converter = None
+        if config_file_path is not None:
+            # Change config file path and force update cache
+            CONFIG_PARSER.file_path = config_file_path
+            CONFIG_PARSER.read_config()
+        if connection_name is not None:
+            connections = CONFIG_PARSER["connections"]
+            if connection_name not in connections:
+                raise ProgrammingError(
```
sfc-gh-yixie, line 324: [ProgrammingError](https://peps.python.org/pep-0249/#programmingerror) is usually for SQL errors. We may need to raise a different error here as this is pure client.

src/snowflake/connector/connection.py
```diff
@@ -307,6 +314,18 @@ def __init__(self, **kwargs) -> None:
                 kwargs["application"] = "streamlit"
 
         self.converter = None
+        if config_file_path is not None:
+            # Change config file path and force update cache
+            CONFIG_PARSER.file_path = config_file_path
+            CONFIG_PARSER.read_config()
+        if connection_name is not None:
```
sfc-gh-yixie, line 321: When `config_file_path` and `connection_name` are not `None`, and the user also provide some kwargs, what will happen?

sfc-gh-yixie: Plus, will `SnowflakeConnection()` use the default connection name?

sfc-gh-mkeller: > When config_file_path and connection_name are not None, and the user also provide some kwargs, what will happen?

All other kwargs will override what we load from env variables, or config file.

sfc-gh-mkeller:
I want to make a note, so that I don't forget the fact that I want to warn users when the config file's permissions are too permissive for security purposes.

src/snowflake/connector/config_parser/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+from tomlkit import parse
+
+from ..constants import config_file
+from .parser import ConfigParser
+
+# Create default configuration file, if it doesn't exist
+if not config_file.exists():
+    # Create default config file
+    default_config = Path(__file__).absolute().parent / "default_config.toml"
```
sfc-gh-lspiegelberg, line 18: just to double-check: This would create a default config file in the folder where connector is deployed? Is this a good idea? Package deployment may be in read-only, protected directories. If creating a default file is really necessary, why not create it in CWD and/or warn the user? I feel default_config files should either be created as part of setup.py during deployment or at most in CWD as it may be fair game to assume CWD is writable. 

Is the existence of this default config file required for later code to run?

src/snowflake/connector/connection.py
sfc-gh-mkeller: > Plus, will `SnowflakeConnection()` use the default connection name?

I played around with the idea of having a default connection_name. I've decided against it for a few reasons:

* There's currently no way to see where a connection loaded its initial values from and which values came from overriding and which came from loading values.
* It could also lead to a forgotten setting put into the default connection and affecting code at run-time on a machine, so an identical script might execute very differently based on its environment.
* If this is necessary I think it's really easy for a user to use [functools.partial](https://docs.python.org/3/library/functools.html#functools.partial) to achieve this behavior.

src/snowflake/connector/config_parser/__init__.py
sfc-gh-lspiegelberg: +1 to refactor into a function and not run it as module import.

Also, I'd maybe recommend to create the default config file as part of setup.py?

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+              nested in. This is supplied by the parent ConfigParser.
+        """
+        self.name = name
+        self.parse_str = parse_str
+        self.choices = choices
+        self._nest_path = _nest_path + [name]
+        self._root_parser: ConfigParser = _root_parser
+        self.env_name = env_name
+
+    def get(self) -> Any:
```
sfc-gh-lspiegelberg, line 78: Nit: If this is an option, why not maybe follow the .get() function layout for dictionaries where a default value can be added when a key is not found?

If such behavior isn't desired, why not use a property `.value` or so? This would be similar to `option_name` below.

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: This line points to [default_config.py](https://github.com/snowflakedb/snowflake-connector-python/blob/mkeller/SNOW-765846-config_parser/src/snowflake/connector/config_parser/default_config.toml). We then copy this file to the user's [resolved config file](https://github.com/snowflakedb/snowflake-connector-python/blob/dfcc728f15fa2627c1f953651d80b94f8f2bcabb/src/snowflake/connector/constants.py#L33).

> I feel default_config files should either be created as part of setup.py during deployment or at most in CWD as it may be fair game to assume CWD is writable.

I don't think this is possible because of wheels, but I agree this would be better!

> Is the existence of this default config file required for later code to run?

It is not, Python connector will run happily without our new config file just like it has previously.

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: Sure, `value` it is then!

src/snowflake/connector/config_parser/parser.py
sfc-gh-lspiegelberg:
```
 # Generate environment name if it was not explicitly supplied,
 #  and isn't disabled
 self._env_name = self._env_name if self.env_name is not None else self.default_env_name
```

I may be even ok with excluding empty string as well
```
 # Generate environment name if it was not explicitly supplied,
 #  and isn't disabled
 self._env_name = self._env_name if self.env_name else self.default_env_name
```

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+        self._options: dict[str, ConfigOption] = dict()
+        self._sub_parsers: dict[str, ConfigParser] = dict()
+        # Dictionary to cache read in config file
+        self.conf_file_cache: TOMLDocument | None = None
+        # Information necessary to be able to nest elements
+        #  and add options in O(1)
+        self._root_parser: ConfigParser = self
+        self._nest_path = [name]
+
+    def read_config(
```
sfc-gh-lspiegelberg, line 199: Does this reload from file/reconfigure the parser each time when invoked?

The comment says something about caching, but the function seems to always reload and reconfigure which probably should be the correct approach.

Maybe remove the cache part from the comment? I think it's implementation detail, and there is not really a cache mechanism in the sense that a config file will be at most read once and then cached (which could be problematic for a couple reasons).

src/snowflake/connector/config_parser/__init__.py
sfc-gh-lspiegelberg: Hm, so wheels won't allow to copy a default config file to some standard dir? Maybe there's a workaround for that, e.g. entry-point?

I may be a bit lost here understanding what points to what. Might be a good idea to put the logic here despite it's just a few lines into a function and add some comments. I find especially mentioning to which file the relative path to __file__ should point to helpful from your comment, as it may help track bugs down if there's some directory setup we couldn't think of yet.

src/snowflake/connector/config_parser/__init__.py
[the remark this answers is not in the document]
sfc-gh-bwarsaw: I figured it was something like that.  Okay, I'll put a pin in this style nit for now :D

src/snowflake/connector/config_parser/__init__.py
sfc-gh-bwarsaw: > Essentially make temp file next to target file, write all content and then overwrite target file with temporary one. Would this be enough?

Yes, and I think you can do all of that a little more cleanly with higher level `tempfile` and `pathlib` APIs.  I've done this like a. million times 😄 

> +1 to refactor into a function and not run it as module import.

+1, or use an [existing library](https://python-atomicwrites.readthedocs.io/en/latest/).

sfc-gh-bwarsaw: > Also, I'd maybe recommend to create the default config file as part of setup.py?

-1; what happens when we replace setuptools with more modern Python packaging based on declarative `pyproject.toml` files?

Is there a top level entry point in the library that everyone needs to call to initialize things?

src/snowflake/connector/config_parser/__init__.py
sfc-gh-bwarsaw: Good points.  I think what @sfc-gh-mkeller is saying is that installing wheels doesn't run any code.  It could contain non-code data files, but that probably doesn't work in a cross platform way.  Do we need `default_config.toml` to hang around after the `copyfile()` to `config_file` location completes?  If not, then just stick the default in a tempfile and let it get unlinked once the real config file is put in place.

src/snowflake/connector/config_parser/parser.py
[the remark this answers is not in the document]
sfc-gh-bwarsaw: I think this can be resolved.

src/snowflake/connector/connection.py
sfc-gh-mkeller: Ahh, I missed the fact that it's not related to the database. Would [Error](https://peps.python.org/pep-0249/#error) make sense?

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: > I may be a bit lost here understanding what points to what. Might be a good idea to put the logic here despite it's just a few lines into a function and add some comments.

I'm also lost and don't really know what logic exactly you think I should put here. Could you take a stab at adding some comment here, maybe I could then expand on it!

> Hm, so wheels won't allow to copy a default config file to some standard dir?

Not as far as I know.

> Maybe there's a workaround for that, e.g. entry-point?

Not for libraries, for an application we could use those. This (import time) is essentially the entry-point for a library

src/snowflake/connector/config_parser/parser.py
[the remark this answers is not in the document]
sfc-gh-bwarsaw: Oops!  I meant [operator.methodcaller](https://docs.python.org/3/library/operator.html#operator.methodcaller)

src/snowflake/connector/config_parser/parser.py
sfc-gh-bwarsaw: I'm cool with leaving it to @sfc-gh-mkeller to decide.  I'll just point out that Guido chose ternary operator syntax deliberately so it would be a little bit ugly and people wouldn't over use it. 😄 .  There's nothing inherently wrong with if-else blocks.

src/snowflake/connector/config_parser/parser.py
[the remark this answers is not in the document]
sfc-gh-bwarsaw: > I don't agree with this. I think it's appropriate this time because the target audience is Snowflake developers for this. I'd definitely not do something like this with Public APIs that are targeted at end-users.

Fair enough.  I was curious because I've seen this principle described in other proposed internal style guides.  I disagreed with it as a general principle there and just wanted to call it out here.  Since it's a specific choice rather than a general principle, it's your call!

test/integ/test_connection.py
sfc-gh-bwarsaw: You're reasoning is sound; I'd just ask that if you keep it this way, please add a comment for your future self because it *looks* odd.  A comment will forestall any future headscratching.

That said, another approach would be to sanitize the exception if you knew where sensitive information could possibly reside.

As a principle, I see `except Exception` mostly on par with bare-`except`.  Generally these should be avoided unless you're catching all exceptions, doing some extra work, and then re-raising the original exception or chaining the original exception to a new one.  The classic example of course is aborting a transaction on an exception.

Thanks for the explanation.  I think whether you keep it or sanitize-and-reraise, please just add a comment explaining your rationale.

src/snowflake/connector/constants.py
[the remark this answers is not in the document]
sfc-gh-jcarroll: In the spec we said the path would be `~/.snowflake/config` and it would be in the ini format (although if it's equivalent to toml in all our docs I think that's fine)

This is the most similar to the current widely used default of `~/.snowsql/config`.

I don't think we should change to a different path now. I think `~/.snowflake/config` is the best option.

I think this will also be 1000x better and more used if it can detect a `~/.snowsql/config` and either accept it or at least give the user some hint to update their path, since every hit when I google "snowflake config file" tells me to put the file at `~/.snowsql/config`.

src/snowflake/connector/config_parser/parser.py
sfc-gh-jcarroll: I wish we had https://github.com/spf13/viper for python

sfc-gh-bwarsaw: I've never used it, but something like https://github.com/alexferl/vyper ?

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: We synced up offline and I've dropped default file generation.

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: We synced up offline and decided not to do default config file creation.

src/snowflake/connector/config_parser/__init__.py
sfc-gh-mkeller: I've now done this because I needed it for testing

src/snowflake/connector/constants.py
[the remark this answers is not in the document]
sfc-gh-mkeller: We synced up offline and came up with a new way described [here](https://github.com/snowflakedb/snowflake-connector-python/blob/e88677010ba460f4894f3d6c5ace8a321d406ace/src/snowflake/connector/constants.py#L18)

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: I have renamed `_type` to `parse_str`, but otherwise kept the API the same. Hopefully my previous comment makes sense and is convincing

src/snowflake/connector/config_parser/parser.py
```diff
@@ -0,0 +1,298 @@
...
+        """User-friendly name of the config option. Includes self._nest_path."""
+        return ".".join(self._nest_path[1:])
+
+    @property
+    def default_env_name(self) -> str:
+        """The default environmental variable name for this option."""
+        pieces = map(lambda e: e.upper(), self._nest_path[1:])
+        return f"SNOWFLAKE_{'_'.join(pieces)}"
+
+    def _get_env(self) -> tuple[bool, str | _T | None]:
```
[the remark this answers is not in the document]
sfc-gh-mkeller: I'm sorry, I'd do this but I keep getting this issue.
```
TypeError: unsupported operand type(s) for |: 'type' and 'TypeVar'
```

Let's follow up on this afterwards!

src/snowflake/connector/connection.py
sfc-gh-mkeller: I've switched to Error

src/snowflake/connector/config_parser/parser.py
sfc-gh-mkeller: I've added a blurb about when I'd expect someone to call this function. Hopefully, this makes things clearer!

src/snowflake/connector/config_manager.py
```diff
@@ -0,0 +1,333 @@
...
+from tomlkit.items import Table
+from typing_extensions import Literal
+
+from snowflake.connector.constants import CONFIG_FILE
+from snowflake.connector.errors import ConfigManagerError, ConfigSourceError
+
+_T = TypeVar("_T")
+
+LOGGER = logging.getLogger(__name__)
+READABLE_BY_OTHERS = stat.S_IRGRP | stat.S_IROTH
```
sfc-gh-bwarsaw, line 26: Do we care if the file is o+wx?  I think ssh wants its directory and config file to be only u+rw but I'm not positive about that.

src/snowflake/connector/config_manager.py
```diff
@@ -0,0 +1,333 @@
...
+        _nest_path: The names of the ConfigManagers that this option is
+          nested in. Used be able to efficiently resolve where to grab
+          value out of configuration file and construct environment
+          variable name. This is supplied by the parent ConfigManager.
+    """
+
+    def __init__(
+        self,
+        *,
+        name: str,
```
sfc-gh-bwarsaw, line 57: Can you make this order match the docstring?

src/snowflake/connector/config_manager.py
```diff
@@ -0,0 +1,333 @@
...
+        self.parse_str = parse_str
+        self.choices = choices
+        self._nest_path = _nest_path + [name]
+        self._root_parser: ConfigManager = _root_parser
+        self.env_name = env_name
+
+    def value(self) -> Any:
+        """Retrieve a value of option.
+
+        This function implements order of precedence between different sources.
```
sfc-gh-bwarsaw, line 86: Could it be a property?

src/snowflake/connector/config_manager.py
```diff
@@ -0,0 +1,333 @@
...
+        """
+        if self.env_name is False:
+            return False, None
+        if self.env_name is not None:
+            env_name = self.env_name
+        else:
+            # Generate environment name if it was not explicitly supplied,
+            #  and isn't disabled
+            env_name = self.default_env_name
+        if env_name not in os.environ:
```
sfc-gh-bwarsaw, line 125: How about not doing a LBYL check and combining these calls, e.g.

```
env_var = os.environ.get(env_name)
if env_var is None:
    return False, None
```

(No need to pass in `None` for the default because that's already the ... um, default!

sfc-gh-bwarsaw (approved):
Questions, but nothing to block this PR

src/snowflake/connector/config_manager.py
sfc-gh-mkeller: I had the same dilemma... I decided to not worry about these really unusual cases, but I'm still not a 100% sure whether we should

src/snowflake/connector/config_manager.py
sfc-gh-mkeller: I explained my reasoning here in another comment https://github.com/snowflakedb/snowflake-connector-python/pull/1549#discussion_r1187962914

src/snowflake/connector/config_parser/parser.py
sfc-gh-bwarsaw: > I don't like property here, because if you're debugging this code in PyCharm (for example) it'll continuously evaluate it.

I don't understand this comment.  What gets continuously evaluated?

src/snowflake/connector/config_manager.py
sfc-gh-hchaturvedi: I would approach it from an allow list perspective then a deny list - can we just come up with a permission that _is okay_ and raise a warning for everything else. That might be simpler.

sfc-gh-bwarsaw: Not sure that will work, but let's at least agree on the permissions that will be allowed or denied.

* u+rw -- this is allowed.  we probably don't care about u+x, though for strictness, probably denied if u+x (check w/ssh?)
* g+rw -- should we allow this?  I'd opt for "no" for now and we can always relax it later.  g+x same treatment as above.
* o+rwx - deny

PR #1551, opened 2023-05-08
Fix 'SOCKSProxyManager' object has no attribute 'proxy_headers' error

### LINKED ISSUE #1550: SNOW-810810: AttributeError when using SOCKS5 Proxy

```
Python 3.9.16 (main, May  8 2023, 20:48:22) 
[GCC 11.3.0]
```

```
Linux-5.19.0-41-generic-x86_64-with-glibc2.35
```

   Replace with the output of `python -m pip freeze`

4. What did you do?

Attempt to connect to Snowflake through a SOCKS5 proxy:

Use a minimal SOCKS5 proxy. From a second computer (`computer2`), open a SOCKS5 proxy on `computer1`

```
user@computer2:~$ ssh -R 1080 user@computer1.local
```

On `computer1`:

```
HTTPS_PROXY="socks5://localhost:1080" python3 -c "import snowflake.connector; snowflake.connector.connect(...)"
```

The above test results in:

```
  File "/home/stan/Ontwikkeling/snowflake-connector-python/src/snowflake/connector/network.py", line 1042, in _request_exec
    raw_ret = session.request(
  File "/home/stan/Ontwikkeling/snowflake-connector-python/src/snowflake/connector/vendored/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/stan/Ontwikkeling/snowflake-connector-python/src/snowflake/connector/vendored/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/home/stan/Ontwikkeling/snowflake-connector-python/src/snowflake/connector/vendored/requests/adapters.py", line 455, in send
    conn = self.get_connection(request.url, proxies)
  File "/home/stan/Ontwikkeling/snowflake-connector-python/src/snowflake/connector/network.py", line 274, in get_connection
    proxy_manager.proxy_headers["Host"] = parsed_url.hostname
AttributeError: 'SOCKSProxyManager' object has no attribute 'proxy_headers'
```

This is because the [SOCKSProxyManager](https://github.com/snowflakedb/snowflake-connector-python/blob/97c25993eb3fb43e0df634fd41f98f9e9ebf20bd/src/snowflake/connector/vendored/urllib3/contrib/socks.py#L159) is a subclass of [PoolManager](https://github.com/snowflakedb/snowflake-connector-python/blob/97c25993eb3fb43e0df634fd41f98f9e9ebf20bd/src/snowflake/connector/vendored/urllib3/poolmanager.py#L140) Which does not have the property `proxy_headers`. It does have a property `headers`.

I modified the code to not set the header when using a SOCKS5 proxy, which seems to work. The connection succeeds over the SOCKS5 proxy. It also works if I set the `headers` property instead of the `proxy_headers` property.

I offer a pull request for the first solution (not setting the `proxy_headers` when using a SOCKS5 proxy), but I leave it up to someone more knowledgable to decide whether this is the desired solution to this.

stan-janssen:
This fixes an error that would arise when using a SOCKS5 proxy.

   Fixes #1550

This code skips setting the `proxy_headers` property of the ProxyManager if we are using a SOCKS5 proxy.

For more details, see the [related issue](https://github.com/snowflakedb/snowflake-connector-python/issues/1550).

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -111,6 +111,7 @@ from .vendored.requests.exceptions import (
 )
 from .vendored.requests.utils import prepend_scheme_if_needed, select_proxy
 from .vendored.urllib3.exceptions import ProtocolError
+from .vendored.urllib3.poolmanager import ProxyManager
 from .vendored.urllib3.util.url import parse_url
 
 if TYPE_CHECKING:
@@ -268,16 +269,14 @@ class ProxySupportAdapter(HTTPAdapter):
                 )
             proxy_manager = self.proxy_manager_for(proxy)
 
-            try:
+            if isinstance(proxy_manager, ProxyManager):
                 # Add Host to proxy header SNOW-232777
                 proxy_manager.proxy_headers["Host"] = parsed_url.hostname
-            except AttributeError:
-                # log that the type proxy_manager not having attribute proxy_headers
+            else:
                 logger.debug(
                     f"Unable to set 'Host' to proxy manager of type {type(proxy_manager)} as"
                     f" it does not have attribute 'proxy_headers'."
                 )
-
             conn = proxy_manager.connection_from_url(url)
         else:
             # Only scheme should be lower case
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -26,7 +26,6 @@ from snowflake.connector.vendored.urllib3.connectionpool import (
     HTTPConnectionPool,
     HTTPSConnectionPool,
 )
-from snowflake.connector.vendored.urllib3.contrib.socks import SOCKSProxyManager
 
 from . import ssl_wrap_socket
 from .compat import (
@@ -269,9 +268,16 @@ class ProxySupportAdapter(HTTPAdapter):
                 )
             proxy_manager = self.proxy_manager_for(proxy)
 
-            if not isinstance(proxy_manager, SOCKSProxyManager):
+            try:
                 # Add Host to proxy header SNOW-232777
                 proxy_manager.proxy_headers["Host"] = parsed_url.hostname
+            except AttributeError:
+                # log that the type proxy_manager not having attribute proxy_headers
+                logger.debug(
+                    f"Unable to set 'Host' to proxy manager of type {type(proxy_manager)} as"
+                    f" it does not have attribute 'proxy_headers'."
+                )
+
             conn = proxy_manager.connection_from_url(url)
         else:
             # Only scheme should be lower case
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -26,6 +26,7 @@ from snowflake.connector.vendored.urllib3.connectionpool import (
     HTTPConnectionPool,
     HTTPSConnectionPool,
 )
+from snowflake.connector.vendored.urllib3.contrib.socks import SOCKSProxyManager
 
 from . import ssl_wrap_socket
 from .compat import (
@@ -268,8 +269,9 @@ class ProxySupportAdapter(HTTPAdapter):
                 )
             proxy_manager = self.proxy_manager_for(proxy)
 
-            # Add Host to proxy header SNOW-232777
-            proxy_manager.proxy_headers["Host"] = parsed_url.hostname
+            if not isinstance(proxy_manager, SOCKSProxyManager):
+                # Add Host to proxy header SNOW-232777
+                proxy_manager.proxy_headers["Host"] = parsed_url.hostname
             conn = proxy_manager.connection_from_url(url)
         else:
             # Only scheme should be lower case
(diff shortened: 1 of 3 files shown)


stan-janssen:

sfc-gh-yixie:
@sfc-gh-jdu @sfc-gh-aling Can you review if this PR makes sense to you?

src/snowflake/connector/network.py
```diff
@@ -265,8 +266,9 @@ def get_connection(
                 )
             proxy_manager = self.proxy_manager_for(proxy)
 
-            # Add Host to proxy header SNOW-232777
-            proxy_manager.proxy_headers["Host"] = parsed_url.hostname
+            if not isinstance(proxy_manager, SOCKSProxyManager):
+                # Add Host to proxy header SNOW-232777
+                proxy_manager.proxy_headers["Host"] = parsed_url.hostname
```
sfc-gh-aling, line 271: I may not have the best knowledge of proxy, but just to be defensive + general and get things work, I will probably go with:

```
try:
    proxy_manager.proxy_headers["Host"] = parsed_url.hostname
except AttributeError:
    # log that proxy_manager not having attribute proxy_headers
```

@sfc-gh-mkeller do you have any knowledge about this?

sfc-gh-aling: @stan-janssen , just want to know how you think about my approach?

stan-janssen: Sure, that would work!

sfc-gh-aling:
testing pr: https://github.com/snowflakedb/snowflake-connector-python/pull/1685/files

sfc-gh-aling (approved):

src/snowflake/connector/network.py
```diff
@@ -268,8 +268,16 @@ def get_connection(
                 )
             proxy_manager = self.proxy_manager_for(proxy)
 
-            # Add Host to proxy header SNOW-232777
-            proxy_manager.proxy_headers["Host"] = parsed_url.hostname
+            try:
```
sfc-gh-stan, line 271: If `SOCKSProxyManager` is the only case this fails, is it better if we just if-else instead of using a try catch?

src/snowflake/connector/network.py
```diff
@@ -268,8 +268,16 @@ def get_connection(
...
+                # Add Host to proxy header SNOW-232777
+                proxy_manager.proxy_headers["Host"] = parsed_url.hostname
+            except AttributeError:
+                # log that the type proxy_manager not having attribute proxy_headers
+                logger.debug(
+                    f"Unable to set 'Host' to proxy manager of type {type(proxy_manager)} as"
+                    f" it does not have attribute 'proxy_headers'."
+                )
```
sfc-gh-aalam, line 279: do you think `logger.warning` is a better log level for this? how common is this issue?

sfc-gh-aalam (approved):

sfc-gh-stan:
This also feels like a problem to be fixed in `requests` since `proxy_manager_for` indicates the return type should be `urllib3.ProxyManager`.

src/snowflake/connector/network.py
sfc-gh-aling: good question, the way I made the change is because `SOCKSProxyManager` is dynamically resolved:

https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/vendored/requests/adapters.py#L56-L61

```python
try:
    from ..urllib3.contrib.socks import SOCKSProxyManager
except ImportError:

    def SOCKSProxyManager(*args, **kwargs):
        raise InvalidSchema("Missing dependencies for SOCKS support.")
```

if `pysocks` is installed,  `SOCKSProxyManager` is a class, however if `pysocks` is not installed, `SOCKSProxyManager` is a function in which case we can not do the instance check `if instance(proxy_manager, SOCKSProxyManager)` which will raise `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union`.

so I avoid importing `SOCKSProxyManager`, otherwise I need to handle the complexity of importing errors.

src/snowflake/connector/network.py
sfc-gh-aling: so far it's only happening to socks5 proxy, and depending on the issue amount (only 1 in our repo). I don't think it's quite common.

depending on user description `I modified the code to not set the header when using a SOCKS5 proxy, which seems to work. The connection succeeds over the SOCKS5 proxy. It also works if I set the headers property instead of the proxy_headers property.` in the [issue](https://github.com/snowflakedb/snowflake-connector-python/issues/1550), not setting the proxy_header in socks5 proxy can also work.

so I treat this as an expected case and use the debug level.

but I'm open to update it to warning as we are coming from throwing error -- if this is a warning level message I will rephrase it as ` f"Unable to set 'Host' to proxy manager of type {type(proxy_manager)} as, it does not have attribute 'proxy_headers', proxy might not work as expected."`

src/snowflake/connector/network.py
sfc-gh-stan: What about checking whether the return type is an instance of `ProxyManager` and only set `proxy_headers` if so? I don't have a preference tho.

src/snowflake/connector/network.py
sfc-gh-aalam: sounds good.

src/snowflake/connector/network.py
sfc-gh-aling: yep I think this would work, code updated!

PR #1556, opened 2023-05-10
[SNOW-812398] Add ability to return raw response json from python-connector

sfc-gh-skumbham:
.

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -265,7 +265,7 @@ class SnowflakeCursor:
         self._inner_cursor: SnowflakeCursor | None = None
         self._prefetch_hook = None
         self._rownumber: int | None = None
-        self._get_last_request_json: dict | None = None
+        self._last_response_dict: dict | None = None
 
         self.reset()
 
@@ -374,17 +374,13 @@ class SnowflakeCursor:
         return None
 
     @property
-    def get_last_request_json(self) -> str:
+    def _last_response_json(self) -> str:
         """
         **** INTERNAL USE ONLY ******
         This method fetches the raw JSON response returned from the backend for the most recently executed call.
         WARNING: This method is primarily used for internal tools. Its implementation may be removed or changed at any time without notice.
         """
-        return json.dumps(self._get_last_request_json)
-
-    @get_last_request_json.setter
-    def get_last_request_json(self, value: dict[str, Any]) -> None:
-        self._get_last_request_json = value
+        return json.dumps(self._last_response_dict)
 
     @overload
     def callproc(self, procname: str) -> tuple:
@@ -557,7 +553,7 @@ class SnowflakeCursor:
                 self._timebomb.cancel()
                 logger.debug("cancelled timebomb in finally")
 
-            self._get_last_request_json = ret
+            self._last_response_dict = ret
 
         if "data" in ret and "parameters" in ret["data"]:
             parameters = ret["data"]["parameters"]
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -371,6 +373,19 @@ def lastrowid(self) -> None:
...
+    @property
+    def get_last_request_json(self) -> str:
+        """
+        **** INTERNAL USE ONLY ******
+        This method fetches the raw JSON response returned from the backend for the most recently executed call.
+        WARNING: This method is primarily used for internal tools. Its implementation may be removed or changed at any time without notice.
+        """
+        return json.dumps(self._get_last_request_json)
+
+    @get_last_request_json.setter
```
sfc-gh-bwarsaw, line 385: What's the setter use case?

sfc-gh-bwarsaw (changes requested):

src/snowflake/connector/cursor.py
```diff
@@ -371,6 +373,19 @@ def lastrowid(self) -> None:
...
         return None
 
+    @property
+    def get_last_request_json(self) -> str:
+        """
+        **** INTERNAL USE ONLY ******
+        This method fetches the raw JSON response returned from the backend for the most recently executed call.
+        WARNING: This method is primarily used for internal tools. Its implementation may be removed or changed at any time without notice.
+        """
+        return json.dumps(self._get_last_request_json)
```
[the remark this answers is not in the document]
sfc-gh-skumbham: Using caching in this scenario could add unnecessary complications. This is intended to be executed only once for each REST call initiated from the client bridge.

src/snowflake/connector/cursor.py
```diff
@@ -371,6 +373,19 @@ def lastrowid(self) -> None:
         """Snowflake does not support lastrowid in which case None should be returned as per PEP249."""
         return None
 
+    @property
+    def get_last_request_json(self) -> str:
```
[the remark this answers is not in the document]
sfc-gh-skumbham: oops. I'm still wearing my java hat so ended up creating getters and setters for a property. I will update it, thanks!

src/snowflake/connector/cursor.py
sfc-gh-skumbham: yea, not required. Removed, thanks!

src/snowflake/connector/cursor.py
```diff
@@ -264,6 +265,7 @@ def __init__(
         self._inner_cursor: SnowflakeCursor | None = None
         self._prefetch_hook = None
         self._rownumber: int | None = None
+        self._last_response_dict: dict | None = None
```
sfc-gh-bwarsaw, line 268: Apologies for thinking out loud again and any churn that might cause.

Generally, and especially with type hints, we don't include the type of the data structure the variable is bound to in the name of the variable.  Now that I'm looking at this again, what I think is interesting is that the property `_last_response_json()` actually returns a string.  In this case though you just need to make sure that the internal variable and the property name don't collide.

What do you think about calling this variable `self._cached_last_result` and the property `_last_result()`?

The other question I have is whether you want to access the last result as a string or as a dictionary.  In the test, you deserialize the string to check its contents.  Why not have the property return the cached last result dictionary directly, rather than serializing it to a string?

sfc-gh-bwarsaw (changes requested):

src/snowflake/connector/cursor.py
sfc-gh-skumbham: yeah, I do think `self._cached_last_result` and `_last_result` makes sense. Also, I have updated the return type to be a dictionary instead of JSON string.

sfc-gh-bwarsaw (approved):
Looks great!  Thanks!

sfc-gh-bwarsaw (approved):

PR #1562, opened 2023-05-15
SNOW-770678: Fix incorrect date conversion in bulk insertion

sfc-gh-aling:

   Fixes SNOW-770678

with the integer representation, the date can only be represented at maximal `31536000000 ` seconds which is `May 3, 2969 12:00:00 AM`. Any date beyond the date will be taken as milliseconds/nanoseconds. leading to incorrect date.

In the PR, we use the simple string form the represent the date

### DIFF
--- a/src/snowflake/connector/converter.py
+++ b/src/snowflake/connector/converter.py
@@ -101,10 +101,6 @@ def _convert_datetime_to_epoch_nanoseconds(dt: datetime) -> str:
     return f"{convert_datetime_to_epoch(dt):f}".replace(".", "") + "000"
 
 
-def _convert_date_to_epoch_milliseconds(dt: datetime) -> str:
-    return f"{(dt - ZERO_EPOCH_DATE).total_seconds():.3f}".replace(".", "")
-
-
 def _convert_time_to_epoch_nanoseconds(tm: dt_t) -> str:
     return (
         str(tm.hour * 3600 + tm.minute * 60 + tm.second)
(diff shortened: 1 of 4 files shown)


src/snowflake/connector/converter.py
```diff
@@ -366,8 +366,8 @@ def _nonetype_to_snowflake_bindings(self, *_) -> None:
         return None
 
     def _date_to_snowflake_bindings(self, _, value: date) -> str:
-        # milliseconds
-        return _convert_date_to_epoch_milliseconds(value)
+        # we are binding "TEXT" value for DATE, check function _adjust_bind_type
+        return value.strftime("%Y-%m-%d")
```
sfc-gh-hachouraria, line 370: minor suggestion: We could rely [on the existing `datetime.date#isoformat(…)` method](https://github.com/python/cpython/blob/023c51d9d8e2fd91069eea2bf12e373f1c71e9d2/Lib/datetime.py#L1046-L1055) which is more explicit about the widths used, just in case there's gotchas with early AD/late BC dates?

src/snowflake/connector/converter.py
```diff
@@ -366,8 +366,8 @@ def _nonetype_to_snowflake_bindings(self, *_) -> None:
         return None
 
     def _date_to_snowflake_bindings(self, _, value: date) -> str:
-        # milliseconds
-        return _convert_date_to_epoch_milliseconds(value)
```
sfc-gh-stan, line 370: If this is the only reference to `_convert_date_to_epoch_milliseconds`, could we remove it?

sfc-gh-stan (approved):

src/snowflake/connector/converter.py
sfc-gh-aling: good callout, I think we can remove it

src/snowflake/connector/converter.py
sfc-gh-aling: thanks for the suggestion, applied your suggestion

PR #1582, opened 2023-05-25
SNOW-799408: Close temp file handle before removing temp cache file

sfc-gh-aling:

   Fixes SNOW-799408

temp file is already an open file handle, instead of opening it again, we directly write the bytes into the file handle and then close the fd


test/extras/run.py
```diff
@@ -23,3 +25,17 @@
             ]
         )
         sub_process.check_returncode()
+        ocsp_cache_dir_path = pathlib.Path(
+            snowflake.connector.ocsp_snowflake.OCSP_RESPONSE_VALIDATION_CACHE.file_path
+        ).parent
+        cache_files = set(os.listdir(ocsp_cache_dir_path))
+        # This is to test SNOW-79940, making sure tmp files are removed
+        # Windows does not have ocsp_response_validation_cache.lock
```
sfc-gh-stan, line 33: Let's explicitly detect the platform and assert separately for windows in an if case.

src/snowflake/connector/cache.py
```diff
@@ -524,7 +524,16 @@ def _save(self, load_first: bool = True) -> bool:
...
-                    with open(tmp_file, "wb") as w_file:
+                    # tmp_file is an opened OS level handle, which means we need to close it manually.
+                    # https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp
+                    # ideally we shall just use the tmp_file fd to write,
+                    # however, using os.write(tmp_file, bytes) causes seg fault during garbage collection when exiting
+                    # python program.
+                    # thus we fall back to the approach using the normal open() method to open a file and write.
+                    os.close(tmp_file)
+                    with open(tmp_file_path, "wb") as w_file:
+                        # note: during garbage collection when exiting python program, open will raise not defined
```
sfc-gh-stan, line 535: I'm a bit confused here. What do you mean by "we depend on this behavior to exit the program gracefully"?

test/extras/run.py
sfc-gh-aling: smart, I will make the change

src/snowflake/connector/cache.py
```diff
@@ -524,7 +524,16 @@ def _save(self, load_first: bool = True) -> bool:
...
                     )
-                    with open(tmp_file, "wb") as w_file:
+                    # tmp_file is an opened OS level handle, which means we need to close it manually.
+                    # https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp
+                    # ideally we shall just use the tmp_file fd to write,
+                    # however, using os.write(tmp_file, bytes) causes seg fault during garbage collection when exiting
+                    # python program.
+                    # thus we fall back to the approach using the normal open() method to open a file and write.
+                    os.close(tmp_file)
+                    with open(tmp_file_path, "wb") as w_file:
```
sfc-gh-mkeller, line 534: I fail to see how this is different from the previous code. The docs [here](https://docs.python.org/3/library/functions.html#open) say:

> (If a file descriptor is given, it is closed when the returned I/O object is closed unless closefd is set to False.)

So to me it sounds like file descriptor, or path. The I/O object returned by `open` should behave identically.

sfc-gh-mkeller (changes requested):

src/snowflake/connector/cache.py
sfc-gh-aling: you are right about the usage of open.

the problem is when exiting the program, during garbage collection, the `open` method is garbage collected, so calling `with open` would raise error, so it didn't get the chance the close the underlying fd.

without `os.close`, the status of a fd during garbage collection is:
opened by mktmpfile -> error raised when calling `open` -> left opened

with `os.close`, the status:
opened -> closed by os.close ->   error raised when calling `open` -> left closed

sfc-gh-mkeller: If we want to protect against this specific scenario, what about doing:
```
try:
	with open(tmp_file, "wb") as w_file:
		...
except AppropiateException:
	# If file descriptor is still open close it
	try:
		os.fstat(tmp_file)
		os.close(tmp_file)
	except OSError:
		pass
```

PR #1617, opened 2023-06-26
SNOW-834366 Fall back to use current schema for temp objects in write…

sfc-gh-sfan:
…_pandas

Description

In 3.0.0 we made a change to use target schema for temp stage and file format, but this breaks user when they only have stage/file format creation privilege in current schema, but not the target schema.

This change adds a fallback mechanism to use current schema if creating temp object fails on target schema.

Testing

integ test

   Fixes SNOW-834366

### DIFF
--- a/src/snowflake/connector/pandas_tools.py
+++ b/src/snowflake/connector/pandas_tools.py
@@ -85,10 +85,11 @@ def _create_temp_stage(
     auto_create_table: bool,
     overwrite: bool,
 ) -> str:
+    stage_name = random_string()
     stage_location = build_location_helper(
         database=database,
         schema=schema,
-        name=random_string(),
+        name=stage_name,
         quote_identifiers=quote_identifiers,
     )
     try:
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/pandas_tools.py
```diff
@@ -62,6 +64,90 @@ def build_location_helper(
...
+    except ProgrammingError as e:
+        # User may not have the privilege to create stage on the target schema, so fall back to use current schema as
+        # the old behavior.
+        logger.debug(
+            f"creating stage {stage_location} failed. Exception {str(e)}. Fall back to use current schema"
+        )
+        stage_location = random_string()
+        _do_create_temp_stage(
+            cursor, stage_location, compression, auto_create_table, overwrite
+        )
```
sfc-gh-aalam, line 107: why are we creating temp stage again if we encountered an exception?

sfc-gh-aalam: ohh, I see. This will use `current_schema` instead of temp schema

sfc-gh-aalam (approved):

sfc-gh-aalam:
Would this change also need to go in sproc-connector as the issue which caused this came from stored procs?

sfc-gh-sfan:
> Would this change also need to go in sproc-connector as the issue which caused this came from stored procs?

Yes. I'm working on the sp connector changes :)

sfc-gh-jdu (approved):

PR #1618, opened 2023-06-27
SNOW-833713 snow sql performance issue with ocsp validation in 1 2 26

sfc-gh-aling:

   Fixes SNOW-833713

   - [x] I am modifying OCSP code

 - lazy evaluation of the module level ocsp cache variable, so that if the connector is in in_secure mode, the ocsp cache doesn't get initialized at all
 - reduce the times to dump cache to disks, if there's no change in the cache, we do not need to re-write file


dataders:
was this an issue surfaced as a result of #1582?

sfc-gh-aling:
@dataders no, they two are irrelevant.
#1582 is to clean up temp file on windows.
this PR tries to improve the OCSP cache performance.

src/snowflake/connector/cache.py
```diff
@@ -58,6 +58,9 @@ def __init__(
         self._cache: dict[K, CacheEntry[V]] = {}
         self._lock = Lock()
         self._reset_telemetry()
+        # indicate whether the cache is modified or not, this variable is for
+        # SFDictFileCache to determine whether to dump cache to file when _save is called
+        self._cache_modified = False
```
sfc-gh-mkeller, line 63: We should move all of these to `SFDictFileCache`, this class has no use for this information!

src/snowflake/connector/cache.py
```diff
@@ -494,7 +514,8 @@ def __getitem__(self, k: K) -> V:
 
     def _setitem(self, k: K, v: V) -> None:
         super()._setitem(k, v)
-        self._save_if_should()
+        if self._try_saving_when_set_item:
```
sfc-gh-mkeller, line 517: I think this check could be moved into `_save_if_should`

src/snowflake/connector/cache.py
```diff
@@ -511,9 +532,12 @@ def _load(self) -> bool:
             logger.debug("Fail to read cache from disk due to error: %s", e)
             return False
 
-    def _save(self, load_first: bool = True) -> bool:
+    def _save(self, load_first: bool = True, force_flush: bool = False) -> bool:
         """Save cache to disk if possible, returns whether it was able to save."""
         self._clear_expired_entries()
+        if not self._cache_modified and not force_flush:
+            # cache is not updated, so there is no need to dump cache to file, we just return
+            return True
```
sfc-gh-mkeller, line 540: It didn't save, that should be `False`
```suggestion
            return False
```

src/snowflake/connector/cache.py
```diff
@@ -572,6 +598,10 @@ def _save_if_should(self) -> bool:
             return self._save()
         return False
 
+    def save(self, load_first: bool = True):
```
sfc-gh-mkeller, line 601: I don't think that we should expose this

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -74,38 +74,50 @@ class OCSPResponseValidationResult(NamedTuple):
...
-    OCSP_RESPONSE_VALIDATION_CACHE: SFDictCache[
-        tuple[bytes, bytes, bytes],
-        OCSPResponseValidationResult,
-    ] = SFDictCache(
-        entry_lifetime=constants.DAY_IN_SECONDS,
-    )
+# use _get_ocsp_response_validation_cache instead of the module variable for lazy evaluation
+OCSP_RESPONSE_VALIDATION_CACHE: SFDictFileCache[
+    tuple[bytes, bytes, bytes], OCSPResponseValidationResult
+] | SFDictCache[tuple[bytes, bytes, bytes], OCSPResponseValidationResult] | None = None
```
sfc-gh-mkeller, line 80: Drop the `None` here and make it be the return value of `_get_ocsp_response_validation_cache`
```suggestion
] | SFDictCache[tuple[bytes, bytes, bytes], OCSPResponseValidationResult]  = _get_ocsp_response_validation_cache()
```
This way you can drop the types and push them onto the function's return type-hint as well.

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -74,38 +74,50 @@ class OCSPResponseValidationResult(NamedTuple):
...
+
+
+def _get_ocsp_response_validation_cache():
+    global OCSP_RESPONSE_VALIDATION_CACHE
```
sfc-gh-mkeller, line 84: This is an invisible side-effect of this function and usually a bad idea!

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -585,9 +599,10 @@ def update_file(ocsp) -> None:
             1. file for OCSP_RESPONSE_VALIDATION_CACHE which keeps ocsp response validation result
             2. ocsp_response_cache.json, the file in the same format as the one downloaded from snowflake cache service
         """
+        ocsp_response_validation_cache = _get_ocsp_response_validation_cache()
         if OCSPCache.CACHE_UPDATED:
-            if isinstance(OCSP_RESPONSE_VALIDATION_CACHE, SFDictFileCache):
```
sfc-gh-mkeller, line 589: The rest of this file should not change. You should be operating on the global cache every time. We resolve it once and then keep using that one!

test/extras/run.py
```diff
@@ -27,7 +27,7 @@
         )
         sub_process.check_returncode()
         ocsp_cache_dir_path = pathlib.Path(
-            snowflake.connector.ocsp_snowflake.OCSP_RESPONSE_VALIDATION_CACHE.file_path
+            snowflake.connector.ocsp_snowflake._get_ocsp_response_validation_cache().file_path
```
sfc-gh-mkeller, line 30: Shouldn't change

sfc-gh-mkeller (changes requested):

src/snowflake/connector/cache.py
sfc-gh-aling: `_save_if_should` is also used by `clear_expired_entries`, so I do want `_try_saving_when_set_item` to impact the behavior of `clear_expired_entries`. or we could introduce another parameter in `_save_if_should` to control the behavior

src/snowflake/connector/cache.py
sfc-gh-aling: yep, you're right!

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-aling: I want this OCSP_RESPONSE_VALIDATION_CACHE to be lazy instantiated. calling the function conflicts my purpose...

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-aling: oh, could you elaborate on this? I might be unfamiliar with the potential pitfalls here

src/snowflake/connector/cache.py
sfc-gh-aling: I will remove this

src/snowflake/connector/cache.py
sfc-gh-aling: updated

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-aling: I have revert the change to lazy initialization, by simply moving the import into the if condition in `ssl_wrap_socket`:
```python
    if FEATURE_OCSP_MODE != OCSPMode.INSECURE:
        from .ocsp_asn1crypto import SnowflakeOCSPAsn1Crypto as SFOCSP
```

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-aling: see comment above, change reverted

test/extras/run.py
sfc-gh-aling: see comment above, change reverted

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-aling: see comment above, change reverted

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -364,7 +365,8 @@ def download_cache_from_server(self, ocsp):
                 # len(OCSP_RESPONSE_VALIDATION_CACHE) is thread-safe, however, we do not want to
                 # block for logging purpose, thus using len(OCSP_RESPONSE_VALIDATION_CACHE._cache) here.
                 logger.debug(
-                    "# of certificates: %s", len(OCSP_RESPONSE_VALIDATION_CACHE._cache)
+                    "# of certificates: %s",
```
sfc-gh-mkeller, line 368: `%s` is for strings, `%i`, or `%u` might be more appropiate
```suggestion
                    "# of certificates: %u",
```

src/snowflake/connector/ocsp_snowflake.py
```diff
@@ -524,7 +526,8 @@ def reset_ocsp_response_cache_uri(ocsp_response_cache_uri) -> None:
         # len(OCSP_RESPONSE_VALIDATION_CACHE) is thread-safe, however, we do not want to
         # block for logging purpose, thus using len(OCSP_RESPONSE_VALIDATION_CACHE._cache) here.
         logger.debug(
-            "OCSP_VALIDATION_CACHE size: %s", len(OCSP_RESPONSE_VALIDATION_CACHE._cache)
+            "OCSP_VALIDATION_CACHE size: %s",
```
sfc-gh-mkeller, line 529: `%s` is for strings, `%i`, or `%u` might be more appropiate
```suggestion
            "OCSP_VALIDATION_CACHE size: %u",
```

test/integ/test_connection.py
```diff
@@ -1276,3 +1276,21 @@ def test_server_session_keep_alive(conn_cnx):
     with conn_cnx() as conn:
         conn.rest.delete_session = mock_delete_session
     mock_delete_session.assert_called_once()
+
+
+@pytest.mark.skipolddriver
+def test_ocsp_mode_in_secure(db_parameters, is_public_test, caplog):
```
sfc-gh-mkeller, line 1282: Use `conn_cnx` instead of `db_parameters`

sfc-gh-mkeller (changes requested):

src/snowflake/connector/ocsp_snowflake.py
sfc-gh-aling: updated

PR #1635, opened 2023-07-11
SNOW-856569 making save actions thread-safe

### LINKED ISSUE #1627: SNOW-856569: Segmentation Fault in cache.py pickle dump
### Python version

Python 3.8.12 (default, Nov 17 2021, 08:36:07)  [Clang 7.1.0 (tags/RELEASE_710/final)]

### Operating system and processor architecture

Linux _x86_64_

### Installed packages

```python
dbt-core==1.5.2
dbt-snowflake==1.5.2
snowflake-connector-python[secure-local-storage] == 3.0.3
```

### What did you do?

Running DBT on our linux machines with more than 1 thread triggers a segmentation fault in snowflake-connector-python for versions >= 3.x.x
We're not seeing this issue locally on MacOS.

### What did you expect to see?

The offending line is [here](https://github.com/snowflakedb/snowflake-connector-python/blob/v3.0.0/src/snowflake/connector/cache.py#L511C37-L511C37) and the truncated stack trace:
```
Fatal Python error: Segmentation fault
Stack (most recent call first):
File "pypi_snowflake_connector_python/site-packages/snowflake/connector/cache.py", line 511 in _save
```

sfc-gh-mkeller:

   Fixes #1627 

   In this PR I hopefully fix the underlying issue causing `pickle.dump` segfaults.
   The just of what I found is that calls calling `_save` did not hold `self._lock` which could lead to other threads modifying the data in the cache while dumping was going on and more issues were found later. I verified this with `assert` statements.
    I also updated all the unsafe call places that I could find. I left in the assert statements for now to catch other unsafe calls.
    I included a draw.io graph for how the different functions call each other, note that control flow goes left to right.
    [SFDictCache control flow.drawio.zip](https://github.com/snowflakedb/snowflake-connector-python/files/12089342/SFDictCache.control.flow.drawio.zip)

### DIFF
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -211,6 +211,11 @@ class SFDictCache(Generic[K, V]):
         other: dict[K, V] | list[tuple[K, V]] | SFDictCache[K, V],
         update_newer_only: bool = False,
     ) -> bool:
+        """Non-locking version of update.
+
+        This should only be used by internal functions when already
+        holding self._lock and other._lock.
+        """
         assert (
             self._lock.locked()
         ), "The mutex self._lock should be locked by this thread"
@@ -253,6 +258,7 @@ class SFDictCache(Generic[K, V]):
     def update(
         self,
         other: dict[K, V] | list[tuple[K, V]] | SFDictCache[K, V],
+        update_newer_only: bool = False,
     ) -> bool:
         """Insert multiple values at the same time, if self could learn from the other.
 
@@ -270,7 +276,7 @@ class SFDictCache(Generic[K, V]):
         into the other caches.
         """
         with self._lock:
-            return self._update(other)
+            return self._update(other, update_newer_only)
 
     def update_newer(
         self,
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -104,14 +104,8 @@ class SFDictCache(Generic[K, V]):
             self._hit(k)
         return v
 
-    def _getitem_non_locking(
-        self,
-        k: K,
-        *,
-        should_record_hits: bool = True,
-    ) -> V:
-        # aliasing _getitem to unify the api with SFDictFileCache
-        return self._getitem(k, should_record_hits=should_record_hits)
+    # aliasing _getitem to unify the api with SFDictFileCache
+    _getitem_non_locking = _getitem
 
     def _setitem(self, k: K, v: V) -> None:
         """Non-locking version of __setitem__.
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -75,8 +75,7 @@ class SFDictCache(Generic[K, V]):
         the dictionary provided.
         """
         cache = cls(**kw)
-        for k, v in _dict.items():
-            cache[k] = v
+        cache.update(_dict)
         return cache
 
     def _getitem(
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -229,10 +229,8 @@ class SFDictCache(Generic[K, V]):
                 g = iter(other.items())
             to_insert = {k: CacheEntry(expiry=expiry, entry=v) for k, v in g}
         elif isinstance(other, SFDictCache):
-            assert (
-                other._lock.locked()
-            ), "The mutex other._lock should be locked by this thread"
-            other._clear_expired_entries()
+            with other._lock:
+                other._clear_expired_entries()
             others_items = list(other._cache.items())
             # Only accept values from another cache if their key is not in self,
             #  or if expiry is later the self known one
@@ -490,7 +488,6 @@ class SFDictFileCache(SFDictCache):
         try:
             if k not in self._cache:
                 loaded = self._load_if_should()
-                currently_holding = True
                 if (not loaded) or k not in self._cache:
                     self._miss(k)
                     raise KeyError
@@ -527,10 +524,12 @@ class SFDictFileCache(SFDictCache):
             #  we have to know whether the file could learn anything from self
             #  so instead of calling self.update we call other.update and swap
             #  the 2 underlying caches after.
+            self._lock.release()
             cache_file_learnt = other.update(
                 self,
                 update_newer_only=True,
             )
+            self._lock.acquire()
             self._cache, other._cache = other._cache, self._cache
             self.telemetry.update(other.telemetry)
             self._cache_modified = cache_file_learnt
--- a/src/snowflake/connector/cache.py
+++ b/src/snowflake/connector/cache.py
@@ -536,6 +536,8 @@ class SFDictFileCache(SFDictCache):
             self._cache_modified = cache_file_learnt
             self.last_loaded = now()
             return True
+        except AssertionError:
+            raise
         except Exception as e:
             logger.debug("Fail to read cache from disk due to error: %s", e)
             return False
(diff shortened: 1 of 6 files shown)


src/snowflake/connector/cache.py
```diff
@@ -508,18 +520,44 @@ def _load(self) -> bool:
...
+            #  so instead of calling self.update we call other.update and swap
+            #  the 2 underlying caches after.
+            self._lock.release()
+            cache_file_learnt = other.update(
+                self,
                 update_newer_only=True,
             )
+            self._lock.acquire()
+            self._cache, other._cache = other._cache, self._cache
+            self.telemetry.update(other.telemetry)
```
sfc-gh-mkeller, line 534: We lose # of hits and misses

sfc-gh-stan (approved):

src/snowflake/connector/cache.py
```diff
@@ -507,19 +527,46 @@ def _load(self) -> bool:
...
             with open(self.file_path, "rb") as r_file:
-                other = pickle.load(r_file)
-            self._update(
-                other,
+                other: SFDictFileCache = pickle.load(r_file)
+            # Since we want to know whether we are dirty after loading
+            #  we have to know whether the file could learn anything from self
+            #  so instead of calling self.update we call other.update and swap
+            #  the 2 underlying caches after.
+            self._lock.release()
```
sfc-gh-aling, line 535: is it possible that `self._load` get called wit an unacquired lock?

src/snowflake/connector/cache.py
```diff
@@ -507,19 +527,46 @@ def _load(self) -> bool:
...
+            cache_file_learnt = other.update(
+                self,
                 update_newer_only=True,
             )
+            self._lock.acquire()
```
sfc-gh-aling, line 540: I can understand that the lock release then reacquire here is because other.update will lock the lock.

I forget why we didn't choose to use a re-entrant lock? those lock operations make me a bit concerned of deadlock. I wonder if RLock could simplify the lock logic here

src/snowflake/connector/cache.py
```diff
@@ -219,7 +236,8 @@ def _update(
                 g = iter(other.items())
             to_insert = {k: CacheEntry(expiry=expiry, entry=v) for k, v in g}
         elif isinstance(other, SFDictCache):
-            other._clear_expired_entries()
+            with other._lock:
+                other._clear_expired_entries()
```
sfc-gh-aling, line 240: just call the public api shall be good?
```suggestion
            other.clear_expired_entries()
```

src/snowflake/connector/cache.py
sfc-gh-mkeller: `RLock` is slower than `Lock` is all.

src/snowflake/connector/cache.py
sfc-gh-mkeller: If that happened an Exception would be raised
> When invoked on an unlocked lock, a [RuntimeError](https://docs.python.org/3/library/exceptions.html#RuntimeError) is raised.
from https://docs.python.org/3/library/threading.html#threading.Lock

src/snowflake/connector/cache.py
```diff
@@ -624,17 +652,24 @@ def __del__(self) -> None:
...
+        with self._lock:
+            self._save(load_first=False, force_flush=True)
+
+    def _update(
+        self,
+        other: dict[K, V] | list[tuple[K, V]] | SFDictCache[K, V],
+        update_newer_only: bool = False,
+    ) -> bool:
+        updated = super()._update(other, update_newer_only=update_newer_only)
+        self._save_if_should()
```
sfc-gh-aling, line 671: I think this `_save_if_should` introduces a new cycle?

test/unit/test_cache.py
```diff
@@ -461,55 +488,54 @@ def test_load_with_expired_entries(self, tmpdir):
...
+        assert not c1.save()
         assert not os.path.exists(cache_path)
 
-        c2 = cache.SFDictFileCache(file_path=cache_path)
-        assert not c2._load()  # load should return false due to no file
+        c2 = NeverSaveSFDictFileCache(file_path=cache_path)
+        assert not c2.load()  # load should return false due to no file
 
 
 def test_file_is_not_updated(tmpdir):
```
sfc-gh-aling, line 515: I think this test becomes flaky because now `update` will try saving

test/unit/test_cache.py
```diff
@@ -461,81 +474,80 @@ def test_load_with_expired_entries(self, tmpdir):
...
     tmp_cache_file = os.path.join(tmpdir, "tmp_cache")
-    sfcache = cache.SFDictFileCache(file_path=tmp_cache_file, entry_lifetime=1)
+    sfcache = NeverSaveSFDictFileCache(file_path=tmp_cache_file, entry_lifetime=1)
```
sfc-gh-aling, line 503: I think we can revert the change here, the purpose of this test is to test that when calling `update`, sfcache doesn't try to save the cache.

test/unit/test_cache.py
```diff
@@ -461,81 +474,80 @@ def test_load_with_expired_entries(self, tmpdir):
...
-    cache3 = cache.SFDictFileCache(file_path=tmp_cache_file, entry_lifetime=1)
+    cache3 = NeverSaveSFDictFileCache(file_path=tmp_cache_file, entry_lifetime=1)
     assert len(cache3) == 0
     os.unlink(tmp_cache_file)
 
 
 def test_cache_do_not_write_while_set_item(tmpdir):
     tmp_cache_file = os.path.join(tmpdir, "tmp_cache")
-    sfcache = cache.SFDictFileCache(tmp_cache_file)
+    sfcache = NeverSaveSFDictFileCache(tmp_cache_file)
```
sfc-gh-aling, line 535: same revert

test/unit/test_cache.py
```diff
@@ -461,81 +474,80 @@ def test_load_with_expired_entries(self, tmpdir):
...
     # there should be no file created as setting item won't trigger creation
     assert sfcache._cache_modified
     assert not os.path.exists(tmp_cache_file)
-    sfcache._save()
+    sfcache.save()
     assert not sfcache._cache_modified
     assert os.path.exists(tmp_cache_file)
     file_modified_time = os.path.getmtime(tmp_cache_file)
-    sfcache2 = cache.SFDictFileCache(tmp_cache_file)
+    sfcache2 = NeverSaveSFDictFileCache(tmp_cache_file)
```
sfc-gh-aling, line 547: same revert

sfc-gh-aling (approved):

test/unit/test_cache.py
sfc-gh-mkeller: I would like to leave it like this, as this makes the test deterministic instead of leaving things up to Python's random number generator

src/snowflake/connector/cache.py
```diff
@@ -219,7 +236,7 @@ def _update(
                 g = iter(other.items())
             to_insert = {k: CacheEntry(expiry=expiry, entry=v) for k, v in g}
         elif isinstance(other, SFDictCache):
-            other._clear_expired_entries()
+            other.clear_expired_entries()
```
sfc-gh-aalam, line 239: here, `self._lock` is locked (line 227), then why are we not using `other._clear_expired_entries()` here? `clear_expired_entries` tries to acquire lock again

sfc-gh-aalam: nvm, other has a different lock

PR #1646, opened 2023-07-17
SNOW-862821 [Python Connector] Update query context cache response body to align with JDBC

sfc-gh-lqin:

   Fixes #SNOW-862821

   When I was helping ODBC team to debug their query context related issue. I found that both ODBC and python connector have an inconsistent query context JSON serialization behavior for HTTP response. To be precise:
   
   1. The json field name should be `queryContextDTO` instead of `queryContext` to keep consistent with JDBC/GS.
   2. The entire `queryContextDTO` field is also not a string, it should be an object (dict in python).
   3. The `context` subfield is not a string, but an `OpaqueContextDTO` object in JDBC, and this object contains a string field which is the context base64 encoded value. 
`JSON: {"sqlText":"insert into t1 values (1, 2), (2, 3), (3, 4)","sequenceId":10,"bindings":{},"bindStage":null,"describeOnly":false,"parameters":{"NEW_SQL_FORMAT":false,"CLIENT_RESULT_CHUNK_SIZE":96},"queryContextDTO":{"entries":[{"id":0,"timestamp":6645635168985,"priority":0,"context":{}}]},"describedJobId":"01ad95f6-0000-0154-0000-0004011a4be2","querySubmissionTime":1689200068021,"isInternal":false,"asyncExec":false} `
    for example, here when context is empty, the value is {}. for string type context, the empty context will be serialized to "". This could cause GS side unable to convert the JSON back to java query context object.
    
    
For references:
1. JDBC https://github.com/snowflakedb/snowflake-jdbc/blob/9d625df285cc9896f3d14c6e1e6eed80cdd30d1b/src/main/java/net/snowflake/client/core/QueryContextDTO.java#L8
2. GS: https://github.com/snowflakedb/snowflake/blob/70d3da634cc30f7a653bf6e6463c63155e7dc786/GlobalServices/src/main/java/com/snowflake/api/monitoring/QueryContextDTO.java#L12

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1111,7 +1111,6 @@ class SnowflakeConnection:
             queryContext = self.get_query_context()
             if(queryContext is not None):
                 data["queryContextDTO"] = queryContext
-            print(data["queryContextDTO"])
         client = "sfsql_file_transfer" if is_file_transfer else "sfsql"
 
         if logger.getEffectiveLevel() <= logging.DEBUG:
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1107,8 +1107,11 @@ class SnowflakeConnection:
             # binding parameters. This is for qmarks paramstyle.
             data["bindings"] = binding_params
         if not _no_results:
-            # not an async query
-            data["queryContext"] = self.get_query_context()
+            # not an async query. Here queryContextDTO should be a dict object field, same with `parameters` field
+            queryContext = self.get_query_context()
+            if(queryContext is not None):
+                data["queryContextDTO"] = queryContext
+            print(data["queryContextDTO"])
         client = "sfsql_file_transfer" if is_file_transfer else "sfsql"
 
         if logger.getEffectiveLevel() <= logging.DEBUG:
@@ -1647,10 +1650,10 @@ class SnowflakeConnection:
         if not self.is_query_context_cache_disabled:
             self.query_context_cache = QueryContextCache(self.query_context_cache_size)
 
-    def get_query_context(self) -> str | None:
+    def get_query_context(self) -> dict | None:
         if self.is_query_context_cache_disabled:
             return None
-        return self.query_context_cache.serialize_to_json()
+        return self.query_context_cache.serialize_to_dict()
 
     def set_query_context(self, data: dict) -> None:
         if not self.is_query_context_cache_disabled:
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1107,9 +1107,10 @@ class SnowflakeConnection:
             # binding parameters. This is for qmarks paramstyle.
             data["bindings"] = binding_params
         if not _no_results:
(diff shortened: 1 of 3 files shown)


test/unit/test_query_context_cache.py
```diff
@@ -51,6 +51,35 @@ def reset_data(self):
         self.priorities = [BASE_PRIORITY + i for i in range(self.capacity)]
 
 
+# This function is used for testing the deserialize_to_dict function is correct. Note that `QueryContextCache::serialize_to_dict` function is used in sending the query context back to GS.
+# It has slight difference with this function on `context` field.
+def serialize_to_json(qcc: QueryContextCache) -> str:
```
sfc-gh-aalam, line 56:
```suggestion
def serialize_to_json(qcc: QueryContextCache) -> str:
"""This function is used for testing the deserialize_to_dict function is correct. Note that `QueryContextCache::serialize_to_dict` function is used in sending the query context back to GS.

It has slight difference with this function on `context` field.
"""
```

src/snowflake/connector/connection.py
```diff
@@ -1107,8 +1107,10 @@ def cmd_query(
             # binding parameters. This is for qmarks paramstyle.
             data["bindings"] = binding_params
         if not _no_results:
-            # not an async query
-            data["queryContext"] = self.get_query_context()
+            # not an async query. Here queryContextDTO should be a dict object field, same with `parameters` field
+            queryContext = self.get_query_context()
+            if queryContext is not None:
+                data["queryContextDTO"] = queryContext
```
sfc-gh-aalam, line 1113: this change means that some responses to GS will not contain `queryContextDTO` field. Is that okay?

sfc-gh-aalam, line 1113:
```suggestion
            # not an async query.
            queryContext = self.get_query_context()
            if queryContext is not None:
                #  Here queryContextDTO should be a dict object field, same with `parameters` field
                data["queryContextDTO"] = queryContext
```

sfc-gh-aalam (commented):
Minor comments. Overall change looks good to me

test/unit/test_query_context_cache.py
```diff
@@ -51,6 +51,35 @@ def reset_data(self):
...
+                    }
+                    for qce in qcc._tree_set
+                ]
+            }
+            # Serialize the data to JSON
+            serialized_data = json.dumps(data)
+
+            return serialized_data
+        except Exception as e:
+            return ""
```
sfc-gh-aalam, line 80:
```suggestion
        except Exception:
            return ""
```
This is causing lint issues. check with this change

sfc-gh-lqin: Yeah. This should be removed since `e` is not used here.

src/snowflake/connector/connection.py
sfc-gh-lqin: I think they are equivalent when GS convert JSON back to the query context object. But I restored to the original version to avoid unnecessary changes. I have tested using my e2e queries and their query context entries look good.

sfc-gh-aalam (approved):

PR #1649, opened 2023-07-18
SNOW-870322 Allow pass type_mapper to arrow to_pandas()

sfc-gh-azhan:

   SNOW-870322 Allow pass type_mapper to arrow to_pandas()

This PR allows fetch_pandas_batches() and fetch_pandas_all() to bypass kwargs to arrow.to_pandas(), e.g., we can pass type_mapper to it.

### DIFF
--- a/src/snowflake/connector/result_set.py
+++ b/src/snowflake/connector/result_set.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import inspect
 from collections import deque
 from concurrent.futures import Future
 from concurrent.futures.thread import ThreadPoolExecutor
@@ -184,17 +185,20 @@ class ResultSet(Iterable[list]):
 
     def _fetch_pandas_all(self, **kwargs) -> DataFrame:
         """Fetches a single Pandas dataframe."""
-        dataframes = list(self._fetch_pandas_batches(**kwargs))
+        concat_args = list(inspect.signature(pandas.concat).parameters)
+        concat_kwargs = {k: kwargs.pop(k) for k in dict(kwargs) if k in concat_args}
+        to_pandas_kwargs = dict(kwargs.items() - concat_kwargs.items())
+        dataframes = list(self._fetch_pandas_batches(**to_pandas_kwargs))
         if dataframes:
-            concat_kwargs = kwargs
-            concat_kwargs.pop("types_mapper", None)
+            # concat_kwargs = kwargs
+            # concat_kwargs.pop("types_mapper", None)
             return pandas.concat(
                 dataframes,
                 ignore_index=True,  # Don't keep in result batch indexes
                 **concat_kwargs,
             )
         # Empty dataframe
-        return self.batches[0].to_pandas(**kwargs)
+        return self.batches[0].to_pandas(**to_pandas_kwargs)
 
     def _get_metrics(self) -> dict[str, int]:
         """Sum up all the chunks' metrics and show them together."""
--- a/src/snowflake/connector/result_set.py
+++ b/src/snowflake/connector/result_set.py
@@ -178,19 +178,23 @@ class ResultSet(Iterable[list]):
         Snowflake's back-end.
         """
         self._can_create_arrow_iter()
-        return self._create_iter(iter_unit=IterUnit.TABLE_UNIT, structure="pandas")
+        return self._create_iter(
+            iter_unit=IterUnit.TABLE_UNIT, structure="pandas", **kwargs
+        )
 
     def _fetch_pandas_all(self, **kwargs) -> DataFrame:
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/result_set.py
```diff
@@ -178,19 +178,23 @@ def _fetch_pandas_batches(self, **kwargs) -> Iterator[DataFrame]:
...
         """Fetches a single Pandas dataframe."""
-        dataframes = list(self._fetch_pandas_batches())
+        dataframes = list(self._fetch_pandas_batches(**kwargs))
         if dataframes:
+            concat_kwargs = kwargs
+            concat_kwargs.pop("types_mapper")
             return pandas.concat(
                 dataframes,
                 ignore_index=True,  # Don't keep in result batch indexes
-                **kwargs,
```
sfc-gh-jdu, line 190: You should be able to pass in **kwargs because non-used arguments will be ignored

sfc-gh-jdu (approved):

src/snowflake/connector/result_set.py
sfc-gh-azhan: It will throw a VauleError unfortunately.

src/snowflake/connector/result_set.py
```diff
@@ -178,19 +178,23 @@ def _fetch_pandas_batches(self, **kwargs) -> Iterator[DataFrame]:
...
+            iter_unit=IterUnit.TABLE_UNIT, structure="pandas", **kwargs
+        )
 
     def _fetch_pandas_all(self, **kwargs) -> DataFrame:
         """Fetches a single Pandas dataframe."""
-        dataframes = list(self._fetch_pandas_batches())
+        dataframes = list(self._fetch_pandas_batches(**kwargs))
         if dataframes:
+            concat_kwargs = kwargs
+            concat_kwargs.pop("types_mapper", None)
```
sfc-gh-mkeller, line 190: I think this is going to be error-prone. What if we added `concat_kwargs` differently?
```
    def _fetch_pandas_all(self, concat_kwargs: dict[str, Any], **kwargs) -> DataFrame:
        ...
            **concat_kwargs,
        )
    	# Empty dataframe
    	return self.batches[0].to_pandas(**kwargs)
```
Just be careful, I think `concat_kwargs` were the original kwargs so the API shouldn't change, so maybe add `to_pandas_kwargs` in this manner.

sfc-gh-mkeller (changes requested):

src/snowflake/connector/result_set.py
sfc-gh-azhan: Can you take a look the new approach? I tried to avoid changing the customer facing API but also make sure the kwargs will be delivered to the right methods.

src/snowflake/connector/result_set.py
```diff
@@ -178,19 +179,26 @@ def _fetch_pandas_batches(self, **kwargs) -> Iterator[DataFrame]:
...
     def _fetch_pandas_all(self, **kwargs) -> DataFrame:
         """Fetches a single Pandas dataframe."""
-        dataframes = list(self._fetch_pandas_batches())
+        concat_args = list(inspect.signature(pandas.concat).parameters)
+        concat_kwargs = {k: kwargs.pop(k) for k in dict(kwargs) if k in concat_args}
+        to_pandas_kwargs = dict(kwargs.items() - concat_kwargs.items())
+        dataframes = list(self._fetch_pandas_batches(**to_pandas_kwargs))
         if dataframes:
+            # concat_kwargs = kwargs
+            # concat_kwargs.pop("types_mapper", None)
```
sfc-gh-mkeller, line 194:
```suggestion
```

src/snowflake/connector/result_set.py
```diff
@@ -178,19 +179,26 @@ def _fetch_pandas_batches(self, **kwargs) -> Iterator[DataFrame]:
...
         self._can_create_arrow_iter()
-        return self._create_iter(iter_unit=IterUnit.TABLE_UNIT, structure="pandas")
+        return self._create_iter(
+            iter_unit=IterUnit.TABLE_UNIT, structure="pandas", **kwargs
+        )
 
     def _fetch_pandas_all(self, **kwargs) -> DataFrame:
         """Fetches a single Pandas dataframe."""
-        dataframes = list(self._fetch_pandas_batches())
+        concat_args = list(inspect.signature(pandas.concat).parameters)
```
sfc-gh-mkeller, line 188: I think we could get away with something simpler, but this will work. Hopefully `pandas.concat` will never introduce an argument that conflicts with our `to_pandas` function

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1692, opened 2023-08-10
SNOW-874210 Fix localfs chunked upload/download

sfc-gh-sfan:
Description

wb means open the file in write mode with truncation. What we
actually want is rb+

Testing

Tested locally

   Fixes SNOW-874210

### DIFF
--- a/src/snowflake/connector/local_storage_client.py
+++ b/src/snowflake/connector/local_storage_client.py
@@ -29,10 +29,11 @@ class SnowflakeLocalStorageClient(SnowflakeStorageClient):
     ) -> None:
         super().__init__(meta, stage_info, chunk_size)
         self.data_file = meta.src_file_name
-        self.stage_file_name: str = os.path.join(
+        self.full_dst_file_name: str = os.path.join(
             stage_info["location"], os.path.basename(meta.dst_file_name)
         )
         if meta.local_location:
+            self.stage_file_name = self.full_dst_file_name
             self.full_dst_file_name = os.path.join(
                 meta.local_location, meta.dst_file_name
             )
@@ -72,11 +73,11 @@ class SnowflakeLocalStorageClient(SnowflakeStorageClient):
 
     def prepare_upload(self) -> None:
         super().prepare_upload()
-        with open(self.stage_file_name, "wb+") as fd:
+        with open(self.full_dst_file_name, "wb+") as fd:
             fd.truncate(self.meta.upload_size)
 
     def _upload_chunk(self, chunk_id: int, chunk: bytes) -> None:
-        with open(self.stage_file_name, "rb+") as tfd:
+        with open(self.full_dst_file_name, "rb+") as tfd:
             tfd.seek(chunk_id * self.chunk_size)
             tfd.write(chunk)
 
--- a/src/snowflake/connector/local_storage_client.py
+++ b/src/snowflake/connector/local_storage_client.py
@@ -6,11 +6,11 @@
 from __future__ import annotations
 
 import os
+import shutil
 from logging import getLogger
-from math import ceil
 from typing import TYPE_CHECKING, Any
 
-from .constants import ResultStatus
+from .constants import FileHeader, ResultStatus
 from .storage_client import SnowflakeStorageClient
 from .vendored import requests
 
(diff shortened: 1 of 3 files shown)


sfc-gh-sfan:
I'm curious how to test this

src/snowflake/connector/local_storage_client.py
```diff
@@ -28,7 +28,7 @@ def __init__(
         chunk_size: int,
         use_s3_regional_url: bool = False,
     ) -> None:
-        super().__init__(meta, stage_info, chunk_size)
+        super().__init__(meta, stage_info, chunk_size, chunked_transfer=False)
```
sfc-gh-aalam, line 31: how is this new parameter used?

src/snowflake/connector/local_storage_client.py
```diff
@@ -55,7 +55,7 @@ def download_chunk(self, chunk_id: int) -> None:
                 os.path.join(
                     self.meta.local_location, os.path.basename(self.meta.dst_file_name)
                 ),
-                "wb",
```
sfc-gh-stan, line 58: Is this because we wanna append instead of truncate?

src/snowflake/connector/local_storage_client.py
sfc-gh-sfan: Oh sorry this should be removed :(

src/snowflake/connector/local_storage_client.py
sfc-gh-sfan: Not really append. We want to open the file for update. See: https://docs.python.org/3/library/functions.html#open

> The default mode is 'r' (open for reading text, a synonym of 'rt'). Modes 'w+' and 'w+b' open and truncate the file. Modes 'r+' and 'r+b' open the file with no truncation.

sfc-gh-sfan: We are using `rb+` for other storage client in https://github.com/snowflakedb/snowflake-connector-python/blob/ea8d733752b2efbcbe712fe4cd9a47cce0f88776/src/snowflake/connector/storage_client.py#L354

sfc-gh-sfan:
I'll try to see if I could come up with a unit test

sfc-gh-sfan:
While unit testing I found the local fs PUT/GET is almost completely broken. Made quite some changes here to make it work.

src/snowflake/connector/local_storage_client.py
```diff
@@ -26,42 +26,45 @@ def __init__(
...
+                    self.meta.local_location,
+                    os.path.basename(self.intermediate_dst_path),
                 ),
-                "wb",
+                "rb+",
             ) as tfd:
-                tfd.seek(chunk_id * self.chunk_size)
-                sfd.seek(chunk_id * self.chunk_size)
-                tfd.write(sfd.read(self.chunk_size))
+                if self.num_of_chunks == 1:
```
sfc-gh-stan, line 59: 🤔 Why do we need a special case here?

sfc-gh-stan (approved):
Thank you for fixing this 🎉 ! I had 2 questions, otherwise LGTM.

test/unit/test_local_storage_client.py
```diff
@@ -0,0 +1,89 @@
...
+
+import pytest
+
+try:
+    from snowflake.connector.constants import LOCAL_FS
+except ImportError:
+    LOCAL_FS = None
+
+from snowflake.connector.file_transfer_agent import SnowflakeFileMeta
+from snowflake.connector.local_storage_client import SnowflakeLocalStorageClient
```
sfc-gh-mkeller, line 19: These should move into the try-clause above too

sfc-gh-mkeller (approved):
Good by me once you address comments!

:ship: :ship:

src/snowflake/connector/local_storage_client.py
sfc-gh-sfan: Ha this is kinda interesting because `num_of_chunks == 1` could mean: 

- We use chunked upload/download and there is only one chunk
- We don't use chunked upload/download

The second case is a bit tricky here because if we decided to not use chunked upload/download because of `multipart_threshold`, and the total data size is bigger than one chunk, then we will miss data.

sfc-gh-sfan:
> Question: Does localfs PUT/GET not use compression/encryption?

I don't think we support encryption, but we do support compression in local fs. I tried to put a compressed file to stage, and unzip the `.gz` stage file. The shasum is the same so it is probably working :)

sfc-gh-stan:
> > Question: Does localfs PUT/GET not use compression/encryption?
> 
> I don't think we support encryption, but we do support compression in local fs. I tried to put a compressed file to stage, and unzip the `.gz` stage file. The shasum is the same so it is probably working :)

I see, I was wondering if we need to call `super().finish_download`, which is needed for decryption. This makes sense then.

PR #1721, opened 2023-09-01
SNOW-891470 adding implementation for default_connection_name

sfc-gh-mkeller:

   Fixes SNOW-891470

   Now that we have flushed out how the configuration file option `default_connection_name` should work I add actual support for it.

### DIFF
--- a/src/snowflake/connector/__init__.py
+++ b/src/snowflake/connector/__init__.py
@@ -7,6 +7,8 @@
 #
 from __future__ import annotations
 
+from functools import wraps
+
 apilevel = "2.0"
 threadsafety = 2
 paramstyle = "pyformat"
@@ -47,6 +49,7 @@ from .version import VERSION
 logging.getLogger(__name__).addHandler(NullHandler())
 
 
+@wraps(SnowflakeConnection.__init__)
 def Connect(**kwargs) -> SnowflakeConnection:
     return SnowflakeConnection(**kwargs)
 
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -302,6 +302,21 @@ class SnowflakeConnection:
         connections_file_path: pathlib.Path | None = None,
         **kwargs,
     ) -> None:
+        """Create a new SnowflakeConnection.
+
+        Connections can be loaded from the TOML file located at
+        snowflake.connector.constants.CONNECTIONS_FILE.
+
+        When connection_name is supplied we will first load that connection
+        and then overwrite any other values supplied.
+
+        When no arguments are given (other than connection_file_path) the
+        default connection will be loaded first. Note that no overwriting is
+        supported in this case.
+
+        If overwriting values from the default connection is desirable, supply
+        the name explicitly.
+        """
         self._lock_sequence_counter = Lock()
         self.sequence_counter = 0
         self._errorhandler = Error.default_errorhandler
@@ -324,6 +339,7 @@ class SnowflakeConnection:
             setattr(self, f"_{name}", value)
 
         self.heartbeat_thread = None
+        empty_kwargs = not bool(kwargs)
 
         if "application" not in kwargs:
             if ENV_VAR_PARTNER in os.environ.keys():
@@ -349,6 +365,17 @@ class SnowflakeConnection:
                     f" known ones are {list(connections.keys())}"
                 )
             kwargs = {**connections[connection_name], **kwargs}
+        elif empty_kwargs:
+            # connection_name is None and kwargs was empty when called
+            def_connection_name = CONFIG_MANAGER["default_connection_name"]
+            connections = CONFIG_MANAGER["connections"]
+            if def_connection_name not in connections:
+                raise Error(
+                    f"Default connection with name '{def_connection_name}' "
+                    "cannot be found, known ones are "
+                    f"{list(connections.keys())}"
+                )
+            kwargs = {**connections[def_connection_name]}
         self.__set_error_attributes()
         self.connect(**kwargs)
         self._telemetry = TelemetryClient(self._rest)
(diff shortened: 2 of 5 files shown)


sfc-gh-jdu (approved):
Thanks!

src/snowflake/connector/__init__.py
```diff
@@ -47,6 +49,7 @@
 logging.getLogger(__name__).addHandler(NullHandler())
 
 
+@wraps(SnowflakeConnection.__init__)
```
sfc-gh-yixie, line 52: For my education, what is this for?

src/snowflake/connector/connection.py
```diff
@@ -349,6 +365,17 @@ def __init__(
                     f" known ones are {list(connections.keys())}"
                 )
             kwargs = {**connections[connection_name], **kwargs}
+        elif empty_kwargs:
+            # connection_name is None and kwargs was empty when called
+            def_connection_name = CONFIG_MANAGER["default_connection_name"]
```
sfc-gh-yixie, line 370: Do we expect "default_connection_name" a must in the config file? If no, this may raise an KeyError. Suggest we raise a meaningful error message to users.

src/snowflake/connector/connection.py
```diff
@@ -349,6 +365,17 @@ def __init__(
...
+            connections = CONFIG_MANAGER["connections"]
```
sfc-gh-yixie, line 371: Similarly, if I happen to have a config file that has no "connections" section, this will raise a KeyError. A meaningful error is better.

src/snowflake/connector/__init__.py
sfc-gh-mkeller: It copies docstrings and other things to the annotated function https://docs.python.org/3/library/functools.html#functools.wraps

src/snowflake/connector/connection.py
```diff
@@ -302,6 +302,21 @@ def __init__(
         connections_file_path: pathlib.Path | None = None,
         **kwargs,
     ) -> None:
+        """Create a new SnowflakeConnection.
+
+        Connections can be loaded from the TOML file located at
+        snowflake.connector.constants.CONNECTIONS_FILE.
+
+        When connection_name is supplied we will first load that connection
+        and then overwrite any other values supplied.
```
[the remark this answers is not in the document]
sfc-gh-mkeller: How do you suggest we add code example here?

src/snowflake/connector/connection.py
sfc-gh-mkeller: It's not, it has a default value

src/snowflake/connector/connection.py
sfc-gh-mkeller: Okay, I tried this out manually. I found 2 bugs that made things not work how I thought they should.
When I remove my `connections.toml` file I now get:
```
  File "/tmp/autoconnect.py", line 3, in <module>
    with connect() as conn:
         ^^^^^^^^^
  File "/Users/mkeller/snowflakedb/snowflake-connector-python/src/snowflake/connector/__init__.py", line 54, in Connect
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mkeller/snowflakedb/snowflake-connector-python/src/snowflake/connector/connection.py", line 373, in __init__
    raise Error(
snowflake.connector.errors.Error: Default connection with name 'mkeller_sfctest0_aws' cannot be found, known ones are []
```
I also assigned a default value to `connections`, so hopefully that will cover all the use-cases and give better error messages.

src/snowflake/connector/connection.py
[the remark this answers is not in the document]
sfc-gh-bwarsaw: Maybe "override" is a better term?

src/snowflake/connector/connection.py
```diff
@@ -324,6 +339,7 @@ def __init__(
             setattr(self, f"_{name}", value)
 
         self.heartbeat_thread = None
+        empty_kwargs = not bool(kwargs)
```
sfc-gh-bwarsaw, line 342: What's the benefit of calling `bool()` here given that `not some_dict` will always give you a boolean?  Also, I personally like using names like `is_kwargs_empty` for a more descriptive variable name.

src/snowflake/connector/connection.py
sfc-gh-bwarsaw: Just to be clear, it's not possible for this to ever be a `KeyError`?

src/snowflake/connector/connection.py
```diff
@@ -349,6 +365,17 @@ def __init__(
...
             kwargs = {**connections[connection_name], **kwargs}
+        elif empty_kwargs:
+            # connection_name is None and kwargs was empty when called
+            def_connection_name = CONFIG_MANAGER["default_connection_name"]
+            connections = CONFIG_MANAGER["connections"]
+            if def_connection_name not in connections:
+                raise Error(
+                    f"Default connection with name '{def_connection_name}' "
+                    "cannot be found, known ones are "
+                    f"{list(connections.keys())}"
```
sfc-gh-bwarsaw, line 376: You don't need `.keys()` here since iteration over a dictionary is always by key by default.  Also, I'm not sure the list view will be super helpful.  Yeah, okay, it's just in an error message, but something like this might read better:

```python
COMMASPACE = ', '
...
f"{COMMASPACE.join(connections)}"
```

src/snowflake/connector/connection.py
sfc-gh-mkeller: You know that I always prefer explicit over implicit, I will change it for you though

src/snowflake/connector/connection.py
sfc-gh-mkeller: Sorry, that is exactly how I leaked the testing account password last time, when the error a few lines above was introduced. On some Python version and OS what you suggest also inserted the values and not only the keys!

src/snowflake/connector/connection.py
sfc-gh-mkeller: Based on my testing it's not

sfc-gh-yixie (approved):

PR #1759, opened 2023-10-09
Snow 916662 design retry timeout config in python connector

sfc-gh-yqiu:

   Changes implementing [SNOW-912528: Python Connector retry & timeout policy configuration](https://snowflakecomputing.atlassian.net/browse/SNOW-912528)

   [Lucidchart for new retry & timeout behaviour
](https://lucid.app/lucidspark/4e9f472c-f9c7-47a4-a01a-89d30f3e01d3/edit?viewport_loc=-3775%2C-767%2C5294%2C3045%2C0_0&invitationId=inv_a4c3556c-b815-4512-8afb-1d30b046b92a)

   [Design document](https://docs.google.com/document/d/16Bxmjpj0t7GpfJKG3DnkhlC_39qwzpnO3V7B1ZK7JDk/edit?usp=sharing)


src/snowflake/connector/auth/keypair.py
```diff
@@ -193,21 +190,24 @@ def handle_timeout(
...
+            delete_params=False,
+        )
+
+        logger.debug("Base timeout handler passed, checking for JWT expiry")
+        jwt_timed_out = False
+        if self._retry_ctx.remaining_time_millis(jwt_timed_out) is not None:
+            jwt_timed_out = self._retry_ctx.remaining_time_millis(jwt_timed_out) < 0
 
-        self.prepare(account=account, user=user)
+        if jwt_timed_out:
```
sfc-gh-aling, line 208: dumb question: it looks like previously we refresh token regardless of the jwt timed out or now, but now we only refresh is jwt timed out.

will there be a case that jwt token not timed out, but treated as invalid by the server?

src/snowflake/connector/auth/keypair.py
```diff
@@ -193,21 +190,24 @@ def handle_timeout(
...
+            account=account,
+            user=user,
+            password=password,
+            delete_params=False,
+        )
+
+        logger.debug("Base timeout handler passed, checking for JWT expiry")
+        jwt_timed_out = False
+        if self._retry_ctx.remaining_time_millis(jwt_timed_out) is not None:
+            jwt_timed_out = self._retry_ctx.remaining_time_millis(jwt_timed_out) < 0
```
sfc-gh-aling, line 206: you can do it in one line:
```suggestion
        jwt_timed_out = False if self._retry_ctx.remaining_time_millis(jwt_timed_out) is None else self._retry_ctx.remaining_time_millis(jwt_timed_out) < 0
```

src/snowflake/connector/connection.py
```diff
@@ -145,11 +146,13 @@ def DefaultConverterClass() -> type:
     "schema": (None, (type(None), str)),  # snowflake
     "role": (None, (type(None), str)),  # snowflake
     "session_id": (None, (type(None), str)),  # snowflake
-    "login_timeout": (120, int),  # login timeout
+    "login_timeout": (None, (type(None), int)),  # login timeout
```
sfc-gh-aling, line 149: what will be default timeout if we change this to None?

src/snowflake/connector/connection.py
```diff
@@ -275,8 +278,17 @@ class SnowflakeConnection:
...
         role: Role in use on Snowflake.
-        login_timeout: Login timeout in seconds. Used while authenticating.
-        network_timeout: Network timeout. Used for general purpose.
+        login_timeout: Login timeout in seconds. Login requests will not be retried after this timeout expires.
+            Note that the login attempt may still take more than login_timeout seconds as an ongoing login request
+            cannot be canceled even upon login timeout expiry. The login timeout only prevents further retries.
+        network_timeout: Network timeout in seconds. Network requests besides login requests will not be retried
+            after this timeout expires. Overriden in cursor query execution if timeout is passed to cursor.execute.
+            Note that an operation may still take more than network_timeout seconds for the same reason as above.
+        socket_timeout: Socket timeout in seconds. Sets both socket connect and read timeout.
```
sfc-gh-aling, line 287: I'm wondering do we really need to introduce socket_timeout. it feels to me that `network_timeout` ~= `socket_timeout` conceptually

src/snowflake/connector/connection.py
```diff
@@ -1233,6 +1288,8 @@ def _authenticate(self, auth_instance: AuthByPlugin):
         )
 
         auth = Auth(self.rest)
+        # record start time for computing timeout
+        auth_instance._retry_ctx.set_start_time()
```
sfc-gh-aling, line 1292: how do we record the start time in the previous implementation?
also if there is a `set_start_time`, will there be a `unset` method to be called?

src/snowflake/connector/network.py
```diff
@@ -247,7 +248,7 @@ def raise_failed_request_error(
 
 
 def is_login_request(url: str) -> bool:
-    return "/session/v1/login-request" in url
+    return "login-request" in parse_url(url).path
```
sfc-gh-aling, line 251: may I ask why this is changed?

src/snowflake/connector/network.py
```diff
@@ -477,9 +478,6 @@ def request(
         else:
             accept_type = CONTENT_TYPE_APPLICATION_JSON
 
-        if timeout is None:
-            timeout = self._connection.network_timeout
-
```
sfc-gh-aling, line 482: from the type hint timeout can still be none, so are we not specifying timeout anymore when it's a `ProxySupportAdapter`?

src/snowflake/connector/network.py
```diff
@@ -689,7 +691,6 @@ def _get_request(
             headers,
             timeout=timeout,
             token=token,
-            socket_timeout=socket_timeout,
```
sfc-gh-aling, line 692: we no longer set socket timeout in `_get_request`?

src/snowflake/connector/network.py
```diff
@@ -793,47 +792,29 @@ def fetch(
...
-                """
-                self.cnt += 1
-                self.reason = reason
-
-            def next_sleep(self) -> int:
-                self.sleeping_time = self._backoff.next_sleep(
-                    self.cnt, self.sleeping_time
-                )
-                return self.sleeping_time
+            def set_retry_reason(self, reason: int = 0):
```
sfc-gh-aling, line 807: I will go with python property `@setter/@getter` instead of method

src/snowflake/connector/time_util.py
```diff
@@ -94,3 +152,86 @@ def get_timing_millis(self) -> int:
                 "Trying to get timing before TimerContextManager has finished"
             )
         return self._end - self._start
+
+
+class TimeoutBackoffCtx:
+    """Base context for handling timeouts and backoffs on retries"""
+
+    DEFAULT_BACKOFF_POLICY = RecursiveMixedBackoff
```
sfc-gh-aling, line 160: the default behavior'd better be documented in the `connection` module

src/snowflake/connector/time_util.py
```diff
@@ -94,3 +152,86 @@ def get_timing_millis(self) -> int:
...
+
+    @property
+    def current_retry_count(self) -> int:
+        return int(self._current_retry_count)
+
+    @property
+    def current_sleep_time(self) -> int:
+        return int(self._current_sleep_time)
+
+    def set_start_time(self) -> None:
```
sfc-gh-aling, line 196: I prefer a property `setter` approach

src/snowflake/connector/time_util.py
```diff
@@ -94,3 +152,86 @@ def get_timing_millis(self) -> int:
...
+        self._start_time_millis = get_time_millis()
+
+    def remaining_time_millis(self, timeout: int | None) -> int | None:
```
sfc-gh-aling, line 199: this can be a getter

sfc-gh-aling (commented):
amazing work, thanks, left some comments, let me know if you have any questions

src/snowflake/connector/auth/_auth.py
```diff
@@ -166,7 +168,7 @@ def authenticate(
         mfa_callback: Callable[[], None] | None = None,
         password_callback: Callable[[], str] | None = None,
         session_parameters: dict[Any, Any] | None = None,
-        timeout: int = 120,
+        mfa_timeout: int | None = None,  # max time waiting for MFA response
```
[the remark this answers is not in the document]
sfc-gh-yqiu: `Auth.authenticate` does not seem to be exposed to the public. The documentation shows that [authentication is only done through the connector](https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api). Furthermore, even internally, the `timeout` argument to `Auth.authenticate` is never used.

I changed the name because this argument only affects the `join` timeout while waiting for MFA authentication, so I found `timeout` to be a strange name as it implies something more general.

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -197,6 +180,7 @@ def handle_timeout(
         account: str,
         user: str,
         password: str,
+        delete_params: bool = True,
```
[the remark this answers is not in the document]
sfc-gh-yqiu: Because subclasses of `AuthByPlugin` (in my case `AuthByKeyPair`) may not want to delete `authenticator, service_name, account, user, password` while handling timeout. I added this so I could call `super().handle_timeout` in `AuthByKeyPair.handle_timeout` while preserving the original function's behaviour.

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -207,20 +191,22 @@ def handle_timeout(
...
 
-        del authenticator, service_name, account, user, password
+        if delete_params:
+            del authenticator, service_name, account, user, password
+
         logger.debug("Default timeout handler invoked for authenticator")
         if not self._retry_ctx.should_retry():
             error = OperationalError(
-                msg=f"Could not connect to Snowflake backend after {self._retry_ctx.get_current_retry_count()} attempt(s)."
+                msg=f"Could not connect to Snowflake backend after {self._retry_ctx.current_retry_count + 1} attempt(s)."
```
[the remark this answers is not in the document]
sfc-gh-yqiu: As is, `current_retry_count` tracks the number of retries that has already happened, so total attempts is the number of retries +1. I could fiddle around with `retry_ctx` to make it so that we don't need this +1 here if necessary.

A note here is that the old code (without the +1) was simply wrong. After the first failed connection attempt it would say "could not connect... after 0 attempts".

However, we can also change the message to `Could not connect to Snowflake backend after X retrie(s)` to get rid of the +1, which is much easier.

src/snowflake/connector/errorcode.py
```diff
@@ -16,6 +16,7 @@
 ER_UNABLE_TO_OPEN_BROWSER = 250008
 ER_UNABLE_TO_START_WEBSERVER = 250009
 ER_INVALID_CERTIFICATE = 250011  # not used but keep here to reserve errno
+ER_INVALID_BACKOFF_POLICY = 250012
```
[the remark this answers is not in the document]
sfc-gh-yqiu: I've searched for these error codes throughout confluence and on public documentation, and I haven't been able to find anything. I also noticed some errors like:

```
ER_NOT_SUPPORT_DATA_TYPE = 255001
ER_NO_PYARROW = 255002
ER_NO_ARROW_RESULT = 255003
ER_NO_PYARROW_SNOWSQL = 255004
ER_FAILED_TO_READ_ARROW_STREAM = 255005
ER_NO_NUMPY = 255006
```

Which seem Python specific. Also, errors like `ER_NO_ACCOUNT_NAME` will never reach the server or come back from the server since it is just a client side check:

image

So I figured I could add some codes of my own since none of the existing ones really make sense for my purposes.

src/snowflake/connector/auth/keypair.py
```diff
@@ -57,19 +54,20 @@ def __init__(
...
         self._lifetime = timedelta(
             seconds=int(os.getenv("JWT_LIFETIME_IN_SECONDS", lifetime_in_seconds))
         )
-        self._jwt_retry_attempts = int(
-            os.getenv(
-                "JWT_CNXN_RETRY_ATTEMPTS", AuthByKeyPair.DEFAULT_JWT_RETRY_ATTEMPTS
-            )
-        )
-        self._timeout = int(
+        self._jwt_timeout = int(
```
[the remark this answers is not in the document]
sfc-gh-yqiu: It is supposed to be used here:

`jwt_timed_out = self._retry_ctx.remaining_time_millis(self._jwt_timeout) < 0`

But I made a typo, so it was:

`jwt_timed_out = self._retry_ctx.remaining_time_millis(jwt_timed_out) < 0`

Which wasn't caught by the test because `jwt_timed_out = False` and `False * <any integer> = 0`... Thanks for catching this!

src/snowflake/connector/auth/keypair.py
```diff
@@ -193,21 +190,24 @@ def handle_timeout(
...
-            )
-            self._retry_ctx.increment_retry()
+        logger.debug("Invoking base timeout handler")
+        super().handle_timeout(
+            authenticator=authenticator,
+            service_name=service_name,
+            account=account,
+            user=user,
+            password=password,
+            delete_params=False,
```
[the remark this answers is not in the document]
sfc-gh-yqiu: `AuthByKeyPair` is the only auth class that defined its own `handle_timeout` while all others previously called `AuthByPlugin.handle_timeout`. That means that all others were previously deleting params, while `AuthByKeyPair` wasn't. This is just to preserve that behaviour.

src/snowflake/connector/auth/keypair.py
sfc-gh-yqiu: To answer this question directly:

> will there be a case that jwt token not timed out, but treated as invalid by the server?

Almost certainly not. Harsh's notes say:

> This timeout is currently set to 10 sec which is considerably less than the GS JWT timeout of 60 seconds.

So the overall goal, both previously and after my changes, is to make sure that an `authenticate` call isn't made without refreshing the token if `JWT_CNXN_WAIT_TIME` seconds has passed.

[See the code here.](https://github.com/snowflakedb/snowflake-connector-python/blob/4777d83f8cac79194f890b34e50162e35478ee00/src/snowflake/connector/auth/_auth.py#L246) Previously, this was achieved by setting all relevant timeouts to `auth_instance.timeout = JWT_CNXN_WAIT_TIME` (slight simplification). This is basically trying to cap the time of every `authenticate` call by capping the time taken in the post request; this does not work fully, as a wait of `socket_timeout=auth_timeout` seconds can happen multiple times at the socket level (a similar issue remains now and should be documented). So, previously, if `handle_timeout` is called, the token has probably timed out anyways - it's just implicit.

A potential problem with the current implementation is that because socket timeout is decoupled from authenticator timeout, users could set socket timeout to much higher than `JWT_CNXN_WAIT_TIME`. We can either document this or cap the socket timeout at `JWT_CNXN_WAIT_TIME` (though this might be confusing if users can't figure out why socket timeout isn't what they set). Also, consider the example:

- each `authenticate` request takes 11 seconds
- `JWT_CNXN_WAIT_TIME = 10`

Then on every other `authenticate` retry that takes 10 seconds, there is technically only 1 second remaining on the token lifetime from the client's POV, and the request will spend 9 seconds past `JWT_CNXN_WAIT_TIME` without renewing the token. Practically, because 10 is much less than 60 seconds, this is a non issue.

src/snowflake/connector/connection.py
sfc-gh-yqiu: In `by_plugin` I defined `DEFAULT_AUTH_CLASS_TIMEOUT = 120`. I though it made the arguments to `connect` a bit more consistent.

src/snowflake/connector/connection.py
sfc-gh-yqiu: I think there are some differences but I am not confident. Let's discuss this on Monday?

src/snowflake/connector/connection.py
sfc-gh-yqiu: [Start time was previously set like this.](https://github.com/snowflakedb/snowflake-connector-python/blob/4777d83f8cac79194f890b34e50162e35478ee00/src/snowflake/connector/network.py#L887)

I added a reset on `start_time_millis` in the `reset` method of the context.

src/snowflake/connector/network.py
sfc-gh-yqiu: [I just used Harsh's implementation.](https://github.com/snowflakedb/snowflake-connector-python/blob/4777d83f8cac79194f890b34e50162e35478ee00/src/snowflake/connector/network.py#L1162) I though it looked a little more robust.

src/snowflake/connector/network.py
sfc-gh-yqiu: I catch this in `fetch`:

```
timeout = (
    timeout if timeout is not None else self._connection.network_timeout
)
```

src/snowflake/connector/network.py
sfc-gh-yqiu: It is never used before, and the previously default socket timeout gets set below in the call stack anyways.

src/snowflake/connector/time_util.py
sfc-gh-yqiu: I am not sure about this as `_start_time_millis` is used internally only, and people using this class should not be able to set it to arbitrary values.

src/snowflake/connector/time_util.py
sfc-gh-yqiu: `remaining_time_millis` outputs the remaining time given an arbitrary timeout value, relative to the context's start time. I don't think getters take arguments so that would break some code.

src/snowflake/connector/auth/keypair.py
EachOneChew: Update: added an override so that authenticators can set socket_timeout if they have any specific requirements.

DESCRIPTION.md
```diff
@@ -14,6 +14,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Fixed issue with connection diagnostics failing to complete certificate checks.
   - Fixed issue that arrow iterator causes `ImportError` when the c extensions are not compiled.
   - Removed dependencies on Cryptodome and oscrypto and removed the `use_openssl_only` parameter. All connections now go through OpenSSL via the cryptography library, which was already a dependency.
+  - Added the `backoff` argument to `snowflake.connector.connect` allowing for configurable backoff policy between retries of failed requests. See available implementations in the `time_util` module.
+  - Added the `socket_timeout` argument to `snowflake.connector.connect` specifying socket read and connect timeout.
+  - Fixed `login_timeout` and `network_timeout` (renamed to `request_timeout` for clarity) behaviour. Retries of login and network requests are now properly halted after these timeouts expire.
```
sfc-gh-mkeller, line 19: This will need docs change https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#functions

src/snowflake/connector/auth/_auth.py
```diff
@@ -166,10 +168,14 @@ def authenticate(
         mfa_callback: Callable[[], None] | None = None,
         password_callback: Callable[[], str] | None = None,
         session_parameters: dict[Any, Any] | None = None,
-        timeout: int = 120,
+        timeout: int
+        | None = None,  # max time waiting for MFA response, currently unused
```
sfc-gh-mkeller, line 172:
```suggestion
        timeout: int | None = None,  # max time waiting for MFA response, currently unused
```

src/snowflake/connector/auth/_auth.py
```diff
@@ -127,8 +127,9 @@ def base_auth_data(
         internal_application_name,
         internal_application_version,
         ocsp_mode,
-        login_timeout,
-        network_timeout=None,
+        login_timeout=None,
+        request_timeout=None,
+        socket_timeout=None,
```
sfc-gh-mkeller, line 132: Type-hint at least your new arguments, but if you could do all of them that'd be great

src/snowflake/connector/time_util.py
```diff
@@ -47,17 +50,72 @@ def get_time_millis() -> int:
...
+    DEFAULT_BACKOFF_FACTOR = 2
+    DEFAULT_ENABLE_JITTER = True
+
+    def __init__(
+        self,
+        base: int = DEFAULT_BACKOFF_BASE,
+        cap: int = DEFAULT_BACKOFF_CAP,
+        factor: int = DEFAULT_BACKOFF_FACTOR,
+        enable_jitter: bool = DEFAULT_ENABLE_JITTER,
+    ):
```
sfc-gh-mkeller, line 65:
```suggestion
    ) -> None:
```

src/snowflake/connector/time_util.py
```diff
@@ -47,17 +50,72 @@ def get_time_millis() -> int:
     return int(time.time() * 1000)
 
 
-class DecorrelateJitterBackoff:
-    # Decorrelate Jitter backoff
-    # https://www.awsarchitectureblog.com/2015/03/backoff.html
-    def __init__(self, base: int, cap: int) -> None:
+class BackoffPolicy(ABC):
```
sfc-gh-mkeller, line 53: I still think that we should at least explore using Generators instead of classes. There are so few `super` calls that I'm not sure if using OOP here is warranted

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,37 @@ class AuthType(Enum):
...
+        timeout: int | None = None,
+        max_retry_attempts: int | None = None,
+        socket_timeout_override: int | None = None,
+        backoff_policy: BackoffPolicy | None = None,
+    ) -> None:
         self.consent_cache_id_token = False
-        self._timeout: int = 120
+
+        if timeout is None:
+            timeout = DEFAULT_AUTH_CLASS_TIMEOUT
```
sfc-gh-mkeller, line 72: Then we could just default to this instead of `None`

src/snowflake/connector/connection.py
sfc-gh-mkeller: Please post what you agreed on

src/snowflake/connector/network.py
```diff
@@ -715,9 +714,9 @@ def _post_request(
         body,
         token=None,
         timeout=None,
+        socket_timeout=None,
```
sfc-gh-mkeller, line 717: Type-hint

src/snowflake/connector/network.py
```diff
@@ -793,47 +792,34 @@ def fetch(
...
-            def increment_retry(self, reason: int = 0) -> None:
-                """Increment retry count and update the reason for retry
 
-                Args:
-                    reason: HTTP response code that caused retry. Defaults to 0.
-                """
-                self.cnt += 1
-                self.reason = reason
+            @property
+            def retry_reason(self) -> int:
```
sfc-gh-mkeller, line 808: If these setters and getters don't implement anything special then could we just use  `self.retry_reason` instead of `self._reason`

src/snowflake/connector/time_util.py
```diff
@@ -47,17 +50,72 @@ def get_time_millis() -> int:
...
+        cap: int = DEFAULT_BACKOFF_CAP,
+        factor: int = DEFAULT_BACKOFF_FACTOR,
+        enable_jitter: bool = DEFAULT_ENABLE_JITTER,
+    ):
+        """Initialize a Backoff
+        backoff_base: Integer constant term used in backoff computations. Usage depends on implementation.
+        backoff_factor: Integer constant term used in backoff computations. Usage depends on implementation.
+        backoff_cap: Maximum backoff time in integer seconds.
+        backoff_enable_jitter: Boolean specifying whether to enable randomized jitter on computed backoff times.
+        """
```
sfc-gh-mkeller, line 71: We use the Google doc-string format, please adhere to it

src/snowflake/connector/time_util.py
```diff
@@ -94,3 +152,82 @@ def get_timing_millis(self) -> int:
...
+    @property
+    def current_sleep_time(self) -> int:
+        return int(self._current_sleep_time)
+
+    def set_start_time(self) -> None:
+        self._start_time_millis = get_time_millis()
+
+    def remaining_time_millis(self, timeout: int | None) -> int | None:
+        if timeout is None or self._start_time_millis is None:
+            return None
```
sfc-gh-mkeller, line 195: Maybe an Exception would be better here? I think it would be an issue if we misused this class

src/snowflake/connector/time_util.py
sfc-gh-mkeller, line 1: Make sure that `mypy` checks this file

src/snowflake/connector/time_util.py
```diff
@@ -94,3 +152,82 @@ def get_timing_millis(self) -> int:
...
+        """Updates retry count and sleep time for another retry"""
+        self._current_retry_count += 1
+        if self._backoff_policy is not None:
+            self._current_sleep_time = self._backoff_policy.next_sleep(
+                self._current_retry_count, self._current_sleep_time
+            )
+        logger.debug(f"Update retry count to {self._current_retry_count}")
+        logger.debug(f"Update sleep time to {self._current_sleep_time} seconds")
+
+    def reset(self) -> None:
```
sfc-gh-mkeller, line 230: This should be called form `__init__` too

test/unit/mock_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    connection._request_timeout = request_timeout
+    connection.request_timeout = request_timeout
+
+    connection._socket_timeout = socket_timeout
+    connection.socket_timeout = socket_timeout
+
+    connection._backoff_policy = backoff_policy
+    connection.backoff_policy = backoff_policy
+
+    return connection
```
sfc-gh-mkeller, line 34:
```suggestion
    return MagicMock(
        _login_timeout = login_timeout
        login_timeout = login_timeout
        _request_timeout = request_timeout
        request_timeout = request_timeout

        _socket_timeout = socket_timeout
        socket_timeout = socket_timeout

        _backoff_policy = backoff_policy
        backoff_policy = backoff_policy
    )
```

test/unit/mock_utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+def mock_request_with_action(next_action, sleep=None):
+    def mock_request(*args, **kwargs):
+        if sleep is not None:
+            time.sleep(sleep)
+        if next_action == "RETRY":
+            response = MagicMock()
+            response.status_code = 503
+            response.close = lambda: None
+            return response
```
sfc-gh-mkeller, line 45:
```suggestion
            return MagicMock(
                status_code = 503
                close = lambda: None
            )
```

test/unit/test_retry_network.py
```diff
@@ -412,3 +426,53 @@ def error_recv_into(*args, **kwargs):
...
+            method="post",
+            full_url="https://testaccount.snowflakecomputing.com/session/v1/login-request",
+            headers=dict(),
+        )
+
+
+@pytest.mark.flaky(reruns=3)
+@pytest.mark.parametrize(
+    "next_action_result",
+    (("RETRY", ServiceUnavailableError), ("ERROR", OperationalError)),
```
sfc-gh-mkeller, line 453:
```suggestion
    (
        ("RETRY", ServiceUnavailableError),
        ("ERROR", OperationalError),
    ),
```

test/unit/test_retry_network.py
```diff
@@ -412,3 +426,53 @@ def error_recv_into(*args, **kwargs):
...
+
+    with pytest.raises(OperationalError):
+        rest.fetch(
+            method="post",
+            full_url="https://testaccount.snowflakecomputing.com/session/v1/login-request",
+            headers=dict(),
+        )
+
+
+@pytest.mark.flaky(reruns=3)
```
sfc-gh-mkeller, line 450: Why is a unit test flaky? Could you make it not flaky, please?

sfc-gh-mkeller (changes requested):

src/snowflake/connector/auth/_auth.py
EachOneChew: Added type hints for my own, but I'm not entirely sure about all the existing ones.

src/snowflake/connector/auth/_auth.py
EachOneChew: Precommit checks keep reformatting that line automatically D:<

src/snowflake/connector/time_util.py
EachOneChew: I've been learning a lot about generators and coroutines while learning about asyncio. I can definitely look into working in generators then (right now) if that sounds good?

src/snowflake/connector/auth/by_plugin.py
EachOneChew: The problem is that the argument might be `None`, so if we default to `DEFAULT_AUTH_CLASS_TIMEOUT` then `timeout = None` will overwrite it.

test/unit/test_retry_network.py
EachOneChew: I made it flaky because very occasionally, a test on the server side would fail but succeed on retry. Maybe because of thread scheduling? In hindsight though, I see why we wouldn't want to mark unit tests like these flaky.

src/snowflake/connector/connection.py
sfc-gh-yqiu: We retained `socket_timeout` but agreed that `network_timeout` and `socket_timeout` sounded too similar. Decided that `request_timeout` was more clear than `network_timeout`. Rebase overwrote my edit to the documentation string, but I've fixed that.

src/snowflake/connector/time_util.py
sfc-gh-yqiu: Makes sense, made the edit.

DESCRIPTION.md
sfc-gh-yqiu: You mentioned there was someone I should talk to for docs changes.

src/snowflake/connector/time_util.py
sfc-gh-mkeller: You cannot change the public API once this gets released!

DESCRIPTION.md
sfc-gh-mkeller: Just pinged you about this privately

sfc-gh-mkeller (approved):
Please make sure that the docs get updated and all the other outstanding comments from the others get resolved. About the generators I sent you a calendar invite, let's decide on them at once 👍 

:ship: :ship:

src/snowflake/connector/backoff_policies.py
```diff
@@ -0,0 +1,145 @@
...
+DEFAULT_BACKOFF_CAP = 16
+DEFAULT_ENABLE_JITTER = True
+
+
+def mixed_backoff(
+    factor: int = DEFAULT_BACKOFF_FACTOR,
+    base: int = DEFAULT_BACKOFF_BASE,
+    cap: int = DEFAULT_BACKOFF_CAP,
+    enable_jitter: bool = DEFAULT_ENABLE_JITTER,
+) -> Callable[..., Iterator[int]]:
```
sfc-gh-mkeller, line 45: `time.sleep` takes a [float](https://github.com/python/typeshed/blob/ba3f23b2040927cb278dcb4d5ea5f8c43eb5c8b3/stdlib/time.pyi#L75). Maybe these should all return floats

sfc-gh-mkeller (approved):
:ship: :ship:

sfc-gh-yqiu:
[Sent this to @sfc-gh-mheavin](https://snowflakecomputing.atlassian.net/browse/SNOW-948813) for documentation purposes.

src/snowflake/connector/connection.py
```diff
@@ -475,8 +488,16 @@ def login_timeout(self) -> int | None:
         return int(self._login_timeout) if self._login_timeout is not None else None
 
     @property
-    def network_timeout(self) -> int | None:
-        return int(self._network_timeout) if self._network_timeout is not None else None
+    def request_timeout(self) -> int | None:
+        return int(self._request_timeout) if self._request_timeout is not None else None
```
sfc-gh-aling, line 492: well, changing the property name is a BCR change.
honestly I don't think changing the name of a property is worth a major version bump
so we probably should just keep the `network_timeout`..
sorry I didn't realize you were about to to rename the property and should call it out sooner

DESCRIPTION.md
```diff
@@ -12,6 +12,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
   - Added support for `use_logical_type` in `write_pandas`.
   - Removed dependencies on pycryptodomex and oscrypto. All connections now go through OpenSSL via the cryptography library, which was already a dependency.
+  - Added the `backoff` argument to `snowflake.connector.connect` allowing for configurable backoff policy between retries of failed requests. See available implementations in the `time_util` module.
+  - Added the `socket_timeout` argument to `snowflake.connector.connect` specifying socket read and connect timeout.
+  - Fixed `login_timeout` and `network_timeout` (renamed to `request_timeout` for clarity) behaviour. Retries of login and network requests are now properly halted after these timeouts expire.
```
sfc-gh-aling, line 17: this renaming is a BCR, see my comment below

DESCRIPTION.md
```diff
@@ -12,6 +12,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
   - Added support for `use_logical_type` in `write_pandas`.
   - Removed dependencies on pycryptodomex and oscrypto. All connections now go through OpenSSL via the cryptography library, which was already a dependency.
+  - Added the `backoff` argument to `snowflake.connector.connect` allowing for configurable backoff policy between retries of failed requests. See available implementations in the `time_util` module.
```
sfc-gh-aling, line 15: I only see `TimeoutBackoffCtx` in the `time_util`, I assume this is the base class, are we missing other implementations in the module?

also I think we'd better call it out explicitly what's the implementations here because `time_util` has some other classes, module variables, methods, etc.

Ideally we can say, available implementations of the backoff policy are:
- snowflake.connector.time_util.XXX
- snowflake.connector.time_util.YYY

this gives us 2 benefits:
-  being clear on what can be used by users
- only the classes called out the public apis, we leave us some room for future change for other apis in the module

src/snowflake/connector/auth/_auth.py
```diff
@@ -239,20 +246,12 @@ def authenticate(
             {k: v for (k, v) in body["data"].items() if k != "PASSWORD"},
         )
 
-        # accommodate any authenticator specific timeout requirements here.
-        # login_timeout comes from user configuration.
-        # Between login timeout and auth specific
-        # timeout use whichever value is smaller
-        auth_timeout = min(self._rest._connection.login_timeout, auth_instance.timeout)
-        logger.debug(f"Timeout set to {auth_timeout}")
```
sfc-gh-aling, line 247: we can add back the debug log for `timeout`, in the code change which should be `auth_instance.socket_timeout_override` I think?

src/snowflake/connector/auth/_auth.py
```diff
@@ -239,20 +246,12 @@ def authenticate(
...
-
         try:
             ret = self._rest._post_request(
                 url,
                 headers,
                 json.dumps(body),
-                timeout=auth_timeout,
-                socket_timeout=auth_timeout,
+                socket_timeout=auth_instance.socket_timeout_override,
```
sfc-gh-aling, line 254: nit: this `_override` seems a bit verbose, do you think we can just name is `auth_instance.socket_timeout`?

src/snowflake/connector/auth/_auth.py
```diff
@@ -307,6 +309,7 @@ def post_request_wrapper(self, url, headers, body) -> None:
                 while not self.ret or self.ret.get("message") == "Timeout":
                     next(c)
             else:
+                # no need to set a timeout on join as _post_request will terminate on timeout
```
sfc-gh-aling, line 312: I would rather change the comment to something like "timeout safeguard", otherwise the comment seems to conflict the implementation of next line

src/snowflake/connector/auth/_auth.py
```diff
@@ -166,10 +168,14 @@ def authenticate(
         mfa_callback: Callable[[], None] | None = None,
         password_callback: Callable[[], str] | None = None,
         session_parameters: dict[Any, Any] | None = None,
-        timeout: int = 120,
+        # max time waiting for MFA response, currently unused
```
sfc-gh-aling, line 171: this comment is a bit confusing, can you point me to where it's used in the MFA reponse?

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,37 @@ class AuthType(Enum):
...
+        )
+
+        # some authenticators may want to override socket level timeout
+        # for example, AuthByKeyPair will do this to ensure JWT tokens are refreshed in time
+        self._socket_timeout_override = socket_timeout_override
 
     @property
-    def timeout(self) -> int:
-        return self._timeout
+    def timeout(self) -> int | None:
```
sfc-gh-aling, line 87: this is also a BCR, previously the assumption is that `timeout` is always returning an integer, but now it can return None.

code like `AuthByPlugin.timeout + 10` will break now when `AuthByPlugin.timeout` returns None

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,37 @@ class AuthType(Enum):
...
+        return self._retry_ctx.timeout
 
-    @timeout.setter
-    def timeout(self, value: Any) -> None:
-        self._timeout = int(value)
+    @property
+    def socket_timeout_override(self) -> int | None:
```
sfc-gh-aling, line 91: I think we can remove the `_override`, and add a comment in the docstring explaining that this will override the socket_timeout.

also do we need a corresponding setting for the property?

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -197,6 +186,7 @@ def handle_timeout(
         account: str,
         user: str,
         password: str,
+        delete_params: bool = True,
```
sfc-gh-aling, line 189: I'm thinking of hiding it from the public apis, moving it into kwargs.
let's discuss about it

DESCRIPTION.md
sfc-gh-aling, line 15: should we name it name `backoff_policy`?

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,37 @@ class AuthType(Enum):
 class AuthByPlugin(ABC):
     """External Authenticator interface."""
 
-    def __init__(self) -> None:
-        self._retry_ctx = AuthRetryCtx()
+    def __init__(
+        self,
+        timeout: int | None = None,
+        max_retry_attempts: int | None = None,
+        socket_timeout_override: int | None = None,
```
sfc-gh-aling, line 66: let's rename this to `socket_timeout` and add docstring to describe the override behavior

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,37 @@ class AuthType(Enum):
...
     @property
-    def timeout(self) -> int:
-        return self._timeout
+    def timeout(self) -> int | None:
+        return self._retry_ctx.timeout
 
-    @timeout.setter
-    def timeout(self, value: Any) -> None:
-        self._timeout = int(value)
+    @property
```
sfc-gh-aling, line 90: let's remove this property

src/snowflake/connector/auth/by_plugin.py
sfc-gh-aling: we can guarantee this will not be none

sfc-gh-aling, line 87:
```suggestion
    def timeout(self) -> int:
```

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -205,22 +195,28 @@ def handle_timeout(
...
         time ranges between 1 and 16 seconds.
+
+        Note that the delete_params parameter is introduced because some
+        authenticators may not want to delete the parameters to this function.
+        Currently, the only authenticator where this is the case is
+        AuthByKeyPair.
         """
 
-        del authenticator, service_name, account, user, password
+        if delete_params:
```
sfc-gh-aling, line 205:
```suggestion
        if kwargs.get("delete_params", True):
```

src/snowflake/connector/auth/keypair.py
```diff
@@ -57,28 +54,31 @@ def __init__(
...
             lifetime_in_seconds: number of seconds the JWT token will be valid
         """
-        super().__init__()
+        super().__init__(
+            max_retry_attempts=int(
+                os.getenv(
+                    "JWT_CNXN_RETRY_ATTEMPTS", AuthByKeyPair.DEFAULT_JWT_RETRY_ATTEMPTS
+                )
+            ),
+            socket_timeout_override=int(
```
sfc-gh-aling, line 63: let's rename to `socket_timeout`

src/snowflake/connector/auth/keypair.py
```diff
@@ -180,9 +180,6 @@ def update_body(self, body: dict[Any, Any]) -> None:
     def assertion_content(self) -> str:
         return self._jwt_token
 
-    def should_retry(self, count: int) -> bool:
-        return count < self._jwt_retry_attempts
-
```
sfc-gh-aling, line 185: let's bring it back for now and create a JIRA for future BCR changes regarding the retry

src/snowflake/connector/backoff_policies.py
```diff
@@ -0,0 +1,145 @@
...
+    This is to enable customization of the constants used in backoff computations.
+"""
+
+DEFAULT_BACKOFF_FACTOR = 2
+DEFAULT_BACKOFF_BASE = 1
+DEFAULT_BACKOFF_CAP = 16
+DEFAULT_ENABLE_JITTER = True
+
+
+def mixed_backoff(
```
sfc-gh-aling, line 40: let's add some unit tests for those backoff policy

src/snowflake/connector/connection.py
```diff
@@ -475,8 +488,16 @@ def login_timeout(self) -> int | None:
...
+
+    @property
+    def socket_timeout(self) -> int | None:
+        return int(self._socket_timeout) if self._socket_timeout is not None else None
+
+    @property
+    def backoff_generator(self) -> Iterator:
```
sfc-gh-aling, line 499: let's make it private if this is internally used only
```suggestion
    def _backoff_generator(self) -> Iterator:
```

src/snowflake/connector/connection.py
```diff
@@ -1099,6 +1142,19 @@ def __config(self, **kwargs):
         if "." in self._account:
             self._account = parse_account(self._account)
 
+        if not isinstance(self._backoff_policy, Callable) or not not isinstance(
```
sfc-gh-aling, line 1145: make sure we have test coverage for this error behavior

src/snowflake/connector/network.py
```diff
@@ -453,8 +454,8 @@ def request(
         body=None,
         method: str = "post",
         client: str = "sfsql",
+        timeout: int | None = None,
         _no_results: bool = False,
-        timeout=None,
```
sfc-gh-aling, line 457: if we want to make the change, we need to make sure all the callers switch the order as well

src/snowflake/connector/time_util.py
```diff
@@ -20,6 +19,8 @@
 
 DEFAULT_MASTER_VALIDITY_IN_SECONDS = 4 * 60 * 60  # seconds
 
+INITIAL_TIMEOUT_SLEEP_TIME = 1
+
```
sfc-gh-aling, line 23:
```suggestion
```

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,36 @@ class AuthType(Enum):
...
 
-    def __init__(self) -> None:
-        self._retry_ctx = AuthRetryCtx()
+    def __init__(
+        self,
+        timeout: int | None = None,
+        max_retry_attempts: int | None = None,
+        socket_timeout: int
+        | None = None,  # if specified, this will take precedence over all other socket timeouts
+        backoff_generator: Iterator | None = None,
```
sfc-gh-aling, line 68: is this only used by us?

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,36 @@ class AuthType(Enum):
...
+        # for example, AuthByKeyPair will do this to ensure JWT tokens are refreshed in time
+        # if not None, this will override socket_timeout specified in connection
+        self._socket_timeout = socket_timeout
 
     @property
     def timeout(self) -> int:
-        return self._timeout
-
-    @timeout.setter
-    def timeout(self, value: Any) -> None:
```
sfc-gh-aling, line 102: I'm afraid removing the setter will break the code..

src/snowflake/connector/network.py
```diff
@@ -1061,15 +1046,17 @@ def _request_exec(
...
         binary_data_handler=None,
-        socket_timeout=DEFAULT_SOCKET_CONNECT_TIMEOUT,
+        socket_timeout: int | None = None,
         is_okta_authentication: bool = False,
     ):
-        if socket_timeout > DEFAULT_SOCKET_CONNECT_TIMEOUT:
-            # socket timeout should not be more than the default.
-            # A shorter timeout may be specified for login time, but
-            # for query, it should be at least 45 seconds.
-            socket_timeout = DEFAULT_SOCKET_CONNECT_TIMEOUT
```
sfc-gh-aling, line 1071: do we need to keep the logic here?

test/unit/test_retry_network.py
```diff
@@ -412,3 +421,51 @@ def error_recv_into(*args, **kwargs):
...
+    )
+
+    with pytest.raises(next_result):
+        rest.fetch(
+            method="post",
+            full_url="https://testaccount.snowflakecomputing.com/queries/v1/query-request",
+            headers=dict(),
+        )
+
+    # 13 seconds should be enough for authenticator to attempt thrice
```
sfc-gh-aling, line 470: typo?
```suggestion
    # 13 seconds should be enough for authenticator to attempt twice
```

test/unit/test_retry_network.py
```diff
@@ -163,7 +166,13 @@ def mock_exec(session, method, full_url, headers, data, token, **kwargs):
 
 
 def test_request_exec():
-    rest = SnowflakeRestful(host="testaccount.snowflakecomputing.com", port=443)
+    connection = mock_connection()
+    connection.errorhandler = Error.default_errorhandler
+    rest = SnowflakeRestful(
+        host="testaccount.snowflakecomputing.com",
+        port=443,
+        connection=connection,
```
sfc-gh-aling, line 174: why we pass connection now?

test/unit/test_retry_network.py
```diff
@@ -116,7 +120,6 @@ def mock_exec(session, method, full_url, headers, data, token, **kwargs):
         return success_result
 
     conn = fake_connector()
-    mockNextSleep.side_effect = lambda cnt, sleep: 0
```
sfc-gh-aling, line 119: why is this removed?

test/unit/test_connection.py
```diff
@@ -316,3 +321,34 @@ def test_missing_default_connection_conf_conn_file(monkeypatch, tmp_path):
...
+
+    with pytest.raises(ForbiddenError):
+        # passing a generator function should make it pass config and error during connection
+        _ = fake_connector(backoff_policy=zero_backoff)
+
+
+@pytest.mark.parametrize("next_action", ("RETRY", "ERROR"))
+@patch("snowflake.connector.vendored.requests.sessions.Session.request")
+def test_handle_timeout(mockSessionRequest, next_action):
+    mockSessionRequest.side_effect = mock_request_with_action(next_action, sleep=5)
```
sfc-gh-aling, line 343: how does this test work? so it will first sleep 5 seconds and then return 503 error code?

src/snowflake/connector/auth/by_plugin.py
```diff
@@ -89,18 +59,39 @@ class AuthType(Enum):
...
+
+        max_retry_attempts = kwargs.get(
+            "max_retry_attempts",
+            int(getenv("MAX_CON_RETRY_ATTEMPTS", DEFAULT_MAX_CON_RETRY_ATTEMPTS)),
+        )
+        self._retry_ctx = TimeoutBackoffCtx(
+            timeout=timeout if timeout is not None else DEFAULT_AUTH_CLASS_TIMEOUT,
+            max_retry_attempts=max_retry_attempts,
+            backoff_generator=backoff_generator,
+        )
```
sfc-gh-aling, line 78: nit: I think the code can be further simplified to:
```suggestion
        self._retry_ctx = TimeoutBackoffCtx(
            timeout=timeout if timeout is not None else DEFAULT_AUTH_CLASS_TIMEOUT,
            max_retry_attempts=kwargs.get(
            "max_retry_attempts",
            int(getenv("MAX_CON_RETRY_ATTEMPTS", DEFAULT_MAX_CON_RETRY_ATTEMPTS)),
            ),
            backoff_generator=backoff_generator,
        )
```

src/snowflake/connector/connection.py
```diff
@@ -276,8 +281,18 @@ class SnowflakeConnection:
         schema: Schema in use on Snowflake.
         warehouse: Warehouse to be used on Snowflake.
         role: Role in use on Snowflake.
-        login_timeout: Login timeout in seconds. Used while authenticating.
-        network_timeout: Network timeout. Used for general purpose.
+        login_timeout: Login timeout in seconds. Login requests will not be retried after this timeout expires.
+            Note that the login attempt may still take more than login_timeout seconds as an ongoing login request
+            cannot be canceled even upon login timeout expiry. The login timeout only prevents further retries.
+            If not specified, login_timeout is set to DEFAULT_AUTH_CLASS_TIMEOUT in the by_plugin module.
```
sfc-gh-aling, line 287: nit: let's be clear where the default value is
```suggestion
            If not specified, login_timeout is set to `snowflake.connector.auth.by_plugin.DEFAULT_AUTH_CLASS_TIMEOUT`.
```

src/snowflake/connector/connection.py
```diff
@@ -276,8 +281,18 @@ class SnowflakeConnection:
...
+        network_timeout: Network timeout in seconds. Network requests besides login requests will not be retried
+            after this timeout expires. Overriden in cursor query execution if timeout is passed to cursor.execute.
+            Note that an operation may still take more than network_timeout seconds for the same reason as above.
+            If not specified, network_timeout is infinite,
+        socket_timeout: Socket timeout in seconds. Sets both socket connect and read timeout.
+        backoff_policy: Backoff policy to use for login and network requests. Must be a callable generator function.
+            Standard implementations of linear and exponential backoff are included in the backoff_policies module.
```
sfc-gh-aling, line 294:
```suggestion
            Standard implementations of linear and exponential backoff are included in the `snowflake.connector.backoff_policies` module.
```

test/unit/test_auth_keypair.py
```diff
@@ -106,21 +110,34 @@ class Bad:
...
         auth_instance = AuthByKeyPair(private_key=bad_private_key)
         with raises(TypeError) as ex:
-            auth_instance.handle_timeout(
-                authenticator="SNOWFLAKE_JWT",
-                service_name=None,
-                account=account,
-                user=user,
-                password=None,
-            )
+            auth_instance.prepare(account=account, user=user)
```
sfc-gh-aling, line 113: handle_timeout won't throw TypeError now?
I think `handle_timeout` calls the `prepare` in `AuthByKeyPair right?

sfc-gh-aling:
awesome work.

looks like the test `test/unit/test_retry_network.py::test_retry_request_timeout[next_action_result1] FAILED [ 78%]` is flaky.
can we improve the stability before checking in?

sfc-gh-aling (approved):

PR #1777, opened 2023-10-17
SNOW-922991: fix file chunks size for files over 80GB

sfc-gh-abenzaoui:

   Fixes #NNNN

The hard-coded S3_CHUNK_SIZE set to 8388608 bytes causes files over 80GB to be split into chunks over the Amazon S3 maximum number of parts per upload of 10000 documented [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html) causing uploads to fail after uploading 10000 chunks.


sfc-gh-abenzaoui:
recheck

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class ChunkSizeCalculator:
+    def __init__(
+        self,
+        current_chunk_size: int = constants["CURRENT_CHUNK_SIZE"],
+        max_object_size: int = constants["MAX_OBJECT_SIZE"],
+        max_part_size: int = constants["MAX_PART_SIZE"],
+        min_part_size: int = constants["MIN_PART_SIZE"],
+        max_parts: int = constants["MAX_PARTS"],
```
sfc-gh-aling, line 28: can those parameter be of type float?
I saw that in the tests there is a float `sample_file_size_5tb = 4.9 * 1024 * 1024 * 1024 * 1024`

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+
+logger = logging.getLogger(__name__)
+
+constants = {
+    "CURRENT_CHUNK_SIZE": 8388608,
+    "MAX_OBJECT_SIZE": 5497558138880,
+    "MAX_PART_SIZE": 5368709120,
+    "MIN_PART_SIZE": 5242880,
+    "MAX_PARTS": 10000,
+}
```
sfc-gh-aling, line 18: they can be flatten out, no need to be put into a dictionary.

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def compute_chunk_size(self, file_size: int = None) -> int:
+        chunk_size = self.current_chunk_size
+        # check if we don"t exceed the allowed max file size 5 TiB
+        if file_size is not None and file_size < self.max_object_size:
+            chunk_size = self._check_max_parts(chunk_size, file_size)
+        else:
+            error_message = f"File size {file_size} exceeds the maximum file size {self.max_object_size}."
+            logger.error(error_message)
+            raise Exception(error_message)
```
sfc-gh-aling, line 44: when users hit the error, what are they supposed to do with the error?

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+        self.current_chunk_size = current_chunk_size
+        self.max_object_size = max_object_size
+        self.max_part_size = max_part_size
+        self.min_part_size = min_part_size
+        self.max_parts = max_parts
+
+    def compute_chunk_size(self, file_size: int = None) -> int:
+        chunk_size = self.current_chunk_size
+        # check if we don"t exceed the allowed max file size 5 TiB
+        if file_size is not None and file_size < self.max_object_size:
```
sfc-gh-aling, line 39: reading from the code, the file_size can not be None, otherwise we will simply throw an error.
in this case this parameter should be a normal positional parameter without default value being None

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+import math
+
+logger = logging.getLogger(__name__)
+
+constants = {
+    "CURRENT_CHUNK_SIZE": 8388608,
+    "MAX_OBJECT_SIZE": 5497558138880,
+    "MAX_PART_SIZE": 5368709120,
+    "MIN_PART_SIZE": 5242880,
+    "MAX_PARTS": 10000,
```
sfc-gh-aling, line 17: is there a link that we can put in our code as comment describing where those constants come from?

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: The maximum file size we can upload via S3 per request is 5 TiB, customers would have to split their files over this limit to upload them in multiple requests.

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: Since those are declared by us, I believe they can be of type int, however, I think `file_size` indeed should be of type float in both places: 
```python
    def compute_chunk_size(self, file_size: float) -> int:
``` 
and
```python
    def _check_max_parts(self, current_chunk_size: int, file_size: float) -> int:
``` 
Or should all of them be of type float?

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+# Amazon S3 multipart upload limits
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
+CURRENT_CHUNK_SIZE = 8 * 1024 ** 2
+MAX_OBJECT_SIZE = 5 * 1024 ** 4
+MAX_PART_SIZE = 5 * 1024 ** 3
+MIN_PART_SIZE = 5 * 1024 ** 2
+MAX_PARTS = 10000
+
+
+class ChunkSizeCalculator:
```
sfc-gh-stan, line 21: Why do we need a class instead of just a function?

DESCRIPTION.md
```diff
@@ -13,6 +13,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added for non-Windows platforms command suggestions (chown/chmod) for insufficient file permissions of config files.
   - Fixed issue with connection diagnostics failing to complete certificate checks.
   - Fixed issue that arrow iterator causes `ImportError` when the c extensions are not compiled.
+  - Fixed an issue with ingesting files over 80 GB due to constant chunk size.
+  - Removed dependencies on Cryptodome and oscrypto and removed the `use_openssl_only` parameter. All connections now go through OpenSSL via the cryptography library, which was already a dependency.
```
sfc-gh-stan, line 17: Is this related?

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(
+        self,
+        current_chunk_size: int = CURRENT_CHUNK_SIZE,
```
sfc-gh-stan, line 24: Optional[int] since it has default value.

src/snowflake/connector/compute_chunk_size.py
sfc-gh-stan: We should not raise the base class `Exception` and use something like `RuntimeError` at least.

src/snowflake/connector/compute_chunk_size.py
```diff
@@ -0,0 +1,69 @@
...
+                f"Setting chunksize to {self.current_chunk_size} instead of the default {current_chunk_size}."
+            )
+            return self.min_part_size
+        else:
+            return current_chunk_size
+
+    # check max chunks number 10k
+    def _check_max_parts(self, current_chunk_size: int, file_size: int) -> int:
+        chunk_size = current_chunk_size
+        num_parts = int(math.ceil(file_size / float(current_chunk_size)))
```
sfc-gh-stan, line 61: we shouldn't need `int` after `ceil`

sfc-gh-stan (changes requested):
Thanks for fixing this, I left some comments.

src/snowflake/connector/compute_chunk_size.py
sfc-gh-aling: @sfc-gh-abenzaoui you can raise the basic connector error: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/errors.py#L32

src/snowflake/connector/compute_chunk_size.py
sfc-gh-aling: depending on usage of it, or even could just be some util functions

src/snowflake/connector/compute_chunk_size.py
sfc-gh-aling: yes, in this case we'd better make `sample_file_size_5tb` an integer

sfc-gh-aling:
to fix the lint issue, you need to install pre-commit and run precommit command locally against the files: https://pre-commit.com/

```
pip install pre-commit
pre-commit run --all-files
```

after the command, it will either update the file for you or point out the places you need to update manually

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -630,7 +630,11 @@ def function_and_callback_wrapper(
     def _create_file_transfer_client(
         self, meta: SnowflakeFileMeta
     ) -> SnowflakeStorageClient:
-        from .constants import AZURE_CHUNK_SIZE, S3_CHUNK_SIZE
+        from .compute_chunk_size import ChunkSizeCalculator
+        from .constants import AZURE_CHUNK_SIZE
+
+        chunk_size_calculator = ChunkSizeCalculator()
+        S3_CHUNK_SIZE = chunk_size_calculator.compute_chunk_size(meta.src_file_size)
```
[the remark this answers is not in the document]
sfc-gh-abenzaoui: I used lowercase for the variable

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: Got rid of the `int` and as agreed the `float` since `current_chunk_size` is an `int`.

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: Rewrote the class into a function

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: I added the links

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: Got rid of the default value

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: Done

src/snowflake/connector/compute_chunk_size.py
sfc-gh-abenzaoui: Done

sfc-gh-aling (approved):

test/unit/test_compute_chunk_size.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+
+try:
+    from snowflake.connector.constants import (
+        S3_MAX_OBJECT_SIZE,
+        S3_MAX_PART_SIZE,
+        S3_MIN_PART_SIZE,
+    )
+    from snowflake.connector.file_transfer_agent import _chunk_size_calculator
+except ImportError:
```
sfc-gh-stan, line 14: Why do need this try catch block?

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -183,6 +189,35 @@ def _update_progress(
     return progress == 1.0
 
 
+def _chunk_size_calculator(file_size: int) -> int:
+    # S3 has limitation on the num of parts to be uploaded, this helper method recalculate the num of parts
+    if file_size > S3_MAX_OBJECT_SIZE:
+        # check if we don't exceed the allowed S3 max file size 5 TiB
+        error_message = f"File size {file_size} exceeds the maximum file size {S3_MAX_OBJECT_SIZE} allowed in S3."
+        logger.error(error_message)
+        Error.errorhandler_wrapper(
```
sfc-gh-stan, line 198: What Exception is this raising?

test/unit/test_compute_chunk_size.py
sfc-gh-aling: old driver doesn't have `file_transfer_agent` module

src/snowflake/connector/file_transfer_agent.py
sfc-gh-aling: is it raising the most generic error `Error` defined in line 201.
I don't find an error type that best fits this case, so just putting a generic one, we can improve in the future
wdyt

sfc-gh-stan (approved):

PR #1786, opened 2023-10-31
SNOW-892716 Improve localfs put/get test

sfc-gh-sfan:
Description

This is a long overdue improvement from early slack message from Brandon

Testing

unit test

   Fixes SNOW-892716

Use Pathlib and with-statement to improve test

### DIFF
--- a/test/unit/test_local_storage_client.py
+++ b/test/unit/test_local_storage_client.py
@@ -4,9 +4,9 @@
 
 import os
 import random
-import shutil
 import string
 import tempfile
+from pathlib import Path
 
 import pytest
 
@@ -26,14 +26,10 @@ def test_multi_chunk_upload(multipart_threshold):
         [random.choice(string.ascii_letters) for _ in range(300)]
     ).encode()
     file_name = "test_file"
-    stage_dir = tempfile.mkdtemp()
-    stage_file = os.path.join(stage_dir, file_name)
-    local_dir = tempfile.mkdtemp()
-    local_file = os.path.join(local_dir, file_name)
-
-    try:
-        with open(local_file, "wb+") as fd:
-            fd.write(file_content)
+    with tempfile.TemporaryDirectory() as stage_dir, tempfile.TemporaryDirectory() as local_dir:
+        stage_file = os.path.join(stage_dir, file_name)
+        local_file = os.path.join(local_dir, file_name)
+        Path(local_file).write_bytes(file_content)
 
         meta = SnowflakeFileMeta(
             name=file_name,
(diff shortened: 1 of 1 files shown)


test/unit/test_local_storage_client.py
```diff
@@ -26,14 +26,10 @@ def test_multi_chunk_upload(multipart_threshold):
...
-    stage_file = os.path.join(stage_dir, file_name)
-    local_dir = tempfile.mkdtemp()
-    local_file = os.path.join(local_dir, file_name)
-
-    try:
-        with open(local_file, "wb+") as fd:
-            fd.write(file_content)
+    with tempfile.TemporaryDirectory() as stage_dir, tempfile.TemporaryDirectory() as local_dir:
+        stage_file = os.path.join(stage_dir, file_name)
+        local_file = os.path.join(local_dir, file_name)
```
sfc-gh-bchinn, line 31: Instead of `os.path.join`, do `Path(...) / ...`! Also, why not use the same temp directory?

```python
with tempfile.TemporaryDirectory() as tmpdir:
    tmpdir = Path(tmpdir)
    stage_file = tmpdir / "stage" / file_name
    local_file = tmpdir / "local" / file_name
    local_file.write_bytes(file_content)

    # ...

    assert stage_file.read_bytes() == file_content
```

sfc-gh-bchinn (approved):
Nice!

test/unit/test_local_storage_client.py
sfc-gh-sfan: Ah that is a good point :) Will do

sfc-gh-sfan: It is actually not that straight forward because we also need to do `mkdir`, otherwise the intermediate directory does not exist. This makes the two `TemporaryDirectory` better because we don't need to manage an awkward `mkdir`.

But I'll still do the `os.path.join` change.

sfc-gh-bchinn: Sounds good 😄  IMO a couple mkdirs is cleaner than two context managers, but up to you!

sfc-gh-bwarsaw: Just a quick note.

> two context managers

Are you familiar with [contextlib.ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)?

And here's a blog post I wrote about it many years ago: https://wefearchange.org/2013/05/resource-management-in-python-33-or.html

sfc-gh-bchinn: TIL! Super cool!

sfc-gh-sfan: I cannot access this blog :(

sfc-gh-bwarsaw: @sfc-gh-sfan Oh no!  What happens?  It WFM and it's a pretty simple static site.

sfc-gh-bchinn: It works for me, but it's not working for another teammate. Maybe VPN shenanigans?

sfc-gh-sfan: It's probably due to devVPN. I'll try that when I'm not connected :)

sfc-gh-bwarsaw: Sigh.  Confirmed the VPN doesn't like my blog.  Maybe it fears change?

sfc-gh-stan (approved):

Issue #1788, opened 2023-11-01
SNOW-956671: We should change the user-agent of the diagnostic report so we can track it's usage


sfc-gh-cconner:
### What is the current behavior?

Currently there is no user agent sent with the diagnostic report.

### What is the desired behavior?

Send a user-agent of snowflake-connector-diagnostic so we can track usage.

### How would this improve `snowflake-connector-python`?

This will allow us to tack usage compared to snowcd and push use away from snowcd.

sfc-gh-cconner:
https://github.com/snowflakedb/snowflake-connector-python/pull/1789

sfc-gh-dszmolka:
thanks for the issue and the fix ;) it seems to be approved for quite some time now, could you please consider merging it (or closing it if it's not needed anymore)
thank you !

PR #1789, opened 2023-11-01
SNOW-956671: We should change the user-agent of the diagnostic report…

sfc-gh-cconner:
… so we can track it's usage

   Fixes #SNOW-956671

   I added a user-agent header to our ssl cert checks so we can track the usage of the python connector diagnostic.

### DIFF
--- a/src/snowflake/connector/connection_diagnostic.py
+++ b/src/snowflake/connector/connection_diagnostic.py
@@ -214,6 +214,12 @@ class ConnectionDiagnostic:
                 context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
                 sock = context.wrap_socket(conn, server_hostname=host)
                 certificate = ssl.DER_cert_to_PEM_cert(sock.getpeercert(True))
+                http_request = f"""GET / {host}:{port} HTTP/1.1\r\n
+                                   Host: {host}\r\n
+                                   User-Agent: snowflake-connector-python-diagnostic
+                                   \r\n\r\n"""
+                sock.send(str.encode(http_request))
+                sock.recv(4096).decode("utf-8")
                 conn.close()
                 return certificate
             else:
--- a/src/snowflake/connector/connection_diagnostic.py
+++ b/src/snowflake/connector/connection_diagnostic.py
@@ -218,8 +218,13 @@ class ConnectionDiagnostic:
                                    Host: {host}\r\n
                                    User-Agent: snowflake-connector-python-diagnostic
                                    \r\n\r\n"""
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/connection_diagnostic.py
```diff
@@ -214,6 +214,12 @@ def __test_socket_get_cert(
                 context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
                 sock = context.wrap_socket(conn, server_hostname=host)
                 certificate = ssl.DER_cert_to_PEM_cert(sock.getpeercert(True))
+                http_request = f"""GET / {host}:{port} HTTP/1.1\r\n
+                                   Host: {host}\r\n
+                                   User-Agent: snowflake-connector-python-diagnostic
+                                   \r\n\r\n"""
+                sock.send(str.encode(http_request))
+                sock.recv(4096).decode("utf-8")
```
sfc-gh-aling, line 222: I'm not familiar with the diagnostic workflow, is the `sock.recv(4096).decode("utf-8")` necessary here?
also do you want try-except here in case `sock.send` or `sock.recv` failed but you still the flow keep moving forward even without http agent successfully, or errors already handled in upper stream?

src/snowflake/connector/connection_diagnostic.py
```diff
@@ -214,6 +214,12 @@ def __test_socket_get_cert(
                 context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
                 sock = context.wrap_socket(conn, server_hostname=host)
                 certificate = ssl.DER_cert_to_PEM_cert(sock.getpeercert(True))
+                http_request = f"""GET / {host}:{port} HTTP/1.1\r\n
```
sfc-gh-aling, line 217: is this method `__test_socket_get_cert` being tested?

src/snowflake/connector/connection_diagnostic.py
sfc-gh-cconner: I swore when I wrote this the sock recv was needed, but it does not seem to be the case.  I've removed that.  I do have the try/except in the methods that call this one, but added the extra layer just in case.

src/snowflake/connector/connection_diagnostic.py
sfc-gh-cconner: This is tested indirectly with the http_host_report test, but added a direct test for this as well.

sfc-gh-aling (approved):

sfc-gh-dszmolka:
fixes https://github.com/snowflakedb/snowflake-connector-python/issues/1788

PR #1804, opened 2023-11-13
SNOW-950840 Add support for the vector data type

sfc-gh-rpanchapakesan:

   Fixes SNOW-950840

Add support for vectors (soon in Private Preview) to the connector. This primarily consists of Arrow deserialization logic.

I will follow these changes up with parsing logic for the new result metadata fields so that the vector field type definition in `constants.py` can use the correct pyarrow type. Right now, we return `pa.binary()` for `FieldType.pa_type` since we don't have access to the vector's type and dimension (which are needed by `pa.list_`). This bypass is functionally equivalent since we only use `FieldType.pa_type` to generate pandas type mappings (`object` for vectors), but I will update it once the aforementioned result metadata changes are complete.

### DIFF
--- a/src/snowflake/connector/constants.py
+++ b/src/snowflake/connector/constants.py
@@ -95,6 +95,14 @@ FIELD_TYPES: tuple[FieldType, ...] = (
     FieldType(
         name="GEOMETRY", dbapi_type=[DBAPI_TYPE_STRING], pa_type=lambda: pa.string()
     ),
+    FieldType(
+        # TODO(SNOW-969160): While pa.binary() results in the correct pandas column
+        # type being generated, it should be switched to pa.list_(...) once parsing
+        # for the new result metadata fields is added.
+        name="VECTOR",
+        dbapi_type=[DBAPI_TYPE_BINARY],
+        pa_type=lambda: pa.binary(),
+    ),
 )
 
 FIELD_NAME_TO_ID: DefaultDict[Any, int] = defaultdict(int)
--- a/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
+++ b/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
@@ -44,11 +44,11 @@ PyObject* FixedSizeListConverter::toPyObject(int64_t rowIndex) const
   // the lengths of the fixed size lists in the array.
 
   ArrowArrayView* elements = m_array->children[0];
-  const auto length = elements->length / m_array->length;
-  PyObject* list = PyList_New(length);
+  const auto fixedSizeArrayLength = elements->length / m_array->length;
+  PyObject* list = PyList_New(fixedSizeArrayLength);
 
-  const int64_t startIndexWithoutOffset = rowIndex * length;
-  for (int64_t i = 0; i < length; ++i)
+  const int64_t startIndexWithoutOffset = rowIndex * fixedSizeArrayLength;
+  for (int64_t i = 0; i < fixedSizeArrayLength; ++i)
   {
     const auto bufferIndexWithoutOffset = startIndexWithoutOffset + i;
     // Currently, the backend only sends back INT32 and FLOAT32, but the
@@ -63,8 +63,7 @@ PyObject* FixedSizeListConverter::toPyObject(int64_t rowIndex) const
         const auto value =
             ArrowArrayViewGetIntUnsafe(elements, bufferIndexWithoutOffset);
         PyList_SetItem(list, i, PyLong_FromLongLong(value));
-      }
-      break;
+      } break;
       case NANOARROW_TYPE_HALF_FLOAT:
       case NANOARROW_TYPE_FLOAT:
       case NANOARROW_TYPE_DOUBLE:
@@ -72,8 +71,7 @@ PyObject* FixedSizeListConverter::toPyObject(int64_t rowIndex) const
         const auto value =
             ArrowArrayViewGetDoubleUnsafe(elements, bufferIndexWithoutOffset);
         PyList_SetItem(list, i, PyFloat_FromDouble(value));
-      }
-      break;
+      } break;
       default:
         std::string errorInfo = Logger::formatString(
             "[Snowflake Exception] invalid arrow element type for fixed size "
--- a/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
+++ b/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
@@ -1,0 +1,90 @@
+//
+// Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
+//
+
+#include "FixedSizeListConverter.hpp"
+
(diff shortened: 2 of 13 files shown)


sfc-gh-rpanchapakesan:

src/snowflake/connector/constants.py
```diff
@@ -95,6 +95,14 @@ class FieldType(NamedTuple):
     FieldType(
         name="GEOMETRY", dbapi_type=[DBAPI_TYPE_STRING], pa_type=lambda: pa.string()
     ),
+    FieldType(
+        # TODO(SNOW-969160): While pa.binary() results in the correct pandas column
+        # type being generated, it should be switched to pa.list_(...) once parsing
+        # for the new result metadata fields is added.
+        name="VECTOR",
+        dbapi_type=[DBAPI_TYPE_BINARY],
+        pa_type=lambda: pa.binary(),
```
sfc-gh-mloring, line 104: How would a caller convert this binary value back into a list of floats? Do we need to document endianness here for users to unpack this correctly?

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  const int64_t startIndexWithoutOffset = rowIndex * length;
+  for (int64_t i = 0; i < length; ++i)
+  {
+    const auto bufferIndexWithoutOffset = startIndexWithoutOffset + i;
+    switch (elements->storage_type)
+    {
+      case NANOARROW_TYPE_INT8:
+      case NANOARROW_TYPE_INT16:
+      case NANOARROW_TYPE_INT32:
```
sfc-gh-mloring, line 58: Should we disallow all but 32 bit int/float to be consistent with our support in the runtime or is there any automatic narrowing that occurs causing us to emit 16/8 bit values for arrow?

src/snowflake/connector/constants.py
sfc-gh-rpanchapakesan: This type is only used when we're generating an empty pandas dataframe (to infer the type of the pandas columns). When rows are returned, we can infer the type from the payload itself. I have a test which verifies this pandas handling.

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -686,6 +717,13 @@ def validate_pandas(
                             "values are not equal".format(i, j, cases[i], c_new)
                         )
                         break
+                    elif data_type.startswith("vector"):
+                        c_case = numpy.array(cases[i])
+                        assert numpy.array_equal(c_case, c_new), (
+                            "{} row, {} column: original value is {}, new value is {}, "
+                            "values are not equal".format(i, j, cases[i], c_new)
+                        )
+                        continue
```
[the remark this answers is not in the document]
sfc-gh-rpanchapakesan: I'm not sure why timestamp handling uses `break`, but the cases for other types and NULL implicitly continue (they fall to the end of the loop, without a break).

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
sfc-gh-rpanchapakesan: I kept all the types in for completeness so that we don't have to release a new connector if the supported precision types change (and since they didn't require extra logic). I can remove the non 32-bit variants if that's preferable, however.

src/snowflake/connector/constants.py
sfc-gh-mloring: Sounds good.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
sfc-gh-mloring: Fine to leave them. Maybe add a comment that these are not in use as of <current python connector version> but may be used in the future. Is there any way we can get unit test coverage for these paths until it's possible to test them e2e?

sfc-gh-mloring (approved):
Looks good on the Vector Data Type side.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
sfc-gh-rpanchapakesan: There unfortunately don't seem to be unit tests of the arrow deserialization code, so I'm not sure if there's a simple way to test these right now.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return nullptr;
+  }
+
+  // m_array->length represents the number of fixed size lists in the array
+  // m_array->children[0] has a buffer view that contains the actual data of
+  // each list, back-to-back m_array->children[0]->length represents the sum of
+  // the lengths of the fixed size lists in the array.
+
+  ArrowArrayView* elements = m_array->children[0];
+  const auto length = elements->length / m_array->length;
```
sfc-gh-sfan, line 47: nit: can we rename this to `fixedSizeArrayLength`? There are a lot of lengths already so let's not introduce yet another `length` :p

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/FixedSizeListConverter.cpp
```diff
@@ -0,0 +1,90 @@
...
+      case NANOARROW_TYPE_INT8:
+      case NANOARROW_TYPE_INT16:
+      case NANOARROW_TYPE_INT32:
+      case NANOARROW_TYPE_INT64:
+      {
+        const auto value =
+            ArrowArrayViewGetIntUnsafe(elements, bufferIndexWithoutOffset);
+        PyList_SetItem(list, i, PyLong_FromLongLong(value));
+      }
+      break;
```
sfc-gh-sfan, line 67: nit: it looks more straightforward if we indent these lines? ditto to the double ones below

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -583,6 +583,54 @@ def test_timestampltz(conn_cnx, scale, timezone):
...
+    if is_public_test:
+        pytest.xfail(
+            reason="This feature hasn't been rolled out for public Snowflake deployments yet."
+        )
+    tests = [
+        (
+            "vector(int,3)",
+            [
+                "NULL",
+                "[1,2,3]::vector(int,3)",
```
sfc-gh-sfan, line 600: Just out of my curiosity, do we allow `null` as a vector element?

sfc-gh-sfan (approved):
LGTM % some nits

test/integ/pandas/test_arrow_pandas.py
sfc-gh-rpanchapakesan: Individual vector elements can't be null, but the vector as a whole could be

PR #1806, opened 2023-11-15
SNOW-950840 Parse new result metadata attributes for internal use

sfc-gh-rpanchapakesan:
I added the `NO-CHANGELOG-UPDATES` label since this change does not affect the public API of the connector. Let me know if I should instead update the changelog.

   Fixes SNOW-950840

Created a new version of result metadata that contains new fields that have been added to the response (and are needed to support vector types in Snowpark). To avoid causing a breaking API change, this PR does not modify the existing `ResultMetadata` class (which is a named tuple so field additions could cause user-facing breakages). Instead, this change only adds parsing for a new `ResultMetadataV2` class that internally replaces usages of `ResultMetadata`. This class by default not returned by `Cursor.description` and `Cursor.describe(...)`. Instead, it is only returned here if a newly-added parameter is set to true.

Longer-term, the old class should be deprecated, but, until that point, this serves as a stop-gap to unblock other work.


src/snowflake/connector/cursor.py
```diff
@@ -144,20 +145,212 @@ class ResultMetadata(NamedTuple):
...
+    @property
+    def fields(self) -> list[ResultMetadataV2Field] | None:
+        return self._fields
+
+
+class ResultMetadataV2:
+    """ResultMetadataV2Field represents the type information of a single column.
+
+    It is a replacement for ResultMetadata that contains additional attributes, namely
+    `vector_dimension` and `fields`.
```
sfc-gh-mloring, line 249: s/namely/currently. Also, let's add a note that this will be unified with `ResultMetadata` in the near future.

sfc-gh-mloring (approved):
Looks good on the vector side!

src/snowflake/connector/cursor.py
```diff
@@ -144,20 +144,221 @@ class ResultMetadata(NamedTuple):
...
+
+    def __init__(
+        self,
+        name: str,
+        type_code: int,
+        is_nullable: bool,
+        display_size: int | None = None,
+        internal_size: int | None = None,
+        precision: int | None = None,
+        scale: int | None = None,
```
sfc-gh-aalam, line 243: is there any particular reason why is_nullable moved up in order here compared to ResultMetadata. Does the ordering even matter?

sfc-gh-aalam (approved):

src/snowflake/connector/cursor.py
sfc-gh-aalam: ordering does not matter. changes look good

sfc-gh-rpanchapakesan: Yeah, I moved it up only since it doesn't have a default value. `ResultMetadataV2` isn't a named tuple, so it shouldn't have any other user-visible effect.

src/snowflake/connector/cursor.py
```diff
@@ -929,7 +1141,9 @@ def execute_async(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
         kwargs["_exec_async"] = True
         return self.execute(*args, **kwargs)
 
-    def describe(self, *args: Any, **kwargs: Any) -> list[ResultMetadata]:
+    def describe(
+        self, *args: Any, **kwargs: Any
+    ) -> list[ResultMetadata | ResultMetadataV2]:
```
sfc-gh-stan, line 1146: Shouldn't this be just List[ResultMetadata]?

src/snowflake/connector/cursor.py
```diff
@@ -940,6 +1154,26 @@ def describe(self, *args: Any, **kwargs: Any) -> list[ResultMetadata]:
...
         kwargs["_describe_only"] = kwargs["_is_internal"] = True
         self.execute(*args, **kwargs)
+
+        if self._description is None:
+            return None
+        return [meta._to_result_metadata_v1() for meta in self._description]
+
+    def _describe_internal(
+        self, *args: Any, **kwargs: Any
+    ) -> list[ResultMetadata | ResultMetadataV2]:
```
sfc-gh-stan, line 1164: Is this returning a mix of ResultMetadata and ResultMetadataV2?

src/snowflake/connector/cursor.py
sfc-gh-rpanchapakesan: Thanks for catching this: it should be just `List[ResultMetadata]` since describe only returns the old format. Same for `_describe_internal` in your other comment.

PR #1818, opened 2023-12-05
APPS-32904: Allow reusing a connection from another connector via session and master token

sfc-gh-klin:

   Addresses APPS-32904

   - [ ] I am adding a new automated test(s) to verify correctness of my new code (haven't checked out tests, in-progress)
   - [ ] I am adding new logging messages (shouldn't be needed?) 
   - [ ] I am adding a new telemetry message (shouldn't be needed?) 
   - [x] I am modifying authorization mechanisms (authentication) 

The Snowflake VS Code extension uses the node connector (snowflake-sdk) to create a connection. Users who use Snowpark in conjunction with the VS Code extension would benefit if the extension lets them use the node connector's connection that they've already authenticated with by logging in through the extension UI. Then when they execute Snowpark code, they wouldn't have to create another connection through the Python connector. 

To get started with this flow, the Python connector needs to be able to accept a session and master token from another connection. 

I'm interested in starting some discussion around this via this PR. 

4. Considerations

`update_tokens` takes some other parameters such as id_token and mfa_token. The Node connector's connection, when serialized, does not have that information. I tested with a trial Snowflake account with MFA enabled and I successfully logged in and reused the node connector connection in Python code. I'm not 100% clear on whether omitting these (id_token, mfa_token) could affect anything else. In connection.py's open_connection, by using the session and master token in update_tokens, I may be bypassing some validation that occurs in all the other authentication paths. Interested in learning more from the regular contributors to this repository.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -89,13 +89,7 @@ from .errorcode import (
     ER_NO_USER,
     ER_NOT_IMPLICITY_SNOWFLAKE_DATATYPE,
 )
-from .errors import (
-    DatabaseError,
-    Error,
-    OperationalError,
-    ProgrammingError,
-    UnauthorizedError,
-)
+from .errors import DatabaseError, Error, OperationalError, ProgrammingError
 from .network import (
     DEFAULT_AUTHENTICATOR,
     EXTERNAL_BROWSER_AUTHENTICATOR,
@@ -939,8 +933,17 @@ class SnowflakeConnection:
                 self._master_validity_in_seconds,
             )
             heartbeat_ret = auth._rest._heartbeat()
-            if not heartbeat_ret.get("success"):
-                raise UnauthorizedError
+            logger.debug(heartbeat_ret)
+            if not heartbeat_ret or not heartbeat_ret.get("success"):
+                Error.errorhandler_wrapper(
+                    self,
+                    None,
+                    ProgrammingError,
+                    {
+                        "msg": "Session and master tokens invalid",
+                        "errno": ER_INVALID_VALUE,
+                    },
+                )
             else:
                 logger.debug("Session and master token validation successful.")
 
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -932,6 +932,9 @@ class SnowflakeConnection:
                 self._master_token,
                 self._master_validity_in_seconds,
             )
+            auth._rest._heartbeat()
+            logger.debug("Session and master token validation successful.")
+
         else:
             if self.auth_class is not None:
                 if type(
@@ -1135,41 +1138,35 @@ class SnowflakeConnection:
             ]:
                 self._authenticator = auth_tmp
 
-        if (
-            not self._master_token
-            and not self._session_token
-            and not self.user
-            and self._authenticator != OAUTH_AUTHENTICATOR
-        ):
-            # OAuth Authentication does not require a username
-            Error.errorhandler_wrapper(
-                self,
-                None,
-                ProgrammingError,
-                {"msg": "User is empty", "errno": ER_NO_USER},
-            )
+        if not self._master_token and not self._session_token:
+            if not self.user and self._authenticator != OAUTH_AUTHENTICATOR:
+                # OAuth Authentication does not require a username
+                Error.errorhandler_wrapper(
+                    self,
+                    None,
+                    ProgrammingError,
+                    {"msg": "User is empty", "errno": ER_NO_USER},
+                )
 
-        if self._private_key or self._private_key_file:
-            self._authenticator = KEY_PAIR_AUTHENTICATOR
+            if self._private_key or self._private_key_file:
+                self._authenticator = KEY_PAIR_AUTHENTICATOR
 
-        if (
-            self.auth_class is None
-            and self._authenticator
-            not in [
-                EXTERNAL_BROWSER_AUTHENTICATOR,
-                OAUTH_AUTHENTICATOR,
-                KEY_PAIR_AUTHENTICATOR,
-            ]
-            and not self._password
-            and not self._master_token
-            and not self._session_token
-        ):
-            Error.errorhandler_wrapper(
-                self,
-                None,
-                ProgrammingError,
-                {"msg": "Password is empty", "errno": ER_NO_PASSWORD},
-            )
+            if (
+                self.auth_class is None
+                and self._authenticator
+                not in [
+                    EXTERNAL_BROWSER_AUTHENTICATOR,
+                    OAUTH_AUTHENTICATOR,
+                    KEY_PAIR_AUTHENTICATOR,
+                ]
+                and not self._password
+            ):
+                Error.errorhandler_wrapper(
+                    self,
+                    None,
+                    ProgrammingError,
+                    {"msg": "Password is empty", "errno": ER_NO_PASSWORD},
+                )
 
         if not self._account:
             Error.errorhandler_wrapper(
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -89,7 +89,13 @@ from .errorcode import (
     ER_NO_USER,
     ER_NOT_IMPLICITY_SNOWFLAKE_DATATYPE,
 )
-from .errors import DatabaseError, Error, OperationalError, ProgrammingError
+from .errors import (
+    DatabaseError,
+    Error,
+    OperationalError,
+    ProgrammingError,
+    UnauthorizedError,
+)
 from .network import (
     DEFAULT_AUTHENTICATOR,
     EXTERNAL_BROWSER_AUTHENTICATOR,
@@ -932,8 +938,11 @@ class SnowflakeConnection:
                 self._master_token,
                 self._master_validity_in_seconds,
             )
-            auth._rest._heartbeat()
-            logger.debug("Session and master token validation successful.")
+            heartbeat_ret = auth._rest._heartbeat()
+            if not heartbeat_ret.get("success"):
+                raise UnauthorizedError
+            else:
+                logger.debug("Session and master token validation successful.")
 
         else:
             if self.auth_class is not None:
@@ -1138,7 +1147,7 @@ class SnowflakeConnection:
             ]:
                 self._authenticator = auth_tmp
 
-        if not self._master_token and not self._session_token:
+        if not (self._master_token and self._session_token):
             if not self.user and self._authenticator != OAUTH_AUTHENTICATOR:
                 # OAuth Authentication does not require a username
                 Error.errorhandler_wrapper(
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -256,6 +256,18 @@ DEFAULT_CONFIGURATION: dict[str, tuple[Any, type | tuple[type, ...]]] = {
         True,
         bool,
     ),  # Enable sending retryReason in response header for query-requests
+    "session_token": (
+        None,
+        (type(None), str),
+    ),  # session token from another connection, to be provided together with master token
+    "master_token": (
+        None,
+        (type(None), str),
+    ),  # master token from another connection, to be provided together with session token
+    "master_validity_in_seconds": (
+        None,
+        (type(None), int),
+    ),  # master token validity in seconds
 }
 
 APPLICATION_RE = re.compile(r"[\w\d_]+")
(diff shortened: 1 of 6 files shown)


sfc-gh-klin:

sfc-gh-klin:
recheck

src/snowflake/connector/connection.py
```diff
@@ -913,100 +925,108 @@ def __open_connection(self):
...
-            self._session_parameters[PARAMETER_CLIENT_STORE_TEMPORARY_CREDENTIAL] = (
-                self._client_store_temporary_credential if IS_LINUX else True
-            )
-            auth.read_temporary_credentials(
-                self.host,
-                self.user,
-                self._session_parameters,
+
+        if self._session_token and self._master_token:
+            auth._rest.update_tokens(
```
sfc-gh-mkeller, line 930: I think we should test whether the "authentication" was successful somehow

src/snowflake/connector/connection.py
```diff
@@ -1115,7 +1135,12 @@ def __config(self, **kwargs):
             ]:
                 self._authenticator = auth_tmp
 
-        if not self.user and self._authenticator != OAUTH_AUTHENTICATOR:
+        if (
+            not self._master_token
```
sfc-gh-mkeller, line 1139: I feel like instead of modifying this and the next if statements it'd be better to rearrange the code so that error checking (btw you missed the account one) could simply be indented with the other old code.

sfc-gh-mkeller (changes requested):
We should also discuss testing plan for this! I think we could do a integration test where we extract these from a normally created connection.

sfc-gh-klin:
I added an integration test called test_with_tokens and ran it with `python -m pytest test/integ/test_connection.py::test_with_tokens` after adding a parameters.py file filled out. 

However, I'm puzzled as to whether or not my test genuinely tests what I am trying to do, because I can put nonsense in for the session_token and master_token, and as long as I have the account/user/password/host filled out it'll pass. I think this is why you suggested to check that the "authentication" succeeded since otherwise it's just supplying the session/master token and storing those to be used when user actually tries to execute something. 

The other thing I'm confused by is that I would expect to be able to provide only the account, master_token, and session_token, but if in parameters.py I don't provide user/password then the integration test also does not succeed. 

```
ERROR test/integ/test_connection.py::test_with_tokens - snowflake.connector.errors.DatabaseError: 250001 (08001): None: Failed to connect to DB: ig86763.preprod12.us-west-2.aws.snowfl...
```
I see that this is because we do some setup in conftest.py around init_test_schema for example, and that connection uses user/password. Is it reasonable for me to also pass along session_token and master_token to give flexibility with parameters.py usage?

sfc-gh-klin:
Huh, kind of interesting. Trying to use session_token and master_token by editing init_test_schema with 
```
    ret = db_parameters
    with snowflake.connector.connect(
        user=ret["user"],
        password=ret["password"],
        host=ret["host"],
        port=ret["port"],
        database=ret["database"],
        account=ret["account"],
        protocol=ret["protocol"],
        session_token=ret["session_token"],
        master_token=ret["master_token"]
```
doesn't seem to work because it immediately makes the token expired. 

```
ERROR test/integ/test_connection.py::test_with_tokens - snowflake.connector.errors.ProgrammingError: 390111: 390111: Session no longer exists.  New login required to access the service.
```

For some reason it's okay when I reuse the token myself in a Python file I execute (which is how I manually tested my code changes) but trying to do it through conftest.py expires the token and my other uses of the connection also die.

src/snowflake/connector/connection.py
sfc-gh-klin: Okay, if I try to make a heartbeat call there and then it fails, user will get this error in their terminal for example
```
  File "/Users/klin/git/sp-test/src/sp_test.py", line 3, in <module>
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/__init__.py", line 54, in Connect
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/connection.py", line 425, in __init__
    self.connect(**kwargs)
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/connection.py", line 715, in connect
    self.__open_connection()
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/connection.py", line 935, in __open_connection
    ret = auth._rest._heartbeat()
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 609, in _heartbeat
    ret = self._post_request(
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 730, in _post_request
    ret = self.fetch(
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 839, in fetch
    ret = self._request_exec_wrapper(
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 975, in _request_exec_wrapper
    raise e
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 880, in _request_exec_wrapper
    return_object = self._request_exec(
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 1189, in _request_exec
    raise err
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 1131, in _request_exec
    raise_failed_request_error(
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/network.py", line 238, in raise_failed_request_error
    Error.errorhandler_wrapper(
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/errors.py", line 290, in errorhandler_wrapper
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/errors.py", line 348, in hand_to_other_handler
    connection.errorhandler(connection, cursor, error_class, error_value)
  File "/opt/homebrew/lib/python3.9/site-packages/snowflake/connector/errors.py", line 221, in default_errorhandler
    raise error_class(
snowflake.connector.errors.InterfaceError: 250003 (08001): 401 Unauthorized: post https://ig86763.preprod12.us-west-2.aws.snowflakecomputing.com:443/session/heartbeat?requestId=4a63f07e-4928-4f98-b02d-58a4e7151f73&request_guid=148cac21-a4da-4341-bb24-862f8a6211c7
```

Does that seem like a reasonable experience to give a dev for when the tokens are invalid? Should that error get caught and replaced with clearer wording?

src/snowflake/connector/connection.py
sfc-gh-klin: Okay, rearranged! Let me know if that's what you were suggesting

sfc-gh-klin:
Removed my previous comments to avoid confusion, I've rewritten the unit test to basically make a first connection then take the tokens off of it and create a second connection. Currently I'm having a local failure on `python -m pytest test/integ/test_connection.py::test_ocsp_mode_insecure` and I'll look into. Not sure I understand that test yet.

src/snowflake/connector/connection.py
```diff
@@ -1115,34 +1138,35 @@ def __config(self, **kwargs):
...
 
-        if not self.user and self._authenticator != OAUTH_AUTHENTICATOR:
-            # OAuth Authentication does not require a username
-            Error.errorhandler_wrapper(
-                self,
-                None,
-                ProgrammingError,
-                {"msg": "User is empty", "errno": ER_NO_USER},
-            )
+        if not self._master_token and not self._session_token:
```
sfc-gh-aling, line 1141: why do we need check to master and session token here?

src/snowflake/connector/connection.py
```diff
@@ -913,100 +925,111 @@ def __open_connection(self):
...
-                    backoff_generator=self._backoff_generator,
-                )
-            else:
-                self.auth_class = AuthByIdToken(
-                    id_token=self._rest.id_token,
-                    application=self.application,
-                    protocol=self._protocol,
-                    host=self.host,
-                    port=self.port,
+            auth._rest._heartbeat()
```
sfc-gh-aling, line 935: I'm curious what kind of error will be raised if invalid/expired token is provided?

src/snowflake/connector/connection.py
```diff
@@ -256,6 +256,18 @@ def _get_private_bytes_from_file(
...
     ),  # Enable sending retryReason in response header for query-requests
+    "session_token": (
+        None,
+        (type(None), str),
+    ),  # session token from another connection, to be provided together with master token
+    "master_token": (
+        None,
+        (type(None), str),
+    ),  # master token from another connection, to be provided together with session token
+    "master_validity_in_seconds": (
```
sfc-gh-aling, line 267: does the session token has a similar validity in seconds concept?

sfc-gh-klin: I don't think it does--or if it does, it's not something the connector code ever sees. This value I get from the node connector when I serialize the object to get the session, master, and master validity. It has no other info on it. 

Example from the python connector code, it references master validity but nothing about session. 
https://sourcegraph.c2.us-central1.gcp-dev.app.snowflake.com/github.com/snowflakedb/snowflake-connector-python/-/blob/src/snowflake/connector/network.py?L408:16

src/snowflake/connector/connection.py
sfc-gh-klin: This is what it shows:
```snowflake.connector.errors.InterfaceError: 250003 (08001): 401 Unauthorized: post https://ig86763.preprod12.us-west-2.aws.snowflakecomputing.com:443/session/heartbeat?requestId=4a63f07e-4928-4f98-b02d-58a4e7151f73&request_guid=148cac21-a4da-4341-bb24-862f8a6211c7```

I think I can try to write a test for that.

File: src/snowflake/connector/connection.py (hunk 3 above)
Comment (sfc-gh-klin) [reply src/snowflake/connector/connection.py:1141]: In my understanding: Master and session token being provided should override the other authentication methods. 

So we check that if it's not provided, then we do the normal authentication workflow like it was before any code change.

Comment (sfc-gh-aling) [reply src/snowflake/connector/connection.py:1141]: in this case, should the condition be `if not (self._master_token and self._session_token)` in which case we only skip the normal authentication workflow when both tokens are detected?

otherwise if only one token is provided, we shall use the normal authentication workflow

File: src/snowflake/connector/connection.py (hunk 4 above)
Comment (sfc-gh-aling) [reply src/snowflake/connector/connection.py:935]: yeah let's add a test for it -- error behavior is also part of the public apis

**Comment by sfc-gh-aling** [comment]
@sfc-gh-klin  can you also update the changelog file: https://github.com/snowflakedb/snowflake-connector-python/blob/main/DESCRIPTION.md -- this is considered a new feature to the connector.

File: src/snowflake/connector/connection.py (hunk 3 above)
Comment (sfc-gh-klin) [reply src/snowflake/connector/connection.py]: Thanks for catching that, that makes sense to me.

File: src/snowflake/connector/connection.py (hunk 4 above)
Comment (sfc-gh-klin) [reply src/snowflake/connector/connection.py]: I added a test for it and initially I edited errors.py to add another error type (the Unauthorized one) but I noticed it caused issues with the old driver test due to imports, so I instead went with ProgrammingError since technically speaking the tokens aren't valid anymore. What's your take on this?

**Comment by sfc-gh-klin** [comment]
I've updated the changelog, and the failing Jenkins tests are related to a different commit and not my changes--some discussion in this Slack channel https://snowflake.slack.com/archives/C7Y5W9XR6/p1704846477609349?thread_ts=1704840824.860299&cid=C7Y5W9XR6

File: test/integ/test_connection.py [hunk 6]
```diff
@@ -131,6 +131,63 @@ def test_with_config(db_parameters):
...
+        account=db_parameters["account"],
+        protocol=db_parameters["protocol"],
+        timezone="UTC",
+    )
+    assert initial_cnx, "invalid initial cnx"
+    master_token = initial_cnx._rest._master_token
+    session_token = initial_cnx._rest._token
+    initial_cnx.close()
+
+    with pytest.raises(Error):
```
sfc-gh-aling, line 179: is this throwing the `InterfaceError` based upon our previous discussion?

> snowflake.connector.errors.InterfaceError: 250003 (08001): 401 Unauthorized: post https://ig86763.preprod12.us-west-2.aws.snowflakecomputing.com:443/session/heartbeat?requestId=4a63f07e-4928-4f98-b02d-58a4e7151f73&request_guid=148cac21-a4da-4341-bb24-862f8a6211c7

src/snowflake/connector/connection.py
sfc-gh-aling: I don't think we need to introduce another error type, whatever is returned nowadays shall be good

test/integ/test_connection.py
sfc-gh-klin: It is now a ProgrammingError--I have made the test be more specific and changed it to ProgrammingError.

src/snowflake/connector/connection.py
sfc-gh-klin: Okay, roger that.

sfc-gh-aling (approved):
LGTM

src/snowflake/connector/network.py
```diff
@@ -595,7 +595,7 @@ def _token_request(self, request_type):
                 },
             )
 
-    def _heartbeat(self) -> None:
+    def _heartbeat(self) -> Any | dict[Any, Any] | dict[str, Any] | None:
```
sfc-gh-mkeller, line 598: `dict[Any, Any]` includes `dict[str, Any]`. No need to have both!
```suggestion
    def _heartbeat(self) -> Any | dict[Any, Any] | None:
```

sfc-gh-mkeller (approved):
One tiny type-hint thing, please accept that suggestion. Otherwise this looks great!
Thanks for adding tests too! I made them a little safer, hope you don't min!

:ship: :ship:

sfc-gh-klin:
Thank you for making the tests safer, this is useful learning for me. 
Accepted your type hint change, will merge when checks pass.

PR #1828, opened 2023-12-13
SNOW-991012 Cleanup some C++ code warnings and performance issues

sfc-gh-llorimer:

   I can create an issue, if needed.

   - Add `-O0` when building in debug mode.
       - This disables C/C++ optimizations, which makes the debugging experience much better.
       - I've only tested this in GCC/Clang. I doubt this would work in MSVC, but I don't think anyone uses this mode with MSVC.
   - There was a `ArrowDecimal` parsing codepath that seems to create a `shared_ptr` on a per-value basis.
       - I switched this to use a stack-allocated object, which should be "free" to allocate.
   - `ReturnVal` just contains 2 pointers, without any ownership over the pointers.
       - I added a comment to reflect this, so it's more clear.
       - The caller does not need to keep a reference to the returned value, so I removed the `cret` field from `PyArrowIterator`.
       - The object has a small, fixed size, so it does not need to be a `shared_ptr`. I switched this to return-by-value in places where it's used.
           - The pyx file didn't immediately work with this, since it tries to use the default constructor. I added a default constructor for it to use.
   - When `NANOARROW_TYPE_ENUM_STRING` is defined in the header, anyone who doesn't use it has a compiler warning.
       - I have moved the definition to the `.cpp` file to silence the warning.
       - I also took this opportunity to move it into the `sf` namespace.
   - I made the move constructor of `UniqueRef` non-explicit.
       - This allows `UniqueRef` to be returned from functions by value. See [this compiler explorer link](https://godbolt.org/z/1j8Kde1dc) for an example.
       - Note that [`std::unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr) also has an implicit move constructor.
   - I made `UniqueRef::reset()` safer.
       - The documentation for [`Py_SETREF`](https://docs.python.org/3/c-api/refcounting.html#c.Py_SETREF) and [`Py_CLEAR`](https://docs.python.org/3/c-api/refcounting.html#c.Py_CLEAR) mention that if the `XDECREF` causes other code to run which calls into the object, then it might try to use the object that is being deleted, which may lead to issues.
       - The new `UniqueRef::reset()` clears the reference before decrementing the refcount.
           - Note that [`std::unique_ptr::reset()`](https://en.cppreference.com/w/cpp/memory/unique_ptr/reset) also does these operations, in the same order.


src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp
```diff
@@ -305,10 +304,10 @@ void CArrowTableIterator::convertScaledFixedNumberColumnToDecimalColumn_nanoarro
...
     } else {
         auto originalVal = ArrowArrayViewGetIntUnsafe(columnArray, rowIdx);
-        std::shared_ptr<ArrowDecimal> arrowDecimal = std::make_shared<ArrowDecimal>();
-        ArrowDecimalInit(arrowDecimal.get(), 128, 38, scale);
-        ArrowDecimalSetInt(arrowDecimal.get(), originalVal);
-        returnCode = ArrowArrayAppendDecimal(newArray, arrowDecimal.get());
+        ArrowDecimal arrowDecimal;
+        ArrowDecimalInit(&arrowDecimal, 128, 38, scale);
+        ArrowDecimalSetInt(&arrowDecimal, originalVal);
+        returnCode = ArrowArrayAppendDecimal(newArray, &arrowDecimal);
```
sfc-gh-aling, line 310: in this case won't `arrowDecimal` be GCed as it's a local variable?

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/nanoarrow_arrow_iterator.pyx
```diff
@@ -225,17 +223,17 @@ cdef class PyArrowTableIterator(PyArrowIterator):
             self.arrow_bytes_size,
             self.number_to_decimal,
         )
-        self.cret = self.cIterator.checkInitializationStatus()
-        if self.cret.get().exception:
+        cdef ReturnVal cret = self.cIterator.checkInitializationStatus()
+        if self.cret.exception:
```
sfc-gh-aling, line 227:
```suggestion
        if cret.exception:
```

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp
sfc-gh-llorimer: There is no GC in C++. Instead, the memory pattern is well specified:

* Previously, when `std::shared_ptr<ArrowDecimal> arrowDecimal` was created, we would allocate some memory in the heap to back the `arrowDecimal`. When it went out of scope, `shared_ptr` would decrement the reference count, it would hit 0, and we would free the memory that backed the `arrowDecimal`.
* Now, when `ArrowDecimal arrowDecimal` is created, we allocate memory on the stack. When it goes out of scope, the relevant stack memory is freed.

Allocating memory on the stack is more-or-less "free". This variable probably has a fixed location in the stack, so the compiler can emit assembly that just uses `ebp - fixedNumber` for `arrowDecimal`.

Allocating memory on the heap is a lot more complex.

Also, if `ArrowArrayAppendDecimal` _does_ require the variable to stick around, then there's already a bug in the code - since the `shared_ptr` is freed, there is no guarantee that the memory would stay valid.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/nanoarrow_arrow_iterator.pyx
sfc-gh-llorimer: Yes, this helps :P

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp
sfc-gh-aling: thanks for refreshing my memory on c++, yeah you are right.
the change looks good to me

sfc-gh-aling (approved):
thanks for the PR! could you also help update the changelog? just saying what's in the PR title should be enough.

sfc-gh-mkeller (approved):
:ship: :ship:

PR #1829, opened 2023-12-14
Fix type hints of SnowflakeCursor.describe and SnowflakeCursor. _describe_internal

sfc-gh-stan:
This is a small PR with two two type hint fixes.


sfc-gh-rpanchapakesan (approved):
Thanks for catching and fixing this!

PR #1801, opened 2023-11-09
SNOW-966003: Fix Arrow return value for zero-row queries

thomasaarholt:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #1800

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   This PR copies the approach done in #1275 by @jakobgm, but based on the latest main (I'm sorry @jakobgm, but it was easier to just copy rather than rebase your solution, and I hope you don't mind that), with one more type hint added due to some later commits. That approach itself is based on the current implementation of `.fetch_pandas_all()`.

   This is a minor code change, copying methodology already implemented by snowflake on the pandas side, but it makes it a lot easier to work with arrow data from the snowflake-connector-python.

   Please let me know if anything more is required from me.

PR #1832, opened 2023-12-18
SNOW-966003: Fix Arrow return value for zero-row queries

sfc-gh-aling:

   Fixes SNOW-966003, dup PR: https://github.com/snowflakedb/snowflake-connector-python/pull/1801#issuecomment-1861345001

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -33,6 +33,7 @@ from typing import (
 )
 
 from typing_extensions import Self
+from warnings import warn
 
 from snowflake.connector.result_batch import create_batches_from_response
 from snowflake.connector.result_set import ResultSet
@@ -1333,8 +1334,25 @@ class SnowflakeCursor:
         )
         return self._result_set._fetch_arrow_batches()
 
-    def fetch_arrow_all(self) -> Table:
+    @overload
+    def fetch_arrow_all(self, force_return_table: Literal[False] = False) -> None:
+        ...
+
+    @overload
+    def fetch_arrow_all(self, force_return_table: Literal[True] = True) -> Table:
+        ...
+
+    def fetch_arrow_all(self, force_return_table: bool = False) -> Table | None:
+        if not force_return_table:
+            warn(
+                "The behaviour of fetch_arrow_all when the query returns zero rows "
+                "will change from returning None to returning an empty pyarrow table with "
+                "schema using the highest bit length for each column.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
         self.check_can_use_arrow_resultset()
+
         if self._prefetch_hook is not None:
             self._prefetch_hook()
         if self._query_result_format != "arrow":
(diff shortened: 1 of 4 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -1333,14 +1333,29 @@ def fetch_arrow_batches(self) -> Iterator[Table]:
         )
         return self._result_set._fetch_arrow_batches()
 
-    def fetch_arrow_all(self) -> Table | None:
+    @overload
+    def fetch_arrow_all(self, force_return_table: Literal[False] = False) -> None:
```
sfc-gh-mkeller, line 1337: This overloading isn't fully correct, see [this example](https://mypy.readthedocs.io/en/stable/literal_types.html). Also I think instead of `-> None` it should be the old behavior, so `-> Table | None` when `False` is provided.

src/snowflake/connector/result_set.py
```diff
@@ -164,13 +173,21 @@ def _fetch_arrow_batches(
         self._can_create_arrow_iter()
         return self._create_iter(iter_unit=IterUnit.TABLE_UNIT, structure="arrow")
 
-    def _fetch_arrow_all(self) -> Table | None:
+    @overload
+    def _fetch_arrow_all(self, force_return_table: Literal[False] = False) -> None:
```
sfc-gh-mkeller, line 177: Same as in the other file

sfc-gh-mkeller (commented):
Other than the overloading this is good to go by me.

:ship: :ship:

sfc-gh-mkeller (approved):
:ship: :ship:

thomasaarholt:
The type signature came from me writing this in vscode with the pyright type checker.

Looks like mypy and pyright have slightly different behaviour. pyright allows my usage, while mypy does not. Here is a version that satisfies both . I believe the return type of `None` when `Literal[False]` is provided, is correct. 

```python
from typing import Literal, overload, reveal_type, NewType

Table = NewType("Table", object)


class Foo:
    @overload
    def fetch_arrow_all(self, force_return_table: Literal[False]) -> None:
        ...

    @overload
    def fetch_arrow_all(self, force_return_table: Literal[True]) -> Table:
        ...

    def fetch_arrow_all(self, force_return_table: bool = False) -> Table | None:
        return Table(1) if force_return_table else None


fetch_true = Foo().fetch_arrow_all(True)
fetch_false = Foo().fetch_arrow_all(False)
reveal_type(fetch_true)  # Table
reveal_type(fetch_false)  # None
```

sfc-gh-aling:
thanks @sfc-gh-mkeller @thomasaarholt , I have applied the change.
I do agree with Mark that when it's `force_return_table` is force, it should be the old `Table | None` as for non-zero row result, pyarrow.Table is returned.

src/snowflake/connector/result_set.py
sfc-gh-aling: updated

src/snowflake/connector/cursor.py
sfc-gh-aling: you are right, updated

thomasaarholt:
Apologies, I was blinded by what the correct return type. You're absolutely right.

sfc-gh-aling:
no worries @thomasaarholt , you are all good!
I merged this PR and the change will be carried in our 2024 Jan release.
thanks for your contribution!

PR #1835, opened 2023-12-19
Fix pyarrow Table type hinting in cursor.py

### LINKED ISSUE #1834: SNOW-994431: Pyarrow Table has wrong type signature in cursor.py
[In cursor.py, pyarrow's `Table`](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/cursor.py#L88C6-L88C6) is set as `Table = None` when pyarrow is not available. It is then used as a type hint. You can't use expressions as type hints, and the correct way here (especially since Table isn't used anywhere in the file except as a type hint) is to use `if TYPE_CHECKING`.

I have made a PR for this, see link below shortly.

thomasaarholt:

   Fixes #1834

   Table is now correctly typed. Note that pyarrow does not have good typing (Table is typed as "unknown"), but now at least the snowflake implementation is correct.

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -15,6 +15,7 @@ import time
 import uuid
 import warnings
 from enum import Enum
+from importlib import util
 from logging import getLogger
 from threading import Lock, Timer
 from types import TracebackType
@@ -76,16 +77,16 @@ if TYPE_CHECKING:  # pragma: no cover
     from .connection import SnowflakeConnection
     from .file_transfer_agent import SnowflakeProgressPercentage
     from .result_batch import ResultBatch
+    from pyarrow import Table
 
 T = TypeVar("T", bound=collections.abc.Sequence)
 
 logger = getLogger(__name__)
 
-if installed_pandas:
-    from pyarrow import Table
-else:
+
+if not util.find_spec("pyarrow"):
     logger.debug("Failed to import pyarrow. Cannot use pandas fetch API")
-    Table = None
+
 
 try:
     from .nanoarrow_arrow_iterator import PyArrowIterator  # NOQA
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -15,7 +15,6 @@ import time
 import uuid
 import warnings
 from enum import Enum
-from importlib import util
 from logging import getLogger
 from threading import Lock, Timer
 from types import TracebackType
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -73,11 +73,11 @@ from .time_util import get_time_millis
 
 if TYPE_CHECKING:  # pragma: no cover
     from pandas import DataFrame
+    from pyarrow import Table
 
     from .connection import SnowflakeConnection
     from .file_transfer_agent import SnowflakeProgressPercentage
     from .result_batch import ResultBatch
-    from pyarrow import Table
 
 T = TypeVar("T", bound=collections.abc.Sequence)
 
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -84,8 +84,12 @@ T = TypeVar("T", bound=collections.abc.Sequence)
 logger = getLogger(__name__)
 
 
-if not util.find_spec("pyarrow"):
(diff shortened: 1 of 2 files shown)


thomasaarholt:
The following is perhaps better formulated as an issue, but I thought I'd just share the following.

I noticed that inside the `SnowflakeCursor` class, you [initialize lots of variables](https://github.com/snowflakedb/snowflake-connector-python/blob/aa5e50b9004ad38a3607a0e70dfd3445a501457c/src/snowflake/connector/cursor.py#L463) as `self.foo: SomeType | None = None`. For instance `self._result_set: ResultSet | None = None`.

 This pattern makes it very difficult for a type checker to work well, since whenever you do something like `self._result_set._fetch_pandas_all`, you get the type error `"_fetch_pandas_all" is not a known member of "None"`.

Generally, the best solution is to have different classes one for initialization and one for when you have a result - or have a ResultSet that can be empty. The other solution is to liberally use `assert self.foo is not None`, so that the type checker gets this. 

I considered adding a couple of `assert self._result_set is not None` when making this PR, but I decided not to, since doing so properly would mean a lot of assertions.

src/snowflake/connector/cursor.py
```diff
@@ -76,16 +77,16 @@
...
 
 T = TypeVar("T", bound=collections.abc.Sequence)
 
 logger = getLogger(__name__)
 
-if installed_pandas:
-    from pyarrow import Table
-else:
+
+if not util.find_spec("pyarrow"):
```
sfc-gh-aalam, line 87: a check similar to this is done in snowflake.connector.options which tells if pandas/pyarrow are installed.

src/snowflake/connector/cursor.py
```diff
@@ -76,16 +77,16 @@
     from .connection import SnowflakeConnection
     from .file_transfer_agent import SnowflakeProgressPercentage
     from .result_batch import ResultBatch
+    from pyarrow import Table
```
sfc-gh-aalam, line 81: if we directly add it here, I think it would break users who do not have pyarrow installed in their environment.

thomasaarholt: It shouldn't, since it's under the `if TYPE_CHECKING` conditional.  I have just now reordered the imports so that the import from pyarrow is next to the import from pandas, so as to make this more clear. Pandas is another package that is not a dependency of snowflake without extras installed (and both pyarrow and pandas are installed by the `[pandas]` extra).

These imports are only tried by the type checker, not at runtime. The usage of the imported classes in type hints in the function definitions is allowed due to the presence of this import at the top of the file: `from __future__ import annotations`. You can read more about Postponed evaluation of annotations [here on stackoverflow](https://stackoverflow.com/a/61544901/1018861). This feature is already being used for [pandas in this file here](https://github.com/snowflakedb/snowflake-connector-python/blob/990df441805f91907248beb9bef1d06cbc6c8f04/src/snowflake/connector/cursor.py#L1360).

You can also try running the following code block to convince yourself. It won't throw an error about `FooBar` (unless you explicitly call `main()`, but that also applies to calling `fetch_pandas_all` without pandas installed).

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from import_that_doesnt_exist import FooBar

def main() -> FooBar:
    return FooBar()

print("Hello, world!") # this will run
```

src/snowflake/connector/cursor.py
thomasaarholt: Thanks. I initially didn't want to use it, since it's called `installed_pandas` and I couldn't see a `installed_pyarrow`. But I realise that it in the function it comes from, it will only evaluate to `True` if both pandas and pyarrow are installed. I decided to improve the warning a bit, since if the user has pyarrow but not pandas installed, the warning would still show.

src/snowflake/connector/cursor.py
sfc-gh-aalam: Thanks. Testing the PR here: https://github.com/snowflakedb/snowflake-connector-python/pull/1838

sfc-gh-sfan (approved):

sfc-gh-mkeller (approved):
Ohh, haha. This is from when I was first learning about type-hints! 😅

:ship: :ship:

sfc-gh-mkeller:
@thomasaarholt please rebase your PR add a changelog entry. We can merge this right after!

thomasaarholt:
Done!

PR #1838, opened 2023-12-20
Test pr 1835

sfc-gh-aalam:
Tests #1835

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -15,6 +15,7 @@ import time
 import uuid
 import warnings
 from enum import Enum
+from importlib import util
 from logging import getLogger
 from threading import Lock, Timer
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -15,6 +15,7 @@
 import uuid
 import warnings
 from enum import Enum
+from importlib import util
```
thomasaarholt, line 18: Looks like I didn’t remove the import that we decided was unnecessary. I’m used to Ruff removing these automatically on save in my projects, sorry about that!

sfc-gh-aalam: could you remove it and add a changelog entry. I'll merge and publish this again

thomasaarholt: I just removed it now, in #1835.

Happy new year!

sfc-gh-mkeller (approved):
Ohh, haha. This is from I was first learning about type-hints! 😅 

:ship: :ship:

PR #1846, opened 2024-01-04
SNOW-999576: Fix several local test problems

sfc-gh-bwarsaw:

   Fixes #NNNN

   * In several cases, we need to dig the `warehouse` parameter out of the connection dictionary, and use that explicitly rather than rely on an implicit user default warehouse.
   * Add `setuptools` as an explicit tox provisioning requirement.
   * Don't assume that unsetting the `SNOWFLAKE_HOME` environment variable is enough to pass the fallback test, since the local user may have `~/.snowflake`, and if they do, this is used as the implicit fallback.  Use a fake home in a temporary directory instead, which is guaranteed not to exist.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -14,19 +14,23 @@ CONNECTION_PARAMETERS = {
 }
 ```
 
-### Running a single test
-
-Assuming that all dependencies are installed, running a single test is as simple as:
-`python -m pytest test/integ/test_connection.py::test_basic`.
-
 ### Running a suite of tests
 
-We use `tox` to run test suites and other utilities.
+We use `tox` version 4 to run test suites and other utilities.
 
 To run the most important tests, execute:
 
 ```shell
 tox -e "fix_lint,py37{,-pandas,-sso}"
+```
+
+### Running a single test
+
+Enter the tox environment you want (e.g. `py38`) and run `pytest` from there:
+
+```shell
+source .tox/py38/bin/activate
+pytest -v test/integ/test_connection.py::test_basic
 ```
 
 ## Test types
--- a/test/README.md
+++ b/test/README.md
@@ -24,6 +24,16 @@ To run the most important tests, execute:
 tox -e "fix_lint,py37{,-pandas,-sso}"
 ```
 
+**NOTE** Some integration tests may be sensitive to the cloud provider of the
+account that the test suite connects to.  By default, when testing locally,
+the "provider" is called `dev`, but you may see some test failures with this
+provider (which isn't really a provider).  To eliminate this false failure,
+set the environment variable `cloud_provider` to one of `aws`, `gcp`, or
+`azure` as appropriate for the account you're running integration tests
+against.  This is handled correctly in CI so should only affect local
+developers.  In the future, we'll try to make this automatic by querying the
+account after the connection is made.
+
 ### Running a single test
 
 Enter the tox environment you want (e.g. `py38`) and run `pytest` from there:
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -22,6 +22,7 @@ envlist = fix_lint,
           coverage
 skip_missing_interpreters = true
 requires =
+    setuptools>=40.6.0
```
sfc-gh-mkeller, line 25: This shouldn't be required. It's only required at build time, for which we use PEP-517 and this is defined here: https://github.com/snowflakedb/snowflake-connector-python/blob/93c9b69f9e2d7d8b159c24c85ab775e54bc1be0c/pyproject.toml#L5

sfc-gh-bwarsaw: Yes, but I think the problem is that tox tries to build the package first, and it uses its `.tox/.tox` provisioned environment to do this.  However, that environment doesn't have setuptools installed by default, so if you do a clean git clone of the repository and then try e.g. `tox -e py38`, it fails very early.  That's why this is added to the `[tox]` environment.  If that doesn't happen for you then it's worth looking at why it differs.  It could be that the Python you're using for the tox environment outside of the provisioned one already has setuptools installed.  Or could you be using tox v3 where maybe things work differently?  I'm using tox 4.11.4 from homebrew.

sfc-gh-bwarsaw: I solved this by removing the dependency on `tox-external-wheels`, which I believe pins tox's provisioned version to tox 3 and requires setuptools because only tox 4 knows about pyproject.toml.

sfc-gh-bwarsaw:

test/README.md
```diff
@@ -14,21 +14,35 @@ CONNECTION_PARAMETERS = {
...
 
 To run the most important tests, execute:
 
 ```shell
 tox -e "fix_lint,py37{,-pandas,-sso}"
 ```
 
+**NOTE** Some integration tests may be sensitive to the cloud provider of the
+account that the test suite connects to.  By default, when testing locally,
+the "provider" is called `dev`, but you may see some test failures with this
```
sfc-gh-mkeller, line 29: I'd word this a little differently. `dev` is simply a made up provider that acts as all cloud providers and it's intended so that devs do not need to set `cloud_provider`

test/README.md
```diff
@@ -14,21 +14,35 @@ CONNECTION_PARAMETERS = {
...
+against.  This is handled correctly in CI so should only affect local
+developers.  In the future, we'll try to make this automatic by querying the
+account after the connection is made.
+
+### Running a single test
+
+Enter the tox environment you want (e.g. `py38`) and run `pytest` from there:
+
+```shell
+source .tox/py38/bin/activate
```
sfc-gh-mkeller, line 42: So `source` is a bashism, while most things support bash syntax [dot](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot) should work on all modern shells (`dash` is the only one that comes to my mind that is a strict sh compatible shell).
```suggestion
. .tox/py38/bin/activate
```

tox.ini
```diff
@@ -21,23 +21,14 @@ envlist = fix_lint,
...
     sso: secure-local-storage
-install_command = python -m pip install -U {opts} {packages}
-external_wheels =
-    py37-ci: dist/*cp37*.whl
-    py38-ci: dist/*cp38*.whl
-    py39-ci: dist/*cp39*.whl
-    py310-ci: dist/*cp310*.whl
-    py311-ci: dist/*cp311*.whl
-    py312-ci: dist/*cp312*.whl
+package = wheel
```
sfc-gh-mkeller, line 31: Now you are introducing dependency on a tox4 feature, please update `minversion = 3.7` above in this file.

sfc-gh-mkeller (approved):
:ship: :ship:

Issue #1870, opened 2024-02-01
SNOW-1038158: Python Connector Diagnostic fails because system$whitelist is deprecated


sfc-gh-cconner:
### Python version

3.10.12

### Operating system and processor architecture

macOS-14.3-arm64-arm-64bit

### Installed packages

```python
asn1crypto==1.5.1
build==1.0.3
certifi==2023.7.22
cffi==1.15.1
charset-normalizer==3.2.0
coverage==7.4.1
cryptography==41.0.3
Cython==3.0.8
exceptiongroup==1.2.0
execnet==2.0.2
filelock==3.12.4
idna==3.4
iniconfig==2.0.0
more-itertools==10.2.0
numpy==1.26.3
oscrypto==1.3.0
packaging==23.1
pendulum==3.0.0
pexpect==4.9.0
platformdirs==3.8.1
pluggy==1.4.0
ptyprocess==0.7.0
pycparser==2.21
pycryptodomex==3.18.0
PyJWT==2.8.0
pyOpenSSL==23.2.0
pyproject_hooks==1.0.0
pytest==7.4.4
pytest-cov==4.1.0
pytest-rerunfailures==13.0
pytest-timeout==2.2.0
pytest-xdist==3.5.0
python-dateutil==2.8.2
pytz==2023.3.post1
pytzdata==2020.1
requests==2.31.0
six==1.16.0
snowflake-connector-python @ file:///Users/cconner/snowflake/source/forks/snowflake-connector-python/dist/snowflake_connector_python-3.7.0-cp310-cp310-macosx_13_0_arm64.whl#sha256=4c12190e7bc2312ec49bbf0d054b475b15e2a081796ce8edc46b5f5063deb6a4
sortedcontainers==2.4.0
time-machine==2.13.0
tomli==2.0.1
tomlkit==0.12.1
typing_extensions==4.7.1
tzdata==2023.4
urllib3==1.26.16
```

### What did you do?

```python
Tried to run the python connector diagnostic and it failed because the result of the system$whitelist is now:

This function has been deprecated. Use SYSTEM$ALLOWLIST instead.
```

### What did you expect to see?

I expected to see the diagnostic report.

### Can you set logging to DEBUG and collect the logs?

```bash
N/A
```

sfc-gh-dszmolka:
thanks for raising this issue ;) and the https://github.com/snowflakedb/snowflake-connector-python/pull/1871 too !

[...]

PR #1680, opened 2023-08-03
SNOW-878611 renaming and aliasing CONFIG_PARSER to CONFIG_MANAGER

sfc-gh-mkeller:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-878611

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   In this PR I modify doc-strings and rename all parser occurrences to manager while being backwards compatible.

PR #1871, opened 2024-02-01
Updated connection diagnostic to use system$allowlist instead of the …

sfc-gh-cconner:
…deprecated system$whitelist

   Fixes #[1870](https://github.com/snowflakedb/snowflake-connector-python/issues/1870) 

   - [ X ] I am adding a new automated test(s) to verify correctness of my new code

   I updated all calls to whitelist to be allowlist.  I left whitelist parameter and check for both allowlist and whitelist to determine which parameter to use for backward compatibility.  I also added integration tests to validate backward compatibility and validate allowlist check failure still allows the report to run.


src/snowflake/connector/connection.py
```diff
@@ -235,7 +235,11 @@ def _get_private_bytes_from_file(
     "connection_diag_whitelist_path": (
         None,
         (type(None), str),
-    ),  # Path to connection diag whitelist json
+    ),  # Path to connection diag whitelist json - Deprecated remove in future
```
sfc-gh-aling, line 238: I have one question and one request:

- is `connection_diag_whitelist_path` only internally used by us (not external customer facing)? if this api keeps change, shall we consider making it private by naming like `_connection_diag_whitelist_path` (add a underscore at the front)
- can you create a JIRA for the future deprecation?

src/snowflake/connector/connection.py
```diff
@@ -657,6 +662,10 @@ def connection_diag_log_path(self):
     def connection_diag_whitelist_path(self):
         return self._connection_diag_whitelist_path
 
+    @property
+    def connection_diag_allowlist_path(self):
+        return self._connection_diag_allowlist_path
```
sfc-gh-aling, line 667: see my previous question about making it private.

also you can check this PR: https://github.com/snowflakedb/snowflake-connector-python/pull/1680/files
to see how we want to deprecate a public API

src/snowflake/connector/connection.py
sfc-gh-cconner: connection_diag_whitelist_path and connection_diag_allowlist_path are the public parameter names for a path to the allowlist/whitelist if they need to provide one.  So it should stay public, right?

Filed a jira for future deprecation.

src/snowflake/connector/connection.py
sfc-gh-cconner: See above.

src/snowflake/connector/connection.py
sfc-gh-cconner: I think these are resolved.

src/snowflake/connector/connection_diagnostic.py
```diff
@@ -43,6 +43,18 @@ def _decode_dict(d):
     return result
 
 
+def _is_list_of_json_objects(allowlist: str | List[Dict[str, Any]]):
+    if isinstance(allowlist, list) and all(
```
sfc-gh-aling, line 47: nit: the type hint `allowlist: str ` doesn't match the `isinstance` check

src/snowflake/connector/connection_diagnostic.py
```diff
@@ -43,6 +43,18 @@ def _decode_dict(d):
     return result
 
 
+def _is_list_of_json_objects(allowlist: str | List[Dict[str, Any]]):
```
sfc-gh-aling, line 46: nit: the type hint `allowlist: str`  doesn't match the `isinstance(allowlist, list)` check.
```suggestion
def _is_list_of_json_objects(allowlist: List[Dict[str, Any]]):
```

src/snowflake/connector/connection_diagnostic.py
```diff
@@ -212,6 +224,10 @@ def __test_socket_get_cert(
                     conn.recv(4096).decode("utf-8")
 
                 context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+                # For some reason SSL checks fail now
+                # the test does not care if cert is valid
+                context.check_hostname = False
+                context.verify_mode = ssl.CERT_NONE
```
sfc-gh-aling, line 230: although the test doesn't care, but in production do we care? this seems dangerous.
do we know why ssl checks fail?

sfc-gh-cconner: We don't care because we want to get the cert whether it's valid or not.  Customers have a tendency to have SSL inspecting proxies with certs that fail.  So we want this check to succeed so we can report what's in the cert.

sfc-gh-cconner: I went with certifi here instead.  So this is gone.  Since this is a diag, we might want to capture that with bad certs for ssl inspection, however, I'll wait and see how it plays out in customer usage.

sfc-gh-aling (approved):

PR #1879, opened 2024-02-08
SNOW-1039216: add default memory pool for pyarrow

sfc-gh-aalam:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/options.py
+++ b/src/snowflake/connector/options.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import importlib
+import os
 import warnings
 from importlib.metadata import distributions
 from logging import getLogger
@@ -78,6 +79,9 @@ def _import_or_missing_pandas_option() -> (
         from pandas import DataFrame  # NOQA
 
         pyarrow = importlib.import_module("pyarrow")
+
+        # set default memory pool to system for pyarrow to_pandas conversion
+        os.environ["ARROW_DEFAULT_MEMORY_POOL"] = "system"
 
         # Check whether we have the currently supported pyarrow installed
         installed_packages = {


sfc-gh-aalam:
paired sp-connector change: https://github.com/snowflakedb/Stored-Proc-Python-Connector/pull/163

DESCRIPTION.md
```diff
@@ -15,6 +15,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Bumped pyOpenSSL dependency from >=16.2.0,<24.0.0 to >=16.2.0,<25.0.0.
   - Fixed a memory leak in decimal data conversion.
   - Fixed a bug where `write_pandas` wasn't truncating the target table.
+  - Set `ARROW_DEFAULT_MEMORY_POOL=system` as default memory allocator for `pyarrow.Table.to_pandas`.
```
sfc-gh-aling, line 18: may I ask what's the effect of this parameter?

sfc-gh-aalam: This forces choosing `system` memory allocator which is better than `mimalloc` in terms of memory allocation. Voltron team found some bugs when using mimalloc covered [here](https://www.support.voltrondata.com/servicedesk/customer/portal/1/VESA-156) - for instance they found there is additional memory blooming (a 1 MB requests from pyarrow ends up allocating 64MB memory block) and once conversion is done, the allocator does not return memory back to os. In our tests with `system` allocator, we noted it saved additional 10% memory in sproc environment so this is a good choice as a default allocator.

sfc-gh-aling: thanks for the explanation. can I understand this as a completely internal implementation detail to improve user experience? from feature wise users won't sense any difference

sfc-gh-stan (approved):
I agree with Adam that this is not necessary to include in CHANGELOG.

DESCRIPTION.md
sfc-gh-aalam: I would set and unset it around the specific to pandas conversion.

sfc-gh-aalam: >  from feature wise users won't sense any difference, or only improvement they can sense.
From feature wise there is no difference, they will only choose a `system` memory allocator if no other default is set  for pyarrow operations.

> also since it's an env var, I'm thinking of whether other python packages (not controlled by us) that customers use will be impacted by this change? is there a different way to set this parameter to scope the impact
`ARROW_DEFAULT_MEMORY_POOL` only affects pyarrow an related functions but you do raise a good point here. In general, pyarrow team encourages every customer to test their workload with the 3 available allocators and choose the best. In sproc environment we have noticed `system` to be better in the scenarios ML has been testing. The default preference in sproc makes sense. Since this is a low impact change, I think we should be okay setting the default value to system here. Once the python program execution is done, this setting is erased so user environment is returned back to a clean environment.

sfc-gh-aling (approved):

PR #1888, opened 2024-02-26
SNOW-1000284: Refactor MetadataV2 ahead of structured types.

sfc-gh-jrose:

   Fixes part of #1000284

   This PR removes the ResultMetadataV2Field class. This class is functionally the same as the ResultMetadataV2 class with the one difference being it can be created without a name. With structured types incoming the metadata can be arbitrarily nested so I don't think it makes sense to have separate classes for the top level metadata and the child metadatas. 

  The only place this class is referenced is the ResultMetadataV2.from_column and I've replaced that with an equivalent initializer for a ResultMetadataV2 object instead. This class was not referenced by any tests or any other repo so I think it should be safe to remove without a breaking-change api version bump.

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -165,71 +165,8 @@ class ResultMetadata(NamedTuple):
         )
 
 
-class ResultMetadataV2Field:
-    """ResultMetadataV2Field represents the type information of one sub-type in a nested type."""
-
-    def __init__(
-        self,
-        type_code: int,
-        is_nullable: bool,
-        internal_size: int | None = None,
-        precision: int | None = None,
-        scale: int | None = None,
-        fields: list[ResultMetadataV2] | None = None,
-    ):
-        self._type_code = type_code
-        self._is_nullable = is_nullable
-        self._internal_size = internal_size
-        self._precision = precision
-        self._scale = scale
-        self._fields = fields
-
-    @classmethod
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -272,7 +209,9 @@ def from_column(cls, col: dict[str, Any]) -> ResultMetadataV2:
 
         fields = None
         if type_code == FIELD_NAME_TO_ID["VECTOR"] and col.get("fields") is not None:
-            fields = [ResultMetadataV2Field.from_column(f) for f in col["fields"]]
+            fields = [
+                ResultMetadata.from_column({"name": None, **f}) for f in col["fields"]
```
sfc-gh-aling, line 213: should this be `ResultMetadataV2`?
```suggestion
                ResultMetadataV2.from_column({"name": None, **f}) for f in col["fields"]
```

sfc-gh-aling (commented):
thanks for the refactor, I'm good with the change if it's not public, and there is no api usage.

btw is it that the `MetadataV2` is not used in production yet? I'm surprised that removing a class doesn't break anything which means we do not have tests for this before.

I think we can add some basic unit test or mock test for it

sfc-gh-jrose:
MetadataV2 is currently used in several places, it's just the MetadataV2Field class that isn't directly referenced elsewhere.

sfc-gh-aling (approved):

PR #1890, opened 2024-02-29
SNOW-1000284: Add support for structured types to ResultMetadataV2

sfc-gh-jrose:

   Implements: #1000284

   This change adds support for structured types to ResultMetadataV2 which should allow for accurate schemas to be created in the snowpark python repo. Notable changes:
   - Changed the from_column handler to support other structured types as well as nested structured types. 
   - Added MAP as a supported metadata type.
   - Added relevant default initializers for pyarrow in order to support casting result batches to pyarrow tables.
   - Added a test that tests that the default initializers for pyarrow work as expected. I did not see an existing one. This test also happens to hit the new code paths for the ResultMetadataV2 changes.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -20,6 +20,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
     - Add `SNOWFLAKE_AUTH_SOCKET_MSG_DONTWAIT` flag (usage: `SNOWFLAKE_AUTH_SOCKET_MSG_DONTWAIT=true`) to make a non-blocking socket.recv call and retry on Error
       - Consider using this if running in a containerized environment and externalbrowser auth frequently hangs while waiting for callback
       - NOTE: this has not been tested extensively, but has been shown to improve the experience when using WSL
+  - Added support for strucutured type information in ResultMetadataV2
 
 - v3.7.1(February 21, 2024)
 
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -20,7 +20,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
     - Add `SNOWFLAKE_AUTH_SOCKET_MSG_DONTWAIT` flag (usage: `SNOWFLAKE_AUTH_SOCKET_MSG_DONTWAIT=true`) to make a non-blocking socket.recv call and retry on Error
       - Consider using this if running in a containerized environment and externalbrowser auth frequently hangs while waiting for callback
       - NOTE: this has not been tested extensively, but has been shown to improve the experience when using WSL
-  - Added support for strucutured type information in ResultMetadataV2
+  - Added support for parsing structured type information in schema queries.
 
 - v3.7.1(February 21, 2024)
 
--- a/src/snowflake/connector/constants.py
+++ b/src/snowflake/connector/constants.py
@@ -44,38 +44,75 @@ class FieldType(NamedTuple):
 
 
 def vector_pa_type(metadata: ResultMetadataV2) -> DataType:
-    """Generate the Arrow type represented by the given vector column metadata.
-
+    """
+    Generate the Arrow type represented by the given vector column metadata.
     Vectors are represented as Arrow fixed-size lists.
     """
+    assert (
+        metadata.fields is not None and len(metadata.fields) == 1
+    ), "Invalid result metadata for vector type: expected a single field to be defined"
+    assert (
+        metadata.vector_dimension or 0
+    ) > 0, "Invalid result metadata for vector type: expected a positive dimension"
 
+    field_type = FIELD_TYPES[metadata.fields[0].type_code]
+    return pa.list_(field_type.pa_type(metadata.fields[0]), metadata.vector_dimension)
+
+
+def array_pa_type(metadata: ResultMetadataV2) -> DataType:
+    """
+    Generate the Arrow type represented by the given array column metadata.
+    """
+    # If fields is missing then structured types are not enabled.
+    # Fallback to json encoded string
     if metadata.fields is None:
-        raise ValueError(
-            "Invalid result metadata for vector type: expected sub-field metadata"
-        )
-    if len(metadata.fields) != 1:
-        raise ValueError(
-            "Invalid result metadata for vector type: expected a single sub-field metadata"
-        )
-    field_type = FIELD_ID_TO_NAME[metadata.fields[0].type_code]
+        return pa.string()
 
-    if metadata.vector_dimension is None:
-        raise ValueError(
-            "Invalid result metadata for vector type: expected a dimension"
-        )
-    elif metadata.vector_dimension <= 0:
-        raise ValueError(
-            "Invalid result metadata for vector type: expected a positive dimension"
-        )
+    assert (
+        len(metadata.fields) == 1
+    ), "Invalid result metadata for array type: expected a single field to be defined"
 
-    if field_type == "FIXED":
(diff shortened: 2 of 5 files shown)


test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1202,6 +1202,89 @@ def test_batch_to_pandas_arrow(conn_cnx, result_format):
...
+            '2020-03-12 01:02:03.123456789' :: TIMESTAMP as TIMESTAMP_type,
+            '{"foo": "bar"}' :: VARIANT as VARIANT_type,
+            '2020-03-12 01:02:03.123456789' :: TIMESTAMP_LTZ as TIMESTAMP_LTZ_type,
+            '2020-03-12 01:02:03.123456789' :: TIMESTAMP_TZ as TIMESTAMP_TZ_type,
+            '2020-03-12 01:02:03.123456789' :: TIMESTAMP_NTZ as TIMESTAMP_NTZ_type,
+            '0xAAAA' :: BINARY as BINARY_type,
+            '01:02:03.123456789' :: TIME as TIME_type,
+            true :: BOOLEAN as BOOLEAN_type,
+            TO_GEOGRAPHY('LINESTRING(13.4814 52.5015, -121.8212 36.8252)') as GEOGRAPHY_typejk,
+            TO_GEOMETRY('LINESTRING(13.4814 52.5015, -121.8212 36.8252)') as GEOMETRY_type,
```
sfc-gh-aalam, line 1269:
```suggestion
            TO_GEOGRAPHY('LINESTRING(13.4814 52.5015, -121.8212 36.8252)') as GEOGRAPHY_type,
            TO_GEOMETRY('LINESTRING(13.4814 52.5015, -121.8212 36.8252)') as GEOMETRY_type,
```
might be a minor typo

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1202,6 +1202,89 @@ def test_batch_to_pandas_arrow(conn_cnx, result_format):
...
+            )
+
+            batches = cur.get_result_batches()
+            assert len(batches) == 1
+            batch = batches[0]
+            arrow_table = batch.to_arrow()
+            for actual, expected in zip(arrow_table.schema, expected_types):
+                assert (
+                    actual.type == expected
+                ), f"Expected {actual.name} :: {actual.type} column to be of type {expected}"
```
sfc-gh-aalam, line 1285: would it be a good idea to also match data values along with types

src/snowflake/connector/constants.py
```diff
@@ -44,38 +44,75 @@ class FieldType(NamedTuple):
...
+    """
+    Generate the Arrow type represented by the given vector column metadata.
     Vectors are represented as Arrow fixed-size lists.
     """
+    assert (
+        metadata.fields is not None and len(metadata.fields) == 1
+    ), "Invalid result metadata for vector type: expected a single field to be defined"
+    assert (
+        metadata.vector_dimension or 0
+    ) > 0, "Invalid result metadata for vector type: expected a positive dimension"
```
sfc-gh-aling, line 56: I think I have this question in the other mypy PR for snowpark-python, is assertion a good practice in production code? I personally prefer raising explicit error (e.g., ValueError here) if some error needs to be called out explicitly and should halt the program.

@sfc-gh-mkeller @sfc-gh-bwarsaw what are your thoughts?

test/integ/pandas/test_arrow_pandas.py
```diff
@@ -1202,6 +1202,89 @@ def test_batch_to_pandas_arrow(conn_cnx, result_format):
...
+                pyarrow.string(),
+                pyarrow.list_(pyarrow.float64(), 5),
+            )
+
+            if enable_structured_types:
+                for param in {
+                    "ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE",
+                    "IGNORE_CLIENT_VESRION_IN_STRUCTURED_TYPES_RESPONSE",
+                    "FORCE_ENABLE_STRUCTURED_TYPES_NATIVE_ARROW_FORMAT",
+                }:
```
sfc-gh-aalam, line 1235: can we unset them afterwards or use `_statement_params` to populate these values so that this change is contained in here

sfc-gh-aalam (approved):
minor comments, overall looks good

DESCRIPTION.md
```diff
@@ -20,6 +20,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
     - Add `SNOWFLAKE_AUTH_SOCKET_MSG_DONTWAIT` flag (usage: `SNOWFLAKE_AUTH_SOCKET_MSG_DONTWAIT=true`) to make a non-blocking socket.recv call and retry on Error
       - Consider using this if running in a containerized environment and externalbrowser auth frequently hangs while waiting for callback
       - NOTE: this has not been tested extensively, but has been shown to improve the experience when using WSL
+  - Added support for strucutured type information in ResultMetadataV2
```
sfc-gh-aling, line 23: I'm wondering if this should be called out in the changelog or rephrased as `ResultMetadataV2` seems to be internal implementation details

src/snowflake/connector/cursor.py
```diff
@@ -201,28 +201,35 @@ def from_column(cls, col: dict[str, Any]) -> ResultMetadataV2:
...
-        if type_code == FIELD_NAME_TO_ID["VECTOR"] and col.get("fields") is not None:
-            fields = [
-                ResultMetadataV2.from_column({"name": None, **f}) for f in col["fields"]
-            ]
+        fields = col.get("fields")
+        processed_fields: Optional[List[ResultMetadataV2]] = None
+        if fields is not None:
+            if col_type in {"VECTOR", "ARRAY", "OBJECT", "MAP"}:
+                processed_fields = [
+                    ResultMetadata.from_column({"name": None, **f})
```
sfc-gh-aling, line 215: should be ResultMetadataV2?
```suggestion
                    ResultMetadataV2.from_column({"name": None, **f})
```

test/integ/pandas/test_arrow_pandas.py
sfc-gh-aling, line 1285: shall we also compare the values?

sfc-gh-aling (changes requested):
implementation looks good to me, but have some questions on the error experience

src/snowflake/connector/constants.py
sfc-gh-jrose: In my opinion assertions are useful in places where we have a soft guarantee that something is already true. Here for example we would never expect our API to have returned this data without the proper fields. The user can't really do anything that would ever trip this unless we change our APIs so in theory it's safe to run this code without the assertions.

test/integ/pandas/test_arrow_pandas.py
sfc-gh-jrose: The query is that creates the table intentionally includes no rows so comparing values is not possible. The various `*_pa_type` functions added above are only called when the schema has to be created from an empty table. In other cases pyarrow figures out the schema based on the data present instead which may use other pyarrow types than those enumerated in the expected tuple.

sfc-gh-jrose: See my comment on Adam's comment.

src/snowflake/connector/constants.py
sfc-gh-aling: I see, it sounds more like a guarantee of the backend.

so users are not expected to encounter this, if they ever encounter the issue, then it means something goes wrong and we probably need a fix.

PR #1898, opened 2024-03-07
SNOW-996732: improve clean up logic of connection

sfc-gh-aling:

  the `__del__` method is unstable and gets called at random timing.
in one of the recent report, we found the `__del__` is causing circular import issue.

as a short term fix, we fix by moving away from `__del__` method to using `atexit` to clean up connection at intepreter shutdown time.

long term fix plan here: SNOW-1218851

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import atexit
 import logging
 import os
 import pathlib
@@ -434,12 +435,7 @@ class SnowflakeConnection:
 
         # get the imported modules from sys.modules
         self._log_telemetry_imported_packages()
-
-    def __del__(self) -> None:  # pragma: no cover
-        try:
-            self.close(retry=False)
-        except Exception:
-            pass
+        atexit.register(self._close_at_exit)
 
     @property
     def insecure_mode(self) -> bool:
@@ -1748,6 +1744,12 @@ class SnowflakeConnection:
                 sf_qid, None
             )  # Prevent KeyError when multiple threads try to remove the same query id
             self._done_async_sfqids[sf_qid] = None
+
+    def _close_at_exit(self):
+        try:
+            self.close(retry=False)
+        except Exception:
+            pass
 
     def get_query_status(self, sf_qid: str) -> QueryStatus:
         """Retrieves the status of query with sf_qid.
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import atexit
+from contextlib import suppress
 import logging
 import os
 import pathlib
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/connection.py
```diff
@@ -1749,6 +1747,12 @@ def _cache_query_status(self, sf_qid: str, status_ret: QueryStatus) -> None:
             )  # Prevent KeyError when multiple threads try to remove the same query id
             self._done_async_sfqids[sf_qid] = None
 
+    def _close_at_exit(self):
+        try:
```
sfc-gh-bwarsaw, line 1751: Maybe use [contextlib.suppress()](https://docs.python.org/3/library/contextlib.html#contextlib.suppress)?

sfc-gh-bwarsaw:
As mentioned in Slack: One thing to keep in mind is that `atexit` handlers don't get run if `os._exit()` is called or the process crashes.

src/snowflake/connector/connection.py
sfc-gh-aling: this is elegant

sfc-gh-aling:
thanks @sfc-gh-bwarsaw , this should not be a concern to our connector as we didn't perform critical job during the closing stage.

src/snowflake/connector/connection.py
```diff
@@ -439,12 +441,7 @@ def __init__(
 
         # get the imported modules from sys.modules
         self._log_telemetry_imported_packages()
-
-    def __del__(self) -> None:  # pragma: no cover
-        try:
-            self.close(retry=False)
-        except Exception:
-            pass
+        atexit.register(self._close_at_exit)
```
sfc-gh-jrose, line 444: Should there be a comment with the long term ticket number so it's clear this is a temporary solution?

sfc-gh-jrose (approved):

src/snowflake/connector/connection.py
sfc-gh-aling, line 444:
```suggestion
        # check SNOW-1218851 for long term improvement plan
        atexit.register(self._close_at_exit)
```

sfc-gh-aling: updated

test/integ/test_connection.py
```diff
@@ -1342,3 +1342,20 @@ def test_ocsp_mode_insecure(conn_cnx, is_public_test, caplog):
...
+        password=db_parameters["password"],
+        host=db_parameters["host"],
+        port=db_parameters["port"],
+        account=db_parameters["account"],
+        database=db_parameters["database"],
+        protocol=db_parameters["protocol"],
+        timezone="UTC",
+    )
+    cnx._close_at_exit()
+    assert cnx.is_closed()
```
sfc-gh-mkeller, line 1361: I don't like when individual tests handle `db_parameters`, use `conn_cnx` instead! There's a smaller chance to leak the dr aprameters that way.

test/integ/test_connection.py
```diff
@@ -1342,3 +1342,20 @@ def test_ocsp_mode_insecure(conn_cnx, is_public_test, caplog):
             assert "snowflake.connector.ocsp_snowflake" in caplog.text
         else:
             assert "snowflake.connector.ocsp_snowflake" not in caplog.text
+
+
+@pytest.mark.skipolddriver
+def test_connection_atexit_close(db_parameters, capsys):
```
sfc-gh-mkeller, line 1348: `capsys` is unused
```suggestion
def test_connection_atexit_close(db_parameters):
```

test/integ/test_connection.py
sfc-gh-aling: updated, thanks for the feedback!

PR #1907, opened 2024-03-26
SNOW-1269820: Structured Type support for pandas and iceberg.

sfc-gh-jrose:

   Fixes #1269820

   The main purpose of this change is to add tests that show that structured types work with iceberg tables and conversion to pandas. While implementing the tests I found a number of gaps in that functionality which I've either patched to add support, or opened relevant tickets to do so in the future.

  Review Notes:

- CArrowTableIterator.cpp
  - I pulled most of the logic out from `reconstructRecordBatches_nanoarrow` into a new function `convertIfNeeded`. This allows me to call it recursively when parsing nested types. I've added logic to support ARRAY, MAP, and OBJECT to the switch in this function.
  -  Most of the converters in this file have ben slightly modified to no longer rely on indexes, but to instead use pointers. This allows calling them from an arbirary nesting level when parsing structured types.
  - There's a couple spots in the converters where I simplified the code slightly to make it more readable.
- result_batch.py
  - By default arrow would cast maps to something like `[('k1', 1), ('k2', 2)]` which is not the most intuitive thing to work with. pyspark would return `{'k1': 1, 'k2': 2}` which is a better form in my opinion. The parameter added in this file will default to the pyspark style return. This should not be a BCR because this map type is not publicly available yet.
- test_arrow_result.py
  - I've refactored the tests so that they parameterize on the matrix of [datatype, iceberg, pandas]. Each datatype is tested against FDN and iceberg tables and collected to either python objects or pandas dataframes.


test/integ/test_arrow_result.py
```diff
@@ -82,6 +96,87 @@
...
+    )
+    for iceberg, pandas, datatype in itertools.product(
+        [True, False],
+        [True, False] if pandas_available else [False],
+        PRIMITIVE_DATATYPE_EXAMPLES,
+    )
+    # Run all tests when not converting to pandas or using iceberg
+    if iceberg is False
+    # Only run iceberg tests on applicable types
+    or (ICEBERG_SUPPORTED and iceberg and datatype not in ICEBERG_UNSUPPORTED_TYPES)
```
sfc-gh-aling, line 169: what's the connector behavior in case of encountering unsupported types?

src/snowflake/connector/result_batch.py
```diff
@@ -703,7 +703,10 @@ def to_pandas(
         """Returns this batch as a pandas DataFrame"""
         self._check_can_use_pandas()
         table = self.to_arrow(connection=connection)
-        return table.to_pandas(**kwargs)
+        # By default arrow returns maps as kv tuples in order to allow duplicate keys.
+        # Snowflake does not support duplicate keys in maps. In order to better match
+        # pyspark this will return maps as dictionaries which would remove duplicate keys.
+        return table.to_pandas(**{"maps_as_pydicts": "strict", **kwargs})
```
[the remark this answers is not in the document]
sfc-gh-jrose: See the PR notes above for `result_batch.py` for why this does not break the user as well as why this makes user experience better.

test/integ/test_arrow_result.py
sfc-gh-jrose: If an unsupported iceberg type is queried then there will be a server-side error from iceberg.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp
```diff
@@ -15,6 +15,246 @@
...
+    }
+
+    case SnowflakeType::Type::ANY:
+    case SnowflakeType::Type::BINARY:
+    case SnowflakeType::Type::BOOLEAN:
+    case SnowflakeType::Type::CHAR:
+    case SnowflakeType::Type::DATE:
+    case SnowflakeType::Type::REAL:
+    case SnowflakeType::Type::TEXT:
+    case SnowflakeType::Type::VARIANT:
```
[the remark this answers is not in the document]
sfc-gh-jrose: VARIANT and other semi-structured types will continue to be passed back to the client as json encoded strings. Object has both a structured and semi-structured type definition, but as far as I'm aware variant is only semi-structured.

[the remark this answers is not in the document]
sfc-gh-aling: I see, as long as we testes that Variant is still passed back as json string, I'm good with that.

test/integ/test_arrow_result.py
sfc-gh-aling: gotcha, can we add negative test case for those ICEBERG_UNSUPPORTED_TYPES?

test/integ/test_arrow_result.py
```diff
@@ -82,6 +96,91 @@
...
+    os.getenv("cloud_provider", "dev") in STRUCTRED_TYPE_ENVIRONMENTS
+)
+
+PANDAS_UNSUPPORTED_NESTED_TYPES = {
+    # SNOW-1326075: Timestamp types drop information when converted to pandas
+    "TIMESTAMP_LTZ",
+    "TIMESTAMP_NTZ",
+    "TIMESTAMP_TZ",
+    "TIME",
+}
```
sfc-gh-aling, line 148: shall we keep testing those types because it's something that users still encounter.
to users, it's not "unsupported" or would we block the call?

DESCRIPTION.md
```diff
@@ -12,6 +12,8 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
   - Added easy logging configuration so that users can easily generate log file by setup log config in `$SNOWFLAKE_HOME/config.toml`.
   - Improved s3 acceleration logic when connecting to China endpoint.
+  - Added support for structured types to fetch_pandas_all
+  - Added additional testing around structured type use in iceberg tables.
```
sfc-gh-aling, line 16: do we need to mention this in changelog?

DESCRIPTION.md
```diff
@@ -12,6 +12,8 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
   - Added easy logging configuration so that users can easily generate log file by setup log config in `$SNOWFLAKE_HOME/config.toml`.
   - Improved s3 acceleration logic when connecting to China endpoint.
+  - Added support for structured types to fetch_pandas_all
```
sfc-gh-aling, line 15: actually I realize this can be a breaking change to customer.
for example, for a column of Object type, previously we get json string, but now it's the python object so users have to rewrite their code?

are you aware if this is part of a snowflake BCR?

src/snowflake/connector/result_batch.py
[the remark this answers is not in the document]
sfc-gh-aling: thanks.
> this feature is not publicly available yet

just to clarify here by this feature you mean returning a dict inside the pandas DF?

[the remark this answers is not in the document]
sfc-gh-jrose: The feature being structured type returns. As of right now we never return a map to the customer. If they query a map column it is currently returned as string.

test/integ/test_arrow_result.py
```diff
@@ -80,9 +95,108 @@
         ),
     ],
     "VARCHAR": [None, "ascii_test", b"\xf0\x9f\x98\x80".decode("utf-8")],
+    "VARIANT": [None, 1.1, [2.2, 3.3], {"foo": "bar"}],
+}
+
+# Some datatypes do not match 1:1 with the original data
+PANDAS_REPRS = dict()
+PANDAS_ARRAY_REPRS = dict()
```
sfc-gh-aling, line 103: is this used?

src/snowflake/connector/cursor.py
```diff
@@ -1314,7 +1314,16 @@ def fetch_pandas_batches(self, **kwargs: Any) -> Iterator[DataFrame]:
...
-        """Fetch Pandas dataframes in batches, where 'batch' refers to Snowflake Chunk."""
+        """
+        Fetch Pandas dataframes in batches, where 'batch' refers to Snowflake Chunk.
+
+        Returns:
+            A pandas dataframe containing the cursors query results.
+
+        Note:
+            Timestamp types that are nested in structured types may not be accurately represented
+            due to a limitation in the pyarrow to pandas conversion.
```
sfc-gh-aling, line 1325: I'm wondering is there a link we can provide to which points to a well-known issue or official doc from pyarrow about this part?

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp
```diff
@@ -15,6 +15,288 @@
...
+          if (columnSchemaView.schema->n_children != 1) {
+            std::string errorInfo = Logger::formatString(
+                "[Snowflake Exception] invalid arrow schema for map entries "
+                "expected 1 "
+                "schema child, but got %d",
+                columnSchemaView.schema->n_children);
+            logger->error(__FILE__, __func__, __LINE__, errorInfo.c_str());
+            PyErr_SetString(PyExc_Exception, errorInfo.c_str());
+            break;
+          }
```
sfc-gh-aling, line 133: I'm comparing the code with CArrowChunkIterator: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowChunkIterator.cpp#L440-L444

do we need to add the `case NANOARROW_TYPE_STRING: {` into the CArrowChunkIterator.cpp when handling the `Map` case?

test/integ/test_arrow_result.py
```diff
@@ -53,8 +68,8 @@
     "INTEGER": [
         None,
         0,
-        99999999999999999999999999999999999999,
-        -99999999999999999999999999999999999999,
+        999999999,
+        -999999999,
     ],
```
sfc-gh-aling, line 73: I'm curious why we changed the test value here

sfc-gh-jrose: Iceberg does not support such large numbers in integer types. I could add another exception for this case if you think it's important for the base test.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp
sfc-gh-jrose: Currently the table iterator does no conversion for String columns along with most other primitives:
https://github.com/snowflakedb/snowflake-connector-python/blob/v3.9.0/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.cpp#L109-L119
This case is basically saying that we trust pyarrow to convert a string correctly by default.

test/integ/test_arrow_result.py
sfc-gh-aling: yeah my concern was we lack coverage for large ints, but as long as it's covered by other tests we should be good

sfc-gh-aling (approved):

PR #1909, opened 2024-04-01
SNOW-1292905:Implement configuration log parser 

sfc-gh-yuwang:

   Fixes #SNOW-1292905

   Design doc of easy logging for python connector is here: https://docs.google.com/document/d/17DLoaICHX1wU0qYTvJpQ4r1qKxGeJQYMZJvyrZi3Ctk/edit?usp=sharing


sfc-gh-yuwang:
Hi @sfc-gh-bwarsaw I remembered that you said we could use toml file instead of a json file, can you give some suggestions here?

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,113 @@
...
+            connection_parameters["CLIENT_CONFIG_FILE"]
+            if connection_parameters and "CLIENT_CONFIG_FILE" in connection_parameters
+            else None
+        )
+        self.SF_CLIENT_CONFIG_FILE = (
+            os.environ["SF_CLIENT_CONFIG_FILE"]
+            if "SF_CLIENT_CONFIG_FILE" in os.environ
+            else None
+        )
+        self.config_file_name = "sf_client_config.json"
```
sfc-gh-igarish, line 35: Python connector has toml file, still we need a new file?

sfc-gh-yuwang: Let's wait for @sfc-gh-bwarsaw  's input on this topic

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,92 @@
+#
+# Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
+#
+
+from __future__ import annotations
```
sfc-gh-bwarsaw, line 5: Also unnecessary.

src/snowflake/connector/log_configuration.py
sfc-gh-bwarsaw: Agree with @sfc-gh-igarish - is there a requirement that the log config file must be JSON, or some other reason this can't be a section in the `~/snowflake/config.toml` file?   If there isn't then I suggest adding a `[logging]` section with `level` and `path` keys to that TOML file.  Then you wouldn't have to also add support for another environment variable because we already have env overload of the location of the TOML file.

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,113 @@
...
+        self.log_file_name = "python_connector.log"
+        self.log_levels = logging._nameToLevel
```
sfc-gh-bwarsaw, line 37: It's not great to use a non-public attribute of the `logging` module because that API isn't guaranteed to be stable. 

`logging.setLevel()` will throw a `ValueError` if given a bogus level string, so can we just rely on that to validate the level?  The other approach would be to do something like `getattr(logging, level)` on the string value and catch the resulting `AttributeError` when the level string is invalid.

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,113 @@
...
+        self.parse_config_file()
+
+    def parse_config_file(self):
```
sfc-gh-bwarsaw, line 40: I think both `parse_config_file()` and `search_config_file()` go away if we use the TOML file.

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,113 @@
...
+        if self.SF_CLIENT_CONFIG_FILE:
+            if is_full_path(self.SF_CLIENT_CONFIG_FILE):
+                return self.SF_CLIENT_CONFIG_FILE
+            else:
+                raise FileNotFoundError(
+                    f"given file path {self.SF_CLIENT_CONFIG_FILE} is not full path"
+                )
+
+        # search under driver directory
+        drive_directory = os.path.dirname(inspect.getfile(SnowflakeConnection))
```
sfc-gh-bwarsaw, line 85: While I think all this goes away with TOML file approach, if not, let's discuss this search algorithm.

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,113 @@
...
+        return ""
+
+    def create_log(self):
+        logging.basicConfig(
+            filename=self.log_path, level=logging.getLevelName(self.log_level)
+        )
+        for logger_name in ["snowflake.connector", "botocore", "boto3"]:
+            logger = logging.getLogger(logger_name)
+            logger.setLevel(logging.getLevelName(self.log_level))
+            ch = logging.FileHandler(os.path.join(self.log_path, self.log_file_name))
```
sfc-gh-bwarsaw, line 106: I think if we use the TOML file approach, the `[logging]path` value should be the full path to the log file, and it must be an absolute path.  If we allow for relative paths, then we have to think about where that path is relative to, and I'd rather not guess that, or put it in the driver directory, etc.  I suggest a default path of `$SNOWFLAKE_HOME/logs` and any relative path would be relative to that default.  (`$SNOWFLAKE_HOME` of course defaults to `~/.snowflake/`.

sfc-gh-bwarsaw (changes requested):
Thanks for working on this!  TL;DR: we should use the TOML file to configure the log location and level.  See detailed suggestions inline.

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,64 @@
...
+        self.log_level = None
+        self.log_file_name = None
+        self.parse_config_file()
+
+    def parse_config_file(self):
+        CONFIG_MANAGER.read_config()
+        data = CONFIG_MANAGER["connections"]
+        if "common" in data:
+            if "log_level" not in data["common"] or "log_path" not in data["common"]:
+                raise ValueError(
```
sfc-gh-bwarsaw, line 29: Do you think it should be an error if they don't have logging configured, or do you think that we should just disable logging if it's not configured?  Won't this break existing `config.toml` files?

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,64 @@
...
+class EasyLoggingConfigPython:
+    def __init__(self):
+        self.log_path = None
+        self.log_level = None
+        self.log_file_name = None
+        self.parse_config_file()
+
+    def parse_config_file(self):
+        CONFIG_MANAGER.read_config()
+        data = CONFIG_MANAGER["connections"]
```
sfc-gh-bwarsaw, line 26: I'd like @sfc-gh-mkeller and @sfc-gh-turbaszek to weigh in on where we want to put the logging settings.  A couple of things to consider:

* Should each connection be able to log in a different place?
* We have precedence in how the [CLI sets up logging](https://docs.snowflake.com/en/developer-guide/snowflake-cli-v2/connecting/configure-logging) so we should at least be consistent with that approach.  I'm not saying that connector logging should go in `[cli.logs]` but we want at least the same key names and setting semantics (unless, if we want to change it we should work with that team to keep consistent).

My gut feeling is that we want a top level `[logs]` section in `config.toml` with the same settings and semantics as in the `[cli.logs]` section.  `[cli.logs]` could inherit or override this (eventually) and if we want per-connection logging, they would alos inherit or override.  We also need to update or write any public documentation about the format for `config.toml`.

Maybe a separate 1-pager to define the semantics and syntax would help, with at least Mark, Tomasz, and myself as reviewers?

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,64 @@
...
+        if "common" in data:
+            if "log_level" not in data["common"] or "log_path" not in data["common"]:
+                raise ValueError(
+                    f"config file at {CONNECTIONS_FILE} is not in correct form, please verify your config file"
+                )
+            data = data["common"]
```
sfc-gh-bwarsaw, line 32: How about something like this for a more efficient approach?  (Keep in mind `:=` operator was added in Python 3.8).

```python
if common := data.get('common'):
    self.log_level = data.get('level')
    self.path = data.get('path')
    self.save_logs = data.get('save_logs')
    # ... do value sanity checking
```

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,64 @@
...
+            # if log path does not exist, create it, else check accessibility
+            if not os.path.exists(self.log_path):
+                os.makedirs(self.log_path, exist_ok=True)
+            elif not os.access(self.log_path, os.R_OK | os.W_OK):
+                raise PermissionError(
+                    f"log path: {self.log_path} is not accessible, please verify your config file"
+                )
+            if not os.path.isabs(self.log_path):
+                raise FileNotFoundError(
+                    f"given log path {self.log_path} is not full path"
```
sfc-gh-bwarsaw, line 45: Suggest: `f"Log path must be an absolute file path: {self.log_path}"`

sfc-gh-bwarsaw (changes requested):

src/snowflake/connector/log_configuration.py
sfc-gh-yuwang: The `common` area in config.toml is meant for easy logging. So if you look at the code, it is disabled when `common` not in connections.toml. 
When `common` in config.toml, it means users want to generate logs, so we should warn them when it is not correctly configured.
This should not break config.toml since dictionaries are read from it and we just add one key-value pair into it

src/snowflake/connector/log_configuration.py
sfc-gh-yuwang: I think we should keep all logs at the same place.
For semantics of config file, I think we should follow the design doc of easy logging and keep it the same as other drivers, otherwise wouldn't it make this project unnecessary? Correct me if I understand your comment wrong

sfc-gh-bwarsaw: > design doc of easy logging

Do you mean: https://docs.google.com/document/d/1zV3E-288HxJKJv5X0ct1lxXcuYGNMcF3--P30xdxJZk/edit#heading=h.5o4hope6ubn0 ?

If so, I'm concerned that we don't have alignment on that yet, and also that SnowCLI has forged a different path, which -- if that's temporary then fine, but if not won't it be confusing that our client side tools and libraries all have different ways to specify log configuration?

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,80 @@
...
+    else:
+        os.makedirs(file_path, exist_ok=True)
+        with open(CONFIG_FILE, "w") as f:
+            connection_content = {
+                "log": {
+                    "level": "wrong",
+                }
+            }
+            f.write(tomlkit.dumps(connection_content))
+    yield
```
sfc-gh-bwarsaw, line 33: I've talked about this in other PRs, but you have to wrap the `yield` in a `try`/`finally`.  The reason being that the `yield` hands execution over to the context manager calling this fixture, but if any exceptions occur in the both of the context manager, it will pop up out of this function and code after the `yield` will not get executed.  I'm not 100% sure how fixtures are invoked, but it's good practice anyway.  E.g.:

```python
def fixture():
    # Set everything up.
    try:
        yield
    finally:
        # Clean everything up.
```

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,80 @@
...
+    with open(CONFIG_FILE, "w") as f:
+        connection_content.pop("log")
+        f.write(tomlkit.dumps(connection_content))
+
+
+@pytest.fixture(scope="function")
+def setup2():
```
sfc-gh-bwarsaw, line 40: Similarly, a more descriptive name would be helpful.  Also, is there any code you can share between the two fixtures, and can you think of a way to refactor them so as not to have to duplicate a bunch of code?

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,80 @@
...
+    file_path = os.path.dirname(CONFIG_FILE)
```
sfc-gh-bwarsaw, line 41: Similar to above, please use a temp file.

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,80 @@
...
+        os.makedirs(file_path, exist_ok=True)
+        with open(CONFIG_FILE, "w") as f:
+            connection_content = {
+                "log": {
+                    "level": "DEBUG",
+                    "path": "incomplete_path/",
+                }
+            }
+            f.write(tomlkit.dumps(connection_content))
+    yield
```
sfc-gh-bwarsaw, line 61: Similar to above, please wrap the `yield`.

sfc-gh-bwarsaw (changes requested):
Thanks again for working on this.  You're definitely in the right direction, and there are just a few things to clean up.

Also, while I don't think there's a changelog in this repo, this feature is significant enough to warrant a changelog and documentation updates.

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,68 @@
...
+        self.level = None
+        self.save_logs = None
+        self.log_file_name = None
+        self.parse_config_file()
+
+    def parse_config_file(self):
+        CONFIG_MANAGER.read_config()
+        data = CONFIG_MANAGER.conf_file_cache
+        if log := data.get("log"):
+            self.save_logs = log.get("save_logs") if log.get("save_logs") else False
```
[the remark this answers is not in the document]
sfc-gh-yuwang: Sure, will change it. Toml file support bool value so I think we don't need to do a conversion

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+
+            # if log path does not exist, create it, else check accessibility
+            if not os.path.exists(self.path):
+                os.makedirs(self.path, exist_ok=True)
+            elif not os.access(self.path, os.R_OK | os.W_OK):
+                raise PermissionError(
+                    f"log path: {self.path} is not accessible, please verify your config file"
+                )
+            if not os.path.isabs(self.path):
```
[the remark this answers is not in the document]
sfc-gh-yuwang: got it

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,65 @@
...
+                os.makedirs(self.path, exist_ok=True)
+            elif not os.access(self.path, os.R_OK | os.W_OK):
+                raise PermissionError(
+                    f"log path: {self.path} is not accessible, please verify your config file"
+                )
+
+    # create_log() is called outside __init__() so that it can be easily turned off
+    def create_log(self):
+        if self.save_logs:
+            self.log_file_name = "python-connector.log"
```
sfc-gh-bwarsaw, line 48: I would make this a module global constant, e.g. at the top level:

```python
LOG_FILE_NAME = 'python-connector.log'
```

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,83 @@
...
+from pathlib import Path
+
+import pytest
+import tomlkit
+
+from snowflake.connector import EasyLoggingConfigPython
+from snowflake.connector.config_manager import CONFIG_MANAGER
+from snowflake.connector.constants import CONFIG_FILE
+
+THIS_DIR = os.path.dirname(os.path.realpath(__file__))
```
sfc-gh-bwarsaw, line 16: pytest already provides a [temp file fixture](https://docs.pytest.org/en/latest/reference/reference.html#tmp-path) you should use.  In general it's not a good idea to use directories inside the source tree for temp files and directories.  See also [the tutorial](https://docs.pytest.org/en/latest/how-to/tmp_path.html#tmp-path).  This should simplify a bunch of the test setups.

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,83 @@
...
+    EasyLoggingConfigPython()
+    assert os.path.exists(NONEXIST_PATH)
+
+
+@pytest.mark.parametrize("config_file_setup", ["inabsolute_path"], indirect=True)
+@pytest.mark.skipolddriver
+def test_config_file_inabsolute_path(config_file_setup):
+    try:
+        EasyLoggingConfigPython()
+    except FileNotFoundError as e:
```
sfc-gh-bwarsaw, line 70: Should this use [pytest.raises()](https://docs.pytest.org/en/latest/reference/reference.html#pytest.raises)?

test/unit/test_easy_logging.py
```diff
@@ -0,0 +1,83 @@
...
+            os.rmdir(NONEXIST_PATH)
+        if os.path.exists(INACCESSIBLE_PATH):
+            os.rmdir(INACCESSIBLE_PATH)
+
+
+@pytest.mark.parametrize("config_file_setup", ["nonexist_path"], indirect=True)
+@pytest.mark.skipolddriver
+def test_config_file_nonexist_path(config_file_setup):
+    assert not os.path.exists(NONEXIST_PATH)
+    EasyLoggingConfigPython()
```
sfc-gh-bwarsaw, line 61: If I'm reading the code right, doesn't this have to call `.create_log()` on the `EasyLoggingConfigPython` instance?  If not, then I think some other test does need to cover `.create_log()`.

sfc-gh-yuwang:
Test older driver is expected to fail since I added a new module that does not exist in older drivers

sfc-gh-bwarsaw (changes requested):
Very close now!  Just a couple of more suggestions.

test/unit/test_easy_logging.py
sfc-gh-yuwang: I planned to cover `.create_log()` in next step, which will initiate easy logging in snowflake connection and a integ test that test is as a whole

sfc-gh-bwarsaw: By "next step" do you mean a follow up PR, or another commit in this PR?  A wise person once said "untested code is broken code!" 😄

sfc-gh-yuwang: In the next follow up PR, I'll definitely add some test on that code

sfc-gh-bwarsaw: I think two simple unit tests could be added here.  One that sets `save_log = True`, writes a log message, and verifies that the log message was written. You can use the [caplog](https://docs.pytest.org/en/latest/reference/reference.html#caplog) fixture.  The second would be similar but would set `save_log = False` and it would check that the log entry was _not_ written.

sfc-gh-yuwang: I did some test, turns out caplog is not suitable here, since python connector will always produce log that caplog can capture. Instead, I am going to check whether there is log in `path` provided by config.toml. I think this is enough for testing `create_log()`

sfc-gh-bwarsaw (approved):

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,65 @@
...
+            self.save_logs = log.get("save_logs", False)
+            self.level = log.get("level", "INFO")
+            self.path = log.get("path", os.path.join(DIRS.user_config_path, "logs"))
+
+            if not os.path.isabs(self.path):
+                raise FileNotFoundError(
+                    f"Log path must be an absolute file path: {self.path}"
+                )
+            # if log path does not exist, create it, else check accessibility
+            if not os.path.exists(self.path):
```
sfc-gh-turbaszek, line 39: Wouldn't it be clearer if we would use `pathlib.Path` instead of `os.path`?

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,65 @@
...
+from snowflake.connector.secret_detector import SecretDetector
+
+LOG_FILE_NAME = "python-connector.log"
+
+
+class EasyLoggingConfigPython:
+    def __init__(self):
+        self.path = None
+        self.level = None
+        self.save_logs = None
```
sfc-gh-turbaszek, line 23: Can we please add type hints?

src/snowflake/connector/log_configuration.py
```diff
@@ -0,0 +1,65 @@
...
+        self.parse_config_file()
+
+    def parse_config_file(self):
+        CONFIG_MANAGER.read_config()
+        data = CONFIG_MANAGER.conf_file_cache
+        if log := data.get("log"):
```
sfc-gh-turbaszek, line 29: My personal preference would be to do `if not log := data.get("log")` and `return`. In this way later code is not indented and feels easier to read

sfc-gh-bwarsaw: I generally prefer positive tests over negative tests, especially if there's an `else` clause, because positive tests are easier to reason about.  This is a good case where the opportunity for an early exit and dedentation overrides that for readability.  Good catch. 

I also generally prefer more narrow conditional tests than false-yness (e.g. `if not`) but in this case because of the assignment operator, testing against `None` doesn't improve readability or understanding.

PR #1919, opened 2024-04-11
SNOW-1292908:Initiatie log parser in SnowflakeConnection

sfc-gh-yuwang:

   Fixes #/SNOW-1292908

### DIFF
--- a/test/integ/test_easy_logging.py
+++ b/test/integ/test_easy_logging.py
@@ -2,16 +2,16 @@
 # Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
 #
 
-from __future__ import annotations
-
-import os
-from logging import getLogger
-from pathlib import Path
 from test.integ.conftest import create_connection
 
 import pytest
 
 pytestmark = pytest.mark.skipolddriver
+
+import os.path
+from logging import getLogger
+from pathlib import Path
+
 try:
     import tomlkit
 
@@ -56,12 +56,13 @@ def test_save_logs(db_parameters, config_file_setup, log_directory):
     assert os.path.exists(os.path.join(log_directory, "python-connector.log"))
     with open(os.path.join(log_directory, "python-connector.log")) as f:
         data = f.read()
+    try:
         assert "Snowflake Connector for Python" in data
-
-    # set logger back to default
-    getLogger("snowflake.connector").setLevel(10)
-    getLogger("botocore").setLevel(0)
-    getLogger("boto3").setLevel(0)
+    finally:
+        # set logger back to default
+        getLogger("snowflake.connector").setLevel(10)
+        getLogger("botocore").setLevel(0)
+        getLogger("boto3").setLevel(0)
 
 
 @pytest.mark.parametrize("config_file_setup", ["no_save_logs"], indirect=True)
--- a/test/integ/test_easy_logging.py
+++ b/test/integ/test_easy_logging.py
@@ -10,10 +10,15 @@ from pathlib import Path
 from test.integ.conftest import create_connection
 
 import pytest
-import tomlkit
 
-from snowflake.connector.config_manager import CONFIG_MANAGER
-from snowflake.connector.constants import CONFIG_FILE
+pytestmark = pytest.mark.skipolddriver
+try:
+    import tomlkit
+
+    from snowflake.connector.config_manager import CONFIG_MANAGER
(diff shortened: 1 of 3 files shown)


test/integ/test_easy_logging.py
```diff
@@ -0,0 +1,71 @@
...
+import os
+from logging import getLogger
+from pathlib import Path
+from test.integ.conftest import create_connection
+
+import pytest
+
+pytestmark = pytest.mark.skipolddriver
+try:
+    import tomlkit
```
sfc-gh-bwarsaw, line 16: Newer Pythons come with [tomllib](https://docs.python.org/3/library/tomllib.html) so we should not add an unnecessary dependency on `tomlkit`.  E.g. the code should read something like:

```python
try:
    from tomllib import dumps
except ModuleNotFoundError:
    from tomlkit import dumps
```

test/integ/test_easy_logging.py
```diff
@@ -0,0 +1,71 @@
...
+
+    from snowflake.connector.config_manager import CONFIG_MANAGER
+    from snowflake.connector.constants import CONFIG_FILE
+except ModuleNotFoundError:
+    pass
```
sfc-gh-bwarsaw, line 21: Is it really best to continue on if these imports fail?  Won't you get name errors in some of the tests if you do?  Also, should the `snowflake.connector` imports actually ever fail given that this is testing that package?

test/integ/test_easy_logging.py
```diff
@@ -0,0 +1,71 @@
...
+def test_save_logs(db_parameters, config_file_setup, log_directory):
+    create_connection("default")
+
+    assert os.path.exists(os.path.join(log_directory, "python-connector.log"))
+    with open(os.path.join(log_directory, "python-connector.log")) as f:
+        data = f.read()
+        assert "Snowflake Connector for Python" in data
+
+    # set logger back to default
+    getLogger("snowflake.connector").setLevel(10)
```
sfc-gh-bwarsaw, line 62: If the `assert` above fails, these resets won't get run, so the test could leave things in a dirty state.

test/integ/test_easy_logging.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.mark.parametrize("config_file_setup", ["save_logs"], indirect=True)
+def test_save_logs(db_parameters, config_file_setup, log_directory):
+    create_connection("default")
+
+    assert os.path.exists(os.path.join(log_directory, "python-connector.log"))
+    with open(os.path.join(log_directory, "python-connector.log")) as f:
+        data = f.read()
+        assert "Snowflake Connector for Python" in data
```
sfc-gh-bwarsaw, line 59: Probably better to move the `assert` outside of the `with` statement.

sfc-gh-bwarsaw (changes requested):

test/integ/test_easy_logging.py
sfc-gh-yuwang: This approach is to pass the old driver test in merge gate, that merge gate will test against a old python connector which does not have this packages

test/integ/test_easy_logging.py
sfc-gh-yuwang: I'll add a try: finally to it

test/integ/test_easy_logging.py
sfc-gh-yuwang: will do

test/integ/test_easy_logging.py
sfc-gh-yuwang: but this only exists in 3.11 right? We still have to use tomlkit if we want to support older pythons?

sfc-gh-bwarsaw: Yes, but we should only be conditionally depending on `tomlkit; python < 3.11` so that we don't pull in that extra dependency when we don't need it.  After chatting in Slack and looking, that's a different issue which should be fixed independently, perhaps when we refactor the config manager into its own library.  For now, we'll leave this as is.

test/integ/test_easy_logging.py
sfc-gh-bwarsaw: Okay, I don't understand why this is necessary.  I'll start a separate conversation with you and Mark

sfc-gh-bwarsaw (approved):

PR #1938, opened 2024-05-03
SNOW-1013595: Add support `token_file_path` in connection parameters

sfc-gh-jdu:

   Fixes #NNNN

   This is for reading the token file in SPCS now.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -8,6 +8,10 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 
+- v3.11.0(TBD)
+
+  - Added support for `token_file_path` connection parameter to read an OAuth token from a file when connecting to Snowflake.
+
 - v3.10.0(April 29,2024)
 
   - Added support for structured types to fetch_pandas_all.
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -1182,6 +1182,11 @@ class SnowflakeConnection:
             ]:
                 self._authenticator = auth_tmp
 
+        # read OAuth token from
+        if "token_file_path" in kwargs:
+            with open(kwargs["token_file_path"]) as f:
+                self._token = f.read()
+
         if not (self._master_token and self._session_token):
             if not self.user and self._authenticator != OAUTH_AUTHENTICATOR:
                 # OAuth Authentication does not require a username
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -359,6 +359,7 @@ class SnowflakeConnection:
           This parameter is only effective when the result format is JSON.
         server_session_keep_alive: When true, the connector does not destroy the session on the Snowflake server side
           before the connector shuts down. Default value is false.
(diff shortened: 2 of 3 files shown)


src/snowflake/connector/connection.py
```diff
@@ -1182,6 +1182,11 @@ def __config(self, **kwargs):
             ]:
                 self._authenticator = auth_tmp
 
+        # read OAuth token from
```
sfc-gh-aling, line 1185: you might want to add doc/default value for it:

doc: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/connection.py#L310

and default value: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/connection.py#L157

src/snowflake/connector/connection.py
```diff
@@ -1182,6 +1182,11 @@ def __config(self, **kwargs):
...
+        if "token_file_path" in kwargs:
+            with open(kwargs["token_file_path"]) as f:
+                self._token = f.read()
```
sfc-gh-aling, line 1188: will `self._token` be overwritten if there is another authentication method being provided?

DESCRIPTION.md
```diff
@@ -8,6 +8,10 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 
+- v3.11.0(TBD)
+
+  - Added support for `token_file_path` connection parameter to read an OAuth token from a file when connecting to Snowflake.
```
sfc-gh-aling, line 13: why we prefer file input rather than a string/byte input in this case?

sfc-gh-jdu: In SPCS, we tried to avoid that because the token file is generated by snowflake and refreshed every 10 minutes

src/snowflake/connector/connection.py
sfc-gh-jdu: from the code, it seems self._token is for OAuth only

src/snowflake/connector/connection.py
sfc-gh-jdu: We can just reuse `token`

sfc-gh-jdu: Default value may not be required because we will reuse `token`

sfc-gh-jdu:
Hey @sfc-gh-sichen, FYI we will use `token_file_path` for config.toml file.

src/snowflake/connector/connection.py
```diff
@@ -1182,6 +1183,11 @@ def __config(self, **kwargs):
             ]:
                 self._authenticator = auth_tmp
 
+        # read OAuth token from
+        if "token_file_path" in kwargs:
```
sfc-gh-aling, line 1187: nit: I think you can do `if kwargs.get("token_file_path"):` to achieve:

1.  check `token_file_path` is in kwargs
2. check if `token_file_path` in the kwargs is None

sfc-gh-jdu: done

sfc-gh-aling (approved):

PR #1964, opened 2024-06-04
Python connector changes for server-side Snowpark

sfc-gh-azwiegincew:
These changes support a wider project of moving Snowpark query processing to the backend.

The overall architecture is documented here: https://docs.google.com/document/d/1SrNl8nnBoixsr503HP3vq0buPpte2E43JbxPx8hqrO4/edit?tab=t.0#heading=h.yxzchrm647id.

Our initial target as well as all connector changes are documented here: https://docs.google.com/document/d/1oAuFspVgqyNnV3n3piKn2W9goUycfjCHQ4unUCOu_rU/edit.

These changes all amount to optionally passing a new argument to the Query API, namely `dataframeAst`, which replaces (Phases 1 – onward) or sits alongside (Phase 0) the `sqlText` argument.

I have confirmed that it's ok to pass the `dataframeAst` value to the server even the server doesn't handle it. We have a separate set of changes planned to actually do something with the value, but these changes aren't blocking this PR or shipping the resulting Connector.


src/snowflake/connector/connection.py
```diff
@@ -1287,6 +1287,7 @@ def cmd_query(
         _update_current_object: bool = True,
```
sfc-gh-lspiegelberg, line 1287: Looking at https://peps.python.org/pep-0249/#connection-objects, this seems like a change which would not affect the pep249 protocol. Should be good to go.

sfc-gh-azwiegincew:
Ping. This has been sitting for 2 weeks.

sfc-gh-aalam (approved):

sfc-gh-lspiegelberg (approved):
lgtm, let's 🚢

DESCRIPTION.md
```diff
@@ -12,6 +12,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added support for `token_file_path` connection parameter to read an OAuth token from a file when connecting to Snowflake.
   - Added support for `debug_arrow_chunk` connection parameter to allow debugging raw arrow data in case of arrow data parsing failure.
   - Added support for `disable_saml_url_check` connection parameter to disable SAML URL check in OKTA authentication.
+  - Added support for the `dataframe_ast` parameter, which allows the caller to propagate dataframe-level AST to Snowflake.
```
sfc-gh-aling, line 15: nit: 3.11.0 is released, this would go to version 3.12.0.

btw is this a public facing feature? if not we probably do not need to call it out, or if it's public facing feature, then it's better to call that out which api this new parameter is added to

sfc-gh-aling (commented):
do not forget to duplicate the change in the stored proc connector: https://github.com/snowflakedb/Stored-Proc-Python-Connector

sfc-gh-aling (approved):

sfc-gh-vbudati (approved):

PR #1987, opened 2024-06-29
SNOW-1296805 Remove option tmpdir from SnowflakeEncryptionUtil.decrypt_file

sfc-gh-stan:

   Fixes SNOW-1296805

 https://github.com/snowflakedb/snowflake-connector-python/blob/972accde29792fb06fbfd068b870022bc7682f82/src/snowflake/connector/storage_client.py#L382 is problematic where self.tmp_dir are on a different device than target directory (`self.full_dst_file_name`) in a network filesystem (e.g. persistent storage mounted to a docker container). Cross device file operations could cause the program to hang (especially on windows machines). We can eliminate this by making the decrypted intermediate file is written to the same directory as target directory (`self.full_dst_file_name`), where we know is writable.

### DIFF
--- a/src/snowflake/connector/encryption_util.py
+++ b/src/snowflake/connector/encryption_util.py
@@ -194,7 +194,7 @@ class SnowflakeEncryptionUtil:
         in_filename: str,
         chunk_size: int = 64 * kilobyte,
     ) -> str:
-        """Decrypts a file and stores the output in the same dierctory of the incoming file.
+        """Decrypts a file and stores the output in the same directory of the incoming file.
 
         Args:
             metadata: The file's metadata input.
--- a/src/snowflake/connector/encryption_util.py
+++ b/src/snowflake/connector/encryption_util.py
@@ -193,23 +193,23 @@ class SnowflakeEncryptionUtil:
         encryption_material: SnowflakeFileEncryptionMaterial,
         in_filename: str,
         chunk_size: int = 64 * kilobyte,
-        tmp_dir: str | None = None,
     ) -> str:
-        """Decrypts a file and stores the output in the temporary directory.
+        """Decrypts a file and stores the output in the same dierctory of the incoming file.
 
         Args:
             metadata: The file's metadata input.
             encryption_material: The file's encryption material.
             in_filename: The name of the input file.
             chunk_size: The size of read chunks (Default value = block_size * 4 * 1024).
-            tmp_dir: Temporary directory to use, optional (Default value = None).
 
         Returns:
             The decrypted file's location.
         """
-        temp_output_file = f"{os.path.basename(in_filename)}#{random_string()}"
-        if tmp_dir:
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/encryption_util.py
```diff
@@ -193,23 +193,23 @@ def decrypt_file(
         encryption_material: SnowflakeFileEncryptionMaterial,
         in_filename: str,
         chunk_size: int = 64 * kilobyte,
-        tmp_dir: str | None = None,
     ) -> str:
-        """Decrypts a file and stores the output in the temporary directory.
+        """Decrypts a file and stores the output in the same dierctory of the incoming file.
```
sfc-gh-mkeller, line 197:
```suggestion
        """Decrypts a file and stores the output in the same directory of the incoming file.
```

sfc-gh-mkeller (commented):
Great job. Same thing should be done to `encrypt_file` and forever put temp directory in the rear-view mirror.
Also, be aware that we started down on the [`.part` naming](https://github.com/snowflakedb/snowflake-connector-python/blob/d7ab2feaaee018c9adec5914433b42b7c00a5af9/src/snowflake/connector/storage_client.py#L99), maybe we should follow that.

sfc-gh-mkeller:
Ohh and unfortunately this is a breaking change that needs to be announced ahead of time!

sfc-gh-stan:
> Great job. Same thing should be done to `encrypt_file` and forever put temp directory in the rear-view mirror. Also, be aware that we started down on the [`.part` naming](https://github.com/snowflakedb/snowflake-connector-python/blob/d7ab2feaaee018c9adec5914433b42b7c00a5af9/src/snowflake/connector/storage_client.py#L99), maybe we should follow that.

`encrypt_file` should be immune from the cross-device file operation issue because it just writes the encrypted file to a temp dir and later deletes that file. It's also more likely that the we don't have write access to the incoming directory. I prefer keeping tmpdir in `encrypt_file` for that reason.

For `.part` naming, what `decrypt_file` returns should look like `<filename>.part<random_suffix>`. Immediately after `decrypt_file` returns, this file gets renamed to just the final filename  and the part file is unlinked: https://github.com/snowflakedb/snowflake-connector-python/blob/706bc2f6ae58da25885064497c217cfedcc19020/src/snowflake/connector/storage_client.py#L389. I don't think we have to worry about the filenames here too much?

sfc-gh-stan:
> Ohh and unfortunately this is a breaking change that needs to be announced ahead of time!

Thanks for pointing this out, I will start the BCR process.

sfc-gh-mkeller:
> For .part naming, what decrypt_file returns should look like <filename>.part<random_suffix>

This was done by design, I wanted to overwrite half-done downloads with new attempts, so a ctrl+c left file would be cleaned up on a second try

sfc-gh-turbaszek:

PR #1989, opened 2024-07-01
SNOW-1514764: Support Tracing in SPCS runtime

sfc-gh-yuwang:

   Fixes #SNOW-1514764

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -487,6 +487,12 @@ class SnowflakeRestful:
             HTTP_HEADER_ACCEPT: accept_type,
             HTTP_HEADER_USER_AGENT: PYTHON_CONNECTOR_USER_AGENT,
         }
+        try:
+            from opentelemetry.propagate import inject
+
+            inject(headers)
+        except ModuleNotFoundError as e:
+            logger.debug(f"Opentelemtry otel injection failed because of: {e}")
         if self._connection.service_name:
             headers[HTTP_HEADER_SERVICE_NAME] = self._connection.service_name
         if method == "post":


src/snowflake/connector/network.py
```diff
@@ -487,6 +487,12 @@ def request(
             HTTP_HEADER_ACCEPT: accept_type,
             HTTP_HEADER_USER_AGENT: PYTHON_CONNECTOR_USER_AGENT,
         }
+        try:
+            from opentelemetry.propagate import inject
+
+            inject(headers)
+        except ModuleNotFoundError as e:
+            logger.debug(f"Opentelemtry otel injection failed because of: {e}")
```
sfc-gh-bdrutu, line 495: Not sure if the log is useful or not...

sfc-gh-bdrutu (approved):

sfc-gh-yuwang:
@sfc-gh-bdrutu Hi, can you give me a brief explanation why this change has to be in python connector? Is there other ways that we can do this?

src/snowflake/connector/network.py
sfc-gh-yuwang: I think it is better than put a `pass` here

src/snowflake/connector/network.py
```diff
@@ -487,6 +487,12 @@ def request(
             HTTP_HEADER_ACCEPT: accept_type,
             HTTP_HEADER_USER_AGENT: PYTHON_CONNECTOR_USER_AGENT,
         }
+        try:
+            from opentelemetry.propagate import inject
+
+            inject(headers)
```
sfc-gh-aling, line 493: 1. `request` method is on the critical path of query execution, what's the overhead of opentelemetry inject added here, e.g., increase in the payload, potential perf regression? also would try-except cause perf regression? I think we need to test perf of this change, and see the payload change
2. should we allow customers to opt out open telemetry collection?
3. how would backend react to request with new header? do we need any backend change for compatibility?

sfc-gh-yuwang: @sfc-gh-bdrutu I think some of this question are answered in slack chat, but can you still take a look?

sfc-gh-aling: just synced up with @sfc-gh-yuwang , I think once we get perf tested (can be as simple as performing "select 1" for 1 million times) and there is no regression, then let's merge it to unblock SPCS scenario first.

sfc-gh-bdrutu: We will add one header "traceparent" which has 55 characters value (always). So the overhead shouldn't be that significant especially it does not happen if the user doesn't configure opentelemetry. Not sure about performance of try/catch.

> how would backend react to request with new header, do we need any backend change?

We are taking care of that.

sfc-gh-aling (approved):

PR #2003, opened 2024-07-16
SNOW-1539600: Refactor structured type tests to explicitly set params

sfc-gh-jrose:

   Fixes #1539600

   This PR adds a wrapper to the connection object in structured type tests that enables required parameters when running in an env that allows structured types. This allows the tests to run when the parameters have not been set at the account level.

  N/A

### DIFF
--- a/test/integ/test_arrow_result.py
+++ b/test/integ/test_arrow_result.py
@@ -201,7 +201,7 @@ DATATYPE_TEST_CONFIGURATIONS = [
 
 @contextmanager
 def structured_type_wrapped_conn(conn_cnx):
-    parameters = {"python_connector_query_result_format": "json"}
+    parameters = {}
     if STRUCTURED_TYPES_SUPPORTED:
         parameters = {
             "python_connector_query_result_format": "arrow",
--- a/test/integ/test_arrow_result.py
+++ b/test/integ/test_arrow_result.py
@@ -176,13 +176,6 @@ STRUCTURED_TYPES_SUPPORTED = (
     CLOUD in STRUCTRED_TYPE_ENVIRONMENTS and RUNNING_ON_GH or CLOUD == "dev"
 )
 
-STRUCTURED_TYPE_PARAMETERS = {
-    "ENABLE_STRUCTURED_TYPES_IN_CLIENT_RESPONSE",
-    "ENABLE_STRUCTURED_TYPES_NATIVE_ARROW_FORMAT",
-    "FORCE_ENABLE_STRUCTURED_TYPES_NATIVE_ARROW_FORMAT",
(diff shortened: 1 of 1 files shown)


test/integ/test_arrow_result.py
```diff
@@ -198,6 +206,15 @@
 ]
 
 
+@contextmanager
+def structured_type_wrapped_conn(conn_cnx):
+    with conn_cnx() as conn:
+        if STRUCTURED_TYPES_SUPPORTED:
+            for param in STRUCTURED_TYPE_PARAMETERS:
+                conn.cursor().execute(f"alter session set {param}=true").fetchall()
+        yield conn
```
sfc-gh-aling, line 215: nit: do you need to unset it afterwards?

sfc-gh-jrose: I believe the session is destroyed once the `with conn_cnx() as conn` scope is left so it should not matter.

sfc-gh-pfus: Out of curiosity - do you test with native arrow only?

sfc-gh-pfus (approved):

test/integ/test_arrow_result.py
sfc-gh-aling: if so, there is an easier way you can do it like this: https://github.com/snowflakedb/snowflake-connector-python/blob/3013fe884a0bf794663ab6e3c18299d90d50095e/test/integ/test_cursor.py#L1566-L1569

sfc-gh-jrose: There are other tests that verify that json results work. In my experiments it looked like a structured response returned in json format and a semi structured response returned in json format were the same so I haven't added tests here. I don't think it would be hard to expand these tests to those cases though.

sfc-gh-pfus: I'm not sure if I agree. We had a discussion with Berlin team about nulls in arrays - currently nulls in arrays are returned as `undefined` instead of `null`. It breaks JSON parsing in drivers, so there is a small chance that for structured types some time in the future it may be returned differently comparing to Arrow.

But the question is still valid - your parameters enabled native arrow, so in that case this is not a JSON anymore. We have three formats to return structured types:
- Response format = JSON (JSON data where each structured type is a separate string containing JSON)
- Response format = Arrow (we have a string column and each row contains a separate string containing JSON - so technically this is similar to JSON)
- Response format = Arrow AND native Arrow enabled (full arrow - if we have a structured object, it is represented as arrow struct column with a pack of arrow columns representing each column)

sfc-gh-jrose: That sounds worth having explicit tests for. I'll look in to those cases and add relevant tests as part of another task.

sfc-gh-aling (approved):

PR #2027, opened 2024-08-16
SNOW-1562604: enhance error handling when polling query result

sfc-gh-aling:

   Fixes SNOW-1562604

add retry when polling query result and the response is None.
this happens when there is a 499 error happening between client and server.
without the fix, the client stopped retrying and errored out.
with the fix, we will keep retrying

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -679,6 +679,7 @@ class SnowflakeRestful:
         headers: dict[str, str],
         token: str = None,
         timeout: int | None = None,
+        is_fetch_query_status: bool = False,
     ) -> dict[str, Any]:
         if "Content-Encoding" in headers:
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/network.py
```diff
@@ -679,6 +679,7 @@ def _get_request(
         headers: dict[str, str],
         token: str = None,
         timeout: int | None = None,
+        is_fetch_query_status: bool = False,
```
sfc-gh-yixie, line 682: I feel we also need this for other queries?

sfc-gh-aling: what are the other queries you mention here? I know there are other GET requests used in the code base. but I prefer to doing this error handling carefully and case by case to avoid unexpected side affect.

sfc-gh-yixie (approved):

PR #2051, opened 2024-09-10
fix mac regression ci

sfc-gh-aling:

   Fixes #NNNN

https://github.com/pandas-dev/pandas/issues/58809
https://github.com/apache/arrow/issues/41696
https://github.com/apache/arrow/pull/43137/files

we should use ideally the latest mac version

python 3.8 is going to be deprecated soon, switching to 3.10 to align with linux test script


sfc-gh-aalam:
I see this change is trying to run tests in a 3.10 python version. Are we doing this because of coming 3.8 deprecation? it is not clear from the link provided in the description

sfc-gh-aling:
@sfc-gh-aalam  yeah Python 3.8 is going to be deprecated soon, also I found the linux test script is using python 3.10 so trying to align them.

tox will create new venv of different Python versions to run so I don't think the system python matters too much here as long as it's not depreciated. but ideally we can adopt the latest python for testing which will last longer -- e.g. if we use python 3.10 we need to update the test script 2 year later. but if it's 3.12, we can use the same script for 4 years

sfc-gh-aling:
I have logged another JIRA to update our tests to use python 3.12 in other tests if possible https://snowflakecomputing.atlassian.net/browse/SNOW-1664019

sfc-gh-aalam (approved):

PR #2057, opened 2024-09-18
SNOW-1572294: connection async api coverage

sfc-gh-aling:

   Fixes #NNNN


src/snowflake/connector/aio/_connection.py
```diff
@@ -87,7 +88,19 @@ def __init__(
         # get the imported modules from sys.modules
         # self._log_telemetry_imported_packages() # TODO: async telemetry support
         # check SNOW-1218851 for long term improvement plan to refactor ocsp code
-        # atexit.register(self._close_at_exit) # TODO: async atexit support/test
+        atexit.register(self._close_at_exit)
+
+    def __enter__(self):
+        # async connection does not support sync context manager
+        raise TypeError(
+            "'SnowflakeConnection' object does not support the context manager protocol"
```
sfc-gh-yuwang, line 96: shoud we raise `NotImplementedError` instead?

src/snowflake/connector/aio/_cursor.py
```diff
@@ -70,6 +70,11 @@ def __init__(
     def __aiter__(self):
         return self
 
+    def __iter__(self):
+        raise TypeError(
```
sfc-gh-yuwang, line 74: should we use `NotImplementedError` here?

src/snowflake/connector/aio/_result_batch.py
```diff
@@ -191,14 +191,27 @@ async def create_iter(
 
     async def _download(
         self, connection: SnowflakeConnection | None = None, **kwargs
-    ) -> aiohttp.ClientResponse:
+    ) -> tuple[bytes, str]:
```
sfc-gh-yuwang, line 194: just curious, what is the reason that we directly return downloaded data instead of response?

test/integ/aio/test_connection_async.py
```diff
@@ -18,8 +73,1443 @@ async def test_basic(db_parameters):
...
+    cnx = snowflake.connector.aio.SnowflakeConnection(**config)
+    try:
+        await cnx.connect()
+        assert cnx.client_session_keep_alive_heartbeat_frequency == 1000
+    finally:
+        await cnx.close()
+
+
+@pytest.mark.skipolddriver
+async def test_keep_alive_heartbeat_frequency_min(db_parameters):
```
sfc-gh-yuwang, line 241: We should add a test that verify heartbeat is working? Maybe you can check whether a heartbeat http request is sent.

src/snowflake/connector/aio/_connection.py
sfc-gh-aling: this is to align the native python error experience if a class does not support context manager

src/snowflake/connector/aio/_cursor.py
sfc-gh-aling: same reason as above, this is to align the native python error experience if a class does not support sync iterator.
iterator operation contains network request (fetching result) so having a sync iterator doesn't make sense for async cursor

src/snowflake/connector/aio/_result_batch.py
sfc-gh-aling: good question, this is because we use `async with connection._rest._use_requests_session() as session:` when issuing the request.

the response object doesn't work if it goes out of the scope of this session -- it will not be able to download data out of the scope of the session

sfc-gh-yuwang (approved):

test/integ/aio/test_connection_async.py
sfc-gh-aling: updated

PR #2060, opened 2024-09-19
SNOW-1650124: Pass Azure SAS Token via params

sfc-gh-jrose:

   Fixes SNOW-1650124

   This PR passes params through to the requests library so that Azure tokens can be passed to requests via params. This allows logging to avoid printing the secret information.

Azure storage client does not exist in stored proc connector.

### DIFF
--- a/src/snowflake/connector/azure_storage_client.py
+++ b/src/snowflake/connector/azure_storage_client.py
@@ -8,7 +8,7 @@ import json
 import os
 import xml.etree.ElementTree as ET
 from datetime import datetime, timezone
-from logging import getLogger
+from logging import Filter, getLogger
 from random import choice
 from string import hexdigits
 from typing import TYPE_CHECKING, Any, NamedTuple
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/azure_storage_client.py
```diff
@@ -95,15 +96,12 @@ def generate_authenticated_url_and_rest_args() -> tuple[bytes, dict[str, Any]]:
             sas_token = self.credentials.creds["AZURE_SAS_TOKEN"]
             if sas_token and sas_token.startswith("?"):
                 sas_token = sas_token[1:]
-            if "?" in url:
-                _url = url + "&" + sas_token
-            else:
-                _url = url + "?" + sas_token
+            params = {k: v for k, v in parse_qsl(sas_token)}
             headers["Date"] = timestamp
```
sfc-gh-aalam, line 100: Is a test for this change already added? Can't tell from the modified files

src/snowflake/connector/network.py
```diff
@@ -451,6 +451,7 @@ def request(
         method: str = "post",
         client: str = "sfsql",
         timeout: int | None = None,
+        params: dict | None = None,
```
sfc-gh-aling, line 454: I think this is technically a BCR since network.request is a public API and we introduced a new positional parameter in the middle of existing params.

what about making it a key word argument?

src/snowflake/connector/network.py
```diff
@@ -678,6 +681,7 @@ def _get_request(
         url: str,
         headers: dict[str, str],
         token: str = None,
+        params: dict | None = None,
```
sfc-gh-aling, line 684: similar comment as the public api one, although this is a private method and we can update every caller, I personally prefer adding as keyword  argument.

test/unit/test_retry_network.py
```diff
@@ -180,6 +180,7 @@ def test_request_exec():
         "headers": {},
         "data": '{"code": 12345}',
         "token": None,
+        "params": None,
```
sfc-gh-aling, line 183: can we add tests to ensure credentials are not leaked in log? we can use pytest caplog to capture logs

sfc-gh-aalam (approved):

PR #2061, opened 2024-09-19
[SNOW-1669128]: Ensure SnowflakeConnection atexit function is registered on import

sfc-gh-rdurrani:

   Fixes #SNOW-1669128

Previously, the close_at_exit function for the SnowflakeConnection object
was registered with the `atexit` module during initialization of a SnowflakeConnection object.
This caused the Snowpark Session object's atexit function to be registered before the
SnowflakeConnection's object when the Session was made without an existing SnowflakeConnection
object, because the SnowflakeConnection object's atexit was only registered after it was initialized
whereas the Snowpark Session object's atexit was registered when the file was imported. During the
atexit for the Snowpark Session, it attempts to send telemetry regarding cleanup operations, but since
it's connection is already closed, it fails to send the telemetry. Moving the registration of the atexit
function here means that it will happen when the SnowflakeConnection object is imported - and so the
Snowpark Session's atexit function will be registered after, and the order will be correct.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -314,6 +314,25 @@ class TypeAndBinding(NamedTuple):
     binding: str | None
 
 
+# SNOW-1669128 previously, the close_at_exit function for the SnowflakeConnection object
+# was registered with the `atexit` module during initialization of a SnowflakeConnection object.
+# This caused the Snowpark Session object's atexit function to be registered before the
+# SnowflakeConnection's object when the Session was made without an existing SnowflakeConnection
+# object, because the SnowflakeConnection object's atexit was only registered after it was initialized
+# whereas the Snowpark Session object's atexit was registered when the file was imported. During the
+# atexit for the Snowpark Session, it attempts to send telemetry regarding cleanup operations, but since
+# it's connection is already closed, it fails to send the telemetry. Moving the registration of the atexit
+# function here means that it will happen when the SnowflakeConnection object is imported - and so the
+# Snowpark Session's atexit function will be registered after, and the order will be correct.
+def _close_connection_at_exit(snowflake_connection: SnowflakeConnection):
+    with suppress(Exception):
+        snowflake_connection.close(retry=False)
+
+
+# check SNOW-1218851 for long term improvement plan to refactor ocsp code
+atexit.register(_close_connection_at_exit)
+
+
 class SnowflakeConnection:
     """Implementation of the connection object for the Snowflake Database.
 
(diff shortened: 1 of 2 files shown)


sfc-gh-rdurrani:
Do folks think that this should be ported to SP? I checked the `connection.py` file there, and there doesn't seem to be an import of `atexit`, so I'm thinking it shouldn't be ported, but wanted to double check!

src/snowflake/connector/connection.py
```diff
@@ -314,6 +314,25 @@ class TypeAndBinding(NamedTuple):
...
+# was registered with the `atexit` module during initialization of a SnowflakeConnection object.
+# This caused the Snowpark Session object's atexit function to be registered before the
+# SnowflakeConnection's object when the Session was made without an existing SnowflakeConnection
+# object, because the SnowflakeConnection object's atexit was only registered after it was initialized
+# whereas the Snowpark Session object's atexit was registered when the file was imported. During the
+# atexit for the Snowpark Session, it attempts to send telemetry regarding cleanup operations, but since
+# it's connection is already closed, it fails to send the telemetry. Moving the registration of the atexit
+# function here means that it will happen when the SnowflakeConnection object is imported - and so the
+# Snowpark Session's atexit function will be registered after, and the order will be correct.
+def _close_connection_at_exit(snowflake_connection: SnowflakeConnection):
```
sfc-gh-jrose, line 327: Where is this parameter passed in from?

sfc-gh-jkew:
I would be hesitant to change the atexit behavior here, as opposed to simply removing the warning to fix a failed telemetry send. This may actually be the root cause; but I also think it's a more dangerous change considering the initial problem ( a warning ).

I might be convinced otherwise though - maybe I'm just being paranoid.

src/snowflake/connector/connection.py
sfc-gh-rdurrani: Good catch, thank you! I've added a variable to track active connections - would this work, or would you suggest another workaround (e.g. perhaps by making a class method to register the atexit function in the SnowflakeConnection object, and calling it from the Snowpark Session object before registering its own atexit function)?

sfc-gh-rdurrani:
> I would be hesitant to change the atexit behavior here, as opposed to simply removing the warning to fix a failed telemetry send. This may actually be the root cause; but I also think it's a more dangerous change considering the initial problem ( a warning ).
> 
> I might be convinced otherwise though - maybe I'm just being paranoid.

I totally get where you're coming from (and am open to removing the warning as well), but like you point out, I think that this is the *root problem*, and further, that the current behavior is *incorrect*. Since the Snowpark session object has a dependency on the SnowflakeConnection object, it makes sense that the snowflake connection object shouldn't be garbage collected before the Session object. I'd argue that while this change is broader, it moves us closer to a more correct garbage collection.

sfc-gh-rdurrani:
Seems that the issue this was trying to resolve has been fixed by another PR!

PR #2068, opened 2024-10-10
SNOW-1325701: Use scoped temp object in write pandas

sfc-gh-yuwang:

   Fixes #NNNN


DESCRIPTION.md
```diff
@@ -7,6 +7,8 @@ https://docs.snowflake.com/
 Source code is also available at: https://github.com/snowflakedb/snowflake-connector-python
 
 # Release Notes
+- v3.13.0(TBD)
+  - Improved `write_pandas` and `BindUploadAgent.upload` to create scoped temporary object instead of temporary object.
```
sfc-gh-aling, line 11: why do we need this change in the public connector?
I thought "scoped object" only applied to the concept of stored proc. is this just for aligning the change with the sproc connector?

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -19,17 +25,13 @@
...
-        f"create or replace temporary stage {_STAGE_NAME} "
-        "file_format=(type=csv field_optionally_enclosed_by='\"')"
-    )
 
     def __init__(
         self,
         cursor: SnowflakeCursor,
         rows: list[bytes],
         stream_buffer_size: int = 1024 * 1024 * 10,
+        use_scoped_temp_object: bool = True,
```
sfc-gh-aling, line 34: do we need this as part of the public api/do we want users to control the behavior?

src/snowflake/connector/pandas_tools.py
```diff
@@ -181,6 +209,7 @@ def write_pandas(
     overwrite: bool = False,
     table_type: Literal["", "temp", "temporary", "transient"] = "",
     use_logical_type: bool | None = None,
+    use_scoped_temp_object: bool = True,
```
sfc-gh-aling, line 212: same question, do we need this to be part of public api?

src/snowflake/connector/utils.py
sfc-gh-aling, line 1: shall we make this module private? I don't think users are going to use those

test/integ/pandas/test_pandas_tools.py
```diff
@@ -602,7 +602,10 @@ def mocked_execute(*args, **kwargs):
             )
             assert m_execute.called and any(
                 map(
-                    lambda e: "CREATE TEMP STAGE" in str(e[0]),
+                    lambda e: (
+                        "CREATE TEMP STAGE" in str(e[0])
+                        or "CREATE SCOPED TEMPORARY STAGE" in str(e[0])
+                    ),
```
sfc-gh-aling, line 608: why it's A or B, I think we should the two condition separately, otherwise we're losing code coverage
use_temp -> CREATE SCOPED TEMPORARY STAGE
not use_temp -> CREATE TEMPORARY STAGE

DESCRIPTION.md
sfc-gh-yuwang: here is some context: https://docs.google.com/document/d/1qfJevIMEZL7O12H5pQK5WZG-a-uwA2K2cmFDusHjdXM/edit?usp=sharing.
I think mostly they want temp object being created as job-scoped object, and they also want to align with sp connector

src/snowflake/connector/bind_upload_agent.py
sfc-gh-yuwang: Will change this to private

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -39,12 +40,24 @@ def __init__(
...
         self._stream_buffer_size = stream_buffer_size
-        self.stage_path = f"@{self._STAGE_NAME}/{uuid.uuid4().hex}"
+        self.stage_path = f"@{self._stage_name}/{uuid.uuid4().hex}"
+        self._use_scoped_temp_object = (
+            cursor.connection._session_parameters.get(
+                _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, True
+            )
+            if cursor.connection._session_parameters
+            else True
+        )
```
sfc-gh-aling, line 53: should we default to `False`? snowflake connector users may not be snowpark users
```suggestion
        self._use_scoped_temp_object = (
            cursor.connection._session_parameters.get(
                _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, False
            )
            if cursor.connection._session_parameters
            else False
        )
```

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -39,12 +40,24 @@ def __init__(
             stream_buffer_size: Size of each file, default to 10MB.
         """
         self.cursor = cursor
+        self._stage_name = random_name_for_temp_object(TempObjectType.STAGE)
```
sfc-gh-aling, line 43: do you think we can set the name based upon self._use_scoped_temp_object ?

sfc-gh-aling (commented):
connector is not only used by snowpark, but also by non-snowpark users. I think we should make the job scoped change only for snowpark users.

DESCRIPTION.md
sfc-gh-zyao: Is my understanding correct?: for regular / non-SP session,  "job" is equivalent to the session itself. So job-scoped objects are de facto session-scoped ones. Therefore this change does not affect behavior of regular sessions.

src/snowflake/connector/pandas_tools.py
```diff
@@ -263,6 +299,14 @@ def write_pandas(
             f"Invalid compression '{compression}', only acceptable values are: {compression_map.keys()}"
         )
 
+    _use_scoped_temp_object = (
+        conn.cursor().connection._session_parameters.get(
+            _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, False
+        )
+        if conn.cursor().connection._session_parameters
+        else False
```
sfc-gh-aling, line 307: do we need to create new cursor here?

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -38,13 +39,29 @@ def __init__(
             rows: Rows of binding parameters in CSV format.
             stream_buffer_size: Size of each file, default to 10MB.
         """
+        self._use_scoped_temp_object = (
+            cursor.connection._session_parameters.get(
+                _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, False
+            )
+            if cursor.connection._session_parameters
```
sfc-gh-zyao, line 46: (not a strong opinion) Will `cursor.connection._session_parameters` ever be None? iirc we always at least initialized to an empty dict, so we can even simplify do 
```python
self._use_scoped_temp_object = cursor.connection._session_parameters.get(
    _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, False
)
```
But to err on the side of caution is also a good idea.

src/snowflake/connector/_utils.py
```diff
@@ -0,0 +1,45 @@
...
+    COLUMN = "COLUMN"
+    PROCEDURE = "PROCEDURE"
+    TABLE_FUNCTION = "TABLE_FUNCTION"
+    DYNAMIC_TABLE = "DYNAMIC_TABLE"
+    AGGREGATE_FUNCTION = "AGGREGATE_FUNCTION"
+    CTE = "CTE"
+
+
+TEMP_OBJECT_NAME_PREFIX = "SNOWPARK_TEMP_"
+ALPHANUMERIC = string.digits + string.ascii_lowercase
```
sfc-gh-zyao, line 28: Is there any special reason that we set the character set to be number + lowercase, but then later uppercase the entire name?

sfc-gh-zyao (approved):
The change looks good to me. Only 2 nitpicking comments. Thanks for looking into this!

src/snowflake/connector/bind_upload_agent.py
sfc-gh-yuwang: I followed the practice in snowpark here, and this approach handled all possible errors too

src/snowflake/connector/_utils.py
sfc-gh-yuwang: same as above, I followed snowpark's practice, I guess because we need a special name format for scoped object

src/snowflake/connector/pandas_tools.py
sfc-gh-yuwang: I changed it

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -38,13 +39,29 @@ def __init__(
...
+            else False
+        )
         self.cursor = cursor
+        self._stage_name = (
+            random_name_for_temp_object(TempObjectType.STAGE)
+            if self._use_scoped_temp_object
+            else "SYSTEMBIND"
+        )
```
sfc-gh-aling, line 54: actually I just realized that if `_use_scoped_temp_object` is True, then each time when we instantiate a BindUploadAgent (we do this when binding parameters), a new temp stage created.

is this the same behavior as the sproc connector change?

sfc-gh-yuwang: This is a tricky problem, the original design is that the temp stage name in bind upload agent is a constant. So that there is only one temp stage created by it.
what i can think of is creating a constant stage name following the naming rule, what do you think?

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -38,13 +38,24 @@ def __init__(
             rows: Rows of binding parameters in CSV format.
             stream_buffer_size: Size of each file, default to 10MB.
         """
+        self._use_scoped_temp_object = (
+            cursor.connection._session_parameters.get(
+                _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, False
+            )
+            if cursor.connection._session_parameters
+            else False
+        )
```
sfc-gh-aling, line 47: what about the following?
```suggestion
        )
        self. _STAGE_NAME = (
            "SNOWPARK_TEMP_STAGE_BIND"
            if self._use_scoped_temp_object
            else "SYSTEMBIND"
        )
```

sfc-gh-aling (approved):

src/snowflake/connector/bind_upload_agent.py
sfc-gh-smao: We hit an issue with this change. The latest `3.12.3` broke our tests.

In our tests, we have a `MockConnection` class which mocks all the public functions of Connection. This change, however, started using a private class member (`connection._session_parameters`), which doesn't seem right.

Issue #2085, opened 2024-10-25
Correctly type CursorDict and add TupleDict


Polandia94:
### What is the current behavior?

cursor.fetchone(), cursor.fetchmany() and cursor.fetchall() return a union type between tuple and dict. That require a lot of casting to work with type checkers.

### What is the desired behavior?

Return a dict(or list of dicts) if cursor is DictCursor, a tuple(or list of tuples) if cursor is a TupleCursor, and keep the current implementation to SnowflakeCursor.

### How would this improve `snowflake-connector-python`?

It will  improve the type hints

### References and other background

This issue was mentioned on:
https://github.com/snowflakedb/snowflake-connector-python/issues/1263
but closed after 6 months
There was a Pull request with some implemntation, with some comments.
https://github.com/snowflakedb/snowflake-connector-python/pull/1264

sfc-gh-sghosh:
Hello @Polandia94 ,

Thanks for raising the issue, will check with team about the current status of the PR and update.

Regards,
Sujan

sfc-gh-sghosh:
Hello @Polandia94 ,

We will work on the fix, at present there is no ETA when it will be fixed or delivered.

Regards.
Sujan

mrufsvold:
@sfc-gh-sghosh It looks like this issue was solved in #2532

sfc-gh-dszmolka:
thank you @mrufsvold  - closing this out.

PR #2086, opened 2024-10-25
add tupleCursor and type dictCursor

Polandia94:
Add tupleCursor like dictCursor but always return Tuple. Modify types of dictCursor to always return dict
related to: https://github.com/snowflakedb/snowflake-connector-python/issues/2085


Polandia94:
recheck

Polandia94:

src/snowflake/connector/cursor.py
```diff
@@ -1724,13 +1724,38 @@ def get_result_batches(self) -> list[ResultBatch] | None:
...
+    def fetchone(self) -> dict | None:
+        return super().fetchone()
+    
+    def fetchmany(self, size: int | None = None) -> list[dict]:
+        return super().fetchmany()
+    
+    def fetchall(self) -> list[dict]:
+        return super().fetchall()
+
+class TupleCursor(SnowflakeCursor):
```
sfc-gh-aling, line 1742: I don't think we need a TupleCursor here -- this is the effectively the default SnowflakeCursor

sfc-gh-aling, line 1742: do we need `TupleCursor`? this is effectively the default SnowflakeCursor.

Polandia94: `SnowflakeCursor `has to return `Union[None|dict|tuple] `because could be instanciated as
`SnowflakeCursor(use_dict_result=True)`.

`TupleCursor `is just a `SnowflakeCursor `that return `Union[None|tuple]` on fetch

On `DictCursor `the better types will not require nothing for the user, on `TupleCursor `it will require that they start to instanciate as `TupleDict`, so maybe its not useful. I let you to decide, let me know if you want and i delete the `TupleCursor `from PR and issue.

sfc-gh-turbaszek:

PR #2089, opened 2024-10-25
SNOW-1708720: Error is raised at debug level when not closing connection

sfc-gh-yuwang:

   Fixes #SNOW-1708720

### DIFF
--- a/src/snowflake/connector/aio/_network.py
+++ b/src/snowflake/connector/aio/_network.py
@@ -827,6 +827,12 @@ class SnowflakeRestful(SnowflakeRestfulSync):
             RuntimeError,
             AttributeError,  # json decoding error
         ) as err:
+            if "Event loop is closed" in str(err):
+                logger.warning(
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/aio/_network.py
```diff
@@ -827,6 +827,12 @@ async def _request_exec(
             RuntimeError,
             AttributeError,  # json decoding error
         ) as err:
+            if "Event loop is closed" in str(err):
+                logger.warning(
+                    f"Snowflake connection is never closed, which is causing {err}"
+                    "please use SnowflakeConnection.close() to close connection and save resources"
+                )
+                return {}
```
sfc-gh-aling, line 835: I'm a bit concerned about handling error here which is on our critical path for every execution, we would need to write a lot of test making sure it doesn't cause side effect.

since the issue we're tackling only happens at atexit->close, I think we can log an info at that time

PR #2090, opened 2024-10-26
SNOW-1708720: Error is raised at debug level when not closing connection

sfc-gh-yuwang:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/aio/_network.py
+++ b/src/snowflake/connector/aio/_network.py
@@ -831,6 +831,7 @@ class SnowflakeRestful(SnowflakeRestfulSync):
                 logger.info(
                     "If you see the logging error message 'RuntimeError: Event loop is closed' during program exit, it probably indicates that the connection was not closed properly before the event loop was shut down. Please use SnowflakeConnection.close() to close connection."
                 )
+                raise err
             if is_login_request(full_url):
                 logger.debug(
                     "Hit a timeout error while logging in. Will be handled by "
--- a/src/snowflake/connector/aio/_network.py
+++ b/src/snowflake/connector/aio/_network.py
@@ -832,7 +832,6 @@ class SnowflakeRestful(SnowflakeRestfulSync):
                     f"Snowflake connection is never closed, which is causing {err}"
                     "please use SnowflakeConnection.close() to close connection and save resources"
                 )
-                return {}
             if is_login_request(full_url):
                 logger.debug(
                     "Hit a timeout error while logging in. Will be handled by "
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/aio/_network.py
```diff
@@ -827,6 +827,11 @@ async def _request_exec(
             RuntimeError,
             AttributeError,  # json decoding error
         ) as err:
+            if "Event loop is closed" in str(err):
```
sfc-gh-aling, line 830: can we also check err type here to be more precise?
```suggestion
            if "Event loop is closed" in str(err) and instance(err, ...):
```

src/snowflake/connector/aio/_network.py
```diff
@@ -827,6 +827,11 @@ async def _request_exec(
...
+                logger.warning(
+                    f"Snowflake connection is never closed, which is causing {err}"
+                    "please use SnowflakeConnection.close() to close connection and save resources"
```
sfc-gh-aling, line 833: I feel we can say something like "event loop is closed, the operation failed, if you see this message during when exiting program, it's likely that the connection doesn't get closed. Please close the connection when exiting.."  (use chatgpt to polish a little bit)

src/snowflake/connector/aio/_network.py
```diff
@@ -827,6 +827,11 @@ async def _request_exec(
             RuntimeError,
             AttributeError,  # json decoding error
         ) as err:
+            if "Event loop is closed" in str(err):
+                logger.warning(
```
sfc-gh-aling, line 831: do you figure it out what session delete does? if it's some critical op that I think this could be a logging.INFO
```suggestion
                logger.info(
```

src/snowflake/connector/aio/_network.py
```diff
@@ -827,6 +827,10 @@ async def _request_exec(
             RuntimeError,
             AttributeError,  # json decoding error
         ) as err:
+            if isinstance(err, RuntimeError) and "Event loop is closed" in str(err):
+                logger.info(
+                    "If you see the logging error message 'RuntimeError: Event loop is closed' during program exit, it probably indicates that the connection was not closed properly before the event loop was shut down. Please use SnowflakeConnection.close() to close connection."
+                )
```
sfc-gh-aling, line 833: do we need to re-raise error here or it doesn't matter?
otherwise the error gets swallowed, and _request_exec returns None, which will be an unknown error handled by the retry logic?

sfc-gh-aling (approved):

PR #2117, opened 2024-12-11
SNOW-1848371 adding connection.is_valid to perform connection validation on TCP/IP and Session levels

sfc-gh-dszmolka:

   Fixes SNOW-1848371

   The PR is for covering a gap where seemingly there's no convenience method to tell 'is this connection still good enough to send queries over it?' . See details in the jira, but basically motivated by the situation where the cloud provider LB tears down the underlying idle connection unilaterally without telling the client, who then tries the submit the next query after the idleness ends, over a socket which is already closed on the other end - ending up with an error. 
   
  Having such a mechanism which the Node.js driver already has, gives the users a tool to conveniently control if the connection/session is still valid or they need to do something about it. 
  
  Plus fixing two very minor typos.


src/snowflake/connector/connection.py
```diff
@@ -1981,3 +1981,23 @@ def _log_telemetry_imported_packages(self) -> None:
...
+        else:
+            try:
+                logger.debug("trying to heartbeat into the session to validate")
+                hb_result = self.rest._heartbeat()
+                session_valid = hb_result.get("success")
+                logger.debug("session still valid? %s", str(session_valid))
+                return session_valid
+            except Exception as e:
+                logger.debug(
+                    "session could not be validated due to exception: %s", str(e)
```
sfc-gh-mkeller, line 2001: Th `%s` should do this automatically
```suggestion
                    "session could not be validated due to exception: %s", e
```

test/integ/test_connection.py
```diff
@@ -1462,3 +1462,13 @@ def test_disable_telemetry(conn_cnx, caplog):
...
+
+
+@pytest.mark.skipolddriver
+def test_is_valid(conn_cnx):
+    """Tests whether connection and session validation happens."""
+    with conn_cnx() as conn:
+        assert conn
+        assert conn.is_valid() is True
+        conn.close()
+        assert conn.is_valid() is False
```
sfc-gh-mkeller, line 1474:
```suggestion
    assert conn.is_valid() is False
```

DESCRIPTION.md
```diff
@@ -10,6 +10,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 - v3.12.5(TBD)
   - Added a feature to limit the sizes of IO-bound ThreadPoolExecutors during PUT and GET commands.
+  - Added a feature to verify if the connection is still good enough to send queries over
```
sfc-gh-mkeller, line 13:
```suggestion
  - Added a feature to verify if the connection is still good enough to send queries over.
```

src/snowflake/connector/connection.py
```diff
@@ -1981,3 +1981,23 @@ def _log_telemetry_imported_packages(self) -> None:
...
             )
+
+    def is_valid(self) -> bool:
+        """This function tries to answer the question: Is this connection still good for sending queries?
+        Attempts to validate the connections both on the TCP/IP and Session levels."""
+        logger.debug("validating connection and session")
+        if self.is_closed():
+            logger.debug("connection is already closed and not valid")
+            return False
+        else:
```
sfc-gh-mkeller, line 1992: You could get rid of this line and dedent the rest of the function

sfc-gh-mkeller (changes requested):

src/snowflake/connector/connection.py
sfc-gh-dszmolka: good one - and done

src/snowflake/connector/connection.py
```diff
@@ -1990,3 +1990,21 @@ def _log_telemetry_imported_packages(self) -> None:
...
+        if self.is_closed():
+            logger.debug("connection is already closed and not valid")
+            return False
+
+        try:
+            logger.debug("trying to heartbeat into the session to validate")
+            hb_result = self.rest._heartbeat()
+            session_valid = hb_result.get("success")
+            logger.debug("session still valid? %s", session_valid)
+            return session_valid
```
sfc-gh-mkeller, line 2007: So, `.get()` I think could return `None` by default and your type-hint says that is not a valid return type. Do either
```suggestion
            return bool(session_valid)
```
Or tell `.get()` to return `False` default value.

sfc-gh-mkeller (approved):
Deal with the small comment first, otherwise looks great. Thank you!

:ship: :ship:

src/snowflake/connector/connection.py
sfc-gh-dszmolka: i like the first version better :) thank you for the suggestion and the review!

PR #2120, opened 2024-12-13
Directly lookup target distributions instead of filtering manually

sfc-gh-rebeling:
We can use `distribution` directly to lookup the distribution we're looking for instead of manually looping through all installed distributions and filtering manually.

In my measurements, this speeds up the runtime of this method by around 100ms.


sfc-gh-mkeller (approved):
:ship: :ship:

sfc-gh-mkeller:
Add changelog entry please before merging!

PR #2122, opened 2024-12-16
SNOW-1825473 adding pat authentication integration

sfc-gh-mkeller:

   Fixes SNOW-1825473

   In this PR I introduce support for the new PAT authentication method.


src/snowflake/connector/auth/oauth.py
```diff
@@ -22,7 +22,7 @@ def type_(self) -> AuthType:
         return AuthType.OAUTH
 
     @property
-    def assertion_content(self) -> str:
+    def assertion_content(self) -> str | None:
```
sfc-gh-igarish, line 25: Why added None here

src/snowflake/connector/network.py
```diff
@@ -187,6 +187,7 @@
 OAUTH_AUTHENTICATOR = "OAUTH"
 ID_TOKEN_AUTHENTICATOR = "ID_TOKEN"
 USR_PWD_MFA_AUTHENTICATOR = "USERNAME_PASSWORD_MFA"
+PROGRAMMATIC_ACCESS_TOKEN = "PROGRAMMATIC_ACCESS_TOKEN"
```
sfc-gh-igarish, line 190: How user will provide authentication name in connection parameter? like "PROGRAMMATIC_ACCESS_TOKEN"? If yes, then can we use "PAT" instead of long name?

src/snowflake/connector/auth/oauth.py
sfc-gh-mkeller: The type-hint was incorrect. It doesn't actually affect correctness, but you can see a few lines below that the function `reset_secrets` sets what this returns to `None`

src/snowflake/connector/network.py
sfc-gh-mkeller: Here's the exact shell command I was using to test the feature: `python -c 'from snowflake.connector import connect; c=connect(account="account", user="user", authenticator="PROGRAMMATIC_ACCESS_TOKEN", token_file_path="/Users/mkeller/pat"); print(c.cursor().execute("select 1, 2").fetchall())'`

I haven't introduced the short form simply because we don't offer short-hands for any of the other authentications. I'm happy to introduce it if you think we should start now though!

PR #2136, opened 2025-01-15
SNOW-1882588 Allow Empty Sql Text when Dataframe Ast is Presented

sfc-gh-bli:

   SNOW-1882588 Allow Empty Sql Text when Dataframe Ast is Presented

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -930,7 +930,7 @@ class SnowflakeCursor:
         if _do_reset:
             self.reset()
         command = command.strip(" \t\n\r") if command else None
-        if not command:
+        if not command and not _dataframe_ast:
             logger.warning("execute: no query is given to execute")
             return None
         logger.debug("query: [%s]", self._format_query_for_log(command))
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -930,14 +930,14 @@ class SnowflakeCursor:
         if _do_reset:
             self.reset()
         command = command.strip(" \t\n\r") if command else ""
-        if not command and not _dataframe_ast:
-            logger.warning("execute: no query is given to execute")
-            return None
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/cursor.py
```diff
@@ -929,8 +929,8 @@ def execute(
 
         if _do_reset:
             self.reset()
-        command = command.strip(" \t\n\r") if command else None
```
sfc-gh-bli, line 932: Change `None` to "", otherwise the following log formatters will fail.

sfc-gh-lspiegelberg: that makes sense.

src/snowflake/connector/cursor.py
```diff
@@ -929,8 +929,8 @@ def execute(
...
-        if not command:
+        command = command.strip(" \t\n\r") if command else ""
+        if not command and not _dataframe_ast:
```
sfc-gh-lspiegelberg, line 933: Can we maybe clarify the logging logic here? If I understand this correctly, basically in the case of _dataframe_ast =True, command will be "".  Should we log out the base64 string in debug mode? I.e.,

```
if not command:
   if _dataframe_ast:
         logger.debug("query[AST base64]: [%s]", _dataframe_ast)
   else:
         logger.warning("execute: no query is given to execute")
         return      
```

sfc-gh-lspiegelberg (approved):
lgtm, some comment re logging. I think we can clear this up a bit.

src/snowflake/connector/cursor.py
sfc-gh-bli: Good Idea, let me change it.

sfc-gh-bli: changed

sfc-gh-mkeller:
Please add tests for this change!

sfc-gh-bli:

PR #2135, opened 2025-01-14
SNOW-1825495 OAuth flows implementation

sfc-gh-mkeller:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-1825621

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [x] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [x] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   In this PR I add support for [OUath authorization code flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow). It's very similar to external browser authentication, but it uses oauth and more options are more better for users.
   This change has been tested manually, as it's fairly complicated to setup and we don't do unit tests for the different authentication methods.

4. (Optional) PR for stored-proc connector: Not applicable

PR #2162, opened 2025-01-30
SNOW-1825621 refresh token implementation and OAuth token cache

sfc-gh-mkeller:

   Fixes SNOW-1825621

   The last piece of the new OAuth code flow implementation.
   The PR builds on top of #2137 and #2135, so those should be merged before this one is reviewed.
   I've added support for refresh token consumption and tested it. Unfortunately this is kind of useless until we introduce a OAUTH token cache, so that needs adding still.


src/snowflake/connector/connection.py
```diff
@@ -335,7 +335,7 @@ def _get_private_bytes_from_file(
         # SNOW-1825621: OAUTH implementation
     ),
     "oauth_security_features": (
-        ("pkce",),
+        ("pkce", "refresh_token"),
```
sfc-gh-eworoshow, line 338: we should omit refresh_token from the default, since the default oauth integration will not have refresh tokens enabled unless the admin turns them on explicitly

src/snowflake/connector/auth/oauth_code.py
```diff
@@ -228,14 +236,20 @@ def prepare(
...
-                    "code": ER_IDP_CONNECTION_ERROR,
-                    "message": "Invalid HTTP request from web browser. Idp "
-                    "authentication could have failed.",
-                },
-            )
+            if self.refresh_token:
+                logger.info(
+                    "oauth refresh token is available, going to try consuming it to recover"
+                )
+                self._consume_refresh_token(conn=conn)
```
sfc-gh-eworoshow, line 243: we only need to use the refresh token after the access token has expired, in order to create a new session without reprompting for credentials

sfc-gh-eworoshow (commented):
Nice!

PR #2167, opened 2025-02-06
SNOW-1915469 Basic support for DECFLOAT type

sfc-gh-tkissinger:

   Implements [SNOW-1915469](https://snowflakecomputing.atlassian.net/browse/SNOW-1915469)

   This PR adds basic decimal float (DECFLOAT) data type support to the python connector.
- Supports ARROW and JSON result format. DECFLOATs are converted to Decimal or numpy.float64
- Server-side binding support using explicit Snowflake type specification as Decimal is already mapped to FIXED by default
- No Pandas support

[SNOW-1915469]: https://snowflakecomputing.atlassian.net/browse/SNOW-1915469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/snowflake/connector/arrow_context.py
+++ b/src/snowflake/connector/arrow_context.py
@@ -159,3 +159,11 @@ class ArrowConverterContext:
         digits = [int(digit) for digit in str(int128) if digit != "-"]
         sign = int128 < 0
         return decimal.Decimal((sign, digits, -scale))
+
+    def DECFLOAT_to_decimal(self, exponent: int, significand: bytes) -> decimal.Decimal:
+        # significand is two's complement big endian.
+        significand = int.from_bytes(significand, byteorder="big", signed=True)
+        return decimal.Decimal(significand).scaleb(exponent)
+
+    def DECFLOAT_to_numpy_float64(self, exponent: int, significand: bytes) -> float64:
+        return numpy.float64(self.DECFLOAT_to_decimal(exponent, significand))
--- a/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/DecFloatConverter.cpp
+++ b/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/DecFloatConverter.cpp
@@ -1,0 +1,87 @@
+
+//
+// Copyright (c) 2012-2023 Snowflake Computing Inc. All rights reserved.
+//
+
+#include "DecFloatConverter.hpp"
+
+#include <cstring>
+#include <memory>
+
+#include "Python/Helpers.hpp"
+
+namespace sf {
+
+Logger* DecFloatConverter::logger =
+    new Logger("snowflake.connector.DecFloatConverter");
+
+const std::string DecFloatConverter::FIELD_NAME_EXPONENT = "exponent";
+const std::string DecFloatConverter::FIELD_NAME_SIGNIFICAND = "significand";
+
+DecFloatConverter::DecFloatConverter(ArrowArrayView& array,
+                                     ArrowSchemaView& schema, PyObject& context,
+                                     bool useNumpy)
+    : m_context(context),
+      m_array(array),
+      m_exponent(nullptr),
+      m_significand(nullptr),
+      m_useNumpy(useNumpy) {
+  if (schema.schema->n_children != 2) {
+    std::string errorInfo = Logger::formatString(
+        "[Snowflake Exception] arrow schema field number does not match, "
+        "expected 2 but got %d instead",
+        schema.schema->n_children);
+    logger->error(__FILE__, __func__, __LINE__, errorInfo.c_str());
+    PyErr_SetString(PyExc_Exception, errorInfo.c_str());
+    return;
+  }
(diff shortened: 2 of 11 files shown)


sfc-gh-tkissinger:

src/snowflake/connector/arrow_context.py
```diff
@@ -159,3 +159,13 @@ def DECIMAL128_to_decimal(self, int128_bytes: bytes, scale: int) -> decimal.Deci
         digits = [int(digit) for digit in str(int128) if digit != "-"]
         sign = int128 < 0
         return decimal.Decimal((sign, digits, -scale))
+
+    def DECFLOAT_to_decimal(self, exponent: int, significand: bytes) -> decimal.Decimal:
+        # significand is two's complement big endian. Convert to target endianess if needed.
+        if byteorder == "little":
+            significand = reversed(significand)
+        significand = int.from_bytes(significand, byteorder=byteorder, signed=True)
```
sfc-gh-pfus, line 167: Is it correct? You reverse byte order if needed, but pass byteorder without changing?

sfc-gh-tkissinger: @sfc-gh-pfus If I got it right, the byteorder indicates the endianess of the machine the driver is running on. The byte array always arrives in big endian and the idea is to convert it to the machines preferred endianess to enable machine-specific optimizations if they are implemented.

sfc-gh-pfus: Ah, I see. So, is Python's endianness OS-specific? Java, for instance, internally always uses big-endian.

sfc-gh-tkissinger: @sfc-gh-pfus I'm not a python expert 🙂 Just know that you need to pass in the array and tell the constructor the byte order of the bytes in that array. If you check the function above, e.g. for parsing our fixed-point number as Decimal (DECIMAL128_to_decimal), we pass the machine endianess in. Arrow uses the machine endianess too when writing the results, but unfortunately GS and driver run on different machines where endianess could differ. So, the function above will return wrong results when we have the driver on a BE machine and GS on a LE machine. Think there is some flag in the arrow metadata telling us about the endianess it was created with, but we ignore this as it looks. So, something we should fix. For the DECFLOAT, we know it is always BE. This means we could tell the constructor to that it is BE w/o reversing the array. However, sometime there is an optimized code path if endianess matches the machine endianess, and that's why we adjust this here. But I don't know whether the is such an optimization in Python

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/DecFloatConverter.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (stringView.size_bytes > 16) {
+    std::string errorInfo = Logger::formatString(
+        "[Snowflake Exception] only precisions up to 38 supported. "
+        "Please update to a newer version of the connector.");
+    logger->error(__FILE__, __func__, __LINE__, errorInfo.c_str());
+    PyErr_SetString(PyExc_Exception, errorInfo.c_str());
+    return nullptr;
+  }
+  PyObject* significand =
+      PyBytes_FromStringAndSize(stringView.data, stringView.size_bytes);
```
sfc-gh-mmishchenko, line 75: Have you considered building a significand as a Py-integer right here? I.e. to `PyLong_FromNativeBytes` instead, it accepts source endianness.

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/SnowflakeType.cpp
```diff
@@ -17,6 +17,7 @@ std::unordered_map<std::string, SnowflakeType::Type>
         {"DOUBLE PRECISION", SnowflakeType::Type::REAL},
         {"DOUBLE", SnowflakeType::Type::REAL},
         {"FIXED", SnowflakeType::Type::FIXED},
+        {"DECFLOAT", SnowflakeType::Type::DECFLOAT},
```
sfc-gh-mmishchenko, line 20: Looks this whole list has been following alphabetic order so far, so this one should be the next after `DATE`.

sfc-gh-mmishchenko (commented):
Just to check - will it also be possible also to cover with test negative scenarios, i.e. a wrong DECFLOAT schema in an input or too long significand?

src/snowflake/connector/nanoarrow_cpp/ArrowIterator/DecFloatConverter.cpp
sfc-gh-tkissinger: we wanted to get out of the C++ world as early as possible and I see no advantage using this function over the python version.

sfc-gh-mmishchenko: Won't it do 1 alloc/copy less, so in general perform faster?

PR #2169, opened 2025-02-10
SNOW-1920533 match privatelink in hostname to setup correct privatelink OCSP Cache url even if user specifies account in UPPERCASE

sfc-gh-dszmolka:

   SNOW-1920533

   Public link OCSP Cache server url is `ocsp.snowflakecomputing.com` but in privatelink environment it works differently, and should look like `ocsp.accountname.region.privatelink.snowflakcomputing.com`. This hostname for OCSP Cache is determined based on the connection configuration (account/host). 
   Since earlier we were looking for pattern `.privatelink.snowflakecomputing.` to be present in account/host, if user specified account as `MYACCOUNT.REGION.PRIVATELINK`, it resulted host becoming partly uppercase and thus not matching the above pattern.
   
   Fix aims to lowercase the hostname before trying to match it to privatelink pattern. Added  test with uppercased account/hostname.

### DIFF
--- a/test/integ/test_connection.py
+++ b/test/integ/test_connection.py
@@ -671,6 +671,17 @@ def test_privatelink_ocsp_url_creation():
         == "http://ocsp.testaccount.us-east-1.privatelink.snowflakecomputing.com/ocsp_response_cache.json"
     )
 
+    del os.environ["SF_OCSP_RESPONSE_CACHE_SERVER_URL"]
+
+    account = "TESTACCOUNT.US-EAST-1.PRIVATELINK"
+    hostname = account + ".snowflakecomputing.com"
+    SnowflakeConnection.setup_ocsp_privatelink(CLIENT_NAME, hostname)
+    ocsp_cache_server = os.getenv("SF_OCSP_RESPONSE_CACHE_SERVER_URL", None)
+    assert (
+        ocsp_cache_server
+        == "http://ocsp.testaccount.us-east-1.privatelink.snowflakecomputing.com/ocsp_response_cache.json"
+    )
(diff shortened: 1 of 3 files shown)


test/integ/test_connection.py
```diff
@@ -671,6 +671,17 @@ def test_privatelink_ocsp_url_creation():
         == "http://ocsp.testaccount.us-east-1.privatelink.snowflakecomputing.com/ocsp_response_cache.json"
     )
 
+    del os.environ["SF_OCSP_RESPONSE_CACHE_SERVER_URL"]
```
sfc-gh-pfus, line 674: Should we restore this value after the test?

sfc-gh-pfus (approved):

sfc-gh-dprzybysz (approved):

test/integ/test_connection.py
sfc-gh-dszmolka: @sfc-gh-pfus `SnowflakeConnection.setup_ocsp_privatelink` populates `SF_OCSP_RESPONSE_CACHE_SERVER_URL` . The `del` before the test is to make sure we're not using some 'leftover' variable before populating it, i guess.

Is this what you had in mind, or something else?

sfc-gh-pfus: Ok then, it's fine.

sfc-gh-dszmolka:
Jenkins fails with the exact same error on a doc-only change too (https://github.com/snowflakedb/snowflake-connector-python/pull/2172), considering a test infra problem

since all other single tests passed, merging this now

PR #2176, opened 2025-02-17
SNOW-1935873 Fix test different key length

sfc-gh-mkubik:
This PR fixes the failing CI tests by ensuring that Base64 encoded private key is decoded to string type as the private key format is strictly coupled with its type.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -12,8 +12,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Removed the workaround for a Python 2.7 bug.
   - Added a <19.0.0 pin to pyarrow as a workaround to a bug affecting Azure Batch.
   - Optimized distribution package lookup to speed up import.
-  - Fixed a bug where privatelink OCSP Cache url could not be determined if privatelink account name was specified in uppercase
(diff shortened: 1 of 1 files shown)


test/integ/test_key_pair_authentication.py
```diff
@@ -128,7 +128,7 @@ def fin():
                 pass
 
             # Ensure the base64-encoded version also works
-            db_config["private_key"] = base64.b64encode(private_key_der)
```
sfc-gh-pfus, line 131: So these are two cases, one with PK encoded one with decoded. Shouldn't we keep both?

sfc-gh-mkubik: the decode that I' adding is not base64 decode. The newly added `decode()` converts the array of bytes to string (https://docs.python.org/3/library/stdtypes.html#bytes.decode). Further in the driver there's the assumption that the `private_key` value must be of type `str` when it's base64 encoded (https://github.com/snowflakedb/snowflake-connector-python/blob/4b3ded11f434d6d930e51ded71153577b57a7827/src/snowflake/connector/auth/keypair.py#L108)

sfc-gh-pfus (approved):

PR #2182, opened 2025-02-21
SNOW-1940996 no-op auth for Stored Proc

sfc-gh-zyao:
### Description
- Introduce a no-op auth for Python Stored Proc (aka Python SP)
- It shall have no effect on non-SP connections
- Proof of the auth working in Stored Procedures (via local testing)
Screenshot 2025-02-20 at 5 34 14 PM
- DO_NOT_PORT_CHANGES_TO_SP
### Testing
- unit test for new auth
- existing test to make sure the SP-specific change does not break existing use cases
- local testing to make sure the change works for SP
  - we are not able add integration test yet for SP use case while it is still work-in-progres

   Fixes SNOW-1940996


sfc-gh-sfan (commented):
Is it possible to not introduce stored proc concept into the connector? Can we not use the HTTP request parser to respond accordingly if we see `/session/heartbeat`?

src/snowflake/connector/connection.py
```diff
@@ -1234,8 +1236,10 @@ def __config(self, **kwargs):
             with open(token_file_path) as f:
                 self._token = f.read()
 
+        tokenless_authenticators = {OAUTH_AUTHENTICATOR, STORED_PROC_AUTHENTICATOR}
+
         if not (self._master_token and self._session_token):
-            if not self.user and self._authenticator != OAUTH_AUTHENTICATOR:
+            if not self.user and self._authenticator not in tokenless_authenticators:
                 # OAuth Authentication does not require a username
```
sfc-gh-pbulawa, line 1243: Can you update the comment to `Tokenless authenticators (...)`

src/snowflake/connector/connection.py
```diff
@@ -1265,14 +1269,15 @@ def __config(self, **kwargs):
                     {"msg": "Password is empty", "errno": ER_NO_PASSWORD},
                 )
 
-        if not self._account:
+        # All connections other than Stored Procedures mandate presence of account name.
```
sfc-gh-pbulawa, line 1272: Could you create a test for that case?

src/snowflake/connector/connection.py
sfc-gh-zyao: Done

src/snowflake/connector/connection.py
sfc-gh-zyao: (I have updated the PR description to mention this as well)
We have existing integration test to make sure the new change does not break existing use cases (i.e. non-Python-Stored-Proc cases). As for the Python SP cases, we can only have local testing (where we do some "hacks" in the local the environment to allow end-to-end testing) at the moment because the entire feature is still work-in-progress

src/snowflake/connector/connection.py
```diff
@@ -1234,9 +1236,11 @@ def __config(self, **kwargs):
             with open(token_file_path) as f:
                 self._token = f.read()
 
+        tokenless_authenticators = {OAUTH_AUTHENTICATOR, STORED_PROC_AUTHENTICATOR}
```
sfc-gh-mmishchenko, line 1239: OAUTH is rather tokenful, so var naming looks a bit confusing here

sfc-gh-pbulawa (approved):

test/unit/test_auth_sp_auth.py
```diff
@@ -0,0 +1,37 @@
...
+
+    reauth_response = auth.reauthenticate()
+    assert (
+        reauth_response == expected_reauth_response
+    ), f"reauthenticate() is expected to return {expected_reauth_response}, but returns {reauth_response}"
+
+    reauth_response = auth.reauthenticate(foo="bar")
+    assert (
+        reauth_response == expected_reauth_response
+    ), f'reauthenticate(foo="bar") is expected to return {expected_reauth_response}, but returns {reauth_response}'
```
sfc-gh-mmishchenko, line 37: Doesn't this test just repeat the implementation of `AuthByStoredProcConnection`?
A positive scenario I believe cannot be tested with an integration test currently? Probably worth adding an integration test to test a negative scenario when non-SP client tries to use this new Auth type, WDYT?

sfc-gh-mmishchenko (commented):
Added some suggestions from my end

sfc-gh-mmishchenko (changes requested):
Please, see the suggestions

src/snowflake/connector/auth/sp_auth.py
```diff
@@ -0,0 +1,47 @@
...
+#
+
+from __future__ import annotations
+
+from typing import Any
+
+from .by_plugin import AuthByPlugin, AuthType
+
+
+class AuthByStoredProcConnection(AuthByPlugin):
```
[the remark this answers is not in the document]
sfc-gh-zyao: I will go with your suggestion, it is a good idea. TLDR is that for the short term we don't plan to change how Stored Proc auth. But this is good, because there is a long-term goal to make SP behave similarly to other connection, which includes how it authenticates.

src/snowflake/connector/connection.py
sfc-gh-zyao: Yeah, you are right, "empty_user_allowed_authenticators" is a better name without causing such confusion

test/unit/test_auth_sp_auth.py
sfc-gh-zyao: Sounds good to add a integ test for negative scenario (and it is done in latest commit).

Regarding the unit test, it is a bit "dummy", but I think it still helps to guard the behavior (and tell other devs that it is intended, and therefore we cover it with unit test cases). What do you think about the unit test?

sfc-gh-mmishchenko (approved):

sfc-gh-sfan (approved):

PR #2183, opened 2025-02-21
SNOW-1944162 Add tests for programmatic access token

sfc-gh-mkubik:
[SNOW-1944162](https://snowflakecomputing.atlassian.net/browse/SNOW-1944162)
Adding mappings and tests for programmatic access token

[SNOW-1944162]: https://snowflakecomputing.atlassian.net/browse/SNOW-1944162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/data/wiremock/mappings/auth/pat/invalid_token.json
```diff
@@ -0,0 +1,39 @@
+{
+  "mappings": [
+    {
+      "scenarioName": "Successful PAT authentication flow",
+      "requiredScenarioState": "Started",
+      "newScenarioState": "Authenticated",
```
sfc-gh-pbulawa, line 6: The scenario state does not correspond with the actual response. Perhaps it should be `Authentication failed`?

test/data/wiremock/mappings/auth/pat/successful_flow.json
```diff
@@ -0,0 +1,70 @@
...
+            "newClientForUpgrade": "3.12.3",
+            "sessionId": 1172562260498,
+            "parameters": [
+              {
+                "name": "CLIENT_PREFETCH_THREADS",
+                "value": 4
+              }
+            ],
+            "sessionInfo": {
+              "databaseName": "TEST_DHEYMAN",
```
sfc-gh-pbulawa, line 52: I think it's better to rename it to `TEST_DB`.

sfc-gh-pbulawa (changes requested):

test/data/wiremock/mappings/auth/pat/successful_flow.json
sfc-gh-mkubik: Agreed. I'll need to take the eternal glory in python driver sources away from Dawid then

test/data/wiremock/mappings/generic/snowflake_disconnect_successful.json
```diff
@@ -0,0 +1,21 @@
+{
```
sfc-gh-pfus, line 1: What is this for? Logout? Do we use POST for logout and not DELETE?

test/unit/test_programmatic_access_token.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@pytest.fixture(scope="session")
+def wiremock_client() -> Generator[Union[WiremockClient, Any], Any, None]:
+    with WiremockClient() as client:
+        yield client
+
+
+@pytest.mark.skipolddriver
+def test_valid_pat(wiremock_client: WiremockClient) -> None:
```
sfc-gh-pfus, line 26: According to what @sfc-gh-dheyman wanted, we should also have such a case: https://github.com/snowflakedb/gosnowflake/blob/master/auth.go#L459

test/data/wiremock/mappings/generic/snowflake_disconnect_successful.json
sfc-gh-mkubik: Yup, that's what we do (https://github.com/snowflakedb/snowflake-connector-python/blob/e52747277dc576997e050657b482a67f87d2a377/src/snowflake/connector/network.py#L662). Otherwise wiremock complains on unmatched requests introducing mess into the logs

test/unit/test_programmatic_access_token.py
sfc-gh-mkubik: Haven't noticed that, will add. Thanks

sfc-gh-pfus (approved):

sfc-gh-pbulawa (approved):

PR #2184, opened 2025-02-25
SNOW-1944208 add unsafe write flag

sfc-gh-mkubik:
This PR adds `unsafe_file_write` flag to connection class. The security patch applied in 3.13.1 has narrowed the permissions of files downloaded with GET to 600. This change seems to be inconvenient for some customers that were relying on the previous behaviour in their workflows so after discussion with the support team we decided to add a flag that restores the previous behaviour of saving files downloaded with GET with 644.

The parameter was added to all cloud interfaces, the default value is False which means the strict permissions will be applied.

The test had to be fixed since due to the compression on PUT the permissions weren't verified on the proper file.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -16,8 +16,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added support for iceberg tables to `write_pandas`.
   - Fixed base64 encoded private key tests.
   - Added Wiremock tests.
-  - Fixed a bug where file permission check happened on Windows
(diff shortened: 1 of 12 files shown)


sfc-gh-mmishchenko:
Wouldn't it be even more convenient (and transparent) to let users decide about exact permissions value to set?

sfc-gh-rsharma:
@sfc-gh-mkubik Do we have an ETA for this fix? If I can get that information, then I can set the customer's expectations accordingly.

sfc-gh-mmishchenko (approved):

PR #2189, opened 2025-02-28
SNOW-1960985 Wiremock tests for authorization_code

sfc-gh-mkubik:
Changes: 
- add wiremock tests for authorization_code
- add `forbidden_ports` parameter to wiremock client to prevent conflicts with oauth redirect server
- remove `client_id` and `client_secret` from oauth token request as they're passed in Authorization header ([RFC-6749/2.3.1](https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1))
- fix Authorization header
- add Accept and Content-Type headers to token request


src/snowflake/connector/auth/_http_server.py
```diff
@@ -134,9 +137,13 @@ def _try_receive_block(
...
     def receive_block(
         self,
-        max_attempts: int = 15,
-        timeout: float | None = 30.0,
+        max_attempts: int = None,
+        timeout: float | None = None,
     ) -> (list[str] | None, socket.socket | None):
+        if max_attempts is None:
+            max_attempts = self.DEFAULT_MAX_ATTEMPTS
+        if timeout is None:
```
sfc-gh-mmishchenko, line 145: Just to be sure - so effectively we prohibit waiting for a callback with no timeout? Makes sense to me.

sfc-gh-mmishchenko (approved):
LGTM

src/snowflake/connector/auth/_http_server.py
sfc-gh-mkubik: Good catch that this behaviour changes but I believe there's no need to allow waiting indefinitely, especially if it's on the socket level where there's no visibility regarding elapsed time or retries

PR #2196, opened 2025-03-05
SNOW-1927118 Enable wiremock tests on Jenkins

sfc-gh-mkubik:
- Use fixed version of manylinux image to enable `yum` while keeping the `latest` tag as a backup.
- Add install openjdk 11 step to Dockerfile to enable running wiremock daemon
- Enable wiremock healthcheck test on Jenkins

### DIFF
--- a/ci/docker/connector_test/Dockerfile
+++ b/ci/docker/connector_test/Dockerfile
@@ -1,6 +1,8 @@
 ARG BASE_IMAGE=quay.io/pypa/manylinux2014_x86_64
 FROM $BASE_IMAGE
 
+RUN yum install -y java
+
 # This is to solve permission issue, read https://denibertovic.com/posts/handling-permissions-with-docker-volumes/
 ARG GOSU_URL=https://github.com/tianon/gosu/releases/download/1.14/gosu-amd64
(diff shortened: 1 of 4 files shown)


ci/docker/connector_test/Dockerfile
```diff
@@ -1,6 +1,8 @@
 ARG BASE_IMAGE=quay.io/pypa/manylinux2014_x86_64
 FROM $BASE_IMAGE
 
+RUN yum install -y java-11-openjdk
```
sfc-gh-pfus, line 4: Why not temurin 17?

sfc-gh-pfus (approved):

sfc-gh-mmishchenko (approved):

ci/docker/connector_test/Dockerfile
sfc-gh-mkubik: It's basically first package with java > 8 that I've found on yum repository would temurin make more sense here?

sfc-gh-pfus: Not temurin itself, but Java 17 is going to be supported longer than 11 :)

PR #2198, opened 2025-03-05
SNOW-1963078 Port _upload / _download / _upload_stream / _download_st…

sfc-gh-zyao:
…ream methods to the public connector

### Description
NO-CHANGELOG-UPDATES
- The change appears large, but it is basically porting the `_upload` / `_download` / `_upload_stream` / `_download_stream` methods (along with `direct_file_operation_utils.py`, which provides needed utils for the implementation of the 4 methods) from SP connector to public connector (reference: a similar change on SP connector: https://github.com/snowflakedb/Stored-Proc-Python-Connector/pull/194). The same implementation has been adopted in SP connector and porting it back to public connector is needed based on [this decision](https://docs.google.com/document/d/1KegQLzyxB5G2Viti5Tr8fD24IPjIKU3cCZtRUUWqm4k)
  - These methods (`_upload` / `_download` / `_upload_stream` / `_download_stream`) are necessary to run Stored Proc with public connector ([this](https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/1e2519f3ddc71547642b007b712aae59ef00dfff/src/snowflake/connector/cursor.py#L808-L833) is how Stored Proc connector implements these methods. The implementation of these methods for public Python connector and SP connector should be exactly the same. It appears a bit different at the moment only because we haven't cleaned up the parameter protection yet)
  - They are helpful for non-SP sessions as well
  - The Python Stored Proc has its own logic of file operation parser and stream downloader, so we abstract away the detailed implementation of FileOperationParser and StreamDownloader on the server side
  - We do not prioritize file operation parser and stream downloader implementation for non-stored-proc here, because it is a non-goal for stored proc fork removal project (this is why the `parse_file_operation` and `download_as_stream` methods are set to raise NotImplemented error

 ### Testing
 - existing unit tests and integ tests to make sure the new change does not break them
 - a new sanity-check test to assert that the needed file utils are present (i.e. parse_file_operation and download_as_stream)
 - local testing to show that the new code works for Python SP connection (we cannot test it formally in a integ test as the entire project is still WIP)
   1. testing `_upload` and `_download_stream`, no compression
upload-then-download-stream

   2. testing `_upload` and `_download_stream`, with compression
compress-upload-then-download-stream

   3. testing `_upload_stream` and `_download`, no compression
upload-stream-then-download

   4. testing `_upload_stream` and `_download`, with compression
compress-upload-stream-then-download

   Fixes #NNNN

7. Fill out the following pre-review checklist:

8. Please describe how your code solves the related issue.

9. (Optional) PR for stored-proc connector:


DESCRIPTION.md
```diff
@@ -8,6 +8,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 - v3.14.1(TBD)
+  - introduce several file operation utilities for easier internal dev
```
sfc-gh-sfan, line 11: I don't think we need release note given the added functions are all internal.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -706,6 +707,12 @@ def _create_file_transfer_client(
                 use_s3_regional_url=self._use_s3_regional_url,
                 unsafe_file_write=self._unsafe_file_write,
             )
+        elif self._stage_location_type == STORED_PROC_FS:
```
sfc-gh-sfan, line 710: Can we have some override mechanism internally so that we don't have to expose this? We could potentially override `_create_file_transfer_client`.

src/snowflake/connector/connection.py
```diff
@@ -472,6 +473,10 @@ def __init__(
         # check SNOW-1218851 for long term improvement plan to refactor ocsp code
         atexit.register(self._close_at_exit)
 
+        # Set up the file operation parser and stream downloader.
```
sfc-gh-zyao, line 476: To reviewers (especially @sfc-gh-sfan ) SP connector will also need to set `connection.rso_id` attribute, in order for SP file operation to work as intended (reference: [SP connector code](https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/1e2519f3ddc71547642b007b712aae59ef00dfff/src/snowflake/connector/connection.py#L202)). I decided to do the attribute setting on the server side (something like [this PoC code](https://github.com/snowflakedb/snowflake/blob/321fe732ef8d7db19a01e1ddc323a33f96e81504/ExecPlatform/src/coprocessor/python/PythonStoredProcSession.cpp#L148-L151)), since rso_id is not a general concept for non-SP connections.

The upside of this approach is that we can abstract away unnecessary/SP-specific concept from public connector

The downside (which I don't think will really happen) is that, it seems as if `rso_id` attribute does not exist in connection (if we only read public connector code). So let's say, if other future developer tries to introduce a seemingly-new attribute `rso_id` on connection, they may think, "okay, `rso_id` is not an existing attribute, so I can add and use it freely, without worrying about side effects". But in fact, `rso_id` would be an existing attribute of connection by that time, and messing the attribute up will affect some Python SP behaviors. But I think the chance is very very slim, because I would be very surprised if any future dev needs to introduce another attribute, and it happens to be literally `rso_id`

DESCRIPTION.md
sfc-gh-zyao: done, reverted the changelog

src/snowflake/connector/file_transfer_agent.py
sfc-gh-zyao: I tested and yes, we can override `_create_file_transfer_client` on the server side for Stored Proc. Does this look good https://github.com/snowflakedb/snowflake-connector-python/pull/2198/commits/088b553599f4bae4455b9b5c07357287bd399502? It stills defines the `STORED_PROC_FS` label though. We need a file system category to bypass the check here https://github.com/snowflakedb/snowflake-connector-python/blob/04d38199120e46519e4785f62956d1eb487961c5/src/snowflake/connector/file_transfer_agent.py#L923

sfc-gh-zyao: Well, I can actually append STORED_PROC_FS to VALID_STORAGE list on the server side. This way we don't need to expose the tag `STORED_PROC_FS` at all. Let me go with this

sfc-gh-zyao: Done

PR #2203, opened 2025-03-07
Add support for workload identity federation

### LINKED ISSUE #2202: SNOW-1967309: SNOW-1927956: Add support for workload identity federation
### What is the current behavior?

Snowflake supports external OAuth but not workload identity federation.

### What is the desired behavior?

Support authenticating to Snowflake using native identity from Azure, AWS, GCP or other platforms that support generic OIDC ID tokens.

### How would this improve `snowflake-connector-python`?

Allow for simpler and more secure configuration without the need to explicitly manage secrets.

### References and other background

go/wif

sfc-gh-pmansour:

   Fixes #2202

   - [x] I am adding new credentials

Adds a new authentication plugin called `AuthByWorkloadIdentity`, and plumbs it through the connection creation. This relies on a few parameters:

* `authenticator`: This is an existing parameter, but now has a new supported value `WORKLOAD_IDENTITY`.
* `workload_identity_provider`: A new optional parameter that can describe how to load the workload identity credential. Can be AWS, AZURE, GCP, OIDC or None. If this is set to None, it will attempt to auto-discover a credential based on the environment.
* `workload_identity_entra_resource`: A new optional parameter, to be used when targeting a custom Azure Entra ID application as the resource, for example in a sovereign cloud.

There is a broad suite of tests for all the new functionality, including a set of new fixtures that fake the behavior of the major CSP metadata services.

Usage examples:
```
# To auto-discover a credential
connect(account="my_account", authenticator="WORKLOAD_IDENTITY")

# To look specifically for an AWS credential
connect(account="my_account", authenticator="WORKLOAD_IDENTITY", workload_identity_provider="AWS")

# To use an OIDC token from the given file path
connect(account="my_account", authenticator="WORKLOAD_IDENTITY", token_file_path="/service-account-token")
```

These changes cannot be used yet as the backend support is not ready. They're gated by a new environment variable `SF_ENABLE_EXPERIMENTAL_AUTHENTICATION` which must be set.

Also note that the default Entra ID application resource is not ready yet, so I used a placeholder value that's guaranteed to fail in reality. This will be updated in a follow-up PR once that application is ready.


src/snowflake/connector/wif_util.py
```diff
@@ -0,0 +1,292 @@
...
+    
+    If this is not populated, returns None.
+    """
+    if not token:
+        logger.debug("No OIDC token was specified.")
+        return None
+
+    try:
+        claims = jwt.decode(token, options={"verify_signature": False})
+        return WorkloadIdentityAttestation(AttestationProvider.OIDC, token, {"iss": claims["iss"], "sub": claims["sub"]})
```
sfc-gh-dbasavin: nit: Same here

test/unit/test_connection.py
```diff
@@ -588,3 +589,35 @@ def test_otel_error_message(caplog, mock_post_requests):
...
+
+
+@patch("snowflake.connector.SnowflakeConnection._authenticate", return_value=None)
+@patch("snowflake.connector.auth.AuthByWorkloadIdentity.__init__", return_value=None)
+def test_connection_params_are_plumbed_into_authbyworkloadidentity(mock_auth_constructor, mock_authenticate):
+    os.environ["SF_ENABLE_EXPERIMENTAL_AUTHENTICATION"] = ""  # we can set this to any value.
+    snowflake.connector.connect(
+        account="my_account_1",
+        workload_identity_provider=AttestationProvider.AWS,
+        workload_identity_entra_resource="api://0b2f151f-09a2-46eb-ad5a-39d5ebef917b",
```
sfc-gh-dbasavin, line 614: This made me wonder whether we want to add extra checks to surface a readable error to users when they set `workload_identity_entra_resource` option but the `workload_identity_provider` is not `AttestationProvider.Azure` or `None`.

I wanted to suggest a similar check for the `token` option, i.e. surface an error when users set the `token` option but the `workload_identity_provider` is not `AttestationProvider.OIDC` or `None`. But then I remembered that we are re-using the `token` option that already existed. Which made me wonder about something else -- would it be better or worse if instead of reusing the existing `token` option we added a new one, e.g. `workload_identity_oidc_token`?

sfc-gh-dbasavin (approved):
Looks great!

test/unit/test_connection.py
sfc-gh-pmansour: #### re. `workload_identity_entra_resource`
That's a fair point and I think we can go either way. I can see a world in which somebody always populates `workload_identity_entra_resource` everywhere for simplicity (e.g. if they always operate in China or Germany), but still uses `workload_identity_provider` to flip between the different CSPs they run on. So there's no reason to force them to change the other parameter as well. OTOH, it could be a mistake and an error would help them catch it. There's a further complication in that they could be using auto-detect (aka `workload_identity_provider=None`) and then the right behavior is unclear.

My personal take is that the main parameter here is `authenticator=WORKLOAD_IDENTITY_FEDERATION`, and every other option is a supporting arg whose value will only be used as needed. But this is a lightly held belief and I'm willing to change it if there's a good reason otherwise.

#### re. `token`:

I debated this briefly... I think it's better to use the existing token option because a) it's used across multiple different authenticators (key-pair, PAT, OAuth), b) `connection.py` already populates it based on the existing `token_file_path` option, and c) customers are used to providing this, either inline or via the file. I think it would be a plus to re-use existing tried-and-tested config paths.

As you said though, if `token` (or `token_file_path`) was specified, we can't assume it was intended for OIDC so we can't fail validation on it.

sfc-gh-dbasavin: SGTM, your reasoning makes sense! Thank you.

src/snowflake/connector/wif_util.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Note: the real token verification (including signature verification) happens on the Snowflake side. The driver doesn't have
+    the keys to verify these JWTs, and in any case that's not where the security boundary is drawn.
+
+    We only decode the JWT here to get some basic claims, which will be used for a) a quick smoke test to ensure we got the right
+    issuer, and b) to find the unique user being asserted and populate assertion_content. The latter may be used for logging
+    and possibly caching.
+    """
+    claims = jwt.decode(jwt_str, options={"verify_signature": False})
+    return claims["iss"], claims["sub"]
```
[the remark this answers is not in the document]
sfc-gh-pmansour: I updated the error handling to check for this, but per the other pattern I've used across this file, I had it `return None` (and later return the generic "No credential was found for {..}") exception instead of an explicit one.

My rationale here is that you could have an invalid token passed in somewhere (e.g. a JWT for keypair auth, or an opaque external OAuth token), but that doesn't matter because `authenticator=WORKLOAD_IDENTITY` and there's another provider that _will_ produce a valid token later down the auto-discovery order, so we should suppress specific exceptions other than "No credential was found".

What do you think?

sfc-gh-pmansour:

src/snowflake/connector/wif_util.py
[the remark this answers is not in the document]
sfc-gh-dbasavin: Works for me, thank you!

src/snowflake/connector/wif_util.py
```diff
@@ -0,0 +1,324 @@
...
+
+@dataclass
+class WorkloadIdentityAttestation:
+    provider: AttestationProvider
+    credential: str
+    user_identifier_components: dict
+
+
+def try_metadata_service_call(
+    method: str, url: str, headers: dict, timeout_sec: int = 3
```
sfc-gh-jkasten, line 53: [no action required] Is there any precedent for the time out value?

On first glance, a timeout of 3 seconds seems pretty tight to me for a request that can't be cached. I am guessing this is for the auto-discovery case and if we expand it we might hang unnecessarily? I would naturally be tempted to expand the timeout for the non-auto discovery case, but perhaps that adds more complexity than it is worth.

I was unable to find any SLOs for latency on the metadata services. I know that this should be a local request and the cloud would break if this were to go down. I am sure it is OK and hopefully there are retries along a real application.

src/snowflake/connector/auth/workload_identity.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @staticmethod
+    def from_attestation(
+        attestation: WorkloadIdentityAttestation,
+    ) -> ApiFederatedAuthenticationType:
+        """Maps the internal / driver-specific attestation providers to API authenticator types.
+
+        The AttestationProvider is related to how the driver fetches the credential, while the API authenticator
+        type is related to how the credential is verified. In most current cases these may be the same, though
+        in the future we could have, for example, multiple AttestationProviders that all fetch an OIDC token.
```
sfc-gh-jkasten, line 37:
```suggestion
        in the future we could have, for example, multiple AttestationProviders that all fetch an OIDC ID token.
```
nit/optional

sfc-gh-jkasten (approved):
Awesome, Peter!

src/snowflake/connector/wif_util.py
sfc-gh-pmansour: This is an interesting case because I'm hearing from both sides of the aisle on this:
- 3s is too long since it's on the critical path of login.
- 3s is too short since it may unnecessary block auth for a real, expected delay.

The specific number is somewhat random here. I also don't know about any SLOs for metadata service latency, but given it's a local service, I suspect we shouldn't see anywhere near 3s latency in any normal code path.

3s is also on the longer side, because if we're doing auto-detect on a supported CSP, we won't actually exhaust the full 3 seconds during attempts to get a credential from the "wrong" CSP [^1]. Still, I'm weary of making this even longer (e.g. 30s or so) because it'll mean that a completely misconfigured workload (e.g. something not running on _any_ supported CSP) will take 3x30s to figure out that it's doomed, which is too long IMO.

Having said that I'm open to further feedback here from the driver folks cc @sfc-gh-mmishchenko for any other thoughts.

[^1]: The major CSPs should immediately reject metadata service requests that don't conform to their specific request signature. For example, Azure requires the `Metadata: True` header while GCP requires the `Metadata-flavor: Google` header. AWS requires neither but uses distinct URL paths that are different from the other two. So in theory, running in "auto-detect" mode on any of the 3 major CSPs shouldn't result in actually exhausting the full 3s timeout.

src/snowflake/connector/auth/workload_identity.py
sfc-gh-pmansour: done

tox.ini
```diff
@@ -86,7 +86,7 @@ skip_install = True
 setenv = {[testenv]setenv}
 passenv = {[testenv]passenv}
 commands =
-    {env:SNOWFLAKE_PYTEST_CMD} -m "not skipolddriver" -vvv {posargs:} test
+    {env:SNOWFLAKE_PYTEST_CMD} --ignore=test/unit --ignore=test/pandas -m "not skipolddriver" -vvv {posargs:} test
```
sfc-gh-pmansour, line 89: Note: unit and pandas tests are already being skipped for the old driver tests anyway ([reference](https://github.com/snowflakedb/snowflake-connector-python/blob/44eb130ed3a4704686032fd53fb59547475f6c60/test/conftest.py#L84-L85)).

This change just avoids traversing those file paths completely when running tests against the old drivers, so that we don't face weird import errors. That means we can confidently import "new" driver code in `test/unit/conftest.py`, knowing that that file won't even be evaluated when running against the old drivers.

src/snowflake/connector/wif_util.py
sfc-gh-jkasten: Yeah, I agree it would only be a possible concern on the 99.9% case... which to be fair is probably around our own regional availability. More generic workload restarts should be built-in for those cases anyway.

src/snowflake/connector/connection.py
```diff
@@ -188,12 +193,14 @@ def _get_private_bytes_from_file(
...
     "private_key_file_pwd": (None, (type(None), str, bytes)),
-    "token": (None, (type(None), str)),  # OAuth/JWT/PAT Token
+    "token": (None, (type(None), str)),  # OAuth/JWT/PAT/OIDC Token
     "token_file_path": (
         None,
         (type(None), str, bytes),
-    ),  # OAuth/JWT/PAT Token file path
+    ),  # OAuth/JWT/PAT/OIDC Token file path
     "authenticator": (DEFAULT_AUTHENTICATOR, (type(None), str)),
+    "workload_identity_provider": (None, (type(None), AttestationProvider)),
```
sfc-gh-pmansour, line 202: Maxim: consider a test that loads this from TOML

src/snowflake/connector/auth/workload_identity.py
```diff
@@ -0,0 +1,101 @@
...
+    def prepare(
+        self,
+        **kwargs: typing.Any,
+    ) -> None:
+        """Fetch the token."""
+        self.attestation = create_attestation(
+            self.provider, self.entra_resource, self.token
+        )
+
+    def reauthenticate(self, **kwargs: typing.Any) -> dict[str, bool]:
```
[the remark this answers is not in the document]
sfc-gh-pmansour: Also see if this is covered by any existing tests.

sfc-gh-pmansour (commented):
Posting on behalf of Maxim

src/snowflake/connector/connection.py
sfc-gh-mmishchenko: Not sure how popular file-based configurations are for cloud-based python apps, but this should ease mapping the implementation on other drivers.

src/snowflake/connector/auth/workload_identity.py
```diff
@@ -0,0 +1,101 @@
...
+            return ApiFederatedAuthenticationType.OIDC
+        return ValueError(f"Unknown attestation provider '{attestation.provider}'")
+
+
+class AuthByWorkloadIdentity(AuthByPlugin):
+    """Plugin to authenticate via workload identity."""
+
+    def __init__(
+        self,
+        provider: AttestationProvider | None = None,
```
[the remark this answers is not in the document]
sfc-gh-pmansour: done

tox.ini
sfc-gh-pmansour: Added this explanation as a comment in tox.ini.

src/snowflake/connector/auth/workload_identity.py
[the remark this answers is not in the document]
sfc-gh-pmansour: Looks like `reauthenticate` is only relevant for the `AuthByIdToken`; for all other auth plugins, the regular `prepare` method is called.
https://github.com/snowflakedb/snowflake-connector-python/blob/7710a8c53b566e896a22719487c586655134af53/src/snowflake/connector/connection.py#L1451-L1456

That's why pretty much all the other plugins return `{"success": False}` in their override of `reauthenticate`. I changed `AuthByWorkloadIdentity` to also do the same.

src/snowflake/connector/connection.py
sfc-gh-pmansour: Good call, turned out I had to explicitly parse a strongly-typed `AttestationProvider` enum value from string when it was provided via TOML. I did this and added a test to verify it works. Thanks!

PR #2240, opened 2025-03-28
SNOW-2011670 Allow requestId to be set with statement params

sfc-gh-lspiegelberg:

   Fixes SNOW-2011670

   As part of the Snowpark IR project, we will send AST as part of the request. To restore them faithfully on the server-side and avoid introducing yet another layer of indirection, we would like to use the HTTP Request Id. Currently, this Request Id is generated on the client as part of issuing the query. With this PR we introduce a new (client-side) statement parameter `requestId` that allows to overwrite the automatically generated request Id. `requestId` must be avalid UUID4 or UUID4 string.

    to be done when interface here is approved.


test/integ/test_cursor.py
```diff
@@ -1967,3 +1968,37 @@ def test_nanoarrow_usage_deprecation():
...
+                cur.execute(
+                    "select seq4() as foo from table(generator(rowcount=>5))",
+                    _statement_params={"requestId": request_id},
+                )
+
+
+def test_custom_request_id(conn_cnx):
+    request_id = uuid.uuid4()
+
+    with conn_cnx() as con:
```
sfc-gh-evandenberg, line 1997: What use case is broken or requires this?  It seems safer to let the client randomly generate the request Id automatically.  How else would the request id be generated?

DESCRIPTION.md
```diff
@@ -18,6 +18,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added `gcs_use_virtual_endpoints` connection property that forces the usage of the virtual GCS usage. See more: https://cloud.google.com/storage/docs/request-endpoints#xml-api
   - Fixed a bug that caused driver to fail silently on `TO_DATE` arrow to python conversion when invalid date was followed by the correct one.
   - Added `check_arrow_conversion_error_on_every_column` connection property that can be set to `False` to restore previous behaviour in which driver will ignore errors until it occurs in the last column. This flag's purpose is to unblock workflows that may be impacted by the bugfix and will be removed in later releases.
+  - Allow to overwrite the automatically generated HTTP Request Id with a custom UUID4 as part of statement parameters (field: `requestId`).
```
sfc-gh-heshah, line 21: nit: Allow the connector to inherit a UUID4 generated upstream, provided in statement parameters (field: `requestId`), rather than automatically generate a UUID4 to use for the HTTP Request ID.

sfc-gh-heshah (approved):
Small nit, otherwise LGTM!

test/integ/test_cursor.py
sfc-gh-heshah: For the Dataframe AST `Request` messages we generate, we will now be including a request ID (generated with `uuid4`) in the Snowpark client. As we plan to include the transitive closure of all Dataframe statements and their dependencies in each `Request` message, statements will be duplicated many times across `Request` messages, so there is a requirement that each statement also carry the "first request ID" of the message they were first included in (to avoid duplication in query history).

We originally would have liked to inherit the request ID generated in the connector (or ideally the SFQID), but it is not possible to retrieve these (and set the "first request ID" field for all statements included for the first time in a `Request` message before sending it to the server) before the query is actually submitted. Both the HTTP request ID, and SFQID are generated at query submission time.

As we would still like to reduce redundancy and yet another set of UUIDs to keep track of for a request, it would be ideal if we could have the connector inherit the request ID generated upstream for the Snowpark Python client's `Request` messages for Dataframe AST.

sfc-gh-lspiegelberg: Thanks Hemit for adding the context!

@sfc-gh-evandenberg @sfc-gh-mkubik Does this sound reasonable so we can proceed with the change?

DESCRIPTION.md
sfc-gh-lspiegelberg: used your suggestion, thanks!

test/integ/test_cursor.py
sfc-gh-evandenberg: Could you explain why we would duplicate sending statements multiple times?  This sounds like something we would like to avoid doing.  Since each dataframe statement has it's own AST ID, there shouldn't be any reason to keep sending the same AST object multiple times as the server can hold onto it for future reference.  Would it be helpful to provide an example to illustrate this case?

sfc-gh-lspiegelberg: I am not sure I fully understand yet whether we're actually sending statements multiple times, potentially the use of the word request here may be confusing. In this PR, we basically add a change to the HTTP request that the snowflake connector sends out. In particular, the connector invokes REST endpoints of the form `<host>/queries/v1/<rest>/?params`, e.g.
```
return self.rest.request(
            "/queries/v1/abort-request?" + urlencode(url_parameters),
            {
                "sqlText": sql,
                REQUEST_ID: str(request_id),
            },
        )
```
In the existing connector the `request_id` refers to an ID that is assigned to each HTTP request so requests can be traced on the HTTP level. This `request_id` is different from the sfqid that GS generates, it's simply a REST id.

Now, for sending the AST, we would like to avoid introducing another layer of IDs. The AST is sent as part of the *payload* of the HTTP request, and contains an id field. To give the ID field a sensible value we would like to reuse the HTTP request ID from the connector. This is a good idea, because server-side we need those IDs to restore order and having the same value (though redundant) simplifies the design. 

However, the connector is currently designed to generate the HTTP request_id within the function body:
```
def connector_foo(...): # <-- some connector API
   request_id = uuid.uuid4()
   ...
   return self.rest.request(
            "/queries/v1/<foo>?" + urlencode(url_parameters),
            {
                "sqlText": sql,
                REQUEST_ID: str(request_id),
            },
        )
```
Therefore, we allow with this PR to pass in a valid (we check for validity) uuid4 ID via statement_params to set the request_id (<-- the HTTP request id). In particular, with AST mode active we 1.) generate id=uuid4 2.) set payload's AST Request->id to id, 3.) Pass to connector via statement_params the id (this code will be part of the snowpark-python repo).

I hope this example helps! lmk if there are other questions or how I could help to clarify.

sfc-gh-evandenberg: Okay, it sounds like you want the payload object and the request itself to share an "id" which is this request id, and then there are some benefits (which i'm not to familiar with but will take your word on.). 

Here we are kind of hijacking the statement parameter to pass this i value even though it doesn't behave like a normal statement parameter that would go to the server.  Should we use a different place holder name to make clearer since it has special behavior:

REQUEST_ID_STATEMENT_PARAM_NAME = "__INTERNAL_OVERRIDE__REQUEST_ID"

sfc-gh-evandenberg (approved):

test/integ/test_cursor.py
sfc-gh-lspiegelberg: Perfect, I'll go ahead then and merge. The suggestion with the `__INTERNAL` prefix I think is good, I want to softly push back a little though in the sense that the REST interface is public. Therefore, the `requestId` url parameter is also technically a public parameter. To be less confusing, I think the best is to keep `requestId` as is.

PR #2247, opened 2025-04-01
SNOW-2019505 fix inconsistent force_put_overwrite value for _upload a…

sfc-gh-zyao:
…nd _download

### Description
- In `_upload` and `_upload_stream` methods, the expected behavior is to respect user decision on overwrite. in Stored Proc connector, `force_put_overwrite` is by default False, so we don't need to explicitly set the argument ([code reference](https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/f22d5625c781f3401c961caa64b0ad6796dd2a6b/src/snowflake/connector/file_transfer_agent.py#L116)) for FileTransferAgent, but it is by default True in public connector ([code reference](https://github.com/snowflakedb/snowflake-connector-python/blob/985ec5e05b619dcbb395bce4232927abe9a48e2a/src/snowflake/connector/file_transfer_agent.py#L348)), so we do need to set `force_put_overwrite=False` here.

   Fixes SNOW-2019505

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -1811,6 +1811,7 @@ class SnowflakeCursor:
             self,
             "",  # empty command because it is triggered by directly calling this util not by a SQL query
             ret,
+            force_put_overwrite=False,  # _upload should respect user decision on overwriting
         )
         file_transfer_agent.execute()
         self._init_result_and_meta(file_transfer_agent.result())
@@ -1878,6 +1879,7 @@ class SnowflakeCursor:
             "",  # empty command because it is triggered by directly calling this util not by a SQL query
             ret,
             source_from_stream=input_stream,
+            force_put_overwrite=False,  # _upload_stream should respect user decision on overwriting
         )
         file_transfer_agent.execute()
         self._init_result_and_meta(file_transfer_agent.result())


src/snowflake/connector/cursor.py
```diff
@@ -1811,6 +1811,7 @@ def _upload(
             self,
             "",  # empty command because it is triggered by directly calling this util not by a SQL query
             ret,
+            force_put_overwrite=False,  # _upload should respect user decision on overwriting
```
sfc-gh-sfan, line 1814: I'm actually a bit confused. Based on the comment, should we read the options to see user's decision on overwriting? why is `False` the decision?

sfc-gh-zyao: When `force_put_overwrite=True`, we will always overwrite, regardless of the value of `overwrite` (here https://github.com/snowflakedb/snowflake-connector-python/blob/985ec5e05b619dcbb395bce4232927abe9a48e2a/src/snowflake/connector/file_transfer_agent.py#L917). So we are disabling the "force overwrite" in this case

sfc-gh-sfan (approved):

sfc-gh-mmishchenko (approved):

PR #2255, opened 2025-04-07
SNOW-2028051 introduce a new client_fetch_threads connection parameter to decouple threads number limitations on fetching and pre-fetching

sfc-gh-mmishchenko:
- [x] a new connection parameter


src/snowflake/connector/cursor.py
```diff
@@ -1208,7 +1208,8 @@ def _init_result_and_meta(self, data: dict[Any, Any]) -> None:
         self._result_set = ResultSet(
             self,
             result_chunks,
-            self._connection.client_prefetch_threads,
+            self._connection.client_fetch_threads
```
sfc-gh-mkubik, line 1211: what is the difference between this one and `client_prefetch_threads`? Would it make sense to allow customers to adjust the max prefetch threads value as well?

sfc-gh-mkubik (approved):

src/snowflake/connector/cursor.py
sfc-gh-mkubik: Decoupled params sound good then!

src/snowflake/connector/connection.py
```diff
@@ -117,6 +117,7 @@
 
 DEFAULT_CLIENT_PREFETCH_THREADS = 4
 MAX_CLIENT_PREFETCH_THREADS = 10
+MAX_CLIENT_FETCH_THREADS = 1024
```
sfc-gh-pfus, line 120: 1024 as a default, isn't it a bit too much? IIUC this has impact on resources consumed by driver (thread number, bandwidth), number of parallel HTTP connections for single (!) query (what if we have multiple parallel queries?), cloud connections etc.
Also, if we run so many downloads, doesn't it mean that each download is slow (because we have to split bandwidth into great number of downloads)?

sfc-gh-mmishchenko: 1024 is the top limit, default doesn't change from current 4 effectively.
The optimal ( (i.e. the one providing the maximum overall bandwidth) number  of fetching threads depends on the workload and the environment. In my environment the connector reached its peak bandwidth at 32 threads.

sfc-gh-pfus (approved):

PR #2292, opened 2025-04-23
SNOW-2057503 allow only whitelisted schemes for OAuth url parameters

sfc-gh-mmishchenko:
- [x] address a security vulnerability by verifying scheme for specified OAuth endpoints
- [x] add unit tests


src/snowflake/connector/connection.py
```diff
@@ -2263,6 +2264,32 @@ def _check_oauth_required_parameters(self) -> None:
                 ProgrammingError,
                 {
                     "msg": "Oauth code flow requirement 'client_secret' is empty",
-                    "errno": ER_NO_CLIENT_ID,
+                    "errno": ER_NO_CLIENT_SECRET,
+                },
+            )
+        if (
+            self._oauth_authorization_url
+            and not self._oauth_authorization_url.startswith("https://")
```
sfc-gh-dheyman, line 2272: perhaps it's better to call `def parsed = urllib.parse(url)` and check `parsed.scheme`? We can validate URL format this way as well.

sfc-gh-dheyman (approved):

sfc-gh-mkubik (approved):

src/snowflake/connector/connection.py
sfc-gh-mmishchenko: It was my original idea, then I remembered that the default value is "https://{host}:{port}/oauth/authorize", which is not a valid url, but now I again think you're right, it is worth formatting and checking the validity - I'll change it.

PR #2294, opened 2025-04-23
Support client-side opt-in of Refresh Token Rotation in Snowflake OAuth

sfc-gh-pmansour:
A client can optionally include `enable_single_use_refresh_tokens=true` in the authorization code flow to opt-in to single-use refresh token semantics. This is part of a new security feature in Snowflake's native OAuth implementation.

To enable this, a user can set the `enable_single_use_refresh_tokens` connection parameter.

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -355,6 +355,11 @@ DEFAULT_CONFIGURATION: dict[str, tuple[Any, type | tuple[type, ...]]] = {
         True,
         bool,
     ),  # SNOW-XXXXX: remove the check_arrow_conversion_error_on_every_column flag
+    # Client-side opt-in to single-use refresh tokens.
+    "enable_single_use_refresh_tokens": (
+        False,
+        bool,
+    )
 }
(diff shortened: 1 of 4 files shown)


src/snowflake/connector/connection.py
```diff
@@ -355,6 +355,11 @@ def _get_private_bytes_from_file(
         True,
         bool,
     ),  # SNOW-XXXXX: remove the check_arrow_conversion_error_on_every_column flag
+    # Client-side opt-in to single-use refresh tokens.
+    "enable_single_use_refresh_tokens": (
```
sfc-gh-mmishchenko, line 359: I'd add `oauth_` prefix to this connection parameter, like for the existing OAuth-related ones.

sfc-gh-mmishchenko (approved):
Looks good, just a connection param name adjustment probably would make sense.

src/snowflake/connector/connection.py
sfc-gh-pmansour: done

PR #2300, opened 2025-04-28
SNOW-2057867 refactor and fixes to make stage related bind and pandas…

sfc-gh-zyao:
… write work for Python sprocs

- relax the condition of using stage solution for array bind (so we will use bind_size >= threshold instead of bind_size > threshold)
- use `_upload_stream` to implement BindUploadAgent
- use `_upload` to implement write_pandas
- correctly use scoped keyword and temp naming pattern for write_pandas
- fix a minor bug where file_transfer_agent returns meta.dst_file_name as file name in download (dst_file_name may contain sub-directories, whereas name will be the base file name without any sub-directory prefixes. We should keep it consistent with upload, where we use meta.name as file name

### Tests
- the new test_direct_file_operation_utils.py to validate parse_file_operation for _upload and _upload_stream
- existing test test_bindings.py to make sure the change does not break existing bind upload logic
- existing test_pandas_tools.py to make sure the change does not break existing write_pandas logic
- existing test_put_get.py to make sure the change does not break file_transfer_agent logic

Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #NNNN

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

4. (Optional) PR for stored-proc connector:

PR #2303, opened 2025-04-28
SNOW-2057867 refactor BindUploadAgent to make it work for Python sprocs

sfc-gh-zyao:
- relax the condition of using stage solution for array bind (so we will use bind_size >= threshold instead of bind_size > threshold)
- use _upload_stream to implement BindUploadAgent

### Tests
- the new test_direct_file_operation_utils.py to validate parse_file_operation for _upload and _upload_stream
- existing test test_bindings.py to make sure the change does not break existing bind upload logic

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/bind_upload_agent.py
+++ b/src/snowflake/connector/bind_upload_agent.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 from __future__ import annotations
 
+import os
 import uuid
 from io import BytesIO
 from logging import getLogger
@@ -76,8 +77,11 @@ class BindUploadAgent:
                 if row_idx >= len(self.rows) or size >= self._stream_buffer_size:
                     break
             try:
-                self.cursor.execute(
-                    f"PUT file://{row_idx}.csv {self.stage_path}", file_stream=f
+                f.seek(0)
+                self.cursor._upload_stream(
+                    input_stream=f,
+                    stage_location=os.path.join(self.stage_path, f"{row_idx}.csv"),
+                    options={},
                 )
             except Error as err:
                 logger.debug("Failed to upload the bindings file to stage.")
--- a/src/snowflake/connector/bind_upload_agent.py
+++ b/src/snowflake/connector/bind_upload_agent.py
@@ -81,7 +81,7 @@ class BindUploadAgent:
                 self.cursor._upload_stream(
                     input_stream=f,
                     stage_location=os.path.join(self.stage_path, f"{row_idx}.csv"),
-                    options={},
+                    options={"source_compression": "auto_detect"},
                 )
             except Error as err:
                 logger.debug("Failed to upload the bindings file to stage.")
--- a/src/snowflake/connector/direct_file_operation_utils.py
+++ b/src/snowflake/connector/direct_file_operation_utils.py
@@ -2,8 +2,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from .errors import NotSupportedError
-
 if TYPE_CHECKING:
     from .connection import SnowflakeConnection
 
@@ -71,7 +69,7 @@ class FileOperationParser(FileOperationParserBase):
             sql = f"PUT {local_file_name} ? {options_in_sql}"
             params = [stage_location]
         else:
-            raise NotSupportedError(f"unsupported command type: {command_type}")
+            raise NotImplementedError(f"unsupported command type: {command_type}")
 
         with self._connection.cursor() as cursor:
             # Send constructed SQL to server and get back parsing result.
@@ -86,4 +84,4 @@ class StreamDownloader(StreamDownloaderBase):
         pass
 
     def download_as_stream(self, ret, decompress=False):
-        raise NotSupportedError("download_as_stream is not yet supported")
+        raise NotImplementedError("download_as_stream is not yet supported")
--- a/src/snowflake/connector/direct_file_operation_utils.py
+++ b/src/snowflake/connector/direct_file_operation_utils.py
@@ -49,29 +49,6 @@ class FileOperationParser(FileOperationParserBase):
     def __init__(self, connection: SnowflakeConnection):
         self._connection = connection
 
-    def _process_options_for_upload(self, options):
-        """Processes the options dict returns the qmark SQL and corresponding bind values.
-        Args:
-            options (dict): the options dict
-        Returns:
-            a tuple of the qmark SQL and corresponding bind values.
-        """
-        options = options or {}
-
-        options_in_sql_raw = []
-        option_bind_values = []
-
-        for k, v in options.items():
-            # Check that all option names are all valid identifiers for better safety.
-            if not k.isidentifier():
-                raise NotSupportedError(f"unsupported option {k}")
-            # Pass option value in binds for better safety.
-            option_bind_values.append(v)
-            options_in_sql_raw.append(f"{k}=?")
-        options_in_sql = " ".join(options_in_sql_raw)
-
-        return options_in_sql, option_bind_values
-
     def parse_file_operation(
         self,
         stage_location,
@@ -82,22 +59,17 @@ class FileOperationParser(FileOperationParserBase):
         has_source_from_stream=False,
     ):
         """Parses a file operation by constructing SQL and getting the SQL parsing result from server."""
-        options_in_sql, option_bind_values = self._process_options_for_upload(options)
+        options = options or {}
+        options_in_sql = " ".join(f"{k}={v}" for k, v in options.items())
 
         if command_type == CMD_TYPE_UPLOAD:
             if has_source_from_stream:
-                assert (
-                    local_file_name is None
-                ), "local_file_name shall be derived from stage_location for stream uploading."
                 stage_location, unprefixed_local_file_name = stage_location.rsplit(
                     "/", maxsplit=1
                 )
                 local_file_name = "file://" + unprefixed_local_file_name
-            # Escape single quotes.
-            local_file_name = local_file_name.replace("'", "''")
-            # Enclose local_file_name with single quotes and pass stage path by a bind for better safety.
-            sql = f"PUT '{local_file_name}' ? {options_in_sql}"
-            params = [stage_location, *option_bind_values]
+            sql = f"PUT {local_file_name} ? {options_in_sql}"
+            params = [stage_location]
         else:
             raise NotSupportedError(f"unsupported command type: {command_type}")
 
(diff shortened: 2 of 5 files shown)


sfc-gh-zyao:
(@sfc-gh-sfan 's comment, copied from the old PR https://github.com/snowflakedb/snowflake-connector-python/pull/2300#discussion_r2064033423): 
```
This does not look like a pure refactor. We do an assertion, a file name split and some escaping. These do not exist in the original code. Why are they required?
```

- for the assertion: it is a fool-proof mechanism, to make sure when users are doing stream upload, they don't pass in a non-empty `local_file_name`. But maybe it is unnecessary? Because when users call `_upload_stream`, there is no way for them to provide a standalone `local_file_name` anyways, by the definition of the interface.
- for the file name splitting: it is [an existing thing](https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/94316b976898ad4403ab67df1b2dc9b363899407/src/snowflake/connector/cursor.py#L792-L793) for sprocs and we want to keep the same behavior here. (I provided more context down below)
- for the escaping: (assuming you mean the escaping for stuff inside 'file://...') it is to avoid the risk of SQL injection without changing what is allowed in local file path. From a high level (forget about stage name biding for now), we pass SQL like `PUT 'file://<some_local_file_path>' @stage_foo` (note the quotes surrounding the entire local file path, as a SQL injection prevention strategy). And we want to make sure the quotes in user-provided local file path still works as intended, by doing the escape (user could have local file name like, `/dir'foo/test.txt`, without escaping this becomes `'file:///dir'foo/test.txt'`, which is incorrect)

Context:
`_upload_stream` method's behavior is that user passes in `input_stream`, `stage_location` as input, where we derive the file name from `stage_location` (We need a file name because stream itself does not contain file name, and we need a name for each file in stage)

For public connector, we have the same interface and we want to keep the same user journey. Note that server is not able to parse and derive the base file name for us. So we do need to do the split.

sfc-gh-zyao:
(@sfc-gh-sfan 's comment, copied from the old PR https://github.com/snowflakedb/snowflake-connector-python/pull/2300#discussion_r2064037753)
```
Is this correct? The original function use PUT which is somewhat equivalent to _upload?
```

Yes, this (https://github.com/snowflakedb/snowflake-connector-python/blob/98381aeba772fed3395081c385dd0f5de0341959/src/snowflake/connector/bind_upload_agent.py#L79-L81) is the existing public connector implementation, note that the `file_stream=f` is passed to `execute`, which does stream uploading instead of file uploading

sfc-gh-zyao:
(@sfc-gh-sfan 's comment, copied from the old PR https://github.com/snowflakedb/snowflake-connector-python/pull/2300#discussion_r2064042281)
```
nit: unintended change
```
Removed

src/snowflake/connector/direct_file_operation_utils.py
```diff
@@ -49,12 +81,37 @@ def parse_file_operation(
...
+            if has_source_from_stream:
+                assert (
+                    local_file_name is None
+                ), "local_file_name shall be derived from stage_location for stream uploading."
+                stage_location, unprefixed_local_file_name = stage_location.rsplit(
+                    "/", maxsplit=1
+                )
+                local_file_name = "file://" + unprefixed_local_file_name
+            # Escape single quotes.
+            local_file_name = local_file_name.replace("'", "''")
```
sfc-gh-sfan, line 97: I guess my question is why the original code does not need to do the single quote escaping? This likely only applies to the `write_pandas` case because bind upload does not allow user-provided path.

Am I missing something? Is `.replace("'", "\\'")` not sufficient?

src/snowflake/connector/direct_file_operation_utils.py
```diff
@@ -49,12 +81,37 @@ def parse_file_operation(
...
+        """Parses a file operation by constructing SQL and getting the SQL parsing result from server."""
+        options_in_sql, option_bind_values = self._process_options_for_upload(options)
+
+        if command_type == CMD_TYPE_UPLOAD:
+            if has_source_from_stream:
+                assert (
+                    local_file_name is None
+                ), "local_file_name shall be derived from stage_location for stream uploading."
+                stage_location, unprefixed_local_file_name = stage_location.rsplit(
+                    "/", maxsplit=1
```
sfc-gh-sfan, line 93: > for the file name splitting: it is [an existing thing](https://github.com/snowflakedb/Stored-Proc-Python-Connector/blob/94316b976898ad4403ab67df1b2dc9b363899407/src/snowflake/connector/cursor.py#L792-L793) for sprocs and we want to keep the same behavior here. (I provided more context down below)

Is this path used in stored proc? I somehow had the impression we override `parse_file_operation`

sfc-gh-zyao: Your impression is right, we don't use this code in sproc execution. But we will still `rsplit` in public connector though, because the interface of `_upload_stream` looks like
```Python
def _upload_stream(
        self,
        input_stream: IO[bytes],
        stage_location: str,
        options: dict[str, Any],
        _do_reset: bool = True,
    ) -> None:
```
We need rsplit `stage_location` to get the file name

src/snowflake/connector/direct_file_operation_utils.py
sfc-gh-zyao: (As discussed offline, we removed the premature "optimization" from this PR. So we don't need those escape altogether)

sfc-gh-zyao:
@sfc-gh-sfan FYI I have removed the premature optimization, to simplify this change

src/snowflake/connector/direct_file_operation_utils.py
```diff
@@ -49,12 +58,32 @@ def parse_file_operation(
...
+            )
 
 
 class StreamDownloader(StreamDownloaderBase):
     def __init__(self, connection):
         pass
 
     def download_as_stream(self, ret, decompress=False):
-        raise NotImplementedError("download_as_stream is not yet supported")
+        raise NotSupportedError("download_as_stream is not yet supported")
```
sfc-gh-sfan, line 89: Why do we not keep using `NotImplementedError`? 🤔

sfc-gh-sfan (approved):

src/snowflake/connector/direct_file_operation_utils.py
sfc-gh-zyao: Actually let me actually keep using the existing `NotImplementedError`. I made this change because I realized that most our code base uses error types derived from [Error class](https://github.com/snowflakedb/snowflake-connector-python/blob/15efed3ac2c2e7a61adf701ba56a307deb1fc4b1/src/snowflake/connector/errors.py#L26), and I tried to adhere to that. But it is better to avoiding doing that in this change, and focus on the key change we need

sfc-gh-zyao:
Hi @sfc-gh-sfan commit [a3c66cf](https://github.com/snowflakedb/snowflake-connector-python/pull/2303/commits/a3c66cfb8748d4620d344037b7253d7eebee4a9c) is the extra fix that I mentioned earlier. Could you please take a look and let me know if it looks good to you

sfc-gh-mmishchenko (commented):
Looks good, besides that private methods/fields naming convention is violated now.

src/snowflake/connector/bind_upload_agent.py
```diff
@@ -76,8 +77,11 @@ def upload(self) -> None:
                 if row_idx >= len(self.rows) or size >= self._stream_buffer_size:
                     break
             try:
-                self.cursor.execute(
-                    f"PUT file://{row_idx}.csv {self.stage_path}", file_stream=f
+                f.seek(0)
+                self.cursor._upload_stream(
```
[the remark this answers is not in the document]
sfc-gh-zyao: We might want to still keep `_upload_stream` as internal utils (i.e. to simplify internal development, but not for external customer to directly consume). What would your suggestion be for this case?

Another reason is that we already have `upload_stream` interface long time ago (way before it was handed over to me) and `snowpark-python` still uses it (for example, [this one](https://github.com/snowflakedb/snowpark-python/blob/06709ec09fc6a535765c9d35925e7102d2b0cea8/src/snowflake/snowpark/session.py#L1391)). So this also prevents me making the renaming.

src/snowflake/connector/direct_file_operation_utils.py
```diff
@@ -49,7 +57,27 @@ def parse_file_operation(
...
+                )
+                local_file_name = "file://" + unprefixed_local_file_name
+            sql = f"PUT {local_file_name} ? {options_in_sql}"
+            params = [stage_location]
+        else:
+            raise NotImplementedError(f"unsupported command type: {command_type}")
+
+        with self._connection.cursor() as cursor:
+            # Send constructed SQL to server and get back parsing result.
+            processed_params = cursor._connection._process_params_qmarks(params, cursor)
```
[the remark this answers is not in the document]
sfc-gh-zyao: I see, I guess your suggestion to do renaming like `_process_params_qmarks  -> process_params_qmarks `? I am a bit worried about confusing external customers, by making `_process_params_qmarks` public. Because normally they will directly use `execute` to run a SQL, without worrying about these binding param utils. 

And here we are more or less using private / internal methods to implement it. Would it be less awkward, if I change to something like
```
processed_params = self._connection._process_params_qmarks(params, cursor)
```
(Such that it is more like we are accessing some internal fields of this current class, and therefore more natural)

src/snowflake/connector/bind_upload_agent.py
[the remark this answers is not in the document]
sfc-gh-mmishchenko: Thanks for the explanation, I see there are cross-dependencies already.

sfc-gh-mmishchenko (approved):
Looks good.

sfc-gh-zyao:
The only failures in "continuous integration" are
- test.unit.test_oauth_token.test_oauth_code_successful_flow
- test.unit.test_oauth_token.test_oauth_code_invalid_state
- test.unit.test_oauth_token.test_oauth_code_token_request_error
- test.unit.test_oauth_token.test_oauth_code_browser_timeout
- test.unit.test_oauth_token.test_oauth_code_custom_urls
- test.unit.test_oauth_token.test_oauth_code_successful_refresh_token_flow
- test.unit.test_oauth_token.test_oauth_code_expired_refresh_token_flow
- test.unit.test_oauth_token.test_client_creds_successful_flow
- test.unit.test_oauth_token.test_client_creds_successful_refresh_token_flow
- test.unit.test_oauth_token.test_client_creds_expired_refresh_token_flow

Those are irrelevant to this change

Issue #2282, opened 2025-04-18
SNOW-2047992: Support for newer pyarrow versions (>19.0.0)


kang8:
### What is the current behavior?

The snowflake-connector-python package currently has a version restriction for pyarrow in its pandas extra requirements: pyarrow<19.0.0 (as seen in setup.cfg). This prevents users from using the latest pyarrow versions with the connector.

### What is the desired behavior?

Update or remove the upper version limit on pyarrow dependency to allow using the latest pyarrow versions with the Snowflake connector.

### How would this improve `snowflake-connector-python`?

This change would allow users to:
* Use the latest pyarrow releases with snowflake-connector-python
* Avoid dependency conflicts in projects that require newer pyarrow versions
* Benefit from performance improvements and bug fixes in newer pyarrow versions
* Simplify dependency management in projects using both snowflake-connector-python and other libraries that depend on pyarrow

### References and other background

I noticed this restriction was added in PR #2163. If there were specific compatibility issues that led to this constraint, it would be helpful to understand if those have been resolved or if there's a timeline for supporting newer versions of pyarrow.

sfc-gh-sghosh:
Hello @kang8 ,

Thanks for raising the issue, we will check and update whether pyarrow versions > 19.0.0 is supported or not, if not why.

Regards,
Sujan

sfc-gh-sghosh:
Hello @kang8 ,

Currently, the fix will tentatively be delivered by mid-May 2025. 

Regards,
Sujan

diwu-sf:
Hi, can you make a new release with pyArrow 19 support please?

sfc-gh-dszmolka:
i see we still have pyarrow<19.0.0 so reopening until it's delivered

sfc-gh-mmishchenko:
@sfc-gh-dszmolka Sure, we'll prioritise this delivery.

jonathanunderwood:
@sfc-gh-dszmolka Can we get an ETA on this please? The fact this isn't fixed in #2310  suggests it's not being looked at.

sfc-gh-dszmolka:
I'm unable to give an ETA at this point, but should I get any closer from internal infos, I'll update here too. Please in the meantime, watch this space as this issue will be updated as events unfold. Thank you for bearing with us while this is delivered. 

**edit**: as a workaround (for non-Windows use-cases); installing the required pyarrow version by explicitly specifying it, should work (e.g. a separate `pip install pyarrow==19`)

[...]

PR #2310, opened 2025-05-05
SNOW-2081683 require pyarrow>=10.0.1

sfc-gh-mkubik:
Add lower threshold for pyarrow version to ensure compatibility with newer pandas releases

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -94,6 +94,6 @@ development =
     pytzdata
 pandas =
     pandas>=2.1.2,<3.0.0
-    pyarrow<19.0.0
+    pyarrow>=10.0.1,<19.0.0
 secure-local-storage =
     keyring>=23.1.0,<26.0.0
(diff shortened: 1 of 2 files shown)


sfc-gh-pbulawa (approved):

setup.cfg
```diff
@@ -94,6 +94,6 @@ development =
     pytzdata
 pandas =
     pandas>=2.1.2,<3.0.0
-    pyarrow<19.0.0
+    pyarrow>=10.0.1,<19.0.0
```
jonathanunderwood, line 97: @sfc-gh-mkubik  The current supported PyArrow release is version 20. Why the restriction to <19 here? That forces an outdated and unsupported PyArrow release. Pandas has no such restriction, so wondering why this is here?

I actually just raised #2349 about this.

jonathanunderwood: In fact, #2349 has been closed as a duplicate of an older issue for this #2282, so it would be good to get some focus here.

sfc-gh-turbaszek:

PR #2322, opened 2025-05-14
SNOW-1947479 Add bulk_upload_chunks parameter to write_pandas

sfc-gh-mkubik:
Add `bulk_upload_chunks` parameter to `write_pandas` function to enable upload of all chunks at once.

### DIFF
--- a/prober/Dockerfile
+++ b/prober/Dockerfile
@@ -1,0 +1,64 @@
+FROM alpine:3.18
+
+# boilerplate labels required by validation when pushing to ACR, ECR & GCR
+LABEL org.opencontainers.image.source="https://github.com/snowflakedb/snowflake-connector-python"
+LABEL com.snowflake.owners.email="triage-snow-drivers-warsaw-dl@snowflake.com"
+LABEL com.snowflake.owners.slack="triage-snow-drivers-warsaw-dl"
+LABEL com.snowflake.owners.team="Snow Drivers"
+LABEL com.snowflake.owners.jira_area="Developer Platform"
+LABEL com.snowflake.owners.jira_component="Python Driver"
+# fake layers label to pass the validation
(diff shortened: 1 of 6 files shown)


test/integ/pandas/test_pandas_tools.py
```diff
@@ -1139,3 +1139,49 @@ def test_pandas_with_single_quote(
...
+    random_table_name = random_string(5, "userspoints_")
+    df_data = [("Dash", 50), ("Luke", 20), ("Mark", 10), ("John", 30)]
+    df = pandas.DataFrame(df_data, columns=["name", "points"])
+
+    table_name = random_table_name
+    col_id = "id"
+    col_name = "name"
+    col_points = "points"
+
+    create_sql = (
```
sfc-gh-fpawlowski, line 1156: nit: could be nice to have a context-manager fixture (maybe factory) that would create table and drop it after leaving the context - it is quite a common thing to do in this tests file.

But it could be considered in the future enhancements as well.

sfc-gh-fpawlowski (changes requested):

src/snowflake/connector/pandas_tools.py
```diff
@@ -437,17 +440,27 @@ def write_pandas(
...
+                # Upload parquet file chunk right away
+                path = chunk_path.replace("\\", "\\\\").replace("'", "\\'")
+                cursor._upload(
+                    local_file_name=f"'file://{path}'",
+                    stage_location="@" + stage_location,
+                    options={"parallel": parallel, "source_compression": "auto_detect"},
+                )
+
+                # Remove chunk file
+                os.remove(chunk_path)
```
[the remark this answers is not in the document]
sfc-gh-mkubik: That's the only reason that I can think of. Sounds like an edge case so possibly we'll be able to make the aggregated upload a default behaviour

test/integ/pandas/test_pandas_tools.py
```diff
@@ -1139,3 +1139,49 @@ def test_pandas_with_single_quote(
             )
         finally:
             cnx.execute_string(f"drop table if exists {table_name}")
+
+
+@pytest.mark.parametrize("write_then_upload", [True, False])
+def test_write_pandas_wildcard_upload(conn_cnx, write_then_upload):
```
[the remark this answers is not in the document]
sfc-gh-mkubik: That's the previous name I've used for this parameter, good catch

test/integ/pandas/test_pandas_tools.py
sfc-gh-mkubik: actually in `test/integ/pandas/test_pandas_tools.py` it seems that the conventions is to drop the test table in finally block but it makes sense to unify this file with other tests. I'd do a separate PR for that though.

sfc-gh-fpawlowski: Yes - that the whole point - the context manager would contain that exact 'finally' block and execute it automatically as fixture (or fixture-produced ContextManager) exits the context of 'with' block.

But I agree - this can be done in a separate PR

sfc-gh-pbulawa (approved):
LGTM

sfc-gh-fpawlowski (approved):

PR #2329, opened 2025-05-21
SNOW-2110470: Implemented local application

sfc-gh-fpawlowski:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/auth/oauth_code.py
+++ b/src/snowflake/connector/auth/oauth_code.py
@@ -45,6 +45,8 @@ def _get_query_params(
 class AuthByOauthCode(AuthByOAuthBase):
     """Authenticates user by OAuth code flow."""
 
+    _LOCAL_APPLICATION_CLIENT_CREDENTIALS = "LOCAL_APPLICATION"
+
     def __init__(
         self,
         application: str,
@@ -54,6 +56,7 @@ class AuthByOauthCode(AuthByOAuthBase):
         token_request_url: str,
         redirect_uri: str,
         scope: str,
+        host: str,
         pkce_enabled: bool = True,
         token_cache: TokenCache | None = None,
         refresh_token_enabled: bool = False,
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/auth/oauth_code.py
```diff
@@ -385,3 +396,27 @@ def _parse_authorization_redirected_request(
...
+        client_id: str,
+        client_secret: str,
+        authentication_url: str,
+        token_request_url: str,
+        host: str,
+    ) -> bool:
+        return (
+            client_id == ""
+            and client_secret == ""
+            and self.__class__._is_snowflake_as_idp(
```
sfc-gh-mkubik, line 411: Should this be possible to override in subclasses?
If not then just a nit:
```suggestion
            and AuthByOauthCode._is_snowflake_as_idp(
```

src/snowflake/connector/auth/oauth_code.py
```diff
@@ -385,3 +396,27 @@ def _parse_authorization_redirected_request(
...
+    def _eligible_for_default_client_credentials(
+        self,
+        client_id: str,
+        client_secret: str,
+        authentication_url: str,
+        token_request_url: str,
+        host: str,
+    ) -> bool:
+        return (
+            client_id == ""
```
sfc-gh-mkubik, line 409: Should this only work when customer explicitly passes `oauth_client_id=""` or None client id should also be allowed? If yes then such cases may be filtered by this check: https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/connection.py#L1207

src/snowflake/connector/auth/oauth_code.py
sfc-gh-fpawlowski: I think that if for some reason a subclass overrides this function it should be reflected in this behaviour, without a need to override also the other methods.

src/snowflake/connector/auth/oauth_code.py
sfc-gh-fpawlowski: Good point, we would be duplicating those checks. I managed to split that logic and refactor it into the base Oauth class.

sfc-gh-mkubik (approved):

PR #2330, opened 2025-05-21
Filter out Deprecation warnings from test_incalid_conection_parameter…

sfc-gh-mkubik:
Temporarily filter out the deprecation warning from asserts in test_invalid_connection_parameters tests. The filtering should be rolled back once the cause of DeprecationWarning is fixed.

### DIFF
--- a/test/integ/test_connection.py
+++ b/test/integ/test_connection.py
@@ -878,7 +878,7 @@ def test_invalid_connection_parameter(db_parameters, name, value, exc_warn):
         try:
             conn = snowflake.connector.connect(**conn_params)
             assert getattr(conn, "_" + name) == value
-            # TODO: remove filtering once the root cause for deprecation warning is fixed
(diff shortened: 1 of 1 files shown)


test/integ/test_connection.py
```diff
@@ -878,8 +878,13 @@ def test_invalid_connection_parameter(db_parameters, name, value, exc_warn):
         try:
             conn = snowflake.connector.connect(**conn_params)
             assert getattr(conn, "_" + name) == value
-            assert len(w) == 1
-            assert str(w[0].message) == str(exc_warn)
+            # TODO: remove filtering once the root cause for deprecation warning is fixed
```
sfc-gh-fpawlowski, line 881: Maybe lets either include some part of the deprecation warning message or create KTLO task and place here its ID - so we know in the future what was it

sfc-gh-mkubik: makes sense, I'll add a jira reference

sfc-gh-fpawlowski (approved):
LGTM

PR #2338, opened 2025-05-27
SNOW-1983343 add timeout for ocsp root certs

sfc-gh-mkubik:
Customers identified that ocsp root certs dictionary access may cause a deadlock when using the driver in multiple threads. We weren't able to identify the reproduction scenario so instead we agreed on introducing a parameter that allows configuring the timeout for the lock instead of waiting indefinitely.

Additionally, new parameter for OCSP revealed a problem with globally defined connection mocks in test_session_manager that got injected into unrelated tests. This was fixed by applying per-test patch instead of the global mock.


sfc-gh-knozderko (approved):

sfc-gh-pbulawa (commented):
Can we have some test for the lock mechanism?

sfc-gh-turbaszek:

PR #2367, opened 2025-06-20
SNOW-2111644 Support sovereign clouds for WIF

sfc-gh-xizhao:

   SNOW-2111644

   This PR adds the support for sovereign clouds for WIF, specifically supporting additional partitions in AWS, and additional issuers in Azure.

### DIFF
--- a/src/snowflake/connector/wif_util.py
+++ b/src/snowflake/connector/wif_util.py
@@ -22,6 +22,19 @@ logger = logging.getLogger(__name__)
 SNOWFLAKE_AUDIENCE = "snowflakecomputing.com"
 DEFAULT_ENTRA_SNOWFLAKE_RESOURCE = "api://fd3f753b-eed3-462c-b6a7-a4b5bb650aad"
 
+"""
+References:
+- https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud#microsoft-entra-authentication-endpoints
+- https://learn.microsoft.com/en-us/answers/questions/1190472/what-are-the-token-issuers-for-the-sovereign-cloud
+"""
+AZURE_ISSUER_PREFIXES = [
+    "https://sts.windows.net/",  # Public and USGov (v1 issuer)
+    "https://sts.chinacloudapi.cn",  # Mooncake (v1 issuer)
+    "https://login.microsoftonline.com/",  # Public (v2 issuer)
+    "https://login.microsoftonline.us",  # USGov (v2 issuer)
+    "https://login.partner.microsoftonline.cn",  # Mooncake (v2 issuer)
+]
+
 
 @unique
 class AttestationProvider(Enum):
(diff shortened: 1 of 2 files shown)


sfc-gh-xizhao:

src/snowflake/connector/wif_util.py
```diff
@@ -22,6 +22,19 @@
...
 DEFAULT_ENTRA_SNOWFLAKE_RESOURCE = "api://fd3f753b-eed3-462c-b6a7-a4b5bb650aad"
 
+"""
+References:
+- https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud#microsoft-entra-authentication-endpoints
+- https://learn.microsoft.com/en-us/answers/questions/1190472/what-are-the-token-issuers-for-the-sovereign-cloud
+"""
+AZURE_ISSUER_PREFIXES = [
+    "https://sts.windows.net/",  # Public and USGov (v1 issuer)
+    "https://sts.chinacloudapi.cn",  # Mooncake (v1 issuer)
```
sfc-gh-pmansour: nit: let's be consistent and either add trailing `/` in all of them or none of them.

src/snowflake/connector/wif_util.py
```diff
@@ -108,6 +121,72 @@ def get_aws_arn() -> str | None:
...
+    elif partition == "aws-cn":
+        # China regions have a different domain suffix
+        return f"sts.{region}.amazonaws.com.cn"
+    elif partition == "aws-us-gov":
+        return (
+            f"sts.{region}.amazonaws.com"  # GovCloud uses .com, but dedicated regions
+        )
+    else:
+        logger.warning("Invalid AWS partition: %s", partition)
+        return None
```
sfc-gh-pmansour: According to Claude, the boto3 SDK can do this directly (though obviously we should verify this):

```
import boto3

# Get the endpoint resolver directly from the session
resolver = boto3.Session().get_component('endpoint_resolver')

# Resolve endpoint for specific service/region/partition
endpoint = resolver.construct_endpoint(
    service_name='sts',
    region_name='ap-southeast-1',
    partition_name='aws'
)

print(endpoint['hostname'])  # sts.ap-southeast-1.amazonaws.com
```

Should we use that instead of hardcoding the endpoint suffixes ourselves?

src/snowflake/connector/wif_util.py
```diff
@@ -108,6 +121,72 @@ def get_aws_arn() -> str | None:
...
+    logger.warning("Invalid AWS ARN: %s", arn)
+    return None
+
+
+def get_aws_sts_hostname(region: str, partition: str) -> str | None:
+    """Constructs the AWS STS hostname for a given region and partition.
+
+    Args:
+        region (str): The AWS region (e.g., 'us-east-1', 'cn-north-1').
+                      Can be an empty string for global STS endpoint if applicable
```
sfc-gh-pmansour: I don't think this is true -- in the current implementation, if this is an empty string, this method will return None.

sfc-gh-pmansour (approved):
Nice!

src/snowflake/connector/wif_util.py
sfc-gh-xizhao: That would be ideal. But the sample code doesn't seem to work since `get_component()` is not supported. Let me know if you have better idea. Would love to get rid of hardcoding and custom logic.

src/snowflake/connector/wif_util.py
sfc-gh-xizhao: Good call, thanks!

src/snowflake/connector/wif_util.py
sfc-gh-pmansour: Claude suggested these alternatives instead:

> #### Method 1: Through botocore session
> ```python
> import boto3
>
> session = boto3.Session()
> # Get the underlying botocore session
> botocore_session = session._session
>
> # Get the endpoint resolver from botocore
> resolver = botocore_session.get_component('endpoint_resolver')
>
> endpoint = resolver.construct_endpoint(
>     service_name='s3',
>     region_name='ap-southeast-1',
>     partition_name='aws'
> )
>
> print(endpoint['hostname'])  # s3.ap-southeast-1.amazonaws.com
> ```
> #### Method 2: Import directly from botocore
> ```python
> import botocore.session
>
> # Create a botocore session directly
> session = botocore.session.get_session()
> resolver = session.get_component('endpoint_resolver')
>
> endpoint = resolver.construct_endpoint(
>     service_name='s3',
>     region_name='ap-southeast-1',
>     partition_name='aws'
> )
>
> print(endpoint['hostname'])  # s3.ap-southeast-1.amazonaws.com
> ```
> #### Method 3: More direct approach using botocore
> ```python
> from botocore.regions import EndpointResolver
> from botocore.loaders import Loader
>
> loader = Loader()
> endpoints_data = loader.load_data('endpoints')
> resolver = EndpointResolver(endpoints_data)
>
> endpoint = resolver.construct_endpoint(
>     service_name='s3',
>     region_name='ap-southeast-1'
> )
>
> print(endpoint['hostname'])  # s3.ap-southeast-1.amazonaws.com
> ```

IMHO, the first one is too hacky to use in product code. The second is fine but unnecessarily does the work to get a session. Is the third one worth a shot?

sfc-gh-xizhao: I'll try them. But the 3rd one doesn't take in partition so I'm afraid it won't work well for sovereign clouds.

sfc-gh-jkasten (approved):
Thanks!

src/snowflake/connector/wif_util.py
sfc-gh-xizhao: Looks like none of them worked. Discussed and decided to use the custom logic for now.

src/snowflake/connector/wif_util.py
sfc-gh-xizhao: updated

PR #2374, opened 2025-06-24
SNOW-2173685 respect existing param control of using SCOPED keyword f…

sfc-gh-zyao:
…or sproc

### Description
- respect existing param control of using SCOPED keyword for sproc
  - currently whether to use scoped keyword for sproc is controlled by the parameter `ENABLE_FIX_1375538`, and this change makes sure we respect that
  - the change is pure no-op for non-sproc because the associated param is not fetched outside sproc, making the exactly the same as existing code (which is not to use scoped) (what's more, for non-sproc, with or without SCOPED keyword will both lead to using session-scoped temp objects when writing pandas. So it is effectively no-op change, even if non-sproc can fetch the param)
  - for Snowflake reviewers, more details about this decision is covered in [this doc](https://docs.google.com/document/d/1nu9Iksk0_6LwCbwCQWpUUA3j7jANXzKprmWNImdlaPg)
### Testing
- existing tests to make sure this change does not break anything
- manual testing to make sure this is working for sproc
  - when we enable `ENABLE_FIX_1375538`, we see a job-scoped temp object is created
     scoped enabled
  - when we disable `ENABLE_FIX_1375538`, we are not seeing job-scoped temp object. Instead, we see session-scoped temp, which remains valid even beyond sproc
     scoped disabled

   Fixes #NNNN


src/snowflake/connector/pandas_tools.py
```diff
@@ -354,19 +353,11 @@ def write_pandas(
         )
 
     _use_scoped_temp_object = (
-        conn._session_parameters.get(
-            _PYTHON_SNOWPARK_USE_SCOPED_TEMP_OBJECTS_STRING, False
-        )
+        conn._session_parameters.get("ENABLE_FIX_1375538", False)
```
sfc-gh-sfan: Let's add a note to remove this with a BCR from the server side to make these temp objects job scoped.

Also, let's use a more meaning variable name to represent `ENABLE_FIX_1375538`. We should verify if this param can be fetched outside of sp. I suspect not and this might just be a pure no-op outside of sp.

sfc-gh-zyao: Done.

- Added the note with the corresponding SNOW ticket.
- Used a constant named `_PARAM_USE_SCOPED_TEMP_FOR_PANDAS_TOOLS` to represent the `ENABLE_FIX_1375538` parameter name
- Verified that the parameter ENABLE_FIX_1375538 is not fetched at all when running in general clients / non-sproc. So good point, this change is pure no-op outside sproc (yay!)

sfc-gh-sfan (approved):

PR #2375, opened 2025-06-25
SNOW-2173966 introduce server DoP cap

sfc-gh-zyao:
### Description
- introduce server DoP cap to connector
  - Currently sproc in fact use serial processing when it comes to file transfer agent execution. We introduce this server DoP cap to allow server side control of such behaviors
  - for Snowflake reviewers, more details are covered in [this doc](https://docs.google.com/document/d/1KegQLzyxB5G2Viti5Tr8fD24IPjIKU3cCZtRUUWqm4k/)
 ### Testing
- Unit tests for the new change
- Existing tests to make sure this change does not break anything
- Manual testing to show that this change works for sproc
Screenshot 2025-06-24 at 6 05 33 PM

   Fixes #NNNN


src/snowflake/connector/cursor.py
```diff
@@ -1086,6 +1088,11 @@ def execute(
                     use_s3_regional_url=self._connection.enable_stage_s3_privatelink_for_us_east_1,
                     iobound_tpe_limit=self._connection.iobound_tpe_limit,
                     unsafe_file_write=self._connection.unsafe_file_write,
+                    snowflake_server_dop_cap_for_file_transfer=getattr(
+                        self.connection,
```
sfc-gh-sfan: Out of curiosity, why is this not `self._connection` like above?

src/snowflake/connector/cursor.py
```diff
@@ -1839,6 +1851,11 @@ def _upload(
             "",  # empty command because it is triggered by directly calling this util not by a SQL query
             ret,
             force_put_overwrite=False,  # _upload should respect user decision on overwriting
+            snowflake_server_dop_cap_for_file_transfer=getattr(
+                self.connection,
+                f"_{_VARIABLE_NAME_SERVER_DOP_CAP_FOR_FILE_TRANSFER}",
+                _DEFAULT_VALUE_SERVER_DOP_CAP_FOR_FILE_TRANSFER,
+            ),
```
sfc-gh-sfan: This is used multiple times. Maybe we could create helper function? 

It might be okay to expose this as a connection property as the name seems legit. This can just be called `_snowflake_max_parallel_for_file_transfer`?

src/snowflake/connector/cursor.py
sfc-gh-zyao: You are right, I think self._connection is more aligned with the existing code here. I ported it as is from the sproc connector in the initial commit (I mean, it looks natural in sproc connector, but not so natural given that all other usages are `self._connection`

I have changed it to make them aligned

src/snowflake/connector/cursor.py
sfc-gh-zyao: Done creating a helper for that. As for exposing it as property, I may be more conservative here. Given that the public connector already has 2 different parallelism bounds/limits, I should probably not adding more there to make it confusing. What do you think?

src/snowflake/connector/cursor.py
sfc-gh-zyao: btw I was referring to these `_parallel` and `_iobound_tpe_limit` things at https://github.com/snowflakedb/snowflake-connector-python/blob/411c973763b84915b1e9e1e52f04df558cde2c38/src/snowflake/connector/file_transfer_agent.py#L450-L451, though they are not property of connection, they are in FileTransferAgent

sfc-gh-sfan (approved):

PR #2387, opened 2025-07-04
SNOW-2171791: Add platform telemetry

sfc-gh-sshetkar:

   [SNOW-2171791](https://snowflakecomputing.atlassian.net/browse/SNOW-2171791)

   - [x] I am adding a new telemetry message

   This PR adds platform detection mechanisms and appends them to the CLIENT_ENVIRONMENT part of the base_auth_data(). This then gets sent to the /login-request endpoint which serializes and stores this data for future use in the SessionDPO.

To avoid bottlenecking the authentication flow, we have parallelized all platform detection I/O calls with a default timeout of 200 milliseconds. Overall testing has found that these tests take on the order of 1-3 milliseconds with more details provided in the design doc. We have also exposed a user parameter so that they can configure how long of a timeout they are ok with.

[Design Doc](https://docs.google.com/document/d/10zjM4nBuXWOcm_OciJzCZ4t9lfrDie9LjMXf7SCS25c/edit?usp=sharing)

[SNOW-2171791]: https://snowflakecomputing.atlassian.net/browse/SNOW-2171791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/csp_helpers.py
+++ b/test/csp_helpers.py
@@ -444,4 +444,4 @@ class FakeAwsLambdaEnvironment(FakeAwsEnvironment):
 
     def __exit__(self, *args, **kwargs):
         self.os_environment_patch.__exit__(*args)
-        super().__exit__(args, **kwargs)
+        super().__exit__(*args, **kwargs)


src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def is_ec2_instance(timeout=0.5):
+    try:
+        fetcher = IMDSFetcher(timeout=timeout, num_attempts=2)
+        document = fetcher._get_request(
+            "/latest/dynamic/instance-identity/document",
+            None,
+            fetcher._fetch_metadata_token(),
+        )
```
sfc-gh-pmansour: I think the answer is yes, but just confirming -- does this work on both IMDSv1 and IMDSv2?

Is there a way to test this, e.g. by packaging a local build of the driver and running it on EC2 VMs? I have a script I use to do that, if it would help :)

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,151 @@
...
+    patterns = [
+        r"^arn:[^:]+:iam::[^:]+:user/.+$",
+        r"^arn:[^:]+:sts::[^:]+:assumed-role/.+$",
+    ]
+    return any(re.match(p, arn) for p in patterns)
+
+
+def has_aws_identity():
+    try:
+        caller_identity = boto3.client("sts").get_caller_identity()
```
sfc-gh-pmansour: can we configure the timeout and max attempts here too?

sfc-gh-sshetkar:

src/snowflake/connector/connection.py
```diff
@@ -197,6 +197,10 @@ def _get_private_bytes_from_file(
     ),  # network timeout (infinite by default)
     "socket_timeout": (None, (type(None), int)),
     "external_browser_timeout": (120, int),
+    "platform_detection_timeout": (
+        None,
+        (type(None), int, float),
```
sfc-gh-pmansour: Why would we ever want this to be a float?

Can we make this an int, and include a unit? e.g. `platform_detection_timeout_milliseconds`

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,233 @@
...
+
+def is_github_action():
+    return (
+        _DetectionState.DETECTED
+        if "GITHUB_ACTIONS" in os.environ
+        else _DetectionState.NOT_DETECTED
+    )
+
+
+def detect_platforms(timeout: int | float | None) -> list[str]:
```
sfc-gh-mmishchenko: Wouldn't it make sense to make use of some cache to detect the platform just once? Probably for a particular setup the environment is not expected to change.
The driver uses file or keyring-based cache for tokens and OCSP check results (cache.py), so that could be leveraged here possibly?

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,233 @@
...
+
+class _DetectionState(Enum):
+    """Internal enum to represent the detection state of a platform."""
+
+    DETECTED = "detected"
+    NOT_DETECTED = "not_detected"
+    TIMEOUT = "timeout"
+
+
+def is_ec2_instance(timeout):
```
sfc-gh-pmansour: Could you please add doc comments for each of these methods, with at least 1 sentence on what and how they do it? Including any assumptions (e.g. if env variable foo is present, we assume driver is running on bar)

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,233 @@
...
+        _DetectionState.DETECTED
+        if all(var in os.environ for var in service_vars)
+        else _DetectionState.NOT_DETECTED
+    )
+
+
+def is_managed_identity_available_on_azure_vm(
+    timeout, resource=DEFAULT_ENTRA_SNOWFLAKE_RESOURCE
+):
+    endpoint = f"http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource={resource}"
```
sfc-gh-pmansour, line 195: This will be mostly false-positives, because we expect most tenants won't have consented to the DEFAULT_ENTRA_SNOWFLAKE_RESOURCE so getting a token will fail even if you have a managed identity available.

We should request a token for some resource that all entra customers don't need consent for. I suggest `https://management.azure.com/`, so `&resource=https://management.azure.com/`.

Can you try this out?

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,233 @@
...
+            "is_gce_cloud_run_job": executor.submit(is_gce_cloud_run_job),
+            "has_gcp_identity": executor.submit(has_gcp_identity, timeout),
+            "is_github_action": executor.submit(is_github_action),
+        }
+
+        platforms = {key: future.result() for key, future in futures.items()}
+
+    platforms["azure_managed_identity"] = has_azure_managed_identity(
+        platforms["is_azure_vm"], platforms["is_azure_function"], timeout
+    )
```
sfc-gh-pmansour: Blocking the `azure_managed_identity` check on completing the `is_azure_vm` and `is_azure_function` checks can result in 2x the max timeout we limit.

Instead, could you make `has_azure_managed_identity` just always try both options above? First check for `IDENTITY_HEADER`, which should be instant, then try to get a token using the metadata service, which is slower but at least done in parallel with the first batch.

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,233 @@
...
+    try:
+        response = requests.get(
+            "http://metadata/computeMetadata/v1/instance/service-accounts/default/email",
+            headers={"Metadata-Flavor": "Google"},
+            timeout=timeout,
+        )
+        response.raise_for_status()
+        return (
+            _DetectionState.DETECTED if response.text else _DetectionState.NOT_DETECTED
+        )
```
sfc-gh-pmansour: nit: instead of raising for status, why not make DETECTED based on the status code?

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,233 @@
...
+    )
+
+    detected_platforms = []
+    for platform_name, detection_state in platforms.items():
+        if detection_state == _DetectionState.DETECTED:
+            detected_platforms.append(platform_name)
+        elif detection_state == _DetectionState.TIMEOUT:
+            detected_platforms.append(f"{platform_name}_timeout")
+        elif detection_state == _DetectionState.NOT_DETECTED:
+            pass
```
sfc-gh-pmansour: nit: useless

src/snowflake/connector/auth/_auth.py
```diff
@@ -101,6 +102,7 @@ def base_auth_data(
         login_timeout: int | None = None,
         network_timeout: int | None = None,
         socket_timeout: int | None = None,
+        platform_detection_timeout: int | float | None = None,
```
sfc-gh-pmansour: I don't see this `base_auth_data` method being explicitly called anywhere. Do you where that happens, and how this new timeout field gets passed?

src/snowflake/connector/platform_detection.py
sfc-gh-pmansour: Could we make this entire method have a general try..catch that logs and swallows failures?

I know the individual REST API calls above have their own try..catch (though some of them only catch specific exceptions), but I think we don't ever want the connection to fail due to any failure to detect the platform (including threading issues).

sfc-gh-pmansour: If this is cheap to do (we guarantee it'll be <0.2s), is there a harm in just doing it every time, rather than introducing another source of complexity and potential bugs in terms of the cache?

test/unit/test_connection.py
```diff
@@ -58,6 +58,14 @@ def __init__(self, password: str, mfa_token: str) -> None:
...
 
 
+@pytest.fixture(autouse=True)
+def mock_detect_platforms():
+    with patch(
+        "snowflake.connector.auth._auth.detect_platforms", return_value=[]
+    ) as mock_detect:
+        yield mock_detect
+
+
```
sfc-gh-pmansour, line 68: Can we add a test that doesn't use this new fixture, and ensures that detect_platforms gets correctly called?

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,489 @@
...
+class MockResponse:
+    def __init__(self, status_code=200, content=b"", text="", headers=None):
+        self.status_code = status_code
+        self.content = content
+        self.text = text
+        self.headers = headers or {}
+
+    def raise_for_status(self):
+        if self.status_code >= 400:
+            raise requests.HTTPError(f"HTTP {self.status_code} Error")
```
sfc-gh-pmansour: Have you considered using the real `Response` class from the requests library? That way you don't need to reimplement `raise_for_status`, or mock any other methods that your product code starts using.

See https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/test/csp_helpers.py#L42-L47

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,489 @@
...
+
+
+class TestDetectPlatforms:
+    @patch.dict(os.environ, {}, clear=True)
+    @patch("snowflake.connector.platform_detection.IMDSFetcher")
+    @patch("snowflake.connector.platform_detection.boto3")
+    @patch("snowflake.connector.platform_detection.requests")
```
sfc-gh-pmansour: Would it be possible to re-use the [no_metadata_service](https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/test/unit/conftest.py#L26-L30) fixture instead of all the custom patches here?

sfc-gh-pmansour: Generally, consider re-using and/or updating the existing mocks for various metadata services (and the [`FakeAwsEnvironment`](https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/test/csp_helpers.py#L246) which mocks the boto library calls) in [csp_helpers.py](https://github.com/snowflakedb/snowflake-connector-python/blob/main/test/csp_helpers.py). You can find usage examples in test_auth_workload_identity.py ([example](https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/test/unit/test_auth_workload_identity.py#L406-L408)).

If those mocks are missing some functionality, consider adding it there. This can make the tests easier to read and maintain, and help future test authors.

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,489 @@
...
+    def test_aws_identity_detection(self, mock_requests, mock_boto3, mock_imds_fetcher):
+        mock_imds_instance = Mock()
+        mock_imds_instance._get_request.side_effect = Exception("No IMDS")
+        mock_imds_fetcher.return_value = mock_imds_instance
+
+        mock_sts_client = Mock()
+        mock_sts_client.get_caller_identity.return_value = {
+            "Arn": "arn:aws:iam::123456789012:user/test-user"
```
sfc-gh-pmansour: Can you parameterize this test to explicitly check for both `iam::..user/` and `sts::...assumed-role/`?

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,489 @@
...
+    @patch("snowflake.connector.platform_detection.IMDSFetcher")
+    @patch("snowflake.connector.platform_detection.boto3")
+    @patch("snowflake.connector.platform_detection.requests")
+    def test_invalid_arn_handling(self, mock_requests, mock_boto3, mock_imds_fetcher):
+        mock_imds_instance = Mock()
+        mock_imds_instance._get_request.side_effect = Exception("No IMDS")
+        mock_imds_fetcher.return_value = mock_imds_instance
+
+        mock_sts_client = Mock()
+        mock_sts_client.get_caller_identity.return_value = {"Arn": "invalid-arn-format"}
```
sfc-gh-pmansour: Suggest parameterizing this test to also try a valid ARN format, but which is unsupported for WIF.

Here's a list you can pick from: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,489 @@
...
+        )
+
+        mock_requests.get.side_effect = requests.RequestException("No metadata")
+        mock_requests.RequestException = requests.RequestException
+        mock_requests.Timeout = requests.Timeout
+
+        result = detect_platforms(timeout=5.0)
+
+        assert "is_aws_lambda" in result
+        assert "is_github_action" in result
```
sfc-gh-pmansour: How does this test validate the 5s timeout? Shouldn't you make the mocked methods have a side effect that waits a certain amount of time (e.g. 10s), then capture timestamps before and after detect_platforms, and ensure that the time spent there is like <6s?

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,489 @@
...
+        self.content = content
+        self.text = text
+        self.headers = headers or {}
+
+    def raise_for_status(self):
+        if self.status_code >= 400:
+            raise requests.HTTPError(f"HTTP {self.status_code} Error")
+
+
+class TestDetectPlatforms:
```
sfc-gh-pmansour, line 30: I wonder if we should add more negative test cases here (e.g. if condition foo doesn't hold, then signal "platform_foo" shouldn't be present). AFAICT, the only negative test cases we add here are a) for invalid or missing ARN, and b) for no platform detected. We should probably increase coverage.

sfc-gh-pmansour (commented):
Thanks Sid!

src/snowflake/connector/platform_detection.py
sfc-gh-pmansour: no longer relevant

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: I was a bit mixed on this. I originally wanted to do a match statement but it is not supported on all the Python versions the driver can run on. I put it here to show more explicitly that how all the states of the enum are being processed. Though, if you think it's clear enough without it, I don't mind removing it. What do you think?

src/snowflake/connector/auth/_auth.py
sfc-gh-sshetkar: The process going from the lowest point of the call stack to the highest point of the call stack looks like:

1. https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/src/snowflake/connector/auth/_auth.py#L170
2. https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/src/snowflake/connector/connection.py#L1670
3. https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/src/snowflake/connector/connection.py#L1638
4. https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/src/snowflake/connector/connection.py#L1318
5. https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/src/snowflake/connector/connection.py#L923
6. https://github.com/snowflakedb/snowflake-connector-python/blob/d89ebeec2c6bf911fff5caa01dbb4d16ae8e720c/src/snowflake/connector/connection.py#L566

The answer for how it gets passed is that the auth object has access to the connection object and passes in the platform_timeout field through the connection object. We see that in step 1 of the call stack listed above.

src/snowflake/connector/platform_detection.py
sfc-gh-pmansour: IMHO it's clear that you're only appending to the list if it's detected or timeout. If you want, you can add a comment too.

src/snowflake/connector/auth/_auth.py
sfc-gh-pmansour: got it, thanks!

src/snowflake/connector/connection.py
sfc-gh-sshetkar: I think float makes sense here since it mirrors how the underlying libraries end up accepting the time. In this case the [requests library ](https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts)and the boto library accept these value types and they are both in units of seconds. I included int as well to be a bit more complete but I think maybe it actually might make more sense to only had float if we had to pick. What do you think?

src/snowflake/connector/platform_detection.py
sfc-gh-mmishchenko: Maybe file-based cache would be an overkill, but putting`@cache` from functools looks a pure gain?
Timeouts here don't give 100% guarantee, some networking ops under the hood don't accept timeouts and may hang for way much longer, for instance, nslookup.

sfc-gh-pmansour: interesting, today I learned that `@functools.cache` was a thing! Seems reasonable to add here

src/snowflake/connector/connection.py
sfc-gh-pmansour: Sounds good.. It's not how I would've designed it (IMHO ints and units are clearer), but fair enough to be compatible with requests and boto.

I'd still suggest allowing a single type rather than a union, and defining the unit.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Added comments

src/snowflake/connector/connection.py
sfc-gh-sshetkar: Yea, I think that makes sense. This also makes it clearer for the end user what value they are setting.

In terms of the union, does this include the None option? I see other timeouts here have None sent in as a representative of "default" behavior and I have seen that type of pattern used in python overall and a lot in this code base for timeouts. I'm not super partial either way.

I think the pro of leaving None are that it lets the detect_platform module encapsulate the idea of what it feels default timeout should be. It also let's the user explicitly opt into default behavior.
The con is that it adds an extra type which is more things to handle and pipe down.

I see the driver has done it both ways before so not sure which way is preferred. @sfc-gh-mmishchenko if the driver team particular about the pattern here passing parameters and how default behavior should be specified?

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: I agree, the data should not change at all and this seems pretty simple and does introduce a lot of extra complexity. Added the functools caching.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Yea, that makes sense. Tried this out and it worked so updated to go against this resource.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Good point. We aren't checking anything special with the text so status code is sufficient.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: I see, if it's clear then I don't see huge benefit to keeping it then. Removed the extra elif statement

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Yea, that's fair. In this case, serialized I/O operations are probably worse than just redoing some work. Updated to try both options and parallelized all the function calls.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Yea, I think that's a good idea. The output here is just for telemetry purposes so shouldn't affect the general driver functionality.

Added.

test/unit/test_connection.py
sfc-gh-sshetkar: I feel like this is tested with the tests in test_connection.py in the integ folder where I think it makes more sense. What do you think?

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Yea that's a good point. The Response class does just work and there's not other aspects I would really want different behavior for here. Switched to using the Response class.

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Was not able to use out of the box since it defaults to connection timeouts which emits a signal. So, I refactored the no_metadata_service to represent unavailable_metadata_service and made a new similar class for broken_metadata_service which threw RequestExceptions. I extended the abstract base class to allow the sub classes to define the behavior when hostnames didn't match to allow for this.

sfc-gh-sshetkar: Refactored to use the existing environments and added new Fake environments and additional methods as needed for platform detection.

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Parameterized the test to check for both cases.

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Parameterized the test to test all the unsupported versions in the list

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Yea, I agree. I added more negative tests to increase coverage for these types of cases.

I'm wondering for a lot of tests, we have default behavior for cases where urls don't match which I think matches the idea that "condition foo" didn't hold because that was checked in the faked classes. Would it make more sense to just add more assertions in those tests that other environment signals don't show up vs adding more negative tests? I added more negative tests for now because I thought it is more explicit and makes the test a little less brittle IMO. What do you think?

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Yea, that's true. I actually don't think this test holds much value since there is already a separate timeout tests that validates we gets those signals. I'm not sure that testing the time spent in a function would give us much value. To me it seems that would be like testing the requests library is doing what it says it will do and then mocking the behavior to validate that. I think what matters more with timeout tests is validating what we do with the exception which is give the correct signals back which is already validated in another test.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Yes, that's correct. I tested this by going on a version that supports IMDSv1 and running the IMDSv2 command (this example given [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/retrieve-iid.html) which is the reference for this code).

Basically the endpoint is the same but IMDSv2 requires a token to be sent. If the token is sent to the IMDSv1 endpoint, then the token is just ignored.

sfc-gh-sshetkar: I actually checked just now, and our workspaces are good examples. They accept IMDSv1 command formats.

sfc-gh-pmansour: awesome, thanks!

src/snowflake/connector/auth/_auth.py
```diff
@@ -177,6 +182,7 @@ def authenticate(
             self._rest._connection.login_timeout,
             self._rest._connection._network_timeout,
             self._rest._connection._socket_timeout,
+            self._rest._connection._platform_detection_timeout_seconds,
```
sfc-gh-pmansour, line 185: hmm, we have 3 different name for this same param: `platform_detection_timeout` (in the `base_auth_data()` param), `timeout_seconds` (in the `detect_platforms()` param) and `platform_detection_timeout_seconds` here. Could we converge? The last one seems the most descriptive.

src/snowflake/connector/connection.py
sfc-gh-pmansour: yes sorry, union with None is fine as a way to make this optional. I meant to avoid a union between int and float. Your new implementation LGTM, thanks!

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+    return (
+        _DetectionState.DETECTED
+        if "LAMBDA_TASK_ROOT" in os.environ
+        else _DetectionState.NOT_DETECTED
+    )
+
+
+def is_valid_arn_for_wif(arn: str) -> bool:
+    """
+    Validate if an AWS ARN is suitable for Web Identity Federation (WIF).
```
sfc-gh-pmansour:
```suggestion
    Validate if an AWS ARN is suitable for use with Snowflake's Workload Identity Federation (WIF).
```

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+    try:
+        config = Config(
+            connect_timeout=timeout_seconds,
+            read_timeout=timeout_seconds,
+            retries={"total_max_attempts": 1},
+        )
+        caller_identity = boto3.client("sts", config=config).get_caller_identity()
+        if not caller_identity or "Arn" not in caller_identity:
+            return _DetectionState.NOT_DETECTED
+        else:
```
sfc-gh-pmansour: nit: unnecessary `else` since the previous branch has a return. Remove it to reduce nesting.

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+    ]
+    return (
+        _DetectionState.DETECTED
+        if all(var in os.environ for var in service_vars)
+        else _DetectionState.NOT_DETECTED
+    )
+
+
+def is_managed_identity_available_on_azure_vm(
+    timeout_seconds, resource="https://management.azure.com/"
```
sfc-gh-pmansour: just triple checking that the format of this resource is exactly correct, including the trailing slash?

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+                        detection timed out, NOT_DETECTED otherwise.
+    """
+    has_azure_function_managed_identity = (
+        _DetectionState.DETECTED
+        if os.environ.get("IDENTITY_HEADER")
+        else _DetectionState.NOT_DETECTED
+    )
+    has_azure_vm_managed_identity = is_managed_identity_available_on_azure_vm(
+        timeout_seconds
+    )
```
sfc-gh-pmansour: can you short-circuit if there's an azure function managed identity before making the IMDS call? That could save on latency and avoid minting an unnecessary token.

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+    if (
+        has_azure_vm_managed_identity == _DetectionState.DETECTED
+        or has_azure_function_managed_identity == _DetectionState.DETECTED
+    ):
+        return _DetectionState.DETECTED
+    if (
+        has_azure_vm_managed_identity == _DetectionState.TIMEOUT
+        or has_azure_function_managed_identity == _DetectionState.TIMEOUT
```
sfc-gh-pmansour: nit: unnecessary / impossible condition, because you only set it to DETECTED / NOT_DETECTED a few lines earlier

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+            if response.headers.get("Metadata-Flavor") == "Google"
+            else _DetectionState.NOT_DETECTED
+        )
+    except requests.Timeout:
+        return _DetectionState.TIMEOUT
+    except requests.RequestException:
+        return _DetectionState.NOT_DETECTED
+
+
+def is_gce_cloud_run_service():
```
sfc-gh-pmansour: nit: GCP Cloud Run. GCE is just the name of GCP's VM platform.
```suggestion
def is_gcp_cloud_run_service():
```

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+    """
+    service_vars = ["K_SERVICE", "K_REVISION", "K_CONFIGURATION"]
+    return (
+        _DetectionState.DETECTED
+        if all(var in os.environ for var in service_vars)
+        else _DetectionState.NOT_DETECTED
+    )
+
+
+def is_gce_cloud_run_job():
```
sfc-gh-pmansour:
```suggestion
def is_gcp_cloud_run_job():
```

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+    Returns:
+        list[str]: List of detected platform names. Platforms that timed out will have
+                  "_timeout" suffix appended to their name. Returns empty list if any
+                  exception occurs during detection.
+    """
+    try:
+        if timeout_seconds is None:
+            timeout_seconds = 0.2
+
+        with ThreadPoolExecutor(max_workers=10) as executor:
```
sfc-gh-pmansour: max_workers is set to 10, but there are 11 tasks in there, meaning at least 1 will block on the others.

Consider taking the checks that don't make network calls out of there (e.g. `is_aws_lambda, is_azure_function, is_gcp_cloud_run_service, is_gcp_cloud_run_job, is_github_action`). This will also use less resources on synchronization.

test/unit/conftest.py
```diff
@@ -24,9 +28,16 @@ def disable_oob_telemetry():
 
 
 @pytest.fixture
-def no_metadata_service():
+def unavailable_metadata_service():
+    """Emulates an environment where all metadata services unavailable."""
```
sfc-gh-pmansour:
```suggestion
    """Emulates an environment where all metadata services are unavailable."""
```

test/unit/conftest.py
```diff
@@ -24,9 +28,16 @@ def disable_oob_telemetry():
...
+    with UnavailableMetadataService() as server:
+        yield server
+
+
+@pytest.fixture
+def broken_metadata_service():
```
sfc-gh-pmansour: What's the difference between this and the one above?

test/unit/conftest.py
```diff
@@ -47,8 +58,43 @@ def fake_azure_metadata_service(request):
...
+    """Emulates the GCE Cloud Job metadata service."""
+    with FakeGceCloudRunJobService() as server:
+        yield server
+
+
+@pytest.fixture
+def fake_github_actions_metadata_service():
+    """Emulates the GitHub Actions metadata service."""
+    with FakeGitHubActionsService() as server:
+        yield server
```
sfc-gh-pmansour, line 100: Is this actually a metadata service?

test/csp_helpers.py
```diff
@@ -63,28 +69,39 @@ def reset_defaults(self):
...
         pass
 
-    @abstractmethod
     def handle_request(self, method, parsed_url, headers, timeout):
-        """Main business logic for handling this request. Should return a Response object."""
-        pass
+        return ConnectTimeout()
+
+    def handle_unexpected_hostname(self):
+        return ConnectTimeout()
```
sfc-gh-pmansour: nit: this method is named `handle_*` but it doesn't handle or throw, it just creates an exception. Suggest throwing the exception directly in there.

test/csp_helpers.py
```diff
@@ -116,21 +142,39 @@ def __exit__(self, *args, **kwargs):
             patcher.__exit__(*args, **kwargs)
 
 
-class NoMetadataService(FakeMetadataService):
-    """Emulates an environment without any metadata service."""
+class UnavailableMetadataService(FakeMetadataService):
+    """Emulates an environment where all metadata services unavailable."""
```
sfc-gh-pmansour:
```suggestion
    """Emulates an environment where all metadata services are unavailable."""
```

test/csp_helpers.py
```diff
@@ -116,21 +142,39 @@ def __exit__(self, *args, **kwargs):
...
-        return None  # Always raise a ConnectTimeout.
+    def expected_hostnames(self):
+        return []  # Always raise a ConnectTimeout.
 
     def handle_request(self, method, parsed_url, headers, timeout):
         # This should never be called because we always raise a ConnectTimeout.
         pass
 
 
+class BrokenMetadataService(FakeMetadataService):
```
sfc-gh-pmansour: Instead of creating a dedicated fake service for this, consider adding a field to the base metadata service class that controls which exception gets thrown, so that individual tests can set it. Each of these metadata services corresponds to a real environment that we're emulating, whereas this `BrokenMetadataService` just seems like a variation that is better served by making the others more configurable.

That way, a test can continue to depend on `no_metadata_service` or `azure_vm_metadata_service`, but just sometimes make it raise a `RequestException` and other times make it raise a `ConnectTimeout`.

test/csp_helpers.py
```diff
@@ -220,31 +273,104 @@ def reset_defaults(self):
...
+        self.k_service = "test-service"
+        self.k_revision = "test-revision"
+        self.k_configuration = "test-configuration"
+
+    def get_environment_variables(self) -> dict[str, str]:
+        return {
+            "K_SERVICE": self.k_service,
+            "K_REVISION": self.k_revision,
+            "K_CONFIGURATION": self.k_configuration,
+        }
```
sfc-gh-pmansour: I think we can make this a simple context manager class rather than inheriting from FakeMetadataService and not supporting most of its intended functionality.

Actually, even better: we could try to match reality by inheriting from `FakeGceMetadataService`, since Cloud Run actually runs a real metadata service with a very similar signature to the GCE one. For the purpose of the Python driver, the only thing it adds is the environment variables, but it doesn't take away the IMDS.

(and same for the CloudRunJobService)

test/csp_helpers.py
```diff
@@ -220,31 +273,104 @@ def reset_defaults(self):
...
+        return []
+
+    def handle_request(self, method, parsed_url, headers, timeout):
+        pass
+
+    def reset_defaults(self):
+        self.github_actions = "github-actions"
+
+    def get_environment_variables(self) -> dict[str, str]:
+        return {"GITHUB_ACTIONS": self.github_actions}
```
sfc-gh-pmansour: Consider making this a basic context manager class rather than an instance of `FakeMetadataService`

test/csp_helpers.py
```diff
@@ -274,7 +406,20 @@ def sign_request(self, request: AWSRequest):
...
+    def fetcher_fetch_metadata_token(self):
+        return self.metadata_token
+
+    def boto3_client(self, *args, **kwargs):
+        mock_client = mock.Mock()
+        mock_client.get_caller_identity.return_value = self.caller_identity
+        return mock_client
+
     def __enter__(self):
+        os.environ["LAMBDA_TASK_ROOT"] = self.lambda_task_root
```
sfc-gh-pmansour: do we want this to be set by default? why not create a separate `FakeAwsLambdaEnvironment` that inherits from this and sets those env variables?

Also, avoid directly setting `os.environ["foo"]` since that doesn't get reverted afterwards. Instead, you can use `self.patchers.append(patch.dict(os.environ, {"LAMBDA_TASK_ROOT": self.lambda_task_root}))`.

(I realize you currently pop it during __exit__ but that seems less idiomatic, given the other patterns here)

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,193 @@
...
+        assert "has_aws_identity" in result
+        assert "is_github_action" in result
+        assert "is_gce_cloud_run_service" in result
+
+    def test_timeout_handling(self, unavailable_metadata_service):
+        result = detect_platforms(timeout_seconds=None)
+        assert "is_azure_vm_timeout" in result
+        assert "is_gce_vm_timeout" in result
+        assert "has_gcp_identity_timeout" in result
+        assert "azure_managed_identity_timeout" in result
```
sfc-gh-pmansour: what does this actually test?
1. It sets `timeout_seconds=None`, which will just use the default anyway
2. It doesn't actually measure any times it took to run

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def test_missing_arn_handling(self, broken_metadata_service, fake_aws_environment):
+        fake_aws_environment.caller_identity = {"UserId": "test-user"}
+        result = detect_platforms(timeout_seconds=None)
+        assert "has_aws_identity" not in result
+
+    def test_azure_managed_identity_wrong_issuer(self, fake_azure_vm_metadata_service):
+        fake_azure_vm_metadata_service.iss = "https://fake-issuer.com"
+        result = detect_platforms(timeout_seconds=None)
+        assert "azure_managed_identity" not in result
```
sfc-gh-pmansour: can you help me understand this test? IIUC, the detection logic doesn't consider the issuer at all. Why would using a fake issuer result in missing this signal?

src/snowflake/connector/platform_detection.py
```diff
@@ -0,0 +1,409 @@
...
+            futures = {
+                "is_ec2_instance": executor.submit(is_ec2_instance, timeout_seconds),
+                "is_aws_lambda": executor.submit(is_aws_lambda),
+                "has_aws_identity": executor.submit(has_aws_identity, timeout_seconds),
+                "is_azure_vm": executor.submit(is_azure_vm, timeout_seconds),
+                "is_azure_function": executor.submit(is_azure_function),
+                "azure_managed_identity": executor.submit(
```
sfc-gh-pmansour:
```suggestion
                "has_azure_managed_identity": executor.submit(
```

test/unit/test_detect_platforms.py
```diff
@@ -0,0 +1,193 @@
...
+@pytest.mark.xdist_group(name="serial_tests")
+class TestDetectPlatforms:
+    @pytest.fixture(autouse=True)
+    def teardown(self):
+        with patch.dict(os.environ, clear=True):
+            detect_platforms.cache_clear()  # clear cache before each test
+            yield
+            detect_platforms.cache_clear()  # clear cache after each test
+
+    def test_no_platforms_detected(self, broken_metadata_service):
```
sfc-gh-pmansour: curious why you use the `broken_metadata_service` here and elsewhere rather than `no_metadata_service`? technically it's more likely for there to not be any metadata service (aka raise `ConnectTimeout`s).

test/unit/test_detect_platforms.py
sfc-gh-pmansour: I do think there's value in having a test that ensures the timeout is respected and the entire method finishes within that (+ some minor error margin).

Currently, if you have a bug that does everything in serial (rather than parallel), or that forgets to pass the timeout, or that passes the wrong timeout variable, you have no way to detect it. In fact, I believe you previously had such a bug (with the azure managed identity), and may currently have another one (with the 10 executors rather than 11), but there are no tests to catch it.

test/unit/test_detect_platforms.py
sfc-gh-pmansour: Generally it's good to [keep tests focused](https://testing.googleblog.com/2018/06/testing-on-toilet-keep-tests-focused.html) and [test behaviors, not methods](https://testing.googleblog.com/2014/04/testing-on-toilet-test-behaviors-not.html), with each test verifying a specific behavior. Otherwise, if you introduce a bug, you'll have to wade through 100 failing tests and try to figure out what went wrong, rather than just 1.

Anyway, I think the negative test coverage you have here is probably sufficient for now. Thanks!

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: yep, double checked this. It works both ways with and without the trailing slash. Removing the trailing slash if that makes it look more correct on first glance.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: I'm a bit confused as to how this is an impossible condition. has_azure_vm_managed_identity can be _DetectionState.TIMEOUT if the minting token request times out. As of now, has_azure_function_managed_identity cannot be _DetectionState.TIMEOUT but I figured this would be general logic if I tested for both. So, if that was possible and we changed the behavior, I think we'd want to do this check here. I am fine with removing the has_azure_function_managed_identity == _DetectionState.TIMEOUT since it's not possible now but I figured keeping the check would avoid unintentional bugs if the behavior of the has_azure_function_managed_identity changed.

sfc-gh-pmansour: I was referring to `has_azure_function_managed_identity` specifically. Just seems like a lot of boilerplate for impossible situations that are clear.

If anything, I'd actually just make `has_azure_function_managed_identity` into a bool so there's no confusion.

test/unit/conftest.py
sfc-gh-sshetkar: This one returns RequestExceptions instead of ConnectTimeout. This makes a difference since platform detection code handles connection timeouts separately and I needed a way to test the scenario where no platforms are detected. These scenarios don't send back timeouts but will send back request exceptions.

test/unit/conftest.py
sfc-gh-sshetkar: I guess it depends on how you look at it. The only thing really needed emulate this for now is the environment variables. However, there are aspects which you might want to emulate like how it interacts with [workflows](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28). These aren't really relevant for WIF though, so I would also be ok with just patching in environment variables since that's all that is really needed. I could see the same argument for GCP Cloud Run Jobs/Services too. Do you think setting up these as metadata_services are overkill for these purposes?

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Yea, that's a fair point. I'm also considering the check for if it is an azure function is just an environment variable read. I think we can just separate it out. It will be redo-ing some work. But I think it'll avoid the cases where we call azure_vm code when it is an azure function but doesn't have a managed identity set.

Something like:
```
# short circuit early to save on latency and avoid minting an unnecessary token
if is_azure_function() == _DetectionState.DETECTED:
    return _DetectionState.DETECTED if is_managed_identity_available_on_azure_function() else _DetectionState.NOT_DETECTED
return is_managed_identity_available_on_azure_vm(platform_detection_timeout_seconds)
```

src/snowflake/connector/auth/_auth.py
sfc-gh-sshetkar: good point, just fixed it so they are all platform_detection_timeout_seconds

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: yea, that's a good point. Separated out the ones that don't require network calls. There are only 6 checks that require network calls so kept those and updated max_workers to 6.

test/unit/conftest.py
sfc-gh-sshetkar: Fixed

test/csp_helpers.py
sfc-gh-sshetkar: Fixed

test/csp_helpers.py
sfc-gh-sshetkar: Fixed

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: In this case, it tests that we collect the timeout signals if a Timeout exception is thrown. I think this could be removed if we include a test which actually mocks the requests calls timing out with a sleep()

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: I think it depends on the setup. I was basing more off how it ran on my dev laptop, cloud workspace, and other VMs (like Azure VM) where I saw that we got NOT_DETECTED instead of TIMEOUTS.

This is because it is more likely to not be able to resolve the name or establish a connection at all rather than timeout.

Here are the exceptions I got on my dev laptop as an example:

Azure VM: HTTPConnectionPool(host='169.254.169.254', port=80): Max retries exceeded with url: /metadata/instance?api-version=2021-02-01 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x103c2acf0>: Failed to establish a new connection: [Errno 65] No route to host'))

EC2: HTTPConnectionPool(host='169.254.169.254', port=80): Max retries exceeded with url: /latest/api/token (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x103acecf0>, 'Connection to 169.254.169.254 timed out. (connect timeout=2)'))

GCE VM: HTTPConnectionPool(host='metadata.google.internal', port=80): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPConnection object at 0x10442b620>: Failed to resolve 'metadata.google.internal' ([Errno 8] nodename nor servname provided, or not known)"))

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

test/csp_helpers.py
sfc-gh-sshetkar: This makes sense to me. Can replicate this behavior for specific tests by just setting the field and I can use a fixture to make it resuable within the test file context.

test/csp_helpers.py
sfc-gh-sshetkar: Yea, that makes sense to me. Updated GCE ones to inherit from `FakeGceMetadataService`.

test/csp_helpers.py
sfc-gh-sshetkar: Updated to be a basic context manager class and just patch the environment variables.

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: Fixed

test/csp_helpers.py
sfc-gh-sshetkar: Agreed, separated out into a new class that inherits. And yea, I forgot to change this one back to mocking instead of manually changing the os.environ.

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Oh this is an older test that was made with cursor that I forgot to remove. It was working since the wrong issuer would cause an exception to be thrown out in the fake code. However, I updated it to just add a boolean flag so that we could enable if the token endpoint code to generate a token would be used or not.

test/unit/test_detect_platforms.py
sfc-gh-sshetkar: Added tests to actually test timeouts across the whole function to ensure it is running I/O operations in parallel.

src/snowflake/connector/platform_detection.py
sfc-gh-pmansour: LGTM

test/unit/conftest.py
sfc-gh-pmansour: I like how you did it in the end. For GH actions we don't know or depend on any of its "service" parts, so you're just doing a context manager. For GCP, we know that it has a full IMDS (and we already have a fully implemented fake one), so you're subclassing it. Looks great!

sfc-gh-pmansour (approved):
LGTM, thanks for all your patience!

Would be good to get @sfc-gh-mmishchenko's LGTM too

PR #2299, opened 2025-04-27
SNOW-2067577 OCSP: stop certificates chain traversal as soon as a trusted one met

sfc-gh-mmishchenko:
- [x] Stop OCSP traversal earlier on a trusted certificate
- [x] Fix standalone OCSP tool

PR #2272, opened 2025-04-16
Snow 1763673 numpy bump-up and windows builds

sfc-gh-fpawlowski:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #NNNN

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

4. (Optional) PR for stored-proc connector:

PR #2239, opened 2025-03-27
Snow-1763673 python3.13 support

sfc-gh-fpawlowski:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.
   Fixes SNOW-1763673

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   These new versions usually need no code-changes. They should just work on new Python versions. I also took this opportunity to update pre-commit hooks and fix all files.
4. (Optional) PR for stored-proc connector:

PR #2389, opened 2025-07-04
SNOW-2187871: Cherry picks to `dev/aio-connector`, part 0 - fix tests

sfc-gh-pczajka:
## What to review?
Most of lines in this change are cherry-picks of commits already on main, majority of which can be skipped.
Below I listed all `aio` changes which reflect changes in sync code. To make the review easier, I arranged the commits in a way that the changes introduced ny hand follow cherry-pick they reflect

### Cherry picks:
 * [x] #2092
   * **changes**: add test in `test/integ/aio/test_bindings_async.py`
 * [x] #2272 
   * **changes**: None
 * [x] #2193 
   * **changes**: None
 * [x] #2107 
   * **changes**: None, changes inherited
 * [x] #2169 
   * **changes**:
     * Apply `.lowercase()` fix to `aio/_connections.py`
     * Add test `test_uppercase_privatelink_ocsp_url_creation` to `test_connection_aio.py`
 * [x] #2165 
   * **changes**:
     * Apply changes to async code
     * fix sync tests to work on non-public environment (#2397)
     * Add async version of the tests
 * [x] #2209
   * **changes**: Apply changes to async tests
 * [x] #2276 
   * **changes**: Apply changes to async tests
 * [x] #2299 
   * **changes**: Apply changes to async code

### Manual changes (review all):
 * fix gihub workflow - change image from `windows-2019` to `windows-latest` (later done in #2239)
 
_Jenkins tests are expected to fail, as there is still misconfiguration between this branch and Jenkins (to be cherry-picked in future)_

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -7,47 +7,6 @@ https://docs.snowflake.com/
 Source code is also available at: https://github.com/snowflakedb/snowflake-connector-python
 
 # Release Notes
-- v3.13.3(TBD)
-- v3.15(TBD)
-  - Bumped up min boto and botocore version to 1.24
-  - OCSP: terminate certificates chain traversal if a trusted certificate already reached
-
-- v3.14.1(April 21, 2025)
-  - Added support for Python 3.13.
-    - NOTE: Windows 64 support is still experimental and should not yet be used for production environments.
-  - Dropped support for Python 3.8.
-  - Added basic decimal floating-point type support.
-  - Added experimental authentication methods.
-  - Added support of GCS regional endpoints.
-  - Added support of GCS virtual urls. See more: https://cloud.google.com/storage/docs/request-endpoints#xml-api
-  - Added `client_fetch_threads` experimental parameter to better utilize threads for fetching query results.
-  - Added `check_arrow_conversion_error_on_every_column` connection property that can be set to `False` to restore previous behaviour in which driver will ignore errors until it occurs in the last column. This flag's purpose is to unblock workflows that may be impacted by the bugfix and will be removed in later releases.
-  - Lowered log levels from info to debug for some of the messages to make the output easier to follow.
-  - Allowed the connector to inherit a UUID4 generated upstream, provided in statement parameters (field: `requestId`), rather than automatically generate a UUID4 to use for the HTTP Request ID.
-  - Improved logging in urllib3, boto3, botocore - assured data masking even after migration to the external owned library in the future.
-  - Improved error message for client-side query cancellations due to timeouts.
-  - Improved security and robustness for the temporary credentials cache storage.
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/aio/_connection.py
```diff
@@ -496,6 +497,26 @@ def _init_connection_parameters(
             elif "streamlit" in sys.modules:
                 connection_init_kwargs["application"] = "streamlit"
 
+        if "insecure_mode" in connection_init_kwargs:
```
sfc-gh-fpawlowski, line 500: Do you think it would be possible to have a separate connection_state class / property - so we do not have to maintain 2 almost identical versions of code (apart from the execution interface)?

Async and Sync classes will share most of their non-I/O code. If we refactor it a bit, we could simply import such 'composites' to both of them, and build up the state using data retrieved either synchronously or asynchronously.

It seems like quite some work, so the question is is it worth the time it can save us in the future?

sfc-gh-pczajka: I think it's worth it - but definitely after merging this code into `main`

sfc-gh-fpawlowski: Sure, here is a follow-up task not to loose it:
https://snowflakecomputing.atlassian.net/browse/SNOW-2203682

src/snowflake/connector/ocsp_asn1crypto.py
```diff
@@ -382,15 +398,22 @@ def extract_certificate_chain(
         from OpenSSL.crypto import FILETYPE_ASN1, dump_certificate
 
         cert_map = OrderedDict()
-        logger.debug("# of certificates: %s", len(connection.get_peer_cert_chain()))
-
-        for cert_openssl in connection.get_peer_cert_chain():
+        cert_chain = connection.get_peer_cert_chain()
+        logger.debug("# of certificates: %s", len(cert_chain))
+        self._lazy_read_ca_bundle()
```
sfc-gh-fpawlowski, line 403: In a few places we use this method (_lazy_read_ca_bundle), which was not reimplemented in a asynchronous way in the original code.
This may be a nit, since our code will work without it and it is used only once (hence the "_lazy_" prefix). But it is a place where we will lock on I/O operation.

sfc-gh-fpawlowski (approved):
Lgtm
There is only non-blocking open question regarding file I/O operations async implementation

src/snowflake/connector/ocsp_asn1crypto.py
sfc-gh-pczajka: I've created a ticket to fix that https://snowflakecomputing.atlassian.net/browse/SNOW-2203709

PR #2138, opened 2025-01-17
SNOW-1882588 Add Test for PR 2136

sfc-gh-bli:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

SNOW-1882588 Add Test for PR https://github.com/snowflakedb/snowflake-connector-python/pull/2136

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

4. (Optional) PR for stored-proc connector:

PR #2134, opened 2025-01-13
Support base64-encoded DER private keys

sfc-gh-eworoshow:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #2133

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [x] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Support base64-encoded DER private keys. We currently support raw DER `bytes`, and now we will also support base64-encoded DER strings. This functionality is useful for passing private keys through environment variables and other configuration mechanisms.

4. (Optional) PR for stored-proc connector:

PR #2404, opened 2025-07-14
Cherrypicks to aio connector part2

sfc-gh-pczajka:
## What to review?
Most of lines in this change are cherry-picks of commits already on main, majority of which can be skipped.
Below I listed all `aio` changes which reflect changes in sync code. To make the review easier, I arranged the commits in a way that the changes introduced ny hand follow cherry-pick they reflect

### Cherry picks:
 * [x] #2106 
   * **changes**: Apply changes to async code
 * [x] #2110 
   * **changes**: None, version bump
 * [x] #2212  
   * **changes**:
      * Fix errors, add tests (#2403)
      * Apply changes to async code
 * [x] #2134 
   * **changes**:
       * Fix tests (#2176)
       * Apply async changes to the code
  * [x] #2136 
   * **changes**:
       * Add tests (#2138)
       * Apply async changes to the code
 

### Tests
 * Applying #2122 was additionally tested with the following script:
 ```
 from snowflake.connector import connect
from snowflake.connector.aio import connect as aconnect

import asyncio

PAT_params = {...}
PAT_params["authenticator"] = "PROGRAMMATIC_ACCESS_TOKEN"
PAT_params["token"] = ...

async def test_pat():
    with connect(**PAT_params) as conn:
        with conn.cursor() as cur:
            cur.execute("SELECT 1")
            print(cur.fetchone())
            print("SYNC - OK")

    connection = await aconnect(**PAT_params)
    cur = connection.cursor()
    await cur.execute("SELECT 1")
    print(await cur.fetchone())
    print("ASYNC - OK")
    await connection.close()

asyncio.run(test_pat())
```

### DIFF
--- a/src/snowflake/connector/aio/_cursor.py
+++ b/src/snowflake/connector/aio/_cursor.py
@@ -532,10 +532,14 @@ class SnowflakeCursor(SnowflakeCursorSync):
 
         if _do_reset:
             self.reset()
-        command = command.strip(" \t\n\r") if command else None
+        command = command.strip(" \t\n\r") if command else ""
         if not command:
-            logger.warning("execute: no query is given to execute")
-            return None
+            if _dataframe_ast:
+                logger.debug("dataframe ast: [%s]", _dataframe_ast)
+            else:
+                logger.warning("execute: no query is given to execute")
+                return None
+
         logger.debug("query: [%s]", self._format_query_for_log(command))
 
         _statement_params = _statement_params or dict()
--- a/test/unit/aio/test_cursor_async_unit.py
+++ b/test/unit/aio/test_cursor_async_unit.py
@@ -62,6 +62,21 @@ def test_cursor_attribute():
     assert cursor.lastrowid is None
 
 
+async def test_query_can_be_empty_with_dataframe_ast():
+    def mock_is_closed(*args, **kwargs):
+        return False
+
+    fake_conn = FakeConnection()
+    fake_conn.is_closed = mock_is_closed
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/snowflake/connector/aio/auth/_pat.py [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+from ._by_plugin import AuthByPlugin as AuthByPluginAsync
+
+
+class AuthByPAT(AuthByPluginAsync, AuthByPATSync):
+    def __init__(self, pat_token: str, **kwargs) -> None:
+        """Initializes an instance with a PAT Token."""
+        AuthByPATSync.__init__(self, pat_token, **kwargs)
+
+    async def reset_secrets(self) -> None:
+        AuthByPATSync.reset_secrets(self)
```
sfc-gh-fpawlowski, line 20: NIT: It may be a design to rethink in the future - that we have all those methods as async.
This is kind of enforced by the 'template-method'-like design in the connection class. However, due to this, we end up calling code like reset_secrets - which probably won't be async for any of the Auth subclasses - in the async coroutine.

sfc-gh-fpawlowski (approved):
Left one NIT, apart from that all looks good 👍

src/snowflake/connector/aio/auth/_pat.py
sfc-gh-pczajka: I've created a jira ticket not to forget about it: https://snowflakecomputing.atlassian.net/browse/SNOW-2216812

PR #2160, opened 2025-01-30
SNOW-1891253 bump pyopenssl upper boundary to <26.0.0 from <25.0.0

sfc-gh-dszmolka:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes SNOW-1891253

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   From the linked issue https://github.com/snowflakedb/snowflake-connector-python/issues/2140 it looks like we're now incompatible with other tools, which require pyopenssl 25.0.0. Checking the CHANGELOG over there, there's no claim for backward-incompatible changes, so bumping the requirement upper boundary. 

4. (Optional) PR for stored-proc connector:

PR #2112, opened 2024-12-03
delete workarround for pyton2.7

Polandia94:
This workarround is only for unsoportted python 2.7

Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #NNNN

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Delete the old workarround. as documented on https://bugs.python.org/issue7980 is only for python2.7

4. (Optional) PR for stored-proc connector:

PR #2408, opened 2025-07-15
Cherrypicks to aio connector part3

sfc-gh-pczajka:
Cherrypicks not requiring changes in the async code:
| PR    | comment |
| -------- | ------- |
| #2151  | workflow update |
| #2160  | dependency update |
| #2112  | No change needed for async code |
| #2120  | Changed file (options.py) is shared by async code |
| #2179  | update workflow |
| #2173  | Changed file (condig_manager.py) is shared by async code |

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -23,7 +23,6 @@ from functools import partial
 from io import StringIO
 from logging import getLogger
 from threading import Lock
-from time import strptime
 from types import TracebackType
 from typing import Any, Callable, Generator, Iterable, Iterator, NamedTuple, Sequence
 from uuid import UUID
@@ -307,9 +306,6 @@ APPLICATION_RE = re.compile(r"[\w\d_]+")
 # adding the exception class to Connection class
 for m in [method for method in dir(errors) if callable(getattr(errors, method))]:
     setattr(sys.modules[__name__], m, getattr(errors, m))
-
-# Workaround for https://bugs.python.org/issue7980
-strptime("20150102030405", "%Y%m%d%H%M%S")
 
 logger = getLogger(__name__)
 
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -95,6 +95,9 @@ jobs:
         if: ${{ matrix.os.id == 'manylinux_aarch64' }}
         uses: docker/setup-qemu-action@v2
         with:
+          # xref https://github.com/docker/setup-qemu-action/issues/188
+          # xref https://github.com/tonistiigi/binfmt/issues/215
+          image: tonistiigi/binfmt:qemu-v8.1.5
           platforms: all
       - uses: actions/checkout@v4
       - name: Building wheel
--- a/.github/workflows/snyk-issue.yml
+++ b/.github/workflows/snyk-issue.yml
@@ -15,19 +15,19 @@ jobs:
   snyk:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout Action
-      uses: actions/checkout@v3
+    - name: checkout action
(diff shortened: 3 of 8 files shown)


src/snowflake/connector/aio/auth/_auth.py
```diff
@@ -76,6 +77,10 @@ async def authenticate(
             )
         logger.debug("authenticate")
 
+        # For no-auth connection, authentication is no-op, and we can return early here.
+        if isinstance(auth_instance, AuthNoAuth):
```
sfc-gh-fpawlowski, line 81: We may need to discuss our naming convention. I see the benefits of using the same names (as @sfc-gh-pczajka mentioned - they are used in big libraries like botocore etc). 
However, in our use case, where we mix inherited synchronous code with asynchronous one, it is easy to get lost.
We may need another task for those readability concerns to discuss and make a decision (I would suggest making a subtask of the design doc task - it should be mentioned in it anyway).

src/snowflake/connector/aio/_connection.py
```diff
@@ -318,6 +320,8 @@ async def __open_connection(self):
                     timeout=self.login_timeout,
                     backoff_generator=self._backoff_generator,
                 )
+            elif self._authenticator == NO_AUTH_AUTHENTICATOR:
```
sfc-gh-fpawlowski: I can't find the same line in the synch code, but I would expect it to be there..
Am I missing smth?

src/snowflake/connector/connection.py
```diff
@@ -300,6 +301,14 @@ def _get_private_bytes_from_file(
         False,
         bool,
     ),  # disable saml url check in okta authentication
+    "iobound_tpe_limit": (
```
sfc-gh-fpawlowski, line 304: We have asyncio Jira for considering reimplementing large datasets downloads using ThreadPoolExxecutor as well for async code. We may want to link this issue (SNOW-1817982) as to be done in its scope - because I feel like we may forget about some parts of it - like adding tests for iobound_tpe_limit (apart from implementing it in async manner).

test/integ/test_put_get.py
```diff
@@ -827,3 +876,21 @@ def test_put_md5(tmp_path, conn_cnx):
                     cur.execute(f"LS @{stage_name}").fetchall(),
                 )
             )
+
+
+@pytest.mark.skipolddriver
+def test_iobound_limit(tmp_path, conn_cnx, caplog):
```
sfc-gh-fpawlowski, line 882: [Same as few lines higher]
We have asyncio Jira for considering reimplementing large datasets downloads using ThreadPoolExxecutor as well for async code. We may want to link this issue (SNOW-1817982) as to be done in its scope - because I feel like we may forget about some parts of it - like adding tests for iobound_tpe_limit (apart from implementing it in async manner).

src/snowflake/connector/aio/_connection.py
sfc-gh-pczajka: No, I've overlooked that this should not be added; removed

src/snowflake/connector/connection.py
sfc-gh-pczajka: Linked: https://snowflakecomputing.atlassian.net/browse/SNOW-2203740

test/integ/test_put_get.py
sfc-gh-pczajka: done

src/snowflake/connector/aio/auth/_auth.py
sfc-gh-pczajka: Added in the description: https://snowflakecomputing.atlassian.net/browse/SNOW-1760278

PR #2413, opened 2025-07-17
Move unsafe_file_write parameter to DEFAULT_CONFIGURATION

sfc-gh-pczajka:

   Fixes SNOW-2207863

   Move `unsafe_file_write` parameter to default configuration. Otherwise accessing this connection property before calling `connect()` might've result in an error


src/snowflake/connector/connection.py
```diff
@@ -1410,11 +1414,6 @@ def __config(self, **kwargs):
             if "host" not in kwargs:
                 self._host = construct_hostname(kwargs.get("region"), self._account)
 
-        if "unsafe_file_write" in kwargs:
```
sfc-gh-fpawlowski, line 1413: Can we add a quick test that by default our driver behaves in the way that unsafe_file_write = False describes.

sfc-gh-pczajka: It's already there :)
https://github.com/snowflakedb/snowflake-connector-python/blob/main/test/integ/test_put_get.py#L747
and
https://github.com/snowflakedb/snowflake-connector-python/blob/main/test/integ/test_put_get.py#L773

sfc-gh-fpawlowski: Great, then we are good 👍 

I found the second test before, but I think we set there explicitly [the value to True](https://github.com/snowflakedb/snowflake-connector-python/blob/c422eb3769f6f4781981f7f687b24cced8eec771/test/integ/test_put_get.py#L778C9-L778C37) so it wasn't preventing changes in the default behaviour in the future.

Nevertheless, the first test You attached is what I needed indeed.

PR #2425, opened 2025-07-23
Cherrypicks to aio connector part4

sfc-gh-pczajka:
In progress

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -46,9 +46,6 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Added a feature to verify if the connection is still good enough to send queries over.
   - Added support for base64-encoded DER private key strings in the `private_key` authentication type.
 
-- v3.12.5(TBD)
-  - Added a feature to limit the sizes of IO-bound ThreadPoolExecutors during PUT and GET commands.
-
 - v3.12.4(December 3,2024)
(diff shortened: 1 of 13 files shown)


sfc-gh-fpawlowski:
Just quick heads-up - there seems to be some conflict between the branches.

test/integ/sso/test_unit_mfa_cache.py
```diff
@@ -20,19 +20,10 @@
...
-try:
-    from snowflake.connector.auth import delete_temporary_credential
-except ImportError:
-    delete_temporary_credential = None
-
-MFA_TOKEN = "MFATOKEN"
 
 
 # Although this is an unit test, we put it under test/integ/sso, since it needs keyring package installed
-@pytest.mark.skipif(
```
sfc-gh-fpawlowski, line 32: That's a great catch. Those tests weren't probably running since that refactor at all..

Only it would be great to move this commit into separate PR into the main branch, to minimise the final difference between the main branch and the dev/aio-connector (apart from the aio module).

sfc-gh-fpawlowski: Apart from this all looks good 👍

sfc-gh-pczajka: This turned out to cause cascade of issues, I've created a separate ticket for that: https://snowflakecomputing.atlassian.net/browse/SNOW-2236300

PR #2441, opened 2025-07-29
Cherrypicks to aio connector part5

sfc-gh-pczajka:
* [x] #2183 
  * changes:
      * Add test in async code
      * Fix case when PAT is passed as "password" in async connection
* [x] #2215 
  * changes: _No changes_


sfc-gh-fpawlowski:
Quick note - it seems like something is wrong with the implementation - we get a lot of:
Error:     raise RuntimeError('Event loop is closed')
cc. @sfc-gh-pczajka

sfc-gh-fpawlowski (approved):
LGTM

PR #2442, opened 2025-07-29
WIF - cherrypick and async implementation

sfc-gh-pczajka:
This change consist of:
 * cherry pick of #2203 
 * async implementation of the same change

Async code should be thoroughly reviewed.


src/snowflake/connector/aio/_wif_util.py
```diff
@@ -0,0 +1,304 @@
...
+    from botocore.utils import InstanceMetadataRegionFetcher
+except ImportError:
+    aioboto3 = None
+    SigV4Auth = None
+    AWSRequest = None
+    InstanceMetadataRegionFetcher = None
+
+from ..errorcode import ER_WIF_CREDENTIALS_NOT_FOUND
+from ..errors import ProgrammingError
+from ..wif_util import (
```
sfc-gh-pczajka, line 21: This code currently contains some code duplication with wif_util.py - this is because part of the code will be modified by further commits, yet to be cherry-picked
Follow-up ticket - https://snowflakecomputing.atlassian.net/browse/SNOW-2234002

src/snowflake/connector/aio/auth/_workload_identity.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import Any
+
+from ...auth.workload_identity import (
+    AuthByWorkloadIdentity as AuthByWorkloadIdentitySync,
+)
+from .._wif_util import AttestationProvider, create_attestation
+from ._by_plugin import AuthByPlugin as AuthByPluginAsync
+
+
+class AuthByWorkloadIdentity(AuthByWorkloadIdentitySync, AuthByPluginAsync):
```
sfc-gh-fpawlowski: Nit: we have triangle inheritance here. Iirc there shouldn't be any problem in Python with it, right?

src/snowflake/connector/aio/_wif_util.py
```diff
@@ -0,0 +1,304 @@
...
+                response._content = content
+                return response
+    except (aiohttp.ClientError, asyncio.TimeoutError):
+        return None
+
+
+async def get_aws_region() -> str | None:
+    """Get the current AWS workload's region, if any."""
+    # Use sync implementation which has proper mocking support
+    from ..wif_util import get_aws_region as sync_get_aws_region
```
sfc-gh-fpawlowski: Should we add TODO with JIRA tikcet to implement this metadatafetcher in async manner and add mocks for the async one in tests as well, or are the problems deeper / won't be possible to fix that?

src/snowflake/connector/aio/_wif_util.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    if aioboto3 is None:
+        logger.debug("aioboto3 not available, falling back to sync implementation")
+        from ..wif_util import create_aws_attestation as sync_create_aws_attestation
+
+        return sync_create_aws_attestation()
+
+    try:
+        # Get credentials using aioboto3
+        session = aioboto3.Session()
```
sfc-gh-fpawlowski: Nit: This might be a low hanging needed for future changes in wif anyway. We could just refactor the method _create_aws_attestation(session) -> we maintain one code and only pass the session as an argument - changing from sync to async - in top level create_aws_attestation. Also we would be able to extend that session with custom logic in the future, which may be needed for Kerberos ticket

src/snowflake/connector/aio/_wif_util.py
```diff
@@ -0,0 +1,304 @@
...
+        logger.debug("Failed to create AWS attestation", exc_info=True)
+        return None
+
+
+async def create_gcp_attestation() -> WorkloadIdentityAttestation | None:
+    """Tries to create a workload identity attestation for GCP.
+
+    If the application isn't running on GCP or no credentials were found, returns None.
+    """
+    res = await try_metadata_service_call(
```
sfc-gh-fpawlowski, line 123: Similar as above - but I am not fully convinced if we should be passing callable here, so I feel free to discuss with it (if it was class we would be able to just redefine _try_metadata_service_call, but here we may need more explicit template-method approach).
It's just easier to reduce duplication now incrementally a bit.

Alternatively all code that goes after that metadata call can be just refactored to method called like create_gcp_attestation_from_response - and it will be defined in sync code.

src/snowflake/connector/aio/_wif_util.py
```diff
@@ -0,0 +1,304 @@
...
+        url_without_query_string = identity_endpoint
+        headers = {"X-IDENTITY-HEADER": identity_header}
+        query_params = f"api-version=2019-08-01&resource={snowflake_entra_resource}"
+
+        # Some Azure Functions environments may require client_id in the URL
+        managed_identity_client_id = os.environ.get("MANAGED_IDENTITY_CLIENT_ID")
+        if managed_identity_client_id:
+            query_params += f"&client_id={managed_identity_client_id}"
+
+    res = await try_metadata_service_call(
```
sfc-gh-fpawlowski, line 180: Same as above

test/unit/aio/test_auth_workload_identity_async.py
```diff
@@ -0,0 +1,382 @@
...
+    assert (
+        '{"_provider":"AWS","arn":"arn:aws:sts::123456789:assumed-role/A-Different-Role/i-34afe100cad287fab"}'
+        == auth_class.assertion_content
+    )
+
+
+# -- GCP Tests --
+
+
+def _mock_aiohttp_exception(exception):
```
sfc-gh-fpawlowski, line 176: nit: maybe we can move this to the mock_utils.py - but won't argue, since here it provides a quick access for the reader, so both sounds good

test/unit/aio/test_auth_workload_identity_async.py
```diff
@@ -0,0 +1,382 @@
...
+        return MockResponse(exception)
+
+    return mock_request
+
+
+@pytest.mark.parametrize(
+    "exception",
+    [
+        aiohttp.ClientError(),
+        asyncio.TimeoutError(),
```
sfc-gh-fpawlowski, line 198: Synchronous version differentiated Timeout from ConnectTimeout. Should we ensure using tests that in the future we are always covering this broader scope of exceptions?

test/unit/aio/test_auth_workload_identity_async.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+# -- Azure Tests --
+
+
+@pytest.mark.parametrize(
+    "exception",
+    [
+        aiohttp.ClientError(),
+        asyncio.TimeoutError(),
```
sfc-gh-fpawlowski, line 256: Same as above

src/snowflake/connector/aio/auth/_workload_identity.py
sfc-gh-pczajka: It's even worse - diamond. Good catch, inheritance order was incorrect here - I've fixed that and added test for every other auth class

src/snowflake/connector/aio/_wif_util.py
sfc-gh-pczajka: Changed

src/snowflake/connector/aio/_wif_util.py
sfc-gh-pczajka: I'd leave this as is for now - until we're cherrypicking other changes (no to cause additional conflicts). I've created follow-up ticket:  https://snowflakecomputing.atlassian.net/browse/SNOW-2234002

test/unit/aio/test_auth_workload_identity_async.py
sfc-gh-pczajka: I'd leave it here - this mock is very implementation-specific

test/unit/aio/test_auth_workload_identity_async.py
sfc-gh-pczajka: Added

test/unit/aio/test_auth_workload_identity_async.py
sfc-gh-pczajka: added

PR #2443, opened 2025-07-29
SNOW-2216803 allow re-raising error in file transfer work function in main thread

sfc-gh-zyao:
The change is protected by the server parameter `PYTHON_RERAISE_FILE_TRANSFER_WORK_FN_ERROR_AS_IS`. And by default it is not populated, unless the workload is Python sproc

   Fixes #NNNN


src/snowflake/connector/file_transfer_agent.py
```diff
@@ -626,6 +628,19 @@ def function_and_callback_wrapper(
                 logger.error(f"An exception was raised in {repr(work)}", exc_info=True)
                 file_meta.error_details = e
                 result = (False, e)
+                # If the parameter is enabled, notify the main thread of work
+                # function error, with the concrete exception stored aside in
+                # exception_caught_in_work, such that towards the end of
+                # the transfer call, we reraise the error as is immediately
+                # instead of continuing the execution after transfer.
+                if self._cursor.connection._session_parameters.get(
```
sfc-gh-zyao: @sfc-gh-sfan would like to discuss you on this. We have 3 options here, one being this code we are seeing here. Two other options are
- only re-raise for `ProgrammingError`  and `SystemError`
  - it makes sense because normally for non-sproc they won't running into error types of these two (ProgrammingError is normally something wrong in the user SQL queries, which would have got caught during SQL parsing for non-sproc; whereas SystemError is something caused by Python C API error, which is unlikely to get triggered by non-sproc
  - but this is not guaranteed to isolate the effect from non-sproc, which is not ideal
- leverage the DoP cap property. We re-raise error if the server DoP cap == 1
  - Because currently only sproc enforces DoP cap == 1, so this will only be applied to sproc
  - But technically users can set DoP cap manually in their config,  and get different behavior in error reporting, which is not ideal
  - Besides that means we need to keep populating the server DoP cap = 1 on the server side forever. And whenever we want to change DoP cap, it will cause unnecessary difficulty.

With the approach chosen by this PR, there still one downside, which is that we need to keep the parameter PYTHON_RERAISE_FILE_TRANSFER_WORK_FN_ERROR_AS_IS on the server side forever. But it is the least obtrusive among the three

sfc-gh-sfan: Can we add a connection level config to control this behavior (default disabled), and use the config when creating the connection for stored proc?

sfc-gh-zyao: I see. That looks more natural. If I understand your idea correctly, in the future we will still need to keep setting that config for sproc, similar to the idea that we need to keep setting a parameter, but a key advantage is that to keep setting a config forever looks more natural than keep setting a parameter forever. And besides after we "revamp" the common sproc infra, we may have an option to stop setting the config for all released connectors

I will redo the change to adopt this idea.

sfc-gh-sfan: Right. The change here is not very controversial for the connector outside of the context of stored proc.

sfc-gh-zyao: Done

src/snowflake/connector/connection.py
```diff
@@ -385,6 +386,10 @@ def _get_private_bytes_from_file(
         _DEFAULT_VALUE_SERVER_DOP_CAP_FOR_FILE_TRANSFER,  # default value
         int,  # type
     ),  # snowflake internal
+    RERAISE_ERROR_IN_FILE_TRANSFER_WORK_FUNCTION: (
```
sfc-gh-sfan: I think using a constant is an overkill here. All the properties will be translated into a field with a underscore prefix. And ideally you should only have this one use of the string.

src/snowflake/connector/file_transfer_agent.py
```diff
@@ -626,6 +628,21 @@ def function_and_callback_wrapper(
...
+                # If the reraise is enabled, notify the main thread of work
+                # function error, with the concrete exception stored aside in
+                # exception_caught_in_work, such that towards the end of
+                # the transfer call, we reraise the error as is immediately
+                # instead of continuing the execution after transfer.
+                if getattr(
+                    self._cursor.connection,
+                    f"_{RERAISE_ERROR_IN_FILE_TRANSFER_WORK_FUNCTION}",
+                    False,
+                ):
```
sfc-gh-sfan: If you have access to `self._cursor`, I wonder if we should pass the value of `reraise_error_in_file_transfer_work_function` into file transfer agent instead.

src/snowflake/connector/connection.py
sfc-gh-zyao: Makes sense. I have changed accordingly

src/snowflake/connector/file_transfer_agent.py
sfc-gh-zyao: Sounds good. Done

sfc-gh-mmishchenko (approved):

PR #2457, opened 2025-08-06
Remove WIF autodetect and all its problems (error messages, issuer checks, explicit timeouts), support explicit client_id for Azure VMs

sfc-gh-pmansour:
SNOW-2246093

This is part of some final pre-GA improvements for WIF:

**1. Remove support for "autodetect" of providers in WIF.**

Previously, you could use wif like this:
```
snowflake.connector.connect(
  account="a1",
  authenticator="WORKLOAD_IDENTITY"
)
```

Now, the `workload_identity_provider` parameter will be required:
```
snowflake.connector.connect(
  account="a1",
  authenticator="WORKLOAD_IDENTITY",
  workload_identity_provider="AWS"
)
```

Previously, this worked by attempting to get an attestation from all available providers, and picking the first one that worked. However, this mode created too many problems (higher latency, lower reliability due to timeouts, opaque error messages), and isn't recommended in production anyway. Better to just remove it before GA.

**2. Improve WIF error messages**

Previously, we swallowed all specific errors and returned `None`, because errors may have been expected in auto-detect mode (e.g. if we get a timeout from trying to get a GCP token, which is expected on Azure). Unfortunately this made it hard to troubleshoot authentication failures.

Since we no longer need to support auto-detect, we can bubble up any direct exceptions that were raised, and raise more specific errors for explicit failures.

**3. Remove issuer checks**

Previously, we couldn't be sure that a token returned from a metadata server was actually a "GCP" or "Azure" one (since they have similar IP addresses, and could support the same request paths in the future), so we parsed the token and had explicit validations on the issuers. Unfortunately, this created brittleness in the face of slightly different issuer formats (e.g. for Azure v1 vs. v2 tokens, or in sovereign clouds).

Since we no longer need to support auto-detect, we don't need to validate the issuer on the client-side.

**4. Remove explicit timeouts for metadata service calls**

Previously, we couldn't be sure which provider was correct, so we set a 3 second timeout for getting a token from each provider to avoid delaying authentication by too long. Unfortunately this created fragility, particularly when talking to metadata services that had slow cold start behavior (e.g. Azure IMDS could take up to 20 seconds for the first request), which would time out and fail authentication.

Since we no longer need to support auto-detect, we don't need to override HTTP timeouts for metadata service calls.

**5. Refactor code and simplify tests**

I updated all code paths that returned `Union[foo, None]` to simply return `foo`, or raise an error if there's an issue.

As a result, a lot of the error handling became more localized and cleaner, and I was able to remove a lot of tests for None-edge cases.

I also got rid of all the auto-detect tests.

**6. Always plumb client_id parameter if MANAGED_IDENTITY_CLIENT_ID is set**

The previous implementation only did this for Azure Functions, but it should be applicable more broadly (including in VMs), especially if user-assigned identities (or multiple identities) are available. See [this doc](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http) for more.

### DIFF
--- a/test/wif/test_wif.py
+++ b/test/wif/test_wif.py
@@ -25,18 +25,6 @@ PROVIDER = os.getenv("SNOWFLAKE_TEST_WIF_PROVIDER")
 
 
 @pytest.mark.wif
-def test_wif_provider_autodetection():
-    connection_params = {
-        "account": ACCOUNT,
-        "authenticator": "WORKLOAD_IDENTITY",
-        "host": HOST,
-    }
-    assert connect_and_execute_simple_query(
-        connection_params
-    ), "Failed to connect with using WIF - automatic provider detection"
-
-
-@pytest.mark.wif
 def test_wif_defined_provider():
     connection_params = {
         "host": HOST,


src/snowflake/connector/connection.py
```diff
@@ -1326,13 +1326,21 @@ def __open_connection(self):
...
-                # Standardize the provider enum.
-                if self._workload_identity_provider and isinstance(
-                    self._workload_identity_provider, str
-                ):
+
+                if isinstance(self._workload_identity_provider, str):
                     self._workload_identity_provider = AttestationProvider.from_string(
                         self._workload_identity_provider
                     )
+                if not self._workload_identity_provider:
```
sfc-gh-dbasavin, line 1334: Optional: Is it worth taking the opportunity to also improve the error message in cases when customers specify `workload_identity_provider='...'` but don't specify `authenticator='WORKLOAD_IDENTITY`?

Currently, the error in such cases is very misleading, e.g.

```
conn = snowflake.connector.connect(account="a1", workload_identity_provider="AWS")
```

Fails with
```
"251005: User is empty"
```

src/snowflake/connector/wif_util.py
```diff
@@ -86,86 +74,90 @@ def extract_iss_and_sub_without_signature_verification(jwt_str: str) -> tuple[st
...
+        raise ProgrammingError(
+            msg=f"No AWS identity was found. Ensure the application is running on AWS with an IAM role attached.",
+            errno=ER_WIF_CREDENTIALS_NOT_FOUND,
+        )
     return caller_identity["Arn"]
 
 
-def get_aws_partition(arn: str) -> str | None:
+def get_aws_partition(arn: str) -> str:
     """Get the current AWS partition from ARN, if any.
```
sfc-gh-dbasavin:
```suggestion
    """Get the current AWS partition from ARN.
```

src/snowflake/connector/wif_util.py
```diff
@@ -62,17 +54,13 @@ class WorkloadIdentityAttestation:
     user_identifier_components: dict
 
 
-def try_metadata_service_call(
-    method: str, url: str, headers: dict, timeout_sec: int = 3
-) -> Response | None:
-    """Tries to make a HTTP request to the metadata service with the given URL, method, headers and timeout.
+def try_metadata_service_call(method: str, url: str, headers: dict) -> Response | None:
```
sfc-gh-dbasavin: `try_metadata_service_call()` doesn't seem to be used anywhere anymore? (as far as I can tell, we call `requests.request()` now directly)

src/snowflake/connector/wif_util.py
```diff
@@ -292,116 +270,64 @@ def create_azure_attestation(
...
     return WorkloadIdentityAttestation(
         AttestationProvider.AZURE, jwt_str, {"iss": issuer, "sub": subject}
     )
 
 
-def create_oidc_attestation(token: str | None) -> WorkloadIdentityAttestation | None:
+def create_oidc_attestation(token: str | None) -> WorkloadIdentityAttestation:
     """Tries to create an attestation using the given token.
 
     If this is not populated, returns None.
```
sfc-gh-dbasavin:
```suggestion
    If this is not populated, raises an error.
```

src/snowflake/connector/connection.py
sfc-gh-pmansour: good call, updated it to say the following:
```
User is empty, but it must be provided unless authenticator is one of OAUTH, NO_AUTH, WORKLOAD_IDENTITY, PROGRAMMATIC_ACCESS_TOKEN, PAT_WITH_EXTERNAL_SESSION.
```

It's possible to improve this even more by moving the "authenticator validation" above the "user validation", but I'd prefer to avoid more structural changes here to reduce risk from this PR.

src/snowflake/connector/wif_util.py
sfc-gh-pmansour: good catch, I meant to delete that but left it till the end, in case I need to consult any code snippets. I just removed it.

sfc-gh-dbasavin (approved):

sfc-gh-pczajka (commented):
Could you remove autodetect check in tests/wif/test_wif.py? Otherwise Jenkins check will keep failing

sfc-gh-pmansour:
> Could you remove autodetect check in tests/wif/test_wif.py? Otherwise Jenkins check will keep failing

Good catch, didn't know that was a thing. Removed!

PR #2463, opened 2025-08-09
Cherrypicks to aio connector part12

sfc-gh-fpawlowski:
* [x] #2364 
  * changes:
    * Renamed directory integ/aio to integ/aio_it
    * 1 import had to be changed
    * No other changes were made

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -55,7 +55,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # temporarily reduce number of jobs: SNOW-2311643
+        # TODO: temporarily reduce number of jobs: SNOW-2311643
         # python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         python-version: ["3.9", "3.13"]
     steps:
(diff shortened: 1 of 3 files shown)


sfc-gh-pczajka (approved):

sfc-gh-pczajka (commented):
Do we want to change aio ot aio_it, or rather directories inside aio? Tests tend to have suffix _async anyway.
I think the second option is less chaotic, WDYT?
currently: `test/integ/aio_it/lambda/some_test.py`
proposition: `test/integ/aio/lambda_it/some_test.py`

sfc-gh-fpawlowski:
> Do we want to change aio ot aio_it, or rather directories inside aio? Tests tend to have suffix _async anyway. I think the second option is less chaotic, WDYT? currently: `test/integ/aio_it/lambda/some_test.py` proposition: `test/integ/aio/lambda_it/some_test.py`

I think we have to rethink well our markers here - do we want this automatic marking (defined in _pytest_collection_modifyitems_) to add both aio and integ/unit/... markers to our tests - or just the latter?

If both, we may actually need to rename both directories - I will look at this closer once we propagate changes that fix the tests - to observe how they are influenced by this change better.

PR #2489, opened 2025-08-19
SNOW-2268606-regression-3.17.0-unexplained-errors-in-connecting-to-IMDS-from-3.17.0-leads-to-connection-failures-and-excessive-logging-works-with-3.16.0

sfc-gh-fpawlowski:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #NNNN

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

4. (Optional) PR for stored-proc connector:

PR #2490, opened 2025-08-19
SNOW-2268606 zero timeout disables endpoint-based cloud platform detection

sfc-gh-mmishchenko:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/auth/_auth.py
+++ b/src/snowflake/connector/auth/_auth.py
@@ -185,7 +185,7 @@ class Auth:
             self._rest._connection.login_timeout,
             self._rest._connection._network_timeout,
             self._rest._connection._socket_timeout,
-            self._rest._connection._platform_detection_timeout_seconds,
+            self._rest._connection.platform_detection_timeout_seconds,
             session_manager=self._rest.session_manager.clone(use_pooling=False),
         )
 
--- a/src/snowflake/connector/platform_detection.py
+++ b/src/snowflake/connector/platform_detection.py
@@ -411,33 +411,36 @@ def detect_platforms(
         }
 
         # Run network-calling functions in parallel
-        with ThreadPoolExecutor(max_workers=6) as executor:
-            futures = {
-                "is_ec2_instance": executor.submit(
-                    is_ec2_instance, platform_detection_timeout_seconds
-                ),
-                "has_aws_identity": executor.submit(
-                    has_aws_identity, platform_detection_timeout_seconds
-                ),
-                "is_azure_vm": executor.submit(
-                    is_azure_vm, platform_detection_timeout_seconds, session_manager
-                ),
-                "has_azure_managed_identity": executor.submit(
-                    has_azure_managed_identity,
-                    platform_detection_timeout_seconds,
-                    session_manager,
-                ),
-                "is_gce_vm": executor.submit(
-                    is_gce_vm, platform_detection_timeout_seconds, session_manager
-                ),
-                "has_gcp_identity": executor.submit(
-                    has_gcp_identity,
-                    platform_detection_timeout_seconds,
-                    session_manager,
-                ),
-            }
+        if platform_detection_timeout_seconds != 0.0:
+            with ThreadPoolExecutor(max_workers=6) as executor:
+                futures = {
+                    "is_ec2_instance": executor.submit(
+                        is_ec2_instance, platform_detection_timeout_seconds
+                    ),
+                    "has_aws_identity": executor.submit(
+                        has_aws_identity, platform_detection_timeout_seconds
+                    ),
+                    "is_azure_vm": executor.submit(
+                        is_azure_vm, platform_detection_timeout_seconds, session_manager
+                    ),
+                    "has_azure_managed_identity": executor.submit(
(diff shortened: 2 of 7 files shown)


src/snowflake/connector/platform_detection.py
```diff
@@ -411,33 +411,36 @@ def detect_platforms(
...
-                ),
-                "has_gcp_identity": executor.submit(
-                    has_gcp_identity,
-                    platform_detection_timeout_seconds,
-                    session_manager,
-                ),
-            }
-
-            platforms.update({key: future.result() for key, future in futures.items()})
+        if platform_detection_timeout_seconds != 0.0:
```
sfc-gh-fpawlowski, line 414: The detect_platforms function is called by the base_auth_data function, which then does not pass platform_detection_timeout_seconds from connection in webbrowser.py and okta.py. Can we add this propagation as well, to be sure users can switch off this traffic fully?

sfc-gh-fpawlowski: Or, if no time for that, can we create a follow up jira?

sfc-gh-fpawlowski, line 414: Should platform_detection_timeout_seconds == None also cause this part being omitted?

test/unit/test_detect_platforms.py
```diff
@@ -288,3 +306,24 @@ def test_gce_cloud_run_job_missing_cloud_run_job(
     ):
         result = detect_platforms(platform_detection_timeout_seconds=None)
         assert "is_gce_cloud_run_job" not in result
+
+    def test_zero_platform_detection_timeout_disables_endpoints_detection_on_cloud(
```
sfc-gh-fpawlowski, line 310: Can we add tests that setting this platform_detection_timeout_seconds = 0 at the connection level makes the ThreadPoolExecutor never called (or those threads never being submitted)? We can do that using MagicMock and assert_not_called.

src/snowflake/connector/platform_detection.py
sfc-gh-mmishchenko: Indeed, fixed that.

sfc-gh-mmishchenko: I believe `None` is reserved as the default value.

test/unit/test_detect_platforms.py
sfc-gh-mmishchenko: A brilliant idea, added.

src/snowflake/connector/platform_detection.py
sfc-gh-fpawlowski: Ah right, we are good then 👍

sfc-gh-sshetkar, line 414: From what I understand, this platform_detection_timeout_seconds = 0 can only happen if the user explicitly passes this value in. Are we planning on asking users to pass in platform_detection_timeout = 0 as a means of disabling these checks?

src/snowflake/connector/auth/_auth.py
```diff
@@ -185,7 +185,7 @@ def authenticate(
             self._rest._connection.login_timeout,
             self._rest._connection._network_timeout,
             self._rest._connection._socket_timeout,
-            self._rest._connection._platform_detection_timeout_seconds,
+            self._rest._connection.platform_detection_timeout_seconds,
```
sfc-gh-sshetkar, line 188: Just wondering for future knowledge, do we prefer using the getter method implementation? I saw that _socket_timeout and _network_timeout used the actual value so was following that pattern before

src/snowflake/connector/platform_detection.py
sfc-gh-sshetkar: One thing I am also slightly worried about doing it this way is that if we roll out this fix. And then make an actual fix for not retrying and setting the log level properly, will we then have to ask users to change their settings back so that platfrom_detection_timeout_seconds to something not 0.0? @sfc-gh-pmansour also wanted to get your thoughts since we would end up wanting this data for WIF.

sfc-gh-sshetkar: I found this [PR](https://github.com/snowflakedb/snowflake-connector-python/pull/2489/files) so I think I might have been misunderstanding the purpose of this PR now. So to clarify, the regression fix was done separately and this PR is just to explicitly give a way for users to not make any network calls for platform detection by setting the platform_detection_timeout to 0 if they want to?

src/snowflake/connector/auth/_auth.py
sfc-gh-mmishchenko: I believe it's a convention - a leading underscore means private, not part of the interface. But it's true the existing code didn't follow that convention.

src/snowflake/connector/platform_detection.py
sfc-gh-mmishchenko: Yes, you get it right, the fix itself was merged already. This just adds an option to disable the networking part in platform detection, as a backup measure.

sfc-gh-fpawlowski:
We may need to fix those errors - MagicMock is passed somewhere, but it is not JSON serializable. Apart from that looks good to me 👍

PR #2496, opened 2025-08-21
initial changes

sfc-gh-eqin:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/connection.py
+++ b/src/snowflake/connector/connection.py
@@ -214,6 +214,7 @@ DEFAULT_CONFIGURATION: dict[str, tuple[Any, type | tuple[type, ...]]] = {
     "authenticator": (DEFAULT_AUTHENTICATOR, (type(None), str)),
     "workload_identity_provider": (None, (type(None), AttestationProvider)),
     "workload_identity_entra_resource": (None, (type(None), str)),
+    "workload_identity_impersonations": (None, (type(None), list)),
     "mfa_callback": (None, (type(None), Callable)),
     "password_callback": (None, (type(None), Callable)),
     "auth_class": (None, (type(None), AuthByPlugin)),
@@ -1359,6 +1360,7 @@ class SnowflakeConnection:
                     provider=self._workload_identity_provider,
                     token=self._token,
                     entra_resource=self._workload_identity_entra_resource,
+                    impersonations=self._workload_identity_impersonations,
                 )
             else:
                 # okta URL, e.g., https://<account>.okta.com/
@@ -1531,6 +1533,7 @@ class SnowflakeConnection:
             workload_identity_dependent_options = [
                 "workload_identity_provider",
                 "workload_identity_entra_resource",
+                "workload_identity_impersonations",
             ]
             for dependent_option in workload_identity_dependent_options:
                 if (
--- a/src/snowflake/connector/wif_util.py
+++ b/src/snowflake/connector/wif_util.py
@@ -12,6 +12,8 @@ import jwt
 from botocore.auth import SigV4Auth
 from botocore.awsrequest import AWSRequest
 from botocore.utils import InstanceMetadataRegionFetcher
+from google.auth import impersonated_credentials
+from google.auth.transport.requests import Request
 
 from .errorcode import ER_INVALID_WIF_SETTINGS, ER_WIF_CREDENTIALS_NOT_FOUND
 from .errors import ProgrammingError
@@ -185,6 +187,7 @@ def create_aws_attestation(
 
 
 def create_gcp_attestation(
+    impersonations: list[str] | None = None,
     session_manager: SessionManager | None = None,
 ) -> WorkloadIdentityAttestation:
     """Tries to create a workload identity attestation for GCP.
@@ -207,6 +210,24 @@ def create_gcp_attestation(
         )
 
     jwt_str = res.content.decode("utf-8")
+    if impersonations:
+        try:
+            for impersonation in impersonations:
+                jwt_str = impersonated_credentials.Credentials(
+                    source_credentials=jwt_str,
+                    target_principal=impersonation,
+                    target_audience=SNOWFLAKE_AUDIENCE,
+                )
+
+            # Refresh the last impersonated credential to get the final token
+            jwt_str.refresh(Request())
+            jwt_str = jwt_str.token
+        except Exception as e:
+            raise ProgrammingError(
+                msg=f"Error impersonating GCP service account: {e}. Ensure the service account has the 'Service Account Token Creator' role.",
+                errno=ER_WIF_CREDENTIALS_NOT_FOUND,
+            )
+
     _, subject = extract_iss_and_sub_without_signature_verification(jwt_str)
     return WorkloadIdentityAttestation(
         AttestationProvider.GCP, jwt_str, {"sub": subject}
@@ -295,6 +316,7 @@ def create_attestation(
     provider: AttestationProvider,
     entra_resource: str | None = None,
     token: str | None = None,
+    impersonations: list[str] | None = None,
     session_manager: SessionManager | None = None,
 ) -> WorkloadIdentityAttestation:
     """Entry point to create an attestation using the given provider.
@@ -313,7 +335,7 @@ def create_attestation(
     elif provider == AttestationProvider.AZURE:
         return create_azure_attestation(entra_resource, session_manager)
     elif provider == AttestationProvider.GCP:
-        return create_gcp_attestation(session_manager)
+        return create_gcp_attestation(impersonations, session_manager)
     elif provider == AttestationProvider.OIDC:
         return create_oidc_attestation(token)
     else:
--- a/src/snowflake/connector/auth/workload_identity.py
+++ b/src/snowflake/connector/auth/workload_identity.py
@@ -55,12 +55,14 @@ class AuthByWorkloadIdentity(AuthByPlugin):
         provider: AttestationProvider | None = None,
         token: str | None = None,
         entra_resource: str | None = None,
+        impersonations: list[str] | None = None,
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
         self.provider = provider
         self.token = token
         self.entra_resource = entra_resource
+        self.impersonations = impersonations
 
         self.attestation: WorkloadIdentityAttestation | None = None
 
@@ -85,6 +87,7 @@ class AuthByWorkloadIdentity(AuthByPlugin):
             self.provider,
             self.entra_resource,
             self.token,
+            self.impersonations,
             session_manager=(
                 conn._session_manager.clone(max_retries=0) if conn else None
             ),


sfc-gh-eqin:
* **#2496** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2496?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2496?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2510** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2510?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/snowflake/connector/wif_util.py
```diff
@@ -207,6 +210,24 @@ def create_gcp_attestation(
         )
 
     jwt_str = res.content.decode("utf-8")
+    if impersonations:
+        try:
+            for impersonation in impersonations:
+                jwt_str = impersonated_credentials.Credentials(
+                    source_credentials=jwt_str,
+                    target_principal=impersonation,
+                    target_audience=SNOWFLAKE_AUDIENCE,
```
sfc-gh-pmansour: I suspect the audience may need to be the GCP account service for all except the last token. But we should validate this.

src/snowflake/connector/wif_util.py
```diff
@@ -207,6 +210,24 @@ def create_gcp_attestation(
         )
 
     jwt_str = res.content.decode("utf-8")
+    if impersonations:
+        try:
+            for impersonation in impersonations:
+                jwt_str = impersonated_credentials.Credentials(
```
sfc-gh-pmansour: This is a Credentials object, not a jwt_str. Suggest making this a different variable from the one above (e.g. current_credentials). You may need to add an extra line before the start of the loop to initialize `current_credentials = jwt_str`.

src/snowflake/connector/wif_util.py
```diff
@@ -207,6 +210,24 @@ def create_gcp_attestation(
...
+                    source_credentials=jwt_str,
+                    target_principal=impersonation,
+                    target_audience=SNOWFLAKE_AUDIENCE,
+                )
+
+            # Refresh the last impersonated credential to get the final token
+            jwt_str.refresh(Request())
```
sfc-gh-pmansour: Do you know what this method does? Does it produce an ID token?

sfc-gh-eqin:

src/snowflake/connector/connection.py
```diff
@@ -1355,10 +1356,24 @@ def __open_connection(self):
...
+                ):
+                    Error.errorhandler_wrapper(
+                        self,
+                        None,
+                        ProgrammingError,
+                        {
+                            "msg": "workload_identity_impersonation_path is only supported for GCP.",
+                            "errno": ER_INVALID_WIF_SETTINGS,
+                        },
+                    )
```
sfc-gh-pmansour: could you please add tests in test/unit/test_connection.py that validate:
- Fails if impersonation_path is set but authenticator != WIF
- Fails if impersonation_path is set and authenticator == WIF, but provider != GCP
- Otherwise, correctly passes impersonation_path into the WorkloadIdentityAuth constructor.

src/snowflake/connector/wif_util.py
```diff
@@ -185,33 +185,74 @@ def create_aws_attestation(
...
     try:
-        res = session_manager.request(
-            method="GET",
-            url=f"http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/identity?audience={SNOWFLAKE_AUDIENCE}",
-            headers={
-                "Metadata-Flavor": "Google",
-            },
-        )
-        res.raise_for_status()
+        if not impersonation_path:
```
sfc-gh-pmansour: A large method with 2 non-overlapping code paths based on if..else is a sign that this method should be broken down into smaller ones :)

Could you extract a `get_gcp_workload_token(audience: str | None = None) -> str` method, which is used in both code paths? Then the `create_gcp_attestation` can call wrap around it:
1. If there is an impersonation_path, it'll call it initially with no audience, then call IAM credentials API.
2. Otherwise, it'll simply call it initially with Snowflake as the audience.

Finally, the last block can be the same in both cases (extract subject and wrap it in a WorkloadIdentityAttestation)

src/snowflake/connector/wif_util.py
```diff
@@ -185,33 +185,74 @@ def create_aws_attestation(
...
+                method="POST",
+                url=f"https://iamcredentials.googleapis.com/v1/{impersonation_path[-1]}:generateIdToken",
+                headers=headers,
+                json=body,
+            )
+            res.raise_for_status()
+
+            token = res.json()["token"]
+            return WorkloadIdentityAttestation(
+                AttestationProvider.GCP, token, {"sub": impersonation_path[-1]}
```
sfc-gh-pmansour: You should parse the sub from the token, similar to what we do in the non-impersonation case

src/snowflake/connector/wif_util.py
```diff
@@ -185,33 +185,74 @@ def create_aws_attestation(
...
+            # Get access token for the current service account
+            res = session_manager.request(
+                method="GET",
+                url="http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token",
+                headers={
+                    "Metadata-Flavor": "Google",
+                },
+            )
+            res.raise_for_status()
+            current_sa_token = res.json()["access_token"]
```
sfc-gh-pmansour: Are you sure this will work? Based on the current code path above, the metadata service should simply return the token as the entire response body, so you'd parse it like this:
```
current_sa_token = res.content.decode("utf-8")
```

test/wif/test_wif.py
sfc-gh-pmansour, line 1: FYI I added a new test doing the same thing (but with the details parameterized) in #2510. Rebasing on that branch will give you the new test

src/snowflake/connector/wif_util.py
sfc-gh-pmansour, line 1: Could you please add unit tests in test/unit/test_auth_workload_identity.py?

src/snowflake/connector/wif_util.py
sfc-gh-eqin: I think this method does work. The difference is that the ID token from the metadata server (the old path) will get a single token from the content. Calling `.json()` does not work.

On the other hand, for the access token from the metadata server (the new path), we will get a json in the body with keys "access_token", "expires_in", "token_type"

src/snowflake/connector/wif_util.py
sfc-gh-eqin: Do you mean this as a call to the metadata server? In which case, we will need to change whether we are requesting an access token or an ID token - is it correct to differentiate this by the audience parameter?

src/snowflake/connector/wif_util.py
sfc-gh-eqin: These 2 are equivalent if we want want to use decoding
```python
json.loads(res.content.decode("utf-8"))['access_token']
res.json()["access_token"]
```

src/snowflake/connector/connection.py
```diff
@@ -1355,10 +1356,24 @@ def __open_connection(self):
...
+                if (
+                    self._workload_identity_impersonation_path
+                    and self._workload_identity_provider != AttestationProvider.GCP
+                ):
+                    Error.errorhandler_wrapper(
+                        self,
+                        None,
+                        ProgrammingError,
+                        {
+                            "msg": "workload_identity_impersonation_path is only supported for GCP.",
```
sfc-gh-pmansour:
```suggestion
                            "msg": "workload_identity_impersonation_path is currently only supported for GCP.",
```

src/snowflake/connector/connection.py
```diff
@@ -1494,6 +1509,16 @@ def __config(self, **kwargs):
...
 
+        # Check that workload_identity_impersonation_path is only set for WORKLOAD_IDENTITY_AUTHENTICATOR
+        if (
+            self._workload_identity_impersonation_path
+            and self._authenticator != WORKLOAD_IDENTITY_AUTHENTICATOR
+        ):
+            raise ProgrammingError(
+                msg="workload_identity_impersonation_path is only supported for Workload Identity Federation.",
+                errno=ER_INVALID_WIF_SETTINGS,
+            )
```
sfc-gh-pmansour: This is already done below, since you added it to `workload_identity_dependent_options`. Suggest removing it from here.

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +184,95 @@ def create_aws_attestation(
...
+            return res.content.decode("utf-8")
+        elif gcp_token_type == GCPTokenType.ACCESS_TOKEN:
+            return res.json()["access_token"]
+            # return json.loads(res.content.decode("utf-8"))["access_token"]
+
     except Exception as e:
         raise ProgrammingError(
             msg=f"Error fetching GCP metadata: {e}. Ensure the application is running on GCP.",
             errno=ER_WIF_CREDENTIALS_NOT_FOUND,
         )
```
sfc-gh-pmansour: Given that the two token requests have different URLs and different response structures, consider two separate small methods: `get_gcp_id_token(audience: str): str` and `get_gcp_access_token(): str`. It seems better to have a tiny amount of duplication (e.g. the base URL) but 2 very small and easy to read methods, than an entire enum, some new branches, and a big method with 2 large branches, etc..

If you really want to avoid duplication, you can just extract the base URL into a constant `GCP_METADATA_SERVICE_ACCOUNT_BASE_URL = 'http://169.254.169.254/computeMetadata/v1/instance/service-accounts/default'`

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +184,95 @@ def create_aws_attestation(
...
+                headers=headers,
+                json=body,
+            )
+            res.raise_for_status()
+
+            jwt_str = res.json()["token"]
+        except Exception as e:
+            raise ProgrammingError(
+                msg=f"Error fetching GCP metadata: {e}. Ensure the application is running on GCP.",
+                errno=ER_WIF_CREDENTIALS_NOT_FOUND,
```
sfc-gh-pmansour: I still think a large branch that goes for most of the method is a sign of something that needs to be refactored. Could you extract the latter branch into some `get_id_token_with_impersonation_path()` method, so that this outer one is simplified to reduce branching?

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
     )
 
 
-def create_gcp_attestation(
-    session_manager: SessionManager | None = None,
-) -> WorkloadIdentityAttestation:
-    """Tries to create a workload identity attestation for GCP.
+def get_gcp_access_token(session_manager: SessionManager | None = None) -> str:
```
sfc-gh-pmansour: session_manager should not be optional

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
+            url=f"{GCP_METADATA_SERVICE_ACCOUNT_BASE_URL}/token",
             headers={
                 "Metadata-Flavor": "Google",
             },
         )
         res.raise_for_status()
+        return res.json()["access_token"]
     except Exception as e:
         raise ProgrammingError(
             msg=f"Error fetching GCP metadata: {e}. Ensure the application is running on GCP.",
```
sfc-gh-pmansour:
```suggestion
            msg=f"Error fetching GCP access token: {e}. Ensure the application is running on GCP.",
```

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
             errno=ER_WIF_CREDENTIALS_NOT_FOUND,
         )
 
-    jwt_str = res.content.decode("utf-8")
+
+def get_gcp_identity_token_via_impersonation(
+    impersonation_path: list[str], session_manager: SessionManager | None = None
```
sfc-gh-pmansour: session_manager should be required

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
+) -> str:
+    """Gets a GCP identity token from the metadata server.
+
+    If the application isn't running on GCP or no credentials were found, raises an error.
+    """
+    current_sa_token = get_gcp_access_token(session_manager)
+    impersonation_path = [
+        f"projects/-/serviceAccounts/{client_id}" for client_id in impersonation_path
+    ]
+    try:
```
sfc-gh-pmansour, line 229: consider a sanity check that `len(impersonation_path)>0`

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
+        res.raise_for_status()
+        return res.json()["token"]
+    except Exception as e:
+        raise ProgrammingError(
+            msg=f"Error fetching GCP metadata: {e}. Ensure the application is running on GCP.",
+            errno=ER_WIF_CREDENTIALS_NOT_FOUND,
+        )
+
+
+def get_gcp_identity_token(session_manager: SessionManager | None = None) -> str:
```
sfc-gh-pmansour: session_manager is required

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
+            url=f"{GCP_METADATA_SERVICE_ACCOUNT_BASE_URL}/identity?audience={SNOWFLAKE_AUDIENCE}",
+            headers={
+                "Metadata-Flavor": "Google",
+            },
+        )
+        res.raise_for_status()
+        return res.content.decode("utf-8")
+    except Exception as e:
+        raise ProgrammingError(
+            msg=f"Error fetching GCP metadata: {e}. Ensure the application is running on GCP.",
```
sfc-gh-pmansour:
```suggestion
            msg=f"Error fetching GCP identity token: {e}. Ensure the application is running on GCP.",
```

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
+            json={
+                "delegates": impersonation_path[:-1],
+                "audience": SNOWFLAKE_AUDIENCE,
+            },
+        )
+        res.raise_for_status()
+        return res.json()["token"]
+    except Exception as e:
+        raise ProgrammingError(
+            msg=f"Error fetching GCP metadata: {e}. Ensure the application is running on GCP.",
```
sfc-gh-pmansour:
```suggestion
            msg=f"Error fetching identity token for impersonated GCP service account '{impersonation_path[-1]}': {e}. Ensure the application is running on GCP.",
```

src/snowflake/connector/wif_util.py
```diff
@@ -184,29 +187,97 @@ def create_aws_attestation(
...
+            errno=ER_WIF_CREDENTIALS_NOT_FOUND,
+        )
+
+
+def create_gcp_attestation(
+    impersonation_path: list[str] | None = None,
+    session_manager: SessionManager | None = None,
```
sfc-gh-pmansour: session_manager is required

test/unit/test_auth_workload_identity.py
```diff
@@ -316,6 +322,39 @@ def test_explicit_gcp_generates_unique_assertion_content(
...
+            "Authorization": f"Bearer {sa1_access_token}",
+            "Content-Type": "application/json",
+        },
+        json={
+            "delegates": ["projects/-/serviceAccounts/sa2"],
+            "audience": "snowflakecomputing.com",
+        },
+    )
+
+    assert auth_class.assertion_content == '{"_provider":"GCP","sub":"sa3"}'
```
sfc-gh-pmansour, line 355: Consider another assert for the authentication request body

```suggestion
    assert auth_class.assertion_content == '{"_provider":"GCP","sub":"sa3"}'
    assert extract_api_data(auth_class) == {
        "AUTHENTICATOR": "WORKLOAD_IDENTITY",
        "PROVIDER": "GCP",
        "TOKEN": sa3_id_token,
    }
```

test/unit/test_auth_workload_identity.py
```diff
@@ -316,6 +322,39 @@ def test_explicit_gcp_generates_unique_assertion_content(
     assert auth_class.assertion_content == '{"_provider":"GCP","sub":"123456"}'
 
 
+@mock.patch("snowflake.connector.session_manager.SessionManager.post")
+def test_gcp_does_impersonation(
```
sfc-gh-pmansour: nit: clearer test method name

```suggestion
def test_gcp_impersonation_calls_correct_apis_and_populates_auth_data_for_final_sa(
```

Perhaps the suggested name is overboard :P but you get the idea -- the test name should say exactly what this test is doing. If this test fails, somebody should be able to look at the test name, figure out what it was supposed to be verifying, and fix it.

test/wif/test_wif.py
sfc-gh-pmansour, line 1: Please rebase on main and revert all changes to this file, except for removing `@pytest.mark.skip` on the new impersonation test

ci/wif/parameters/parameters_wif.json.gpg
sfc-gh-pmansour, line 1: reminder to revert this after rebasing

ci/test_wif.sh
sfc-gh-pmansour, line 1: reminder to revert this after rebasing

src/snowflake/connector/wif_util.py
sfc-gh-eqin: Added above

sfc-gh-pmansour:
@sfc-gh-eqin you may need to update the [release notes](https://github.com/snowflakedb/snowflake-connector-python/blob/8c32f5e29041a4bd5ea870154300b48f7595263c/DESCRIPTION.md?plain=1#L9) to add a new section for vNext, and add a bullet point saying impersonation for workload identity federation is now supported.

DESCRIPTION.md
```diff
@@ -7,6 +7,9 @@ https://docs.snowflake.com/
 Source code is also available at: https://github.com/snowflakedb/snowflake-connector-python
 
 # Release Notes
+- v3.17.4(September 03,2025)
```
sfc-gh-pmansour: Next release definitely won't be today, and probably won't be a patch version release

```suggestion
- v3.18.0(TBD)
```

DESCRIPTION.md
```diff
@@ -7,6 +7,9 @@ https://docs.snowflake.com/
...
+  - Added support for service account impersonation for Workload Identity Federation on GCP workloads only
```
sfc-gh-pmansour:
```suggestion
  - Added the `workload_identity_impersonation_path` parameter to support service account impersonation for Workload Identity Federation on GCP workloads only
```

Issue #2508, opened 2025-08-27
Support for custom oauth authenticators


jochenott:
### What is the current behavior?

Currently, when trying to use `authenticator="OAUTH_CLIENT_CREDENTIALS"`, the implementation (in class `AuthByOauthCredentials`) sends the client secret in basic auth.
Unfortunately, however, our oauth server accepts them only in the request body.  This is rejected by the oauth server and retrieving an access token fails.

While I'm aware the server behavior is not standard compliant, I cannot easily change the server behavior and we need some form of solution.

### What is the desired behavior?

I can think of two solutions:

 1. support sending client secret in the request body; this would solve our immediate issue
 2. more generally, allow the `auth_class` argument of `connect` to be classes other than built-in auth classes or subclasses of `AuthByKeyPair`, which is [the current restriction](https://github.com/snowflakedb/snowflake-connector-python/blob/7985a5fd5349225345afd22d9b976d0d2ef1b33f/src/snowflake/connector/connection.py#L1208-L1216). For example, also allow (customized) subclasses of built-in classes such as `AuthByOauthCredentials`.

### How would this improve `snowflake-connector-python`?

Wider oauth compatibility.

### References and other background

We are forced by snowflake to move away from username/password. Oauth compatibility should be improved accordingly to make this move smoother.

sfc-gh-dszmolka:
hello @jochenott - thank you for filing this enhancement request and also for the details. The team will consider it for future planning. 

If you are already a Snowflake customer, I strongly recommend to reach out to your Snowflake Account Team (not the Support) as they can help bringing such requests forward and with the prioritisation amongst the concurrent incoming requests. Thank you in advance !

sfc-gh-pszpakowski:
@jochenott  The proposed PR was merged in Python v .4.0.0 release, closing this request as done.

sfc-gh-pczajka:
@jochenott ad 2 - could you please open another request for that? It will be easier to track when it won't be mingled with another issue

jochenott:
> [@jochenott](https://github.com/jochenott) ad 2 - could you please open another request for that? It will be easier to track when it won't be mingled with another issue

Not sure what you mean. I think this issue has been fully addressed by merging https://github.com/snowflakedb/snowflake-connector-python/pull/2509 . For what do you require a new issue?

PR #2509, opened 2025-08-27
Add option to send oauth client credentials in body

jochenott:

#2508

   - [X] I am modifying authorization mechanisms

It adds an option to send the oauth client credentials in the request body, as defined in RFC-6749 Section 2.3.1 (as alternative).


src/snowflake/connector/auth/oauth_credentials.py
```diff
@@ -16,6 +17,10 @@
 logger = logging.getLogger(__name__)
 
 
+class TokenRequestMode(StrEnum):
```
sfc-gh-fpawlowski: We support Python versions from 3.9 to 3.13. StrEnum won't work below Python 3.11.

src/snowflake/connector/auth/oauth_credentials.py
```diff
@@ -16,6 +17,10 @@
...
+    header_only = "header_only"
+    header_and_body = "header_and_body"
+
```
sfc-gh-fpawlowski: I thought precommit would fix this to keep 2 empty lines between classes declarations. Maybe we can change that manually

src/snowflake/connector/auth/oauth_credentials.py
```diff
@@ -40,6 +46,7 @@ def __init__(
             **kwargs,
         )
         self._application = application
+        self._token_request_mode = TokenRequestMode[token_request_mode] if token_request_mode else TokenRequestMode.header_only
```
sfc-gh-fpawlowski: What errors will user get if he passes invalid _token_request_mode_? Can we catch that error / check beforehand and handle gracefully?

jochenott:

src/snowflake/connector/auth/oauth_credentials.py
jochenott: I changed it to derive from `str, Enum` in the second commit.

src/snowflake/connector/auth/oauth_credentials.py
jochenott: I moved converting the string to a `TokenRequestMode` to the call site, i.e. to `connection.py`, where an invalid value is now raising a descriptive error and handled via the `Error.errorhandler_wrapper` similar to other invalid connection values.

src/snowflake/connector/auth/oauth_credentials.py
jochenott: Fixed manually.

src/snowflake/connector/connection.py
```diff
@@ -337,6 +338,11 @@ def _get_private_bytes_from_file(
         (type(None), str),
         # SNOW-1825621: OAUTH implementation
     ),
+    "oauth_token_request_mode": (
```
sfc-gh-pczajka: Could you move the default value here? This way logic will stay in the same place and we won't need to validate the 'None' case

src/snowflake/connector/auth/oauth_credentials.py
```diff
@@ -16,6 +17,11 @@
 logger = logging.getLogger(__name__)
 
 
+class TokenRequestMode(str, Enum):
```
sfc-gh-pczajka: Do we expect more cases here? If not I'd change that to less cryptic "oauth_credentials_in_body: bool"

src/snowflake/connector/auth/oauth_credentials.py
```diff
@@ -60,4 +68,7 @@ def _request_tokens(
             "grant_type": "client_credentials",
             "scope": self._scope,
         }
+        if self._token_request_mode is TokenRequestMode.header_and_body:
```
sfc-gh-pczajka: Could you modify `test/unit/test_auth_oauth_credentials.py::test_oauth_client_credentials_authenticator_is_case_insensitive` to ensure that the request body is properly updated?

src/snowflake/connector/auth/oauth_credentials.py
jochenott: Hard to tell without knowing what other implementations are out there. I wanted to allow for that possibility, but if you think this is not needed, I can change it to a boolean.

src/snowflake/connector/auth/oauth_credentials.py
jochenott: Since the parameter has now be changed to a boolean `oauth_credentials_in_body`, this is obsolete.

src/snowflake/connector/auth/oauth_credentials.py
jochenott: I changed it to a boolean in the latest update.

src/snowflake/connector/auth/oauth_credentials.py
jochenott: I adapted the test to check the `fields` passed to `AuthByOauthCredentials._get_request_token_response` for all three cases (`oauth_credential_in_body`is `True` / `False` / not passed).

src/snowflake/connector/connection.py
jochenott: Changed to a bool with a default to keep current behavior, i.e. `oauth_credentials_in_body=False`, so this comment no longer applies.

sfc-gh-pczajka (approved):

sfc-gh-pczajka:
Thanks for the contribution!

PR #2517, opened 2025-09-03
Support WIF Impersonation on AWS workloads

sfc-gh-eqin:

   Fixes SNOW-2183347

The issue is that some use cases have a CSP account that maps to multiple other CSP accounts, each of which may map to a Snowflake user.

The code does the following on AWS workloads:

* Allows users to specify a workload_identity_impersonation_path of impersonated accounts
* Uses the impersonations to get the identity token for the last role in the impersonation path
  * For AWS, this is the GetCallerIdentity request using a session of the last Assumed Role instead of the identity token
* Uses the same verification procedure with the Snowflake backend to get a session token with the user corresponding to the last service account in the impersonation path

### DIFF
--- a/test/csp_helpers.py
+++ b/test/csp_helpers.py
@@ -389,9 +389,9 @@ class FakeAwsEnvironment:
             self.assume_role_call_count += 1
             return {
                 "Credentials": {
-                    "AccessKeyId": gen_dummy_access_token(arn, "access_key"),
-                    "SecretAccessKey": gen_dummy_access_token(arn, "secret_key"),
-                    "SessionToken": gen_dummy_access_token(arn, "session_token"),
+                    "AccessKeyId": "access_key",
+                    "SecretAccessKey": "secret_key",
+                    "SessionToken": "session_token",
                     "Expiration": int(time()) + 60 * 60,
                 },
                 "AssumedRoleUser": {"AssumedRoleId": hash(arn), "Arn": arn},
--- a/test/unit/test_auth_workload_identity.py
+++ b/test/unit/test_auth_workload_identity.py
@@ -274,7 +274,7 @@ def test_get_aws_sts_hostname_invalid_inputs(region, partition):
     assert "Invalid AWS partition" in str(excinfo.value)
 
 
-def test_aws_calls_correct_apis_and_populates_auth_data_for_final_role(
+def test_aws_impersonation_calls_correct_apis_for_each_role_in_impersonation_path(
     fake_aws_environment: FakeAwsEnvironment,
 ):
     impersonation_path = [
--- a/test/unit/test_connection.py
+++ b/test/unit/test_connection.py
@@ -713,24 +713,16 @@ def test_workload_identity_impersonation_path_errors_for_unsupported_providers(
 
 
 @pytest.mark.parametrize(
-    "provider_param,impersonation_path,expected_provider",
+    "provider_param,impersonation_path",
     [
-        (
-            AttestationProvider.GCP,
-            ["sa2@project.iam.gserviceaccount.com"],
-            AttestationProvider.GCP,
-        ),
-        (
-            AttestationProvider.AWS,
-            ["arn:aws:iam::1234567890:role/role2"],
-            AttestationProvider.AWS,
-        ),
-        ("GCP", ["sa2@project.iam.gserviceaccount.com"], AttestationProvider.GCP),
-        ("AWS", ["arn:aws:iam::1234567890:role/role2"], AttestationProvider.AWS),
+        (AttestationProvider.GCP, ["sa2@project.iam.gserviceaccount.com"]),
+        (AttestationProvider.AWS, ["arn:aws:iam::1234567890:role/role2"]),
+        ("GCP", ["sa2@project.iam.gserviceaccount.com"]),
+        ("AWS", ["arn:aws:iam::1234567890:role/role2"]),
     ],
 )
 def test_workload_identity_impersonation_path_populates_auth_class_for_supported_provider(
-    monkeypatch, provider_param, impersonation_path, expected_provider
+    monkeypatch, provider_param, impersonation_path
 ):
     with monkeypatch.context() as m:
         m.setattr(
@@ -743,7 +735,6 @@ def test_workload_identity_impersonation_path_populates_auth_class_for_supported
             workload_identity_provider=provider_param,
             workload_identity_impersonation_path=impersonation_path,
         )
-        assert conn.auth_class.provider == expected_provider
         assert conn.auth_class.impersonation_path == impersonation_path
 
 


sfc-gh-eqin:
* **#2517** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

test/csp_helpers.py
```diff
@@ -376,6 +381,42 @@ def __init__(self):
         )
         self.metadata_token = "test-token"
 
+    def assume_role(self, RoleArn, RoleSessionName):
```
sfc-gh-eqin: Not too sure if there is a better way to verify the session that this is called on. I couldn't figure out a way to reference the values inside the boto3.Session that calls this method.

test/unit/test_auth_workload_identity.py
```diff
@@ -274,6 +274,28 @@ def test_get_aws_sts_hostname_invalid_inputs(region, partition):
...
+    )
+    auth_class.prepare(conn=None)
+
+    assert (
+        auth_class.assertion_content
+        == '{"_provider":"AWS","partition":"aws","region":"us-east-1"}'
+    )
+    api_data = extract_api_data(auth_class)
+    assert api_data["AUTHENTICATOR"] == "WORKLOAD_IDENTITY"
+    assert api_data["PROVIDER"] == "AWS"
```
sfc-gh-pmansour: I don't think this test asserts the most important / relevant parts of the logic here. The authenticator and provider are irrelevant and already verified in a different test above, so we shouldn't verify them again in this test. We need something to show that actual assumption happened.

Normally you should assert that the output (e.g. JWT subject) is the expected one, but this is very difficult here since the output of an AWS assertion isn't a JWT with clearly expected fields, but rather an opaque blob.

Instead, could you at least verify that the correct "assume-role" APIs were called? The test name here includes `calls_correct_apis` but this isn't verified anywhere here.

test/unit/test_connection.py
```diff
@@ -709,20 +707,30 @@ def test_workload_identity_impersonation_path_unsupported_for_non_gcp_providers(
...
+            AttestationProvider.AWS,
+        ),
+        ("GCP", ["sa2@project.iam.gserviceaccount.com"], AttestationProvider.GCP),
+        ("AWS", ["arn:aws:iam::1234567890:role/role2"], AttestationProvider.AWS),
     ],
 )
-def test_workload_identity_impersonation_path_supported_for_gcp_provider(
-    monkeypatch, provider_param
+def test_workload_identity_impersonation_path_populates_auth_class_for_supported_provider(
+    monkeypatch, provider_param, impersonation_path, expected_provider
```
sfc-gh-pmansour: nit: expected_provider is not really relevant for this test and is already verified in other tests, no need to parameterize it and assert its value. You can just focus on impersonation_path.

test/csp_helpers.py
sfc-gh-pmansour: nit: please follow Python snakecase naming convention:

```suggestion
    def assume_role(self, role_arn, role_session_name):
```

test/csp_helpers.py
```diff
@@ -376,6 +381,42 @@ def __init__(self):
...
+                    "SecretAccessKey": gen_dummy_access_token(arn, "secret_key"),
+                    "SessionToken": gen_dummy_access_token(arn, "session_token"),
+                    "Expiration": int(time()) + 60 * 60,
+                },
+                "AssumedRoleUser": {"AssumedRoleId": hash(arn), "Arn": arn},
+                "ResponseMetadata": {},
+            }
+            self.num_assumptions += 1
+            return self.last_assume_role_response
+        return {}
```
sfc-gh-pmansour: not sure I understand this block, but will take a deeper look later

test/csp_helpers.py
sfc-gh-eqin: The code called in the impersonation logic calls the AWS API `assume_role`. The keyword args for this use RoleArn and RoleSessionName, so patching the method with this dummy method requires the same keywords.

If we want, I can try changing this to `**kwargs`

test/csp_helpers.py
sfc-gh-eqin: This is really just a hacky way to verify that the `assume_role` calls are in order. I had wanted to verify that the session calling the client with `assume_role` had been constructed with the correct credentials, but could not figure out a way to do so efficiently.

test/unit/test_connection.py
sfc-gh-eqin: I see, should we still be testing that the WIIP parameter is valid for both AWS and GCP providers then? Or would only GCP work?

test/unit/test_auth_workload_identity.py
```diff
@@ -274,6 +274,22 @@ def test_get_aws_sts_hostname_invalid_inputs(region, partition):
     assert "Invalid AWS partition" in str(excinfo.value)
 
 
+def test_aws_calls_correct_apis_and_populates_auth_data_for_final_role(
```
sfc-gh-pmansour: nit: test name should be descriptive of what it's verifying
```suggestion
def test_aws_with_impersonation_path_calls_correct_apis(
```

test/csp_helpers.py
```diff
@@ -376,6 +380,25 @@ def __init__(self):
...
+            self.assumption_path
+            and kwargs["RoleArn"] == self.assumption_path[self.assume_role_call_count]
+        ):
+            arn = self.assumption_path[self.assume_role_call_count]
+            self.assume_role_call_count += 1
+            return {
+                "Credentials": {
+                    "AccessKeyId": gen_dummy_access_token(arn, "access_key"),
+                    "SecretAccessKey": gen_dummy_access_token(arn, "secret_key"),
+                    "SessionToken": gen_dummy_access_token(arn, "session_token"),
```
sfc-gh-pmansour: tiny nit: those don't need to be "tokens" by the typical definition of the word. You can just use random string placeholders here to simplify and avoid confusion:

```suggestion
                    "AccessKeyId": "access_key_id",
                    "SecretAccessKey": "secret_access_key",
                    "SessionToken": "session_token",
```

test/unit/test_connection.py
sfc-gh-pmansour: Testing for both AWS and GCP is fine, and is already parameterized via the `provider_param` parameter. All I'm saying is you don't need to assert / verify that the parsed provider == expected_provider. That part is irrelevant for this test and is verified elsewhere.

So if you just drop `expected_provider` and its corresponding assertion, this test will be 👌

test/csp_helpers.py
sfc-gh-pmansour: ack, I read it again and I think it makes sense. Thanks!

sfc-gh-pmansour (approved):

PR #2518, opened 2025-09-04
SNOW-2117147 Added certificates revocation checking with revocation lists (CRLs)

sfc-gh-mmishchenko:
[ ] Move CRL cache manager to a singletone
[ ] Add checking if CRL disk cache was cleaned up recently

### DIFF
--- a/src/snowflake/connector/crl.py
+++ b/src/snowflake/connector/crl.py
@@ -1,0 +1,469 @@
+#!/usr/bin/env python
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
+from enum import Enum, unique
+from logging import getLogger
+from pathlib import Path
+from types import TracebackType
+from typing import Any
+
+from cryptography import x509
+from cryptography.hazmat._oid import ExtensionOID
+from cryptography.hazmat.backends import default_backend
+from typing_extensions import Self
+
+from .crl_cache import CRLCacheEntry, CRLCacheManager, CRLFileCache, CRLInMemoryCache
+from .session_manager import SessionManager
+
+logger = getLogger(__name__)
+
+
+@unique
+class CertRevocationCheckMode(Enum):
+    """Certificate revocation check modes based on revocation lists (CRL)
+
+    CRL mode descriptions:
+        DISABLED: No revocation check is done.
+        ENABLED: Revocation check is done in the strictest way. The endpoint must expose at least one fully valid
+            certificate chain. Any check error invalidate the chain.
+        ADVISORY: Revocation check is done in a more relaxed way. Only a revocated certificate can invalidate
+            the chain. An error is treated positively (as a successful check).
+    """
+
+    DISABLED = "DISABLED"
+    ENABLED = "ENABLED"
+    ADVISORY = "ADVISORY"
+
+
+class _CRLValidationResult(Enum):
+    """Certificate revocation validation result statuses"""
+
+    REVOKED = "REVOKED"
+    UNREVOKED = "UNREVOKED"
+    ERROR = "ERROR"
+
+
+@dataclass
+class CRLConfig:
+    """Configuration class for CRL validation settings."""
+
+    cert_revocation_check_mode: CertRevocationCheckMode = (
+        CertRevocationCheckMode.DISABLED
+    )
+    allow_certificates_without_crl_url: bool = False
+    connection_timeout_ms: int = 3000
+    read_timeout_ms: int = 3000
+    cache_validity_time: timedelta = timedelta(hours=24)
+    enable_crl_cache: bool = True
+    enable_crl_file_cache: bool = True
+    crl_cache_dir: Path | str | None = None
+    crl_cache_removal_delay_days: int = 7
+    crl_cache_cleanup_interval_hours: int = 1
+    crl_cache_start_cleanup: bool = False
+
+    @classmethod
+    def from_connection(cls, connection) -> CRLConfig:
+        """
+        Create a CRLConfig instance from a SnowflakeConnection instance.
+
+        This method extracts CRL configuration parameters from the connection's
+        read-only properties and creates a CRLConfig instance.
+
+        Args:
+            connection: SnowflakeConnection instance containing CRL configuration
+
+        Returns:
+            CRLConfig: Configured CRLConfig instance
+
+        Raises:
+            ValueError: If session_manager is not available in the connection
+        """
+        # Extract CRL-specific configuration parameters from connection properties
+        cert_revocation_check_mode_str = connection.cert_revocation_check_mode
+        if isinstance(cert_revocation_check_mode_str, str):
+            try:
+                cert_revocation_check_mode = CertRevocationCheckMode(
+                    cert_revocation_check_mode_str
+                )
+            except ValueError:
+                logger.warning(
+                    f"Invalid cert_revocation_check_mode: {cert_revocation_check_mode_str}, "
+                    "defaulting to DISABLED"
+                )
+                cert_revocation_check_mode = CertRevocationCheckMode.DISABLED
+        else:
+            cert_revocation_check_mode = cert_revocation_check_mode_str
+
+        # Handle cache validity time from hours to timedelta
+        cache_validity_time = timedelta(hours=connection.crl_cache_validity_hours)
+
+        # Handle cache directory path
+        crl_cache_dir = connection.crl_cache_dir
+        if crl_cache_dir is not None and not isinstance(crl_cache_dir, Path):
+            crl_cache_dir = Path(crl_cache_dir)
+
+        return cls(
+            cert_revocation_check_mode=cert_revocation_check_mode,
+            allow_certificates_without_crl_url=connection.allow_certificates_without_crl_url,
+            connection_timeout_ms=connection.crl_connection_timeout_ms,
+            read_timeout_ms=connection.crl_read_timeout_ms,
+            cache_validity_time=cache_validity_time,
+            enable_crl_cache=connection.enable_crl_cache,
+            enable_crl_file_cache=connection.enable_crl_file_cache,
+            crl_cache_dir=crl_cache_dir,
+            crl_cache_removal_delay_days=connection.crl_cache_removal_delay_days,
+            crl_cache_cleanup_interval_hours=connection.crl_cache_cleanup_interval_hours,
+            crl_cache_start_cleanup=connection.crl_cache_start_cleanup,
+        )
+
+
+class CRLValidator:
+    def __init__(
+        self,
+        session_manager: SessionManager | Any,
+        cert_revocation_check_mode: CertRevocationCheckMode = CertRevocationCheckMode.DISABLED,
+        allow_certificates_without_crl_url: bool = False,
+        connection_timeout_ms: int | None = None,
+        read_timeout_ms: int | None = None,
+        cache_validity_time: timedelta | None = None,
+        cache_manager: CRLCacheManager | None = None,
+    ):
+        self._session_manager = session_manager
+        self._cert_revocation_check_mode = cert_revocation_check_mode
+        self._allow_certificates_without_crl_url = allow_certificates_without_crl_url
+        self._connection_timeout_ms = connection_timeout_ms or 3000
+        self._read_timeout_ms = read_timeout_ms or 3000
+        self._cache_validity_time = cache_validity_time or timedelta(hours=24)
+        self._cache_manager = cache_manager or CRLCacheManager.noop()
+
+    @classmethod
+    def from_config(
+        cls, config: CRLConfig, session_manager: SessionManager
+    ) -> CRLValidator:
+        """
+        Create a CRLValidator instance from a CRLConfig.
+
+        This method creates a CRLValidator and its underlying objects (except session_manager)
+        from configuration parameters found in the CRLConfig.
+
+        Args:
+            config: CRLConfig instance containing CRL-related parameters
+            session_manager: SessionManager instance
+
+        Returns:
+            CRLValidator: Configured CRLValidator instance
+        """
+        # Create cache manager if caching is enabled
+        cache_manager = None
+        if config.enable_crl_cache:
+            # Create memory cache
+            memory_cache = CRLInMemoryCache(config.cache_validity_time)
+
+            # Create file cache if enabled
+            if config.enable_crl_file_cache:
+                removal_delay = timedelta(days=config.crl_cache_removal_delay_days)
+                file_cache = CRLFileCache(
+                    cache_dir=config.crl_cache_dir, removal_delay=removal_delay
+                )
+            else:
+                from snowflake.connector.crl_cache import NoopCRLCache
+
+                file_cache = NoopCRLCache()
+
+            # Create cache manager with cleanup
+            cleanup_interval = timedelta(hours=config.crl_cache_cleanup_interval_hours)
+
+            cache_manager = CRLCacheManager(
+                memory_cache=memory_cache,
+                file_cache=file_cache,
+                cleanup_interval=cleanup_interval,
+                start_cleanup=config.crl_cache_start_cleanup,
+            )
+        else:
+            cache_manager = CRLCacheManager.noop()
+
+        return cls(
+            session_manager=session_manager,
+            cert_revocation_check_mode=config.cert_revocation_check_mode,
+            allow_certificates_without_crl_url=config.allow_certificates_without_crl_url,
+            connection_timeout_ms=config.connection_timeout_ms,
+            read_timeout_ms=config.read_timeout_ms,
+            cache_validity_time=config.cache_validity_time,
+            cache_manager=cache_manager,
+        )
+
+    def validate_certificate_chains(
+        self, certificate_chains: list[list[x509.Certificate]]
+    ) -> bool:
+        """
+        Validate certificate chains against CRLs with actual HTTP requests
+
+        Args:
+            certificate_chains: List of certificate chains to validate
+
+        Returns:
+            True if validation passes, False otherwise
+
+        Raises:
+            ValueError: If certificate_chains is None or empty
+        """
+        if self._cert_revocation_check_mode == CertRevocationCheckMode.DISABLED:
+            return True
+
+        if certificate_chains is None or len(certificate_chains) == 0:
+            logger.warning("Certificate chains are empty")
+            if self._cert_revocation_check_mode == CertRevocationCheckMode.ADVISORY:
+                return True
+            return False
+
+        results = []
+        for chain in certificate_chains:
+            result = self._validate_single_chain(chain)
+            # If any of the chains is valid, the whole check is considered positive
+            if result == _CRLValidationResult.UNREVOKED:
+                return True
+            results.append(result)
+
+        # In non-advisory mode we require at least one chain get a clear UNREVOKED status
+        if self._cert_revocation_check_mode != CertRevocationCheckMode.ADVISORY:
+            return False
+
+        # We're in advisory mode, so any error is treated positively
+        return any(result == _CRLValidationResult.ERROR for result in results)
+
+    def _validate_single_chain(
+        self, chain: list[x509.Certificate]
+    ) -> _CRLValidationResult:
+        """Validate a single certificate chain"""
+        # An empty chain is considered an error
+        if len(chain) == 0:
+            return _CRLValidationResult.ERROR
+        # the last certificate of the chain is considered the root and isn't validated
+        results = []
+        for i in range(len(chain) - 1):
+            result = self._validate_certificate(chain[i], chain[i + 1])
+            if result == _CRLValidationResult.REVOKED:
+                return _CRLValidationResult.REVOKED
+            results.append(result)
+
+        if _CRLValidationResult.ERROR in results:
+            return _CRLValidationResult.ERROR
+
+        return _CRLValidationResult.UNREVOKED
+
+    def _validate_certificate(
+        self, cert: x509.Certificate, parent: x509.Certificate
+    ) -> _CRLValidationResult:
+        """Validate a single certificate against CRL"""
+        # Check if certificate is short-lived (skip CRL check)
+        if self._is_short_lived_certificate(cert):
+            return _CRLValidationResult.UNREVOKED
+
+        # Extract CRL distribution points
+        crl_urls = self._extract_crl_distribution_points(cert)
+
+        if not crl_urls:
+            # No CRL URLs found
+            if self._allow_certificates_without_crl_url:
+                return _CRLValidationResult.UNREVOKED
+            return _CRLValidationResult.ERROR
+
+        results: list[_CRLValidationResult] = []
+        # Check against each CRL URL
+        for crl_url in crl_urls:
+            result = self._check_certificate_against_crl(cert, parent, crl_url)
+            if result == _CRLValidationResult.REVOKED:
+                return result
+            results.append(result)
+
+        if all(result == _CRLValidationResult.ERROR for result in results):
+            return _CRLValidationResult.ERROR
+
+        return _CRLValidationResult.UNREVOKED
+
+    @staticmethod
+    def _is_short_lived_certificate(cert: x509.Certificate) -> bool:
+        """Check if certificate is short-lived (validity <= 5 days)"""
+        try:
+            # Use UTC versions to avoid deprecation warnings
+            validity_period = cert.not_valid_after_utc - cert.not_valid_before_utc
+        except AttributeError:
+            # Fallback for older versions
+            validity_period = cert.not_valid_after - cert.not_valid_before
+        return validity_period.days <= 5
+
+    @staticmethod
+    def _extract_crl_distribution_points(cert: x509.Certificate) -> list[str]:
+        """Extract CRL distribution point URLs from certificate"""
+        try:
+            crl_dist_points = cert.extensions.get_extension_for_oid(
+                ExtensionOID.CRL_DISTRIBUTION_POINTS
+            ).value
+
+            urls = []
+            for point in crl_dist_points:
+                if point.full_name:
+                    for name in point.full_name:
+                        if isinstance(name, x509.UniformResourceIdentifier):
+                            urls.append(name.value)
+            return urls
+        except x509.ExtensionNotFound:
+            return []
+
+    def _get_crl_from_cache(self, crl_url: str) -> CRLCacheEntry | None:
+        return self._cache_manager.get(crl_url)
+
+    def _put_crl_to_cache(
+        self, crl_url: str, crl: x509.CertificateRevocationList, ts: datetime
+    ) -> None:
+        self._cache_manager.put(crl_url, crl, ts)
+
+    def _download_crl(
+        self, crl_url: str
+    ) -> tuple[x509.CertificateRevocationList | None, datetime | None]:
+        now = datetime.now(UTC)
+        try:
+            logger.debug("Trying to download CRL from: %s", crl_url)
+            response = self._session_manager.get(crl_url, timeout=30)
+            response.raise_for_status()
+
+        except Exception:
+            # CRL fetch or parsing failed
+            logger.exception("Failed to download CRL from %s", crl_url)
+            return None, None
+
+        try:
+            logger.debug("Trying to parse CRL from: %s", crl_url)
+            crl = x509.load_der_x509_crl(response.content, backend=default_backend())
+
+            # Check if CRL is expired
+            try:
+                next_update = crl.next_update_utc
+            except AttributeError:
+                next_update = crl.next_update
+
+            if next_update and now > next_update:
+                logger.warning(
+                    "The CRL from %s was expired on %s", crl_url, next_update
+                )
+                return None, None
+
+            return crl, now
+        except Exception:
+            logger.exception("Failed to parse CRL from %s", crl_url)
+            return None, None
+
+    def _check_certificate_against_crl(
+        self, cert: x509.Certificate, parent: x509.Certificate, crl_url: str
+    ) -> _CRLValidationResult:
+        """Check if certificate is revoked according to CRL"""
+        now = datetime.now(UTC)
+        logger.debug("Trying to get cached CRL for %s", crl_url)
+        cached_crl = self._get_crl_from_cache(crl_url)
+        if (
+            cached_crl is None
+            or cached_crl.is_crl_expired_by(now)
+            or cached_crl.is_evicted_by(now, self._cache_validity_time)
+        ):
+            crl, ts = self._download_crl(crl_url)
+            if crl and ts:
+                self._put_crl_to_cache(crl_url, crl, ts)
+        else:
+            crl = cached_crl.crl
+
+        # If by some reason we didn't get a valid CRL we consider it a check error
+        if crl is None:
+            return _CRLValidationResult.ERROR
+
+        # Check if certificate is revoked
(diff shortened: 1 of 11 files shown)


src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,575 @@
...
+
+        if cert_revocation_check_mode == CertRevocationCheckMode.DISABLED:
+            # The rest of the parameters don't matter if CRL checking is disabled
+            return cls(cert_revocation_check_mode=cert_revocation_check_mode)
+
+        # Handle cache validity time from hours to timedelta
+        cache_validity_time = (
+            cls.cache_validity_time
+            if connection.crl_cache_validity_hours is None
+            else timedelta(hours=connection.crl_cache_validity_hours)
```
sfc-gh-pczajka: Should we wrap this into more readable message? Currently for incorrect type I've got:
```
...
TypeError: unsupported type for timedelta hours component: str
```

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,575 @@
...
+        connection_timeout_ms = (
+            cls.connection_timeout_ms
+            if connection.crl_connection_timeout_ms is None
+            else connection.crl_connection_timeout_ms
+        )
+        read_timeout_ms = (
+            cls.read_timeout_ms
+            if connection.crl_read_timeout_ms is None
+            else connection.crl_read_timeout_ms
+        )
```
sfc-gh-pczajka, line 137: should we convert these into integers?

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,575 @@
...
+
+    def _put_crl_to_cache(
+        self, crl_url: str, crl: x509.CertificateRevocationList, ts: datetime
+    ) -> None:
+        self._cache_manager.put(crl_url, crl, ts)
+
+    def _fetch_crl_from_url(self, crl_url: str) -> bytes | None:
+        try:
+            logger.debug("Trying to download CRL from: %s", crl_url)
+            response = self._session_manager.get(crl_url, timeout=30)
```
sfc-gh-pczajka: Should this be. `config. connection_timeout_ms`?

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,575 @@
...
+    ) -> CRLValidationResult:
+        """Check if certificate is revoked according to CRL"""
+        revoked_cert = crl.get_revoked_certificate_by_serial_number(cert.serial_number)
+        return (
+            CRLValidationResult.REVOKED
+            if revoked_cert
+            else CRLValidationResult.UNREVOKED
+        )
+
+    def validate_connection(self, connection) -> bool:
```
sfc-gh-pczajka: Can we add connection type?

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Absolutely, good point.

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Do you think it would make sense to make it more tolerable to the the input type and introduce a nested int() conversion? Or just wrap into a better message?

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: It should, thx

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Yes, let me recheck all type hints again.

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Applied a unified approach with a conversion to the target type.

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Applied an explicit conversion to the target type.

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Was addressed.

src/snowflake/connector/crl.py
sfc-gh-mmishchenko: Addressed.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+@dataclass
+class CRLConfig:
+    """Configuration class for CRL validation settings."""
+
+    cert_revocation_check_mode: CertRevocationCheckMode = (
+        CertRevocationCheckMode.DISABLED
+    )
+    allow_certificates_without_crl_url: bool = False
+    connection_timeout_ms: int = 3000
+    read_timeout_ms: int = 3000
```
sfc-gh-jkasten, line 57: This seems really short default given we are dealing with 10MB files in the non-emergency case (DigiCert). Have we tested this and are we confident with this setting? (Performance will also get significantly worse in the mass revocation scenarios)

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+    cache_validity_time: timedelta = timedelta(hours=24)
+    enable_crl_cache: bool = True
+    enable_crl_file_cache: bool = True
+    crl_cache_dir: Path | str | None = None
+    crl_cache_removal_delay_days: int = 7
+    crl_cache_cleanup_interval_hours: int = 1
+    crl_cache_start_cleanup: bool = False
```
sfc-gh-jkasten, line 64: Certificates rotate and CRLs stop being accessed. I would almost think that if you built the complexity into the code, that it would be worth it to turn it on by default? I imagine that removing the files is relatively cheap, but I am sure I am missing a concern.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+        Args:
+            certificate_chains: List of certificate chains to validate
+
+        Returns:
+            True if validation passes, False otherwise
+
+        Raises:
+            ValueError: If certificate_chains is None or empty
+        """
+        if self._cert_revocation_check_mode == CertRevocationCheckMode.DISABLED:
```
sfc-gh-jkasten, line 274: optional / no action required - It seems pretty easy to create a CRLValidator and call validate_certificate_chains thinking that you are validating CRLs and unwittingly just performing NOOPs. The default CRLValidation mode is DISABLED and the individual method's documentation is insufficient to understand what the method does.

I would be tempted to make this more prominent.

My preferred order...
1. Move the configuration to the caller. Don't create a CRLValidator in the first place if CRL checking is disabled. Throw an exception if someone tries to create a disabled CRLValidator. Avoid unnecessary confusion?
2. Log that CRLValidation is disabled to give some breadcrumbs that they are calling a NOOP?
3. Add the caveat to the API documentation?

I don't think it is that problematic a the use of the library is limited, but I do think it would be confusing for new developers.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+            return True
+
+        if certificate_chains is None or len(certificate_chains) == 0:
+            logger.warning("Certificate chains are empty")
+            if self._cert_revocation_check_mode == CertRevocationCheckMode.ADVISORY:
+                return True
+            return False
```
sfc-gh-jkasten, line 281: nit / optional: I think this simplifies to 

`return self._cert_revocation_check_mode == CertRevocationCheckMode.ADVISORY`

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+            if not cert_chain:
+                logger.debug("No certificate chain found in connection")
+                return []
+
+            # Convert OpenSSL certificates to cryptography x509 certificates
+            x509_chain = []
+            for cert_openssl in cert_chain:
+                cert_der = dump_certificate(FILETYPE_ASN1, cert_openssl)
+                cert_x509 = x509.load_der_x509_certificate(cert_der, default_backend())
+                x509_chain.append(cert_x509)
```
sfc-gh-jkasten, line 565: I think you can avoid this conversion by using `get_peer_certificate(true)` ?

https://www.pyopenssl.org/en/latest/api/ssl.html#OpenSSL.SSL.Connection.get_peer_cert_chain

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+            logger.debug(
+                "Extracted %d certificates for CRL validation", len(x509_chain)
+            )
+            return [x509_chain]  # Return as a single chain
+
+        except Exception as e:
+            logger.warning(
+                "Failed to extract certificate chain for CRL validation: %s", e
+            )
+            return []
```
sfc-gh-jkasten, line 576: This doesn't appear to return a robust set of chains? There is very rarely a singular chain for a given website. [[See real world example](https://docs.google.com/document/d/1cgkHlAqFBmF0qrnSgdwvGp1RhJdx0w2nayIl_iJIknE/edit?tab=t.l6soesqfmoiq#heading=h.bouoatrg7sjk)]
I think this code will suffer from problems like the Starfield incident.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+        Args:
+            connection: OpenSSL connection object
+
+        Returns:
+            List of certificate chains, where each chain is a list of x509.Certificate objects
+        """
+        from OpenSSL.crypto import FILETYPE_ASN1, dump_certificate
+
+        try:
+            cert_chain = connection.get_peer_cert_chain()
```
sfc-gh-jkasten, line 555: Elsewhere in the code, it is assumed that this always returns the certificates in order and that the root is included. I don't think that is true though or rather all the documentation I could quickly find for this function indicated it was just the chain as sent by the peer.

For example...
1. Well-configured servers won't send the root.
2. The "chain" might not even be direct [[example](https://docs.google.com/document/d/1cgkHlAqFBmF0qrnSgdwvGp1RhJdx0w2nayIl_iJIknE/edit?tab=t.l6soesqfmoiq#bookmark=id.tdbgfl6yrth1)]

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+    def _validate_single_chain(
+        self, chain: list[x509.Certificate]
+    ) -> CRLValidationResult:
+        """Validate a single certificate chain"""
+        # An empty chain is considered an error
+        if len(chain) == 0:
+            return CRLValidationResult.ERROR
+        # the last certificate of the chain is considered the root and isn't validated
+        results = []
+        for i in range(len(chain) - 1):
```
sfc-gh-jkasten, line 307: FWIW, I don't think this is true based upon how chains are being returned today.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+        for crl_url in crl_urls:
+            result = self._check_certificate_against_crl_url(cert, ca_cert, crl_url)
+            if result == CRLValidationResult.REVOKED:
+                return result
+            results.append(result)
+
+        if all(result == CRLValidationResult.ERROR for result in results):
+            return CRLValidationResult.ERROR
+
+        return CRLValidationResult.UNREVOKED
```
sfc-gh-jkasten, line 346: This only works because we are assuming that each CRL distribution point points to a complete CRL, yet we don't short-circuit the validation when we come across an UNREVOKED result.

If we are going to make the assumption that each CRL is complete, we should return the result as soon as it != ERROR. (The assumption is fine based upon RFC 5280 + Baseline Requirements)

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+    @staticmethod
+    def _is_short_lived_certificate(cert: x509.Certificate) -> bool:
+        """Check if certificate is short-lived (validity <= 5 days)"""
+        try:
+            # Use timezone.utc versions to avoid deprecation warnings
+            validity_period = cert.not_valid_after_utc - cert.not_valid_before_utc
+        except AttributeError:
+            # Fallback for older versions
+            validity_period = cert.not_valid_after - cert.not_valid_before
+        return validity_period.days <= 5
```
sfc-gh-jkasten, line 357: This needs to be redefined. [[Definition of a short-lived cert](https://github.com/cabforum/servercert/blob/main/docs/BR.md#:~:text=Short%2Dlived%20Subscriber,days%20(604%2C800%20seconds).)] Add a little buffer as notAfter is inclusive.

e.g. CAs can issue 7 day certificates (short-lived) and short-lived certs do not need to contain a CRL distribution point. Those certificates would mistakenly fail validation here.
https://github.com/cabforum/servercert/blob/main/docs/BR.md#712112-crl-distribution-points

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+
+            urls = []
+            for point in crl_dist_points:
+                if point.full_name:
+                    for name in point.full_name:
+                        if isinstance(name, x509.UniformResourceIdentifier):
+                            urls.append(name.value)
+            return urls
+        except x509.ExtensionNotFound:
+            return []
```
sfc-gh-jkasten, line 375: nit/optional: Defensive coding would make sure that we are getting back HTTP URLs. Technically all valid certificates should only use HTTP, but not too infrequently CAs fail to follow the rules and RFC 5280 is much more loose on the field. (Mostly, smaller CAs)

https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13
https://github.com/cabforum/servercert/blob/main/docs/BR.md#712112-crl-distribution-points

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+        self, cert: x509.Certificate, ca_cert: x509.Certificate, crl_url: str
+    ) -> CRLValidationResult:
+        """Check if certificate is revoked according to CRL by the provided URL"""
+        now = datetime.now(timezone.utc)
+        logger.debug("Trying to get cached CRL for %s", crl_url)
+        cached_crl = self._get_crl_from_cache(crl_url)
+        if (
+            cached_crl is None
+            or cached_crl.is_crl_expired_by(now)
+            or cached_crl.is_evicted_by(now, self._cache_validity_time)
```
sfc-gh-jkasten, line 432: We shouldn't eject a valid cached CRL before we can guarantee we have downloaded the next one. This significantly increases the brittleness of the implementation.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+        ):
+            crl, ts = self._download_crl(crl_url)
+            if crl and ts:
+                self._put_crl_to_cache(crl_url, crl, ts)
```
sfc-gh-jkasten, line 436: nit: Ideally, we compare the cached version to the downloaded version and use the most recent one. (Avoid 10-day window attacks before the CRL expires.)

The design doc algorithm covers this.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+        # Verify CRL signature with CA public key
+        # Check if the CA certificate is the expected CRL issuer
+        if crl.issuer != ca_cert.subject:
+            logger.warning(
+                "CRL issuer (%s) does not match CA certificate subject (%s) for URL: %s",
+                crl.issuer,
+                ca_cert.subject,
+                crl_url,
+            )
+            # In most cases this indicates a configuration issue, but we'll still try verification
```
sfc-gh-jkasten, line 453: We should return an error here. This is specifically part of the RFC 5280 Section 6.3 algorithm.
It is extremely unlikely that a CA use the same key with different "issuers", but it is possible. 

(Certainly 12 years ago there were several examples of key reuse across different issuers. It would be much more difficult to do today given the recent "key freshness" requirements for new CAs and there might even be an explicit restriction that I just glossed over as the requirement never pertained to me, but we should error out regardless.)

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+
```
sfc-gh-jkasten, line 454: This is missing the IssuingDistributionPoint check. 
All certificates whose CA partitions their CRLs would be vulnerable to substitution attacks.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+                    crl.tbs_certlist_bytes,
+                    ec.ECDSA(hash_algorithm),
+                )
+            else:
+                # For other key types (DSA, etc.), try without padding
+                public_key.verify(
+                    crl.signature,
+                    crl.tbs_certlist_bytes,
+                    hash_algorithm,
+                )
```
sfc-gh-jkasten, line 504: This would not be legal. I think you can safely drop support and throw an exception. It will probably make debugging easier?
(We will deal with post-quantum algorithms before DSA is allowed...)
https://github.com/cabforum/servercert/blob/main/docs/BR.md#7132-signature-algorithmidentifier

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,576 @@
...
+            # Determine the appropriate padding based on the signature algorithm
+            public_key = ca_cert.public_key()
+
+            # Handle different key types with appropriate signature verification
+            if isinstance(public_key, rsa.RSAPublicKey):
+                # For RSA signatures, we need to use PKCS1v15 padding
+                public_key.verify(
+                    crl.signature,
+                    crl.tbs_certlist_bytes,
+                    padding.PKCS1v15(),
```
sfc-gh-jkasten, line 488: Note: Technically, CAs can use RSA-PSS and we are ignoring the signature algorithm specified. I am guessing the rest of our codebase would fail to handle that as well?
https://github.com/cabforum/servercert/blob/main/docs/BR.md#71321-rsa

If it affects everyone, we can leave it for later. I don't think CAs are going to switch before PQ.

src/snowflake/connector/crl_cache.py
```diff
@@ -0,0 +1,643 @@
...
+        Check if the CRL has expired.
+
+        Args:
+            ts: Time to check against
+
+        Returns:
+            True if the CRL has expired, False otherwise
+        """
+        next_update = self._next_update()
+        return next_update is not None and next_update < ts
```
sfc-gh-jkasten, line 48: nit/optional: Can we return expired if nextUpdate is None. This should never happen based upon the requirements?

sfc-gh-jkasten (commented):
Thanks, Maxim and Patryk!

src/snowflake/connector/crl.py
sfc-gh-pczajka: IMO it's quite elegant solution that simplifies the code - whole logic is kept in one place, connection / network does not need to care what option is set in revocation_check_mode. I could add special NoopValidator class to make this more explicit (similarly to cache) - WDYT?

src/snowflake/connector/crl.py
sfc-gh-pczajka: Increased to 90s

src/snowflake/connector/crl.py
sfc-gh-pczajka: Nice catch!

src/snowflake/connector/crl.py
sfc-gh-pczajka: Could you elaborate? How should we behave instead? What is the Starfield incident?

src/snowflake/connector/crl.py
sfc-gh-pczajka: 1. How should we behave if the root is missing? Assume it's ok?

2. What do you mean? Should we sort them before this check? Some of them can be "extras" which can be ignored if we find a valid path? (one on the example has two parents)

src/snowflake/connector/crl.py
sfc-gh-pczajka: Just to be clear: you refer to [the previous comment](https://github.com/snowflakedb/snowflake-connector-python/pull/2518#discussion_r2392368228) :D?

src/snowflake/connector/crl.py
sfc-gh-pczajka: You mean we should return first found "UNREVOKED" as well? What's TLDR behind the assumption (you cannot issue a certificate which is valid longer than yours)?

src/snowflake/connector/crl.py
sfc-gh-pczajka: Should we trust expired crl list if we fail to download a new one?

src/snowflake/connector/crl.py
sfc-gh-pczajka: To make it clear: in case the cache misses / entry is expired, we try to download new one and choose most recent one?

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed #2567

src/snowflake/connector/crl.py
sfc-gh-jkasten: I would opt for a NoopValidator class to make it more explicit, but you don't need to block for this change. 

More background, My interpretation of the code was that...
``` 
validator = CrlValidator()
validator.validate_certificate_chains(chains)
```
is a NOOP, but it doesn't read that way. Thinking people are calling security functions and nothing actually being done is an extremely common security vulnerability.  It actually has an extremely long history in the TLS space itself so I am a little sensitive to it.

From the paper [The Most Dangerous Code in the World: Validating SSL Certificates in Non-Browser Software](https://www.cs.columbia.edu/~suman/docs/suman_ccs12.pdf) - CCS 2012 (CCS is a top tier security conference)
```
Second, many SSL libraries are unsafe by default, requiring higher-level software to correctly set their options, provide hostname
verification functions, and interpret return values. As we show,
software that relies on these libraries is often not up to the task
```

Again though -- we have like 1 caller of this library. Making changes here isn't that important. The optional/no action required header in the original comment was accurate.

src/snowflake/connector/crl.py
sfc-gh-jkasten: Yes - sorry for the confusion.

src/snowflake/connector/crl.py
sfc-gh-jkasten: This is pretty minor and is only a performance optimization. 

If the results are [UNREVOKED, ERROR], my understanding is. that we return UNREVOKED. Therefore, we are assuming that we did not have to download that second CRL distribution point? We are assuming that the CRLs are redundant?

Now why I think this is OK requires a close reading across both the CABForum requirements and RFC 5280. I will try to provide my understanding and viewpoint, but it isn't terribly important. 

Multiple CRL distribution points can either be for redundancy purposes or the CRLs may only contain a subset of the revocation reasons.  For instance, by RFC 5280, you can put keyCompromise revocations in one CRL and all other reasons in another in case a client wants to optimize.

However, CRL Distribution Points reasons are not allowed by the Baseline Requirements, the associated CRLs MUST have all reason codes by RFC 5280." [[BR text](https://github.com/cabforum/servercert/blob/main/docs/BR.md#712112-crl-distribution-points), [RFC 5280 text](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13:~:text=If%20the%20DistributionPoint%20omits%20the%20reasons%20field%2C%20the%20CRL%20MUST%0A%20%20%20include%20revocation%20information%20for%20all%20reasons.)] Every CRL distribution point found in a certificate will have all reasons. The BRs are unlikely to loosen on this requirement as they are currently built for what the browsers want... and no one has the stomach for more complexity.

The Baseline Requirements do not prohibit CAs issuing CRLs with onlySomeReasons, but it essentially can't be a CRL that is referenced inside a certificate due to the other requirements.

src/snowflake/connector/crl.py
sfc-gh-jkasten: So CRLs typically have "lifetimes" of 7 days. Our default here IIRC, is to start trying to fetch a new CRL after an hour?

I love that we try to fetch a new CRL before the old one expires, but if that fetch fails, it would be optimal to use the one fetched slightly before. It may still have several days of lifetime. [[desigh doc discussion](https://docs.google.com/document/d/1Dy4eb8dDV75XIUKBvOMfeZjnxYca6ICKsCtgWwfU9Fs/edit?disco=AAABk6j3w_w)]

You are very likely to not be able to fetch a CRL when you are being MITM'd, so cached values are really where the power comes in. An old CRL is much better than no CRL.

src/snowflake/connector/crl.py
sfc-gh-jkasten: Yes, I think that sounds right. Don't assume anything. You should assume you are being actively MITM'd when you are fetching a CRL. (That is the atttack they are trying to defend against)

src/snowflake/connector/crl.py
sfc-gh-jkasten: Yes, I think longterm we need to build our own chains from the presented peer's chain. This is the part that always trips libraries up and causes large-scale outages when properties of the chain change [[design doc comment](https://docs.google.com/document/d/1Dy4eb8dDV75XIUKBvOMfeZjnxYca6ICKsCtgWwfU9Fs/edit?disco=AAABk5HwcwU)]

1.  For reference - the TLS 1.3 spec for root omission [[RFC 8446 - Section 4.4.2 text](https://datatracker.ietf.org/doc/html/rfc8446#:~:text=Because%0A%20%20%20certificate%20validation%20requires%20that%20trust%20anchors%20be%20distributed%0A%20%20%20independently%2C%20a%20certificate%20that%20specifies%20a%20trust%20anchor%20MAY%20be%0A%20%20%20omitted%20from%20the%20chain%2C%20provided%20that%20supported%20peers%20are%20known%20to%0A%20%20%20possess%20any%20omitted%20certificates.)]

I think you are probably going to need to find the roots from the root store to validate the full chain. Which leans into part 2.

2. Chains are more like graphs. Technically we should find each chain that reaches a root and validate them individually. We did do this with JDBC [[code](https://github.com/sfc-gh-jkasten/snowflake-jdbc/blob/6a4e077460bdcac97c63fe62e181452cc30e8ef0/src/main/java/net/snowflake/client/core/crl/VerifiedCertPathBuilder.java)]

TLS 1.3 spec describing the requirement and complexity [[RFC 8446 - Section 4.4.2 text](https://datatracker.ietf.org/doc/html/rfc8446#:~:text=Note%3A%20Prior%20to,MUST%20be%20first.)]

Simplified background blog: https://medium.com/@sleevi_/path-building-vs-path-verifying-the-chain-of-pain-9fbab861d7d6

src/snowflake/connector/crl.py
sfc-gh-jkasten: Gave some background in the response to the other [comment](https://github.com/snowflakedb/snowflake-connector-python/pull/2518/files#r2402022502).

Longer list of background sources.
1. https://medium.com/@sleevi_/path-building-vs-path-verifying-the-chain-of-pain-9fbab861d7d6
2. https://netflixtechblog.com/revisiting-bettertls-certificate-path-building-4c978b79843f
3. [Recent Starfield Situation](https://docs.google.com/document/d/1RKOgtgn-oVIM0dA1dWYAz1PzOjXiIcmZUwDZDH-9ffw/edit?tab=t.0#heading=h.vnkx2gi28its)

sfc-gh-jkasten (commented):
Thanks @sfc-gh-pczajka for picking this up. Sorry for all of the comments and complexity.

src/snowflake/connector/crl.py
sfc-gh-pfus: Wow, 90s? IMHO it's totally too much. Two reasons:
1) If you need 90s to download CRL, this is 99,99% chance, that it is already hanging and will never be downloaded. But still - the driver hangs for 90s. We had many complains from customers that our connections are very slow due to OCSP and still time out anyway.
2) If there is another "global" timeout for the whole HTTP connection (and I believe there is) and it is lower than this, there's no point setting particular step timeout above this total timeout.
In a normal situation I would say that 3-5s, maybe 10 max is sufficient to download even quite a large CRL. I believe they are pretty good cached. But maybe I'm wrong and @sfc-gh-jkasten has better suggestion.

src/snowflake/connector/crl.py
sfc-gh-pfus: +1, remaining drivers also have it enabled (at least go has it).

src/snowflake/connector/crl.py
sfc-gh-pfus: Also, just to double check - how range behaves if negative number is returned? In case of self signed certificates (used in testing environments) we have no chain.

src/snowflake/connector/crl.py
sfc-gh-pfus: Is it the final decision that any unrevoked status is final? If yes, we have a bug in our [design](https://docs.google.com/document/d/1Dy4eb8dDV75XIUKBvOMfeZjnxYca6ICKsCtgWwfU9Fs/edit?tab=t.0#heading=h.pzq6z9v7gkgt) and all drivers do the same...

src/snowflake/connector/crl_cache.py
sfc-gh-pfus: If we return expired, we would redownload this CRL for every new HTTP connection :) But yes, it should never happen.

src/snowflake/connector/crl.py
sfc-gh-jkasten: I was mostly concerned that I didn't think the current algorithm was consistent between ERROR and UNREVOKED.

The design has...
```
    if (results.contains(CERT_ERROR)) {
        return CERT_ERROR
    } else {
        return CERT_UNREVOKED
    }
```
which is consistent.and conservative.

This Python algorithm only returns ERROR if all the results are ERROR, which I thought was making the redundancy assumption.

I did think that @sfc-gh-pfus brought up the topic a few months ago when we discovered that DigiCert had both extremely large CRLs and was relying upon redundancy. Proving correctness without doing the additional work to parse the reasonCodes required a lot of additional very close reading on my part which I subsequently performed.

Regardless, we should stay consistent. We should follow the design doc. If we need more performance collectively, we should change over everything.

sfc-gh-pczajka: Fixed https://github.com/snowflakedb/snowflake-connector-python/pull/2567

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed https://github.com/snowflakedb/snowflake-connector-python/pull/2567

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed https://github.com/snowflakedb/snowflake-connector-python/pull/2567

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed https://github.com/snowflakedb/snowflake-connector-python/pull/2567

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed https://github.com/snowflakedb/snowflake-connector-python/pull/2567

src/snowflake/connector/crl.py
sfc-gh-jkasten: I don't have relevant latency/throughput numbers. DigiCert is my primary concern as they do not effectively partition their CRL.

I do know that with a 5 second timeout we are requiring the client is able to download the CRL at 2MB per second in the non-emergency revocation scenario. We know that DigiCert has ballooned their CRL to 16MB/20MB with a relatively small mass revocation before and it can easily balloon to be much higher.

The pressure is on CAs (Microsoft is still dealing with that [incident](https://bugzilla.mozilla.org/show_bug.cgi?id=1965612) where their CRL should have ballooned to 100MB.)

I guess this value is configurable though and the client already had to specially configure their code to hit this path anyway. It shouldn't be that big of a deal.

PR #2520, opened 2025-09-05
SNOW-2176203: Support intermediates in trust stores

sfc-gh-snoonan:
Our current python connector does not support a configuration with intermediate certificates in a trust store as roots of trust. This is allowed by RFCs; the root does not need to be self signed, and this is the behavior we have in our Go client.

These changes enable partial chain validation to normalize behavior across clients.

   - [x] I am modifying OCSP code


src/snowflake/connector/ssl_wrap_socket.py
```diff
@@ -38,9 +40,57 @@
...
+        pass
+
+
+def _build_context_with_partial_chain(cafile: str | None) -> PyOpenSSLContext:
+    """Create PyOpenSSL context configured for CERT_REQUIRED and partial-chain trust."""
+    ctx = PyOpenSSLContext(ssl_.PROTOCOL_TLS_CLIENT)
+    try:
+        ctx.verify_mode = ssl.CERT_REQUIRED
+    except Exception:
+        pass
```
sfc-gh-jkasten, line 85: optional: According to the documentation, `PROTOCOL_TLS_CLIENT` enables `CERT_REQUIRED` and `check_hostname` by default.
https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT

I guess I would either be explicit with both options or maybe drop this addtional code?

src/snowflake/connector/ssl_wrap_socket.py
```diff
@@ -38,9 +40,57 @@
...
+            # Leave context unchanged; handshake/validation surfaces failures
+            pass
+    try:
+        store = ctx._ctx.get_cert_store()
+        from OpenSSL import crypto as _crypto
+
+        if hasattr(_crypto, "X509StoreFlags") and hasattr(
+            _crypto.X509StoreFlags, "PARTIAL_CHAIN"
+        ):
+            store.set_flags(_crypto.X509StoreFlags.PARTIAL_CHAIN)
```
sfc-gh-jkasten, line 73: Note/optional: There is an [ssl.VERIFY_X509_PARTIAL_CHAIN](https://docs.python.org/3/library/ssl.html#ssl.VERIFY_X509_PARTIAL_CHAIN) which can be set on the SSLContext directly under [SSLContext.verify_flags](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_flags) I think? (Version 3.10)

It might simplify some of this code. My apologies if I missed something.

sfc-gh-jkasten (approved):
Thanks! 

LGTM - though it has been many years since I wrote production Python.

src/snowflake/connector/ssl_wrap_socket.py
sfc-gh-snoonan: This is PyOpenSSL versus the stdlib. We're working with pretty old code here, so I try all the things :-).

src/snowflake/connector/ssl_wrap_socket.py
sfc-gh-snoonan: As it is today, we do hostname verification in urllib3 rather than the SSLContext. (src/snowflake/connector/vendored/urllib3/util/ssl_.py). We might want to change that, but one thing at a time I think. This is another stdlib vs PyOpenSSL distinction. The context here just maps to the OpenSSL method used; we could probably get away with just PROTOCOL_TLS. They both map to OpenSSL.SSL.SSLv23_METHOD.

Yes, this is janky.

DESCRIPTION.md
```diff
@@ -8,8 +8,9 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 - v3.18.0(TBD)
-  - Added the `workload_identity_impersonation_path` parameter to support service account impersonation for Workload Identity Federation on GCP and AWS workloads only
+  - Added the `workload_identity_impersonation_path` parameter to support service account impersonation for Workload Identity Federation on GCP workloads only
```
sfc-gh-pczajka:
```suggestion
  - Added the `workload_identity_impersonation_path` parameter to support service account impersonation for Workload Identity Federation on GCP and AWS workloads only
```
I believe bad merge :D

test/unit/test_ssl_partial_chain.py
```diff
@@ -0,0 +1,70 @@
...
+        cert_reqs=None,
+        ca_certs=None,
+        server_hostname=None,
+        ssl_version=None,
+        ciphers=None,
+        ssl_context=None,
+        ca_cert_dir=None,
+        key_password=None,
+        ca_cert_data=None,
+        tls_in_tls=False,
```
sfc-gh-pczajka:
```suggestion
        sock,
        ssl_context=None,
        **kwargs
```

sfc-gh-pczajka (commented):
Left two nits

PR #2525, opened 2025-09-08
SNOW-2107016 Add option to exclude boto3 and botocore from dependencies

sfc-gh-pczajka:

   Fixes #SNOW-2107016

This PR introduces an option to install `snowflake-connector-python` without boto3 and botocore dependencies 
```
SNOWFLAKE_NO_BOTO=1 pip install snowflake-connector-python
```
It includes as well:
  * pre-commit check to ensure that all boto imports are included in try..except block
  * github workflow ensuring that the default installation includes boto dependencies, but "slim" one does not.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -12,7 +12,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Fixed `get_results_from_sfqid` when using `DictCursor` and executing multiple statements at once
   - Added the `oauth_credentials_in_body` parameter supporting an option to send the oauth client credentials in the request body
   - Fix retry behavior for `ECONNRESET` error
-  - Added an option to exclude `botocore` and `boto3` dependencies by setting `SNOWFLAKE_NO_BOTO` environment variable
+  - Added an option to exclude `botocore` and `boto3` dependencies by setting `SNOWFLAKE_NO_BOTO` environment variable during installation
 
 - v3.17.4(September 22,2025)
   - Added support for intermediate certificates as roots when they are stored in the trust store


setup.py
```diff
@@ -174,6 +174,34 @@ def new__compile(obj, src: str, ext, cc_args, extra_postargs, pp_opts):
...
+
+class SmartEggInfoCommand(egg_info):
+    """Custom egg_info command that auto-adds AWS extra unless slim is specified."""
+
+    def finalize_options(self):
+        super().finalize_options()
+
+        # Check if slim is being requested via environment variable
+        # (This is the most reliable way to detect slim installation intent)
+        is_slim_install = os.environ.get("SNOWFLAKE_SLIM_INSTALL", "").lower() in (
```
sfc-gh-pczajka: Refactor

```suggestion
        is_slim_install = os.environ.get("SNOWFLAKE_NO_BOTO", "").lower() in (
```

.github/workflows/check_installation.yml
```diff
@@ -0,0 +1,56 @@
...
+          source test_default_env/bin/activate
+
+          python -m pip install .
+          pip freeze | grep boto || exit 1  # boto3/botocore should be installed by default
+
+          # Deactivate and clean up
+          deactivate
+          rm -rf test_default_env
+
+      - name: Test installation with SNOWFLAKE_NO_BOTO=1 (should exclude boto)
```
sfc-gh-pmansour, line 43: Could we make this more generally into a slim package, in case we need to include additional dependencies in the near future?

Example: we're discovering [more gaps](https://snowflake.slack.com/archives/C08EG14LK41/p1757351539482579) in our WIF client-side story on Azure because we're rolling our own client library rather than using Microsoft's MSAL libraries. It would be nice to have the option to bundle MSAL later.

Having a `slim` package would give flexibility to include such dependencies, rather than a "noboto" package which only seems to be related to AWS boto.

src/snowflake/connector/wif_util.py
```diff
@@ -152,6 +158,12 @@ def create_aws_attestation(
 
     If the application isn't running on AWS or no credentials were found, raises an error.
     """
+    if not BOTO_AVAILABLE:
+        raise MissingDependencyError(
+            msg="boto3 or botocore dependency is not installed. [boto] extension is required for the AWS provider.",
```
sfc-gh-pmansour: Suggest a more user-facing and actionable error message:
```suggestion
            msg="AWS Workload Identity Federation is not supported in the slim snowflake-python-connect build. Please update your pip dependencies to use the full connector package instead.",
```

.github/workflows/check_installation.yml
sfc-gh-pczajka: We had a discussion and thought to keep more granular - in this example we would have two flags - "NO_BOTO" and "NO_MSAL". In case we bundled them in a "slim" package, client which would want to include MSAL only would be forced to install boto as well.
We also thought that NO_BOTO/NO_MSAL is more descriprive to the user rather than "slim", which "excludes some dependencies"

WDYTH?

sfc-gh-pmansour: Good point, being granular (e.g. adding specific `no_XX` labels) seems fine to give customers control over what they're installing, as long as the default has all the goodies we want to make broadly available. We should just make sure our messaging is that most people should install the full one, unless they have specific size restrictions and have a different authentication method planned.

In the future, if we end up adding more such "optional" libraries, perhaps we could also add a slim label too? Basically `slim=no_boto + no_msal + etc...`. That way, a size-constrained customers can simply use SLIM and assume it'll avoid large non-required dependencies. But we can cross that bridge once we get there.

sfc-gh-mmishchenko:
Did you consider a reuse of the existing approach for optional libraries implemented in `options.py`? I mean something like in the following draft:
https://github.com/snowflakedb/snowflake-connector-python/pull/2334/files#diff-43c77b79e1c2dc0869808edf380c998d22b3873760cfe40f65553f8f149a3d39
https://github.com/snowflakedb/snowflake-connector-python/pull/2334/files#diff-30b6bcf5e62718f3fad164ca9b81936925dcb03cf31c36418f9001d2581cc315

src/snowflake/connector/wif_util.py
sfc-gh-pczajka: Chenged to "AWS Workload Identity Federation can't be used because boto3 or botocore optional dependency is not installed. Try installing missing dependencies."

DESCRIPTION.md
```diff
@@ -12,6 +12,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Fixed `get_results_from_sfqid` when using `DictCursor` and executing multiple statements at once
   - Added the `oauth_credentials_in_body` parameter supporting an option to send the oauth client credentials in the request body
   - Fix retry behavior for `ECONNRESET` error
+  - Added an option to exclude `botocore` and `boto3` dependencies by setting `SNOWFLAKE_NO_BOTO` environment variable
```
sfc-gh-mmishchenko: It's probably obvious, but maybe worth wording it explicitly: "...environment variable during installation"

sfc-gh-mmishchenko (approved):
Looks good

DESCRIPTION.md
sfc-gh-pczajka: Added :+1:

PR #2531, opened 2025-09-11
Fix get_results_from_sfqid with DictCursor + multi statements

sfc-gh-bchinn:
Port PR: https://github.com/snowflakedb/Stored-Proc-Python-Connector/pull/225

I noticed that get_results_from_sfqid assumes that fetchall returns a tuple when that's not necessarily the case. Fixing it here + adding a test that fails without the fix.

```
E                   AssertionError: assert [{'multiple s...ccessfully.'}] == [{'1': 1}]
E                     At index 0 diff: {'multiple statement execution': 'Multiple statements executed successfully.'} != {'1': 1}
E                     Full diff:
E                     - [{'1': 1}]
E                     + [{'multiple statement execution': 'Multiple statements executed successfully.'}]
```

PR #2532, opened 2025-09-11
SNOW-2333702 Fix types for DictCursor

sfc-gh-bchinn:
Port PR: https://github.com/snowflakedb/Stored-Proc-Python-Connector/pull/225

On main, the following code shows the following mypy errors:

```
from snowflake.connector import connect, DictCursor

def main() -> None:
    conn = connect()

    c1 = conn.cursor()
    print([row[0] for row in c1.fetchall()])

    c2: DictCursor = conn.cursor(DictCursor)
    print([row["COL1"] for row in c2.fetchall()])
```

- `error: Incompatible types in assignment (expression has type "SnowflakeCursor", variable has type "DictCursor")  [assignment]`
- `error: No overload variant of "__getitem__" of "tuple" matches argument type "str"  [call-overload]`

This change fixes both errors.

### DIFF
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -432,6 +432,7 @@ class SnowflakeCursorBase(abc.ABC, Generic[FetchRow]):
     @property
     @abc.abstractmethod
     def _use_dict_result(self) -> bool: ...
+        """Decides whether results from helper functions are returned as a dict."""
 
     @property
     def description(self) -> list[ResultMetadata]:
--- a/src/snowflake/connector/cursor.py
+++ b/src/snowflake/connector/cursor.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 from __future__ import annotations
 
+import abc
 import collections
 import logging
 import os
@@ -19,6 +20,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    Generic,
     Iterator,
     Literal,
     NamedTuple,
@@ -86,6 +88,7 @@ if TYPE_CHECKING:  # pragma: no cover
     from .result_batch import ResultBatch
 
 T = TypeVar("T", bound=collections.abc.Sequence)
+FetchRow = TypeVar("FetchRow", bound=tuple[Any, ...] | dict[str, Any])
 
 logger = getLogger(__name__)
 
@@ -332,29 +335,7 @@ class ResultState(Enum):
     RESET = 3
 
 
-class SnowflakeCursor:
-    """Implementation of Cursor object that is returned from Connection.cursor() method.
-
-    Attributes:
-        description: A list of namedtuples about metadata for all columns.
-        rowcount: The number of records updated or selected. If not clear, -1 is returned.
-        rownumber: The current 0-based index of the cursor in the result set or None if the index cannot be
-            determined.
-        sfqid: Snowflake query id in UUID form. Include this in the problem report to the customer support.
-        sqlstate: Snowflake SQL State code.
-        timestamp_output_format: Snowflake timestamp_output_format for timestamps.
-        timestamp_ltz_output_format: Snowflake output format for LTZ timestamps.
-        timestamp_tz_output_format: Snowflake output format for TZ timestamps.
-        timestamp_ntz_output_format: Snowflake output format for NTZ timestamps.
-        date_output_format: Snowflake output format for dates.
-        time_output_format: Snowflake output format for times.
-        timezone: Snowflake timezone.
-        binary_output_format: Snowflake output format for binary fields.
-        arraysize: The default number of rows fetched by fetchmany.
-        connection: The connection object by which the cursor was created.
-        errorhandle: The class that handles error handling.
-        is_file_transfer: Whether, or not the current command is a put, or get.
-    """
+class SnowflakeCursorBase(abc.ABC, Generic[FetchRow]):
 
     # TODO:
     #    Most of these attributes have no reason to be properties, we could just store them in public variables.
@@ -1514,8 +1495,17 @@ class SnowflakeCursor:
 
         return self
 
-    def fetchone(self) -> dict | tuple | None:
-        """Fetches one row."""
+    @abc.abstractmethod
+    def fetchone(self) -> FetchRow:
+        pass
+
+    def _fetchone(self) -> dict[str, Any] | tuple[Any, ...] | None:
+        """
+        Fetches one row.
+
+        Returns a dict if self._use_dict_result is True, otherwise
+        returns tuple.
+        """
         if self._prefetch_hook is not None:
             self._prefetch_hook()
         if self._result is None and self._result_set is not None:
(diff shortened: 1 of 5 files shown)


sfc-gh-bchinn:
* **#2550** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2550?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2532** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2532?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2532?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2531** <a href="https://app.graphite.dev/github/pr/snowflakedb/snowflake-connector-python/2531?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/snowflake/connector/cursor.py
```diff
@@ -1925,14 +1907,51 @@ def _create_file_transfer_agent(
         )
 
 
-class DictCursor(SnowflakeCursor):
+class SnowflakeCursor(SnowflakeCursorBase[tuple[Any, ...]]):
```
sfc-gh-pczajka, line 1935: Is this separation necessary? I'm worried it might cause a breaking change

sfc-gh-bchinn: Yes, it is a breaking change, and unfortunately the separation is necessary, if we want accurate types. If `DictCursor` subclasses `SnowflakeCursor`, `fetchone` MUST have the same type, which is exactly what we don't want for DictCursor. It is a breaking change, but IMO the Python ecosystem is moving towards being more strongly typed, so this is the forward direction we should go.

sfc-gh-pczajka: After team discussion this does not break [PEP249](https://peps.python.org/pep-0249/#cursor-objects) and does not change driver's behavior, so does not qualify as BCR

src/snowflake/connector/cursor.py
```diff
@@ -1925,14 +1907,51 @@ def _create_file_transfer_agent(
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._use_dict_result = False
+
+    def fetchone(self) -> tuple[Any, ...] | None:
+        row = self._fetchone()
+        assert row is None or isinstance(row, tuple)
+        return row
```
sfc-gh-pczajka, line 1967:
```suggestion
    def fetchone(self) -> tuple[Any, ...] | None:
        return self._fetchone()
```
This assertion will result in unintuitive error for the users. IMO it does not bring much value, I'd remove it

src/snowflake/connector/cursor.py
```diff
@@ -1925,14 +1907,51 @@ def _create_file_transfer_agent(
...
-            use_dict_result=True,
-        )
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._use_dict_result = True
+
+    def fetchone(self) -> dict[str, Any] | None:
+        row = self._fetchone()
+        assert row is None or isinstance(row, dict)
+        return row
```
sfc-gh-pczajka, line 1981:
```suggestion
    def fetchone(self) -> dict[str, Any] | None:
        return self._fetchone()
```
Same here

src/snowflake/connector/cursor.py
```diff
@@ -1925,14 +1907,51 @@ def _create_file_transfer_agent(
...
+        binary_output_format: Snowflake output format for binary fields.
+        arraysize: The default number of rows fetched by fetchmany.
+        connection: The connection object by which the cursor was created.
+        errorhandle: The class that handles error handling.
+        is_file_transfer: Whether, or not the current command is a put, or get.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._use_dict_result = False
```
sfc-gh-pczajka: This field should be a part of `SnowflakeCursorBase` - other functions assume its existence. It's ok to pass it as fixed argument for "__init__"

src/snowflake/connector/cursor.py
```diff
@@ -86,6 +88,7 @@
     from .result_batch import ResultBatch
 
 T = TypeVar("T", bound=collections.abc.Sequence)
+FetchRow = TypeVar("FetchRow", bound=tuple[Any, ...] | dict[str, Any])
```
sfc-gh-pczajka: Some tests are failing
```
.tox/py39-sso-ci/lib/python3.9/site-packages/snowflake/connector/cursor.py:91: in <module>
    FetchRow = TypeVar("FetchRow", bound=tuple[Any, ...] | dict[str, Any])
E   TypeError: unsupported operand type(s) for |: 'types.GenericAlias' and 'types.GenericAlias'
```

src/snowflake/connector/cursor.py
```diff
@@ -332,29 +335,7 @@ class ResultState(Enum):
...
-        date_output_format: Snowflake output format for dates.
-        time_output_format: Snowflake output format for times.
-        timezone: Snowflake timezone.
-        binary_output_format: Snowflake output format for binary fields.
-        arraysize: The default number of rows fetched by fetchmany.
-        connection: The connection object by which the cursor was created.
-        errorhandle: The class that handles error handling.
-        is_file_transfer: Whether, or not the current command is a put, or get.
-    """
+class SnowflakeCursorBase(abc.ABC, Generic[FetchRow]):
```
sfc-gh-pczajka, line 341: Test `test_cursor_execute_timeout` started failing with AttributeError, please take a look

sfc-gh-bchinn: Oh the error is here:
```py
@patch("snowflake.connector.cursor.SnowflakeCursor._SnowflakeCursor__cancel_query")
```
Double underscore methods/attributes get mangled names, and in this case, `__cancel_query` is now defined in `SnowflakeCursorBase`, not `SnowflakeCursor`, so it mangles with `SnowflakeCursorBase`.

I'll fix this, but IMO we shouldn't be accessing mangled names directly...

src/snowflake/connector/cursor.py
sfc-gh-bchinn: `__init__` feels like the wrong place for this. But I'll update this to make sure subclasses define it

src/snowflake/connector/cursor.py
sfc-gh-bchinn: Removing it breaks typechecking. `_fetchone` returns `dict | tuple | None`, but this returns `tuple | None`. We know it's never going to return `dict` because `_use_dict_result = False`, but the typechecker doesn't know that, so we need to add an assert here.

I can change it to throw `TypeError` if we get an unexpected error, instead of `AssertionError`. But we need to add some logic to check this regardless

src/snowflake/connector/cursor.py
```diff
@@ -448,6 +429,10 @@ def __del__(self) -> None:  # pragma: no cover
             if logger.getEffectiveLevel() <= logging.INFO:
                 logger.info(e)
 
+    @property
+    @abc.abstractmethod
+    def _use_dict_result(self) -> bool: ...
```
sfc-gh-pczajka:
```suggestion
    def _use_dict_result(self) -> bool: ...
        """Decides whether results from helper functions are returned as a dict."""
```

PR #2365, opened 2025-06-20
Snow 2062305 process pool batch fetcher

sfc-gh-mmishchenko:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #NNNN

2. Fill out the following pre-review checklist:

   - [ ] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Please write a short description of how your code change solves the related issue.

4. (Optional) PR for stored-proc connector:

PR #2540, opened 2025-09-17
Cherrypicks to aio connector part17

sfc-gh-pczajka:
Cherry picks to aio-connector:

* #2329
  * apply all changes (most Oauth code is iherited from sync version)

----

**Commits applied without changes:**
* #2355 - PAT with external session skipped until GA - https://snowflakecomputing.atlassian.net/browse/SNOW-2344581
* 46b4141f64c37c2eca3685e25cc2a12a12a4f7ad - trivia change, inherited
* #2365 - process pools not supported in async
* f442062d4d4d3568e36110cbb71673c4a68ccac6 - temporarily disable tests
* #2382 - GPG file update

---

* #2417
  * apply all changes

### DIFF
--- a/src/snowflake/connector/aio/_result_set.py
+++ b/src/snowflake/connector/aio/_result_set.py
@@ -174,7 +174,12 @@ class ResultSet(ResultSetSync):
         result_chunks: list[JSONResultBatch] | list[ArrowResultBatch],
         prefetch_thread_num: int,
     ) -> None:
-        super().__init__(cursor, result_chunks, prefetch_thread_num)
+        super().__init__(
+            cursor,
+            result_chunks,
+            prefetch_thread_num,
+            use_mp=False,  # async code depends on aio rather than multiprocessing
+        )
         self.batches = cast(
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/connection.py
```diff
@@ -1271,6 +1276,15 @@ def __open_connection(self):
                 )
             elif self._authenticator == PROGRAMMATIC_ACCESS_TOKEN:
                 self.auth_class = AuthByPAT(self._token)
+            elif self._authenticator == PAT_WITH_EXTERNAL_SESSION:
```
sfc-gh-fpawlowski, line 1279: nit: This may be a low hanging fruit for the future - to at least fallback to synch code.

sfc-gh-fpawlowski, line 1279: Also user will now end up using AuthByOkta (in else block) since there is no match for PAT with external session in async. Do we want better info, or to stay with this one?

sfc-gh-pczajka: https://snowflakecomputing.atlassian.net/browse/SNOW-2363688

PR #2549, opened 2025-09-23
Use python3 for pre-commit hooks

sfc-gh-bchinn:
MacOS no longer provides `/usr/bin/python`, so use `python3` instead.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,7 +51,7 @@ repos:
     hooks:
     -   id: check-no-native-http
         name: Check for native HTTP calls
-        entry: python ci/pre-commit/check_no_native_http.py
+        entry: python3 ci/pre-commit/check_no_native_http.py
         language: system
         files: ^src/snowflake/connector/.*\.py$
         exclude: |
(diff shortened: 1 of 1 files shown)


sfc-gh-bchinn:
* **#2549** <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-connector-python/2549?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snowflakedb/snowflake-connector-python/2549?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

sfc-gh-turbaszek:
Have you had issues with using `python`? I think for development (that's where precommits are used) people should have dedicated, preferably isolated virtual environments. Given that I'm not sure if using python3 is required. I would even consider this as misleading guidance that it's ok to run precommits just on system python

sfc-gh-bchinn:
@sfc-gh-turbaszek It looks like `snowflake-connector-python` only supports 3.9+ anyway, so if development is only expected to occur in virtual environments, what's the harm in switching to `python3`?

While I agree that development should generally occur in a virtual environment,
1. The README doesn't mention anything about creating a virtual environment
2. It's possible to use a virtual environment for development without activating it in your shell

sfc-gh-turbaszek (approved):

PR #2567, opened 2025-10-03
Snow 2388762 crl post review fixes

sfc-gh-pczajka:

   Fixes #NNNN

### DIFF
--- a/src/snowflake/connector/crl.py
+++ b/src/snowflake/connector/crl.py
@@ -54,7 +54,7 @@ class CRLConfig:
     )
     allow_certificates_without_crl_url: bool = False
     connection_timeout_ms: int = 3000
-    read_timeout_ms: int = 3000
+    read_timeout_ms: int = 90000  # 90s
     cache_validity_time: timedelta = timedelta(hours=24)
     enable_crl_cache: bool = True
     enable_crl_file_cache: bool = True
@@ -276,9 +276,7 @@ class CRLValidator:
 
         if certificate_chains is None or len(certificate_chains) == 0:
             logger.warning("Certificate chains are empty")
-            if self._cert_revocation_check_mode == CertRevocationCheckMode.ADVISORY:
-                return True
-            return False
+            return self._cert_revocation_check_mode == CertRevocationCheckMode.ADVISORY
 
         results = []
         for chain in certificate_chains:
@@ -549,25 +547,16 @@ class CRLValidator:
         Returns:
             List of certificate chains, where each chain is a list of x509.Certificate objects
         """
-        from OpenSSL.crypto import FILETYPE_ASN1, dump_certificate
-
         try:
-            cert_chain = connection.get_peer_cert_chain()
+            # Convert OpenSSL certificates to cryptography x509 certificates
+            cert_chain = connection.get_peer_cert_chain(as_cryptography=True)
             if not cert_chain:
                 logger.debug("No certificate chain found in connection")
                 return []
-
-            # Convert OpenSSL certificates to cryptography x509 certificates
-            x509_chain = []
-            for cert_openssl in cert_chain:
-                cert_der = dump_certificate(FILETYPE_ASN1, cert_openssl)
-                cert_x509 = x509.load_der_x509_certificate(cert_der, default_backend())
-                x509_chain.append(cert_x509)
-
             logger.debug(
-                "Extracted %d certificates for CRL validation", len(x509_chain)
+                "Extracted %d certificates for CRL validation", len(cert_chain)
             )
-            return [x509_chain]  # Return as a single chain
+            return [cert_chain]  # Return as a single chain
 
         except Exception as e:
             logger.warning(
--- a/src/snowflake/connector/crl.py
+++ b/src/snowflake/connector/crl.py
@@ -345,14 +345,32 @@ class CRLValidator:
 
     @staticmethod
     def _is_short_lived_certificate(cert: x509.Certificate) -> bool:
-        """Check if certificate is short-lived (validity <= 5 days)"""
+        """Check if certificate is short-lived according to CA/Browser Forum definition:
+        - For certificates issued on or after 15 March 2024 and prior to 15 March 2026:
+          validity period <= 10 days (864,000 seconds)
+        - For certificates issued on or after 15 March 2026:
+          validity period <= 7 days (604,800 seconds)
+        """
         try:
             # Use timezone.utc versions to avoid deprecation warnings
+            issue_date = cert.not_valid_before_utc
             validity_period = cert.not_valid_after_utc - cert.not_valid_before_utc
         except AttributeError:
             # Fallback for older versions
+            issue_date = cert.not_valid_before
             validity_period = cert.not_valid_after - cert.not_valid_before
-        return validity_period.days <= 5
+
+        # Convert issue_date to UTC if it's not timezone-aware
+        if issue_date.tzinfo is None:
+            issue_date = issue_date.replace(tzinfo=timezone.utc)
+
+        march_15_2024 = datetime(2024, 3, 15, tzinfo=timezone.utc)
+        march_15_2026 = datetime(2026, 3, 15, tzinfo=timezone.utc)
+        if issue_date >= march_15_2026:
+            return validity_period.total_seconds() <= 604800  # 7 days in seconds
+        if issue_date >= march_15_2024:
+            return validity_period.total_seconds() <= 864000  # 10 days in seconds
+        return False
 
     @staticmethod
     def _extract_crl_distribution_points(cert: x509.Certificate) -> list[str]:
--- a/test/unit/test_crl.py
+++ b/test/unit/test_crl.py
@@ -1495,3 +1495,67 @@ def test_crl_signature_verification_with_issuer_mismatch_warning(
         assert (
             warning_found
         ), f"Expected warning about CRL issuer mismatch not found in logs: {[r.message for r in caplog.records]}"
+
+
+@pytest.mark.parametrize(
+    "issue_date,validity_days,expected",
+    [
+        (
+            # Issued before March 15, 2024 - not short-lived
+            datetime(2023, 3, 14, tzinfo=timezone.utc),
+            3,
+            False,
+        ),
+        (
+            # Issued on March 15, 2024, should use 10-day rule
+            datetime(2024, 3, 15, tzinfo=timezone.utc),
+            10,
+            True,
+        ),
+        (
+            # Issued on March 15, 2024, should use 10-day rule
+            datetime(2024, 3, 15, tzinfo=timezone.utc),
+            11,
+            False,
+        ),
+        (
+            # Issued on March 15, 2024, should use 10-day rule
+            datetime(2024, 3, 15),
+            10,
+            True,
+        ),
+        (
+            # Issued on March 15, 2024, should use 10-day rule
+            datetime(2024, 3, 15),
+            11,
+            False,
+        ),
+        (
+            # Issued on March 15, 2026, should use 7-day rule
+            datetime(2026, 3, 15, tzinfo=timezone.utc),
+            7,
+            True,
+        ),
+        (
+            # Issued on March 15, 2026, should use 7-day rule
+            datetime(2026, 3, 15, tzinfo=timezone.utc),
+            8,
+            False,
+        ),
+        (
+            # Issued on March 15, 2026, should use 7-day rule
+            datetime(2026, 3, 15),
+            7,
+            True,
+        ),
+        (
+            # Issued on March 15, 2026, should use 7-day rule
+            datetime(2026, 3, 15),
+            8,
+            False,
+        ),
+    ],
+)
+def test_is_short_lived_certificate(cert_gen, issue_date, validity_days, expected):
+    cert = cert_gen.create_short_lived_certificate(validity_days, issue_date)
+    assert CRLValidator._is_short_lived_certificate(cert) == expected


src/snowflake/connector/crl.py
```diff
@@ -347,14 +345,32 @@ def _validate_certificate(
...
+          validity period <= 7 days (604,800 seconds)
+        """
         try:
             # Use timezone.utc versions to avoid deprecation warnings
+            issue_date = cert.not_valid_before_utc
             validity_period = cert.not_valid_after_utc - cert.not_valid_before_utc
         except AttributeError:
             # Fallback for older versions
+            issue_date = cert.not_valid_before
             validity_period = cert.not_valid_after - cert.not_valid_before
```
sfc-gh-jkasten, line 484: nit: The not_valid_after field is inclusive. This will calculate that the lifetime is 1 second shorter than it actually is. I think this falls on the safe side of the equation down below, (it will treat certificates as short-lived which are technically not).

[RFC 5280 Section 4.1.2.5 text](https://datatracker.ietf.org/doc/html/rfc5280#:~:text=The%20validity%20period%20for%20a%20certificate%20is%20the%20period%20of%20time%20from%0A%20%20%20notBefore%20through%20notAfter%2C%20inclusive)

The other thing you have to worry here is an underlying library trying to include leap seconds on you.
It is common in the WebPKI to give a little buffer and to stay away from the edges[[design doc](https://docs.google.com/document/d/1Dy4eb8dDV75XIUKBvOMfeZjnxYca6ICKsCtgWwfU9Fs/edit?tab=t.0#heading=h.dm08p29sbd9q)]. Competent CAs will always give a lot of breathing room away from the edges. 

Edit: There is also some complexity around the definition of "issued_at". Most certificates are actually backdated and so browsers use the earliest SCT contained within the certificate for their definition. We should ignore that complexity though and maybe just give a bit more breathing room.

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
-        return CRLValidationResult.UNREVOKED
+    def _is_certificate_trusted_by_os(self, cert: x509.Certificate) -> bool:
+        if trusted_cert := self._trusted_ca.get(cert.subject):
+            # Compare DER-encoded forms of certificates
+            cert_der = cert.public_bytes(serialization.Encoding.DER)
+            trusted_cert_der = trusted_cert.public_bytes(serialization.Encoding.DER)
+            return cert_der == trusted_cert_der
+        return False
+
+    def _validate_certificate_with_cache(
```
sfc-gh-pczajka: `@functools.cache` is disabled in object methods by precommit due to possible memory leaks - that's why

src/snowflake/connector/crl.py
```diff
@@ -343,14 +436,32 @@ def _validate_certificate(
...
+        if issue_date.tzinfo is None:
+            issue_date = issue_date.replace(tzinfo=timezone.utc)
+
+        march_15_2024 = datetime(2024, 3, 15, tzinfo=timezone.utc)
+        march_15_2026 = datetime(2026, 3, 15, tzinfo=timezone.utc)
+        if issue_date >= march_15_2026:
+            return validity_period.total_seconds() <= 604800  # 7 days in seconds
+        if issue_date >= march_15_2024:
+            return validity_period.total_seconds() <= 864000  # 10 days in seconds
+        return False
```
sfc-gh-jkasten: optional: The max certificate lifetime by the baseline requirements is 398 days. (So should never see a trusted certificate hit the < case. (Certificates that are shorter than 10 days would long be expired)
```suggestion
        return validity_period.total_seconds() <= 864000  # 10 days in seconds
```

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
-                return CRLValidationResult.REVOKED
-            results.append(result)
 
-        if CRLValidationResult.ERROR in results:
+        subject_certificates: dict[x509.Name, list[x509.Certificate]] = defaultdict(
+            list
+        )
+        for cert in chain:
+            subject_certificates[cert.subject].append(cert)
+        is_being_visited: set[x509.Name] = set()
```
sfc-gh-jkasten: no action needed: By the spec, certificate chains are supposed to be created by the `(<issuer>, <authority key identifier (AKID)>)` matching the `(<subject>, <subject key identifier (SKID)>)` of the parent.

Historically, different CAs shared keys between multiple CA certificates. Today there are sometimes multiple CA certificates which share the same subject and key.

I think this is fine, for the space that we are going to be working in. I don't remember any rules preventing the reuse of names today with different keys, but it would certainly be atypical and frowned upon.

My understanding is that the worst thing that would happen is that the wrong CA cert is used to try to verify the certificate, which seems reasonable given the risk.

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
+                    # ignore invalid path result
+                    continue
+                if ca_result == CRLValidationResult.UNREVOKED:
+                    # good path found
+                    return self._validate_certificate_with_cache(cert, ca_cert)
+                valid_results.append((ca_result, ca_cert))
+
+            if len(valid_results) == 0:
+                # "root" certificate not cought by "is_trusted_by_os" check
+                logger.warning("Root certificate not trusted by OS: %s", cert.subject)
```
sfc-gh-jkasten: Would `no path found towards trust anchor` be more accurate?
I think this case hits for any random intermediate that doesn't have a path forward? 

We might consider adding in the cert.issuer as well (and perhaps serial if it is easy)

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
+                        return CRLValidationResult.REVOKED
+                    return CRLValidationResult.ERROR
+
+            # no ERROR result found, all paths are REVOKED
+            return CRLValidationResult.REVOKED
+
+        is_being_visited.add(chain[0].subject)
+        error_result = False
+        revoked_result = False
+        for cert in subject_certificates[chain[0].subject]:
```
sfc-gh-jkasten, line 393: My understanding is that this is allowing for multiple leaf certificates in the chain. (This is outside the spec and fine in theory.)

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
     def _validate_single_chain(
```
sfc-gh-jkasten, line 310: Is _validate_certificate_list more accurate? (Many chains are being checked. `certificate_list` is the name of the field in the TLS 1.3 RFC.)

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
-            result = self._validate_certificate(chain[i], chain[i + 1])
-            if result == CRLValidationResult.REVOKED:
-                return CRLValidationResult.REVOKED
-            results.append(result)
 
-        if CRLValidationResult.ERROR in results:
+        subject_certificates: dict[x509.Name, list[x509.Certificate]] = defaultdict(
+            list
+        )
+        for cert in chain:
```
sfc-gh-jkasten, line 327: nit: Perhaps add a quick check on the chain size? It would be a weird attack, but I get a bit more concerned as we look towards PQ.

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
+            # REVOKED - all paths are revoked
+            # ERROR - some certificates on potentially unrevoked paths can't be verified
+            # None - ignore this path (cycle detected)
+            if self._is_certificate_trusted_by_os(cert):
+                # found a trusted certificate
+                return CRLValidationResult.UNREVOKED
+            if cert.issuer in self._trusted_ca:
+                # issuer is trusted by OS
+                return self._validate_certificate_with_cache(
+                    cert, self._trusted_ca[cert.issuer]
```
sfc-gh-jkasten: I am concerned that since we aren't actually checking the signature of the chain in all cases it opens us up to a subtle attack.

Send
1. revoked, but legitimately signed leaf for `example.com`
2. valid intermediate that signed the revoked leaf and chains to root
3. bogus leaf certificate for `example.com` with 5 day lifetime that says it is signed by a root with the same name as a legitimate root. (It can be signed by any key the attacker has)

The first two certificates make the chain pass normal certificate validation. 
When revocation is checked, it ignores the revoked leaf certificate as L378 will continue to check all certificates with the same name. The code will say that the bogus certificate is not revoked and validation passes. The CRL for the bogus certificate is never checked due to the short lifetime. The code will say that the issuer is "trusted" because the name matches a real root.

sfc-gh-jkasten (commented):
Thanks, Patryk!

src/snowflake/connector/crl.py
sfc-gh-jkasten: I think we need to always check the signature of the certificate against the issuer even if we can skip the revocation checks?

The certificates that we are given are not verified and are just what the client gives us [[SSL_get_peer_cert_chain documentation](https://docs.openssl.org/1.1.1/man3/SSL_get_peer_cert_chain/#synopsis)]

sfc-gh-jkasten: I am adding the rest of the @snowflakedb/pki-oversight team to the review. They work more European-friendly hours.

My understanding is that the drivers team wants to land this feature today and I am concerned about this (possible) bug is a blocker. It is deterministic with no additional privileges needed other than to MITM the connection?

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,101 @@ def validate_certificate_chains(
...
+                )
+            if cert.issuer in is_being_visited:
+                # cycle detected - invalid path
+                return None
+
+            valid_results: list[tuple[CRLValidationResult, x509.Certificate]] = []
+            for ca_cert in subject_certificates[cert.issuer]:
```
sfc-gh-jkasten, line 351: We need to make sure that the "issuer" is an actual issuer (basicConstraints == true).

Consider an attacker sending 
1. Revoked `example.com`
2. Legit intermediate
3. `example.com` signed by a leaf certificate `jdkasten.com`
4. `jdkasten.com` 
5. `jdkasten.com`'s intermediate.

All of the signatures would work out. L378 means that revocation would also work. We need to verify that all CAs are actually CAs. (pathLen is probably not necessary to check)

This attack is related to the [lack of signature attack](https://github.com/snowflakedb/snowflake-connector-python/pull/2567/files#r2412125139). I hope there is a library that will allow you to check both the signature and the CA properties with the same code.

sfc-gh-jkasten: cc @snowflakedb/pki-oversight

src/snowflake/connector/crl.py
sfc-gh-pczajka: It's called "chain" all over the code base (including openSSL and other drivers) - I'd leave it like this.

src/snowflake/connector/crl.py
sfc-gh-pczajka: That's right - just to be safe

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed, I added signature verification logic to traverse algorithm (if there is a siganture mismath the edge is ignored)

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed :)

src/snowflake/connector/crl.py
```diff
@@ -190,10 +193,21 @@ def __init__(
         self._read_timeout_ms = read_timeout_ms
         self._cache_validity_time = cache_validity_time
         self._cache_manager = cache_manager or CRLCacheManager.noop()
+        self._validate_certificate_cache: dict[
```
sfc-gh-fpawlowski: We may want to avoid using verbs in structures' names. If its not callable maybe names like __validated_certifacates_cache_ or __validation_results_cache_

src/snowflake/connector/crl.py
```diff
@@ -294,22 +308,125 @@ def validate_certificate_chains(
...
-                return CRLValidationResult.REVOKED
-            results.append(result)
 
-        if CRLValidationResult.ERROR in results:
+        subject_certificates: dict[x509.Name, list[x509.Certificate]] = defaultdict(
+            list
+        )
+        for cert in chain:
+            subject_certificates[cert.subject].append(cert)
+        is_being_visited: set[x509.Name] = set()
```
sfc-gh-fpawlowski: As discussed in office - we may want to change name to containnig noun like "certs..."

src/snowflake/connector/crl.py
sfc-gh-jkasten: I am so sorry. I think the function you used does not check that the CA certificate is an actual CA certificate? 

Here is the documentation for the function.
https://cryptography.io/en/42.0.2/x509/reference/#cryptography.x509.Certificate.verify_directly_issued_by

Here is the code.
https://github.com/pyca/cryptography/blob/fe8438c62eb470c976114b75581b8c7ea3e9e854/src/rust/src/x509/certificate.rs#L321

Let me know if I completely missed the control. (If so, sorry)

src/snowflake/connector/crl.py
```diff
@@ -294,24 +310,133 @@ def validate_certificate_chains(
...
 
-        if CRLValidationResult.ERROR in results:
+        subject_certificates: dict[x509.Name, list[x509.Certificate]] = defaultdict(
+            list
+        )
+        for cert in chain:
+            subject_certificates[cert.subject].append(cert)
+        currently_visited_subjects: set[x509.Name] = set()
+
+        def traverse_chain(cert: x509.Certificate) -> CRLValidationResult | None:
```
sfc-gh-jkasten, line 331: Sorry - another revocation bypass. We need to fully validate the certificate. In this additional case, the expiration time of the certificate is not being checked.

1. Revoked `target.com`
2. Legit intermediate signing `target.com` that chains to root
3. Expired `target.com` certificate (can be easily retrieved via crt.sh, etc.)
4. Expired `target.com` cert's legit intermediate that chains to root.

1 + 2 satisfies normal certificate validation. 3 + 4 satisfies the revocation check (It is extremely likely the CRL endpoint is still in use)
Sorry for not calling out this additional detail last night.

src/snowflake/connector/crl.py
sfc-gh-jkasten: After thinking about this more, I think you would need to make sure that the public key contained within each of the leaf certificates is equivalent at minimum.

What we are missing is the ability to tie the leaf certificate to the TLS handshake.

In essence consider the attack...

1. Revoked target.com
2. Legit intermediate signing target.com that chains to root
3. New target.com certificate you pulled from CT/crt.sh - it is completely valid, but you don't have the key.
4. New target.com's intermediate

(1 + 2, get past the normal certificate verification. 3+4 bypass revocation checks)

In general, this is outside of the TLS spec and would require that TLS servers get multiple certificates from trusted CAs. I suspect that this capability is not worth the additional trouble. It contributes to the other attacks I listed, but its removal does not prevent them.

sfc-gh-jkasten:
I tried to capture the perceived vulnerabilities in [SNOW-2401045](https://snowflakecomputing.atlassian.net/browse/SNOW-2401045)

PR #2584, opened 2025-10-20
Snow 2401045 fix vulnerabilities in crl

sfc-gh-pczajka:
Fix CRL implementation vulnerabilities described in SNOW-2401045
1. Remove possibility of checking multiple leaf certificates - made sense as SSL connection returns only one peer certificate
2. check `basicConstraints isCA = true` for all certificates except a leaf one
3. validates certificate expritation time
4. check CRL IssuingDistributionPoint against the url
5. refactor validate_connection - remove logic for multiple chains

### DIFF
--- a/test/integ/test_cursor_binding.py
+++ b/test/integ/test_cursor_binding.py
@@ -42,21 +42,38 @@ def test_binding_security(conn_cnx, db_parameters):
 
             # SQL injection safe test
             # Good Example
-            with pytest.raises(ProgrammingError):
-                cnx.cursor().execute(
-                    "SELECT * FROM {name} WHERE aa=%s".format(
-                        name=db_parameters["name"]
-                    ),
-                    ("1 or aa>0",),
+            # server behavior change: this no longer raises an error, but returns an empty result set
+            try:
+                res = (
+                    cnx.cursor()
+                    .execute(
+                        "SELECT * FROM {name} WHERE aa=%s".format(
+                            name=db_parameters["name"]
+                        ),
+                        ("1 or aa>0",),
+                    )
+                    .fetchall()
                 )
+                assert res == []
+            except ProgrammingError:
+                # old server behavior: OK
+                pass
 
-            with pytest.raises(ProgrammingError):
-                cnx.cursor().execute(
-                    "SELECT * FROM {name} WHERE aa=%(aa)s".format(
-                        name=db_parameters["name"]
-                    ),
-                    {"aa": "1 or aa>0"},
+            try:
+                res = (
+                    cnx.cursor()
+                    .execute(
+                        "SELECT * FROM {name} WHERE aa=%(aa)s".format(
+                            name=db_parameters["name"]
+                        ),
+                        {"aa": "1 or aa>0"},
+                    )
+                    .fetchall()
                 )
+                assert res == []
+            except ProgrammingError:
+                # old server behavior: OK
+                pass
 
             # Bad Example in application. DON'T DO THIS
             c = cnx.cursor()


src/snowflake/connector/crl.py
```diff
@@ -270,61 +270,63 @@ def from_config(
...
-            results.append(result)
+        if chain is None:
+            chain = []
+        result = self._validate_chain(peer_cert, chain)
 
-        # In non-advisory mode we require at least one chain get a clear UNREVOKED status
-        if self._cert_revocation_check_mode != CertRevocationCheckMode.ADVISORY:
+        if result == CRLValidationResult.UNREVOKED:
+            return True
+        elif result == CRLValidationResult.REVOKED:
```
sfc-gh-turbaszek:
```suggestion
        if result == CRLValidationResult.REVOKED:
```

src/snowflake/connector/crl.py
```diff
@@ -270,61 +270,63 @@ def from_config(
...
-
-        results = []
-        for chain in certificate_chains:
-            result = self._validate_single_chain(chain)
-            # If any of the chains is valid, the whole check is considered positive
-            if result == CRLValidationResult.UNREVOKED:
-                return True
-            results.append(result)
+        if chain is None:
+            chain = []
```
sfc-gh-turbaszek:
```suggestion
        chain = chain if chain is not None else []
```
Personal preference, but ternary may be simpler

src/snowflake/connector/crl.py
```diff
@@ -270,61 +270,62 @@ def from_config(
...
         Returns:
           UNREVOKED: If there is a path to any trusted certificate where all certificates are unrevoked.
           REVOKED: If all paths to trusted certificates are revoked.
           ERROR: If there is a path to any trusted certificate on which none certificate is revoked,
              but some certificates can't be verified.
         """
-        # An empty chain is considered an error
-        if len(chain) == 0:
+        # Check if start certificate is expired
+        if not self._is_valid(start_cert):
```
sfc-gh-jkasten: nit/optional: I would be tempted to move away from `is_valid` which generally means more than simply having correct dates. The documentation is very helpful regardless.

Maybe `_is_within_lifetime` or `_is_within_validity_dates`?

src/snowflake/connector/crl.py
```diff
@@ -270,61 +270,62 @@ def from_config(
...
+            )
             return CRLValidationResult.ERROR
 
         subject_certificates: dict[x509.Name, list[x509.Certificate]] = defaultdict(
             list
         )
         for cert in chain:
+            if not self._is_ca_certificate(cert):
+                logger.warning("Ignoring non-CA certificate: %s", cert)
+                continue
```
sfc-gh-jkasten, line 328: We should also be checking the lifetimes of the intermediates. 

I realize we are almost certainly not going to get attacked by an expired leaked intermediate, but it should be cheap to add and would be closer to spec.

src/snowflake/connector/crl.py
```diff
@@ -474,18 +508,8 @@ def _is_short_lived_certificate(cert: x509.Certificate) -> bool:
...
-        except AttributeError:
-            # Fallback for older versions
-            issue_date = cert.not_valid_before
-            validity_period = cert.not_valid_after - cert.not_valid_before
-
-        # Convert issue_date to UTC if it's not timezone-aware
-        if issue_date.tzinfo is None:
-            issue_date = issue_date.replace(tzinfo=timezone.utc)
+        issue_date, expiry_date = CRLValidator._get_certificate_validity_dates(cert)
+        validity_period = expiry_date - issue_date
```
sfc-gh-jkasten: nit: the notAfter date is inclusive so this calculation is off by one [[text from RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#:~:text=The%20validity%20period%20for%20a%20certificate%20is%20the%20period%20of%20time%20from%0A%20%20%20notBefore%20through%20notAfter%2C%20inclusive)]. I think your code falls on the safe side and CAs worth their salt say away from the edges, but we should probably fix it to show competence.

sfc-gh-jkasten (approved):
Awesome, thanks @sfc-gh-pczajka!

src/snowflake/connector/crl.py
sfc-gh-pczajka: Nice, changed

src/snowflake/connector/crl.py
sfc-gh-pczajka: Thanks for the catch! I've overlooked this - fixed the check and tests

src/snowflake/connector/crl.py
sfc-gh-pczajka: Fixed

PR #2585, opened 2025-10-21
SNOW-1763096: Add async telemetry support

sfc-gh-turbaszek:
Adds support for async telemetry sending. This change does not iclude support for async errorhandler.

### DIFF
--- a/src/snowflake/connector/errors.py
+++ b/src/snowflake/connector/errors.py
@@ -152,8 +152,6 @@ class Error(Exception):
         telemetry_data: dict[str, Any],
     ) -> None:
         """Send telemetry data by in-band telemetry if it is enabled, otherwise send through out-of-band telemetry."""
-        from .aio._connection import SnowflakeConnection as AsyncSnowflakeConnection
-
         if (
             connection is not None
             and connection.telemetry_enabled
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/errors.py
```diff
@@ -161,19 +165,27 @@ def send_exception_telemetry(
             telemetry_data[TelemetryField.KEY_EXCEPTION.value] = self.__class__.__name__
```
sfc-gh-turbaszek, line 163: @sfc-gh-fpawlowski @sfc-gh-pczajka should we add additional information about aio usage? At this point we can check class of connection to determine if aio is being used or not. WDYT?

sfc-gh-turbaszek:

src/snowflake/connector/errors.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 from __future__ import annotations
 
+import asyncio
```
sfc-gh-pczajka: asyncio is an optional dependency, this might result in import error if "[aio]" extension is not provided in installation

src/snowflake/connector/errors.py
sfc-gh-pczajka: Sounds good to me!

test/unit/test_errors_telemetry.py
```diff
@@ -0,0 +1,52 @@
...
+
+    msg = _extract_message_from_log_call(conn)
+    assert msg[TelemetryField.KEY_TYPE.value] == TelemetryField.SQL_EXCEPTION.value
+    assert msg[TelemetryField.KEY_SOURCE.value] == conn.application
+    assert msg[TelemetryField.KEY_EXCEPTION.value] == "Error"
+    assert TelemetryField.KEY_DRIVER_TYPE.value in msg
+    assert TelemetryField.KEY_DRIVER_VERSION.value in msg
+
+
+async def test_error_telemetry_async_connection():
```
sfc-gh-pczajka: async tests are mirrired in test/unit/aio directory, could you move it there?

sfc-gh-pczajka (approved):

PR #2593, opened 2025-10-23
NO-SNOW: Cherry-picks to aio branch

sfc-gh-turbaszek:
All commits cherry picked. Changes to aio code are limited to:
[Fix get_results_from_sfqid with DictCursor + multi statements](https://github.com/snowflakedb/snowflake-connector-python/pull/2593/commits/fac1b37eb4a7ce8cf57d0e6e2b4240855d1c9d81)

### DIFF
--- a/test/integ/aio_it/test_dbapi_async.py
+++ b/test/integ/aio_it/test_dbapi_async.py
@@ -881,7 +881,8 @@ async def test_callproc_invalid(conn_cnx):
             # stored procedure does not exist
             with pytest.raises(errors.ProgrammingError) as pe:
                 await cur.callproc(name_sp)
-            assert pe.value.errno == 2140
+            # this value might differ between Snowflake environments
+            assert pe.value.errno in [2140, 2139]
 
             await cur.execute(
                 f"""
--- a/test/integ/aio_it/test_load_unload_async.py
+++ b/test/integ/aio_it/test_load_unload_async.py
@@ -39,7 +39,6 @@ def s3_test_data(request, conn_cnx, db_parameters):
         return conn_cnx(
             user=db_parameters["user"],
             account=db_parameters["account"],
-            password=db_parameters["password"],
         )
 
     return create_test_data(request, db_parameters, connection)


test/integ/test_dbapi.py
```diff
@@ -857,7 +857,8 @@ def test_callproc_invalid(conn_cnx):
             # stored procedure does not exist
             with pytest.raises(errors.ProgrammingError) as pe:
                 cur.callproc(name_sp)
-            assert pe.value.errno == 2140
```
sfc-gh-fpawlowski, line 860: We still have test_callproc_invalid in test_dbapi_async.py

Do we want to apply this fix there as well in case issues reoccur in the future?

test/integ/test_load_unload.py
```diff
@@ -35,7 +35,6 @@ def connection():
         return conn_cnx(
             user=db_parameters["user"],
             account=db_parameters["account"],
-            password=db_parameters["password"],
```
sfc-gh-fpawlowski, line 38: This change was not applied to async code - it still passes password param.

In general I can see we have quite many places still using password parameter (maybe lets create Jira for that). While for some tests it makes sense, we may want to investigate it further, since key-pair is now the main approach for integ tests.

test/integ/pandas_it/test_pandas_tools.py
```diff
@@ -972,15 +972,21 @@ def test_all_pandas_types(
     with conn_cnx() as cnx:
         try:
             success, nchunks, nrows, _ = write_pandas(
-                cnx, df, table_name, quote_identifiers=True, auto_create_table=True
+                cnx,
+                df,
+                table_name,
+                quote_identifiers=True,
+                auto_create_table=True,
+                use_logical_type=True,
```
sfc-gh-fpawlowski: Since this change was applied in separate [PR#2605](https://github.com/snowflakedb/snowflake-connector-python/pull/2605) we may want to remove this commit from this branch to keep the rebased dev/aio-connector history clean after the rebase-merge.

PR #2607, opened 2025-10-28
NO-SNOW: Send telemetry when closing connection

sfc-gh-turbaszek:
The following code currently won't emit any telemetry:
```py
conn = connect(connection_name="py")
r = conn.execute_string("select 2;")
print(r[0].fetchall())
```
on the other hand the following code works (sends telemetry)
```py
with connect(connection_name="py") as conn:
    r = conn.execute_string("select 2;")
    print(r[0].fetchall())
```

The difference is that context manager handles connection closing differently than handler registered `atexit`. The handler was configured to basically not send telemetry, causing potential loss of observability.  

However, the handler is already doing additional REST calls (in majority of cases):
https://github.com/snowflakedb/snowflake-connector-python/blob/9c9be3659b9c1b6f213f1cf73293ec1334bf925e/src/snowflake/connector/connection.py#L1166

which under the hood does:
https://github.com/snowflakedb/snowflake-connector-python/blob/9c9be3659b9c1b6f213f1cf73293ec1334bf925e/src/snowflake/connector/network.py#L660-L667

This PR proposes to adopt similar approach (send but not retry) in telemetry close handler.

### DIFF
--- a/test/unit/test_telemetry.py
+++ b/test/unit/test_telemetry.py
@@ -157,6 +157,30 @@ def test_telemetry_send_batch_disabled():
     assert rest_call.call_count == 0
 
 
+def test_telemetry_send_batch_with_retry_flag():
+    """Tests that send_batch respects the retry parameter."""
+    client, rest_call = get_client_and_mock()
+
+    client.add_log_to_batch(snowflake.connector.telemetry.TelemetryData({}, 2000))
+
+    # Test with retry=True
+    client.send_batch(retry=True)
+
+    assert rest_call.call_count == 1
+    # Verify _no_retry parameter is False when retry=True
+    call_kwargs = rest_call.call_args[1]
+    assert call_kwargs["_no_retry"] is False
+
+    # Add another log and test with retry=False (default)
+    client.add_log_to_batch(snowflake.connector.telemetry.TelemetryData({}, 3000))
+    client.send_batch(retry=False)
+
+    assert rest_call.call_count == 2
+    # Verify _no_retry parameter is True when retry=False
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/telemetry.py
```diff
@@ -155,7 +155,7 @@ def try_add_log_to_batch(self, telemetry_data: TelemetryData) -> None:
         except Exception:
             logger.warning("Failed to add log to telemetry.", exc_info=True)
 
-    def send_batch(self) -> None:
+    def send_batch(self, retry: bool = False) -> None:
```
sfc-gh-fpawlowski, line 158: Do we want to add tests - to make sure no additional requests are done if retry is False - e.g. using Wiremock or simpler mocks?

src/snowflake/connector/telemetry.py
```diff
@@ -186,6 +186,7 @@ def send_batch(self) -> None:
                 method="post",
                 client=None,
                 timeout=5,
+                _no_retry=not retry,
```
sfc-gh-fpawlowski, line 189: Do we want to investigate if the underlying requests library won't execute its built in requests in this case? I think by default there is 1, but we decreased it to 0 for e.g. platform detection.

Fyi HttpConfig can propagate max_retries option there.

sfc-gh-turbaszek: Same pattern is used in `self.rest.delete_session(retry=retry)`, the idea here is to "fire and forget" to avoid any blocking behavior during connection closure

PR #2400, opened 2025-07-11
Separate server and redirect URIs in `AuthHttpServer`

npeshkov:
1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #2396

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [x] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [x] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   Allowing redirect URI to be different from server URI allows users to provide non-localhost redirect URIs (common in cloud IDEs) without preventing the server to start up on localhost.
   - Backwards compatible.
     - Default server URI host is localhost, which is the same as what users used to provide (non-localhost does not work).
     - Server's port is the port in the redirect URI (or random), which is the same as before.

PR #2609, opened 2025-10-28
SNOW-2194055: Separate server and redirect URIs in AuthHttpServer

sfc-gh-turbaszek:
Fixes support for `redirect_uri`. Copy of https://github.com/snowflakedb/snowflake-connector-python/pull/2400

### DIFF
--- a/src/snowflake/connector/auth/oauth_code.py
+++ b/src/snowflake/connector/auth/oauth_code.py
@@ -121,7 +121,7 @@ class AuthByOauthCode(AuthByOAuthBase):
         logger.debug("authenticating with OAuth authorization code flow")
         with AuthHttpServer(
             redirect_uri=self._redirect_uri,
-            uri=self._uri,
+            uri=self._uri or self._redirect_uri,  # for backward compatibility
         ) as callback_server:
             code = self._do_authorization_request(callback_server, conn)
             return self._do_token_request(code, callback_server, conn)
--- a/test/unit/test_auth_oauth_auth_code.py
+++ b/test/unit/test_auth_oauth_auth_code.py
@@ -304,6 +304,7 @@ def test_auth_oauth_auth_code_uses_redirect_uri(
         "host",
         pkce_enabled=False,
         enable_single_use_refresh_tokens=rtr_enabled,
+        uri="http://localhost:0",
     )
 
     def fake_get_request_token_response(_, fields: dict[str, str]):
@@ -437,5 +438,5 @@ def test_auth_oauth_auth_code_passes_uri_to_http_server(
             user="user",
         )
         mock_http_server_init.assert_called_once_with(
-            uri=uri, redirect_uri=redirect_uri
+            uri=uri or redirect_uri, redirect_uri=redirect_uri
         )


src/snowflake/connector/auth/oauth_code.py
```diff
@@ -117,10 +122,20 @@ def _request_tokens(
     ) -> (str | None, str | None):
         """Web Browser based Authentication."""
         logger.debug("authenticating with OAuth authorization code flow")
-        with AuthHttpServer(self._redirect_uri) as callback_server:
+        with AuthHttpServer(
+            redirect_uri=self._redirect_uri,
+            uri=self._read_uri_from_env(),
```
sfc-gh-fpawlowski: Do we intend to expose this only through this env variable? Or should customer be able to pass oauth_socket_address through connection params? Is it known at the start time of the application and should be only possible to override with this env var?

Anyway we may need to create a separate Jira ticket to update the docs to include this new parameter in OAuth documentation for Python Driver.

src/snowflake/connector/auth/oauth_code.py
```diff
@@ -117,10 +122,20 @@ def _request_tokens(
...
+        ) as callback_server:
             code = self._do_authorization_request(callback_server, conn)
             return self._do_token_request(code, callback_server, conn)
 
+    def _read_uri_from_env(self) -> str:
+        oauth_socket_address = os.getenv(
+            ENV_VAR_OAUTH_SOCKET_ADDRESS, "http://localhost"
```
sfc-gh-snoonan: localhost, not http://localhost ?

Generally, this separation of host and port in this way rather than having the address specified as a hostport seems weird to me. Can you ensure the description explains why were are doing it this way?

sfc-gh-fpawlowski (commented):
Left minor comment regarding the flow of configuration

src/snowflake/connector/auth/oauth_code.py
sfc-gh-turbaszek: Current default value (`oauth_redirect_uri`) is `http://127.0.0.1` so keeping protocol should be fine. Later on in the code we parse the URI and reference parts of it separately (port, host, etc).

I agree that having address + port is not neccessary, I will change that.

src/snowflake/connector/auth/oauth_code.py
sfc-gh-turbaszek: Done 👌

src/snowflake/connector/auth/_http_server.py
```diff
@@ -123,6 +129,30 @@ def __init__(
             query=parsed_uri.query,
             fragment=parsed_uri.fragment,
         )
+        if parsed_redirect:
+            if self._is_local_uri(parsed_redirect) and port != parsed_redirect.port:
```
sfc-gh-snoonan: Out of curiosity, why are we handling this as a separate case? It'd be helpful to have a comment explaining it.

src/snowflake/connector/auth/oauth_code.py
sfc-gh-snoonan: It looks like they are still separate?

sfc-gh-turbaszek: Forgot to push 🙈

src/snowflake/connector/auth/_http_server.py
sfc-gh-turbaszek: I removed the logic, it was mostly caused by changing port value after socket bind and was part of attempt to preserve backward compatibility.

npeshkov (commented):
Previously, without any configuration on user side, effective redirect uri would have a port added to it. This no longer happens, which will break setups in which the identity provider expects the uri to end with a port

sfc-gh-snoonan (approved):

PR #2610, opened 2025-10-28
CP: Snow 1747564 econnreset error should be retried (#2547)

sfc-gh-turbaszek:
Cherry picks (https://github.com/snowflakedb/snowflake-connector-python/pull/2547). The aio implementation was adjusted  accordingly.

### DIFF
--- a/src/snowflake/connector/network.py
+++ b/src/snowflake/connector/network.py
@@ -238,6 +238,10 @@ def is_login_request(url: str) -> bool:
     return "login-request" in parse_url(url).path
 
 
+def is_econnreset_exception(e: Exception) -> bool:
+    return "ECONNRESET" in repr(e)
+
+
 class RetryRequest(Exception):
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/network.py
```diff
@@ -965,7 +969,7 @@ def _request_exec_wrapper(
                 )
             retry_ctx.retry_reason = reason
 
-            if "Connection aborted" in repr(e) and "ECONNRESET" in repr(e):
+            if is_econnreset_exception(e):
```
sfc-gh-turbaszek, line 972: This change is skipped in aio due to following note
```
# notes: in sync implementation we check ECONNRESET in error message and close low level urllib session
            #  we do not have the logic here because aiohttp handles low level connection close-reopen for us
```

sfc-gh-pczajka (approved):

PR #2613, opened 2025-10-28
Async connect context manager support - approach 2 - aenter refactored into separate prepare method.

sfc-gh-fpawlowski:
Approach 2 - not idempotent aenter but separate prepare method.
Main benefit - not relying on conn.is_closed() in SnowflakeConnection.__aenter__ (no unnecessary logic there).

Main drawback:
- not calling real SnowflakeConnection.__aenter__ in snowflake.connector.aio.connect.aenter - we cannot feel safe from missing some logic there.
- will we be always able to keep all preparation logic in this conn._prepare_aexit() method?

Analogous to: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L1376-L1412

### DIFF
--- a/src/snowflake/connector/aio/__init__.py
+++ b/src/snowflake/connector/aio/__init__.py
@@ -49,16 +49,11 @@ class _AsyncConnectContextManager:
         """Make the wrapper iterable like a coroutine."""
         return self.__await__()
 
-    # TODO: below is okay if we make idempotent __aenter__ of SnowflakeConnection class - so check if connected and do not repeat connecting
-    # async def __aenter__(self) -> SnowflakeConnection:
-    #     """Enable async with connect(...) as conn:"""
-    #     self._conn = await self._coro
-    #     return await self._conn.__aenter__()
-
     async def __aenter__(self) -> SnowflakeConnection:
         """Enable async with connect(...) as conn:"""
         self._conn = await self._coro
         # Connection is already connected by the coroutine
+        self._conn._prepare_aenter()
         return self._conn
 
     async def __aexit__(self, exc_type: Any, exc: Any, tb: Any) -> None:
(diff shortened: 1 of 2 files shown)


src/snowflake/connector/aio/_connection.py
```diff
@@ -165,8 +166,20 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             "'SnowflakeConnection' object does not support the context manager protocol"
         )
 
+    def _prepare_aenter(self) -> None:
+        """
+        All connection changes done before entering connection context have to be done here, as we expose the same api through snowflake.connector.aio.connect() and call this function there at __aenter__ as well.
+        """
+        pass
```
sfc-gh-turbaszek, line 173: Is it needed?

src/snowflake/connector/aio/__init__.py
```diff
@@ -10,7 +13,90 @@
...
+
+    def __init__(self) -> None:
+        self.__wrapped__ = SnowflakeConnection.__init__
+        self.__name__ = "connect"
+        self.__doc__ = SnowflakeConnection.__init__.__doc__
+        self.__module__ = __name__
+        self.__qualname__ = "connect"
+        self.__annotations__ = getattr(
+            SnowflakeConnection.__init__, "__annotations__", {}
+        )
```
sfc-gh-turbaszek, line 82: All of those are by default reassigned when using `@wraps` (in `update_wrapper`)

src/snowflake/connector/aio/_connection.py
sfc-gh-fpawlowski: That was the clue of this (second) approach - to prevent future differences between with SnowflakeConnection() vs with connect(). But I think I am more in favour of the first one so I am closing this today, unless any objections are there.

src/snowflake/connector/aio/__init__.py
sfc-gh-fpawlowski: Correct - I leaned toward this first approach proposal so I have addressed it there - ptal
https://github.com/snowflakedb/snowflake-connector-python/pull/2614

PR #2614, opened 2025-10-28
SNOW-2671717: Async connect context manager support - approach 1 - idempotent __aenter__ through checking if conn closed

sfc-gh-fpawlowski:
Approach 1 - idempotent __aenter__ through checking if conn closed. 
Main benefit - calling SnowflakeConnection.__aenter__ in the snowflake.connector.aio.connect.__aenter__ - we can feel safe from missing some logic there.

Main drawback - will always conn.is_closed() be reliable.

Analogous to: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L1376-L1412

### DIFF
--- a/test/integ/aio_it/conftest.py
+++ b/test/integ/aio_it/conftest.py
@@ -101,10 +101,7 @@ async def create_connection(connection_name: str, **kwargs) -> SnowflakeConnecti
     from parameters.py.
     """
     ret = fill_conn_kwargs_for_tests(connection_name, **kwargs)
-    connection = SnowflakeConnection(**ret)
-    conn = await async_connect(**ret)
-    await connection.connect()
-    return conn
+    return await async_connect(**ret)
 
 
 @asynccontextmanager


src/snowflake/connector/aio/__init__.py
```diff
@@ -10,7 +13,89 @@
...
+        """Send a value into the wrapped coroutine."""
+        return self._coro.send(arg)
+
+    def throw(self, *args: Any, **kwargs: Any) -> Any:
+        """Throw an exception into the wrapped coroutine."""
+        return self._coro.throw(*args, **kwargs)
+
+    def close(self) -> None:
+        """Close the wrapped coroutine."""
+        return self._coro.close()
```
sfc-gh-turbaszek, line 124: Are those methods used somewhere?

src/snowflake/connector/aio/async_connect_implementation.md
```diff
@@ -0,0 +1,97 @@
+# Async Connect Wrapper: Design & Implementation
```
sfc-gh-fpawlowski: Do we want to have such doc in the aio directory, or keep all info in the official docs only?

src/snowflake/connector/aio/_connection.py
```diff
@@ -128,6 +128,7 @@ def __init__(
         if "platform_detection_timeout_seconds" not in kwargs:
             self._platform_detection_timeout_seconds = 0.0
 
+        # TODO: why we have it here if never changed
```
sfc-gh-fpawlowski: To discuss - is it dead code?

src/snowflake/connector/aio/__init__.py
sfc-gh-fpawlowski: We need those for [PEP492 compliance](http://peps.python.org/pep-0492/) - I added [comment that should explain](https://github.com/snowflakedb/snowflake-connector-python/pull/2614/files#diff-cee0e379944a8144da97a08f1c0b649942128eb1a9ea653a003d3a887899e61cR46-R55) the behaviour better and extracted such interface to a separate class (debatable if its a good choice, but seems cleaner to me).

Example would be:
`result = await asyncio.wait_for(aio.connect(...), timeout=5.0)  # May call throw()`

src/snowflake/connector/aio/async_connect_implementation.md
sfc-gh-fpawlowski: If so we need to add it to Manifest.in file

sfc-gh-fpawlowski:
Should we have connect.__name__ == "__init__" - as it is in synch connect?

Also type of asynch connect is_AsyncConnectWrapper, not function.

src/snowflake/connector/aio/__init__.py
```diff
@@ -9,8 +12,152 @@
...
+    async def __aexit__(self, exc_type: Any, exc: Any, tb: Any) -> None:
+        """Exit async context manager."""
+        if self._conn is not None:
+            return await self._conn.__aexit__(exc_type, exc, tb)
+        else:
+            return None
+
+
+@wraps(SnowflakeConnection.__init__)
+def Connect(**kwargs: Any) -> HybridCoroutineContextManager[SnowflakeConnection]:
```
sfc-gh-fpawlowski: Construct similar to the one used in synch code of main __init__.py

src/snowflake/connector/aio/_connection.py
sfc-gh-fpawlowski: Removed - as its a dead code

src/snowflake/connector/aio/__init__.py
```diff
@@ -9,8 +12,155 @@
...
+        return conn
+
+    return _AsyncConnectContextManager(_connect_coro())
+
 
-async def connect(**kwargs) -> SnowflakeConnection:
-    conn = SnowflakeConnection(**kwargs)
-    await conn.connect()
-    return conn
+connect = Connect
```
sfc-gh-fpawlowski: Do we wish to repeat this "pattern" from the synch code? Are there any counterarguments of simply doing `def connect(..):`?

sfc-gh-turbaszek: To me `def connect` is much more cleaner. Current state is caused by `def Connect` not complying to Python naming conventions, thus we have `connect` alias.

sfc-gh-turbaszek (approved):

PR #2620, opened 2025-10-30
SNOW-1759084: Adjust aio telemetry events

sfc-gh-turbaszek:
There's a difference in `fetchall` implementation between sync and aio code. In sync code the `fetchall` calls `fetchone` under the hood. In AIO code the implementation is different and as a result does not call `_report_metrics` after fetching all result.

### DIFF
--- a/src/snowflake/connector/aio/_result_set.py
+++ b/src/snowflake/connector/aio/_result_set.py
@@ -103,8 +103,7 @@ class ResultSetIterator:
                         item,
                     )
             rets.extend(batch)
-            # yield to avoid blocking the event loop for too long when processing large result sets
-            # await asyncio.sleep(0)
+        await self._final()
         return rets
 
     async def generator(self):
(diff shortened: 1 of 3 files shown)


src/snowflake/connector/aio/_result_set.py
```diff
@@ -103,8 +103,7 @@ async def fetch_all_data(self):
                         item,
                     )
             rets.extend(batch)
-            # yield to avoid blocking the event loop for too long when processing large result sets
-            # await asyncio.sleep(0)
+        await self._final()
```
sfc-gh-fpawlowski, line 106: I suppose not caring about the above comment is caused by the fact that user can just use 'async queries' - in Snowflake meaning?

sfc-gh-turbaszek: I removed the comment because imo it does not make any sense. There's no `yield` and `await asyncio.sleep(0)` was already commented

PR #2624, opened 2025-10-31
NO-SNOW: CRL cherry picks to AIO

sfc-gh-turbaszek:


setup.cfg
```diff
@@ -92,12 +95,9 @@ development =
...
-    pytest-asyncio
+    responses
 pandas =
     pandas>=2.1.2,<3.0.0
     pyarrow
 secure-local-storage =
     keyring>=23.1.0,<26.0.0
-aio =
-    aiohttp>=3.12.14
-    aioboto3>=15.0.0
```
sfc-gh-turbaszek, line 107: Revert

src/snowflake/connector/aio/_session_manager.py
```diff
@@ -75,6 +80,33 @@ async def connect(
...
+                            "could not be validated via CRL: hostname={}".format(
+                                req.url.host
+                            )
+                        ),
+                        errno=ER_OCSP_RESPONSE_CERT_STATUS_REVOKED,
+                    )
+                logger.debug(
+                    "The certificate revocation check was successful. No additional checks will be performed."
+                )
+
```
sfc-gh-turbaszek, line 96: CRL glue code in AIO

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,586 @@
...
+
+            certificate_chains = self._extract_certificate_chains_from_connection(
+                cert_der_chain
+            )
+        except Exception as e:
+            logger.warning(
+                "Failed to extract certificate chain for CRL validation: %s", e
+            )
+            certificate_chains = []
+        return self.validate_certificate_chains(certificate_chains)
```
sfc-gh-turbaszek: This method has changed compared to main branch. The if statement handles sync vs aio scenario. We can potentially refactor it by accepting only `certificate_chains` and delegating certificate pre-processing to other places.

src/snowflake/connector/crl.py
```diff
@@ -0,0 +1,586 @@
...
+        """
+
+        if not cert_chain:
+            logger.debug("No certificate chain found in connection")
+            return []
+
+        # Convert OpenSSL certificates to cryptography x509 certificates
+        x509_chain = []
+        for cert_der in cert_chain:
+            cert_x509 = x509.load_der_x509_certificate(cert_der, default_backend())
```
sfc-gh-turbaszek: On main branch there's also cert to bytes conversion. But as it differs for aio and sync scenarios I moved it to earlier stage.

src/snowflake/connector/aio/_session_manager.py
```diff
@@ -90,6 +112,24 @@ async def connect(
                 protocol._snowflake_ocsp_validated = True
         return connection
 
+    def validate_crl(self, protocol: ResponseHandler, req: ClientRequest):
+        # Resolve CA file path from environment variables or use certifi default
+        cafile_for_ctx = resolve_cafile({"ca_certs": certifi.where()})
+        crl_validator = CRLValidator.from_config(
+            FEATURE_CRL_CONFIG,
+            get_current_session_manager(),
```
sfc-gh-fpawlowski: The get_current_session_manager was used in OCSP as there was no easy way to associate it with one particular session_manager instance without drastic code changes. I think we could safely pass 'self._session_manager' here.

src/snowflake/connector/aio/_session_manager.py
```diff
@@ -71,6 +79,20 @@ async def connect(
...
         connection = await super().connect(req, traces, timeout)
         protocol = connection.protocol
+
+        logger.debug(
+            "CRL Check Mode: %s",
+            FEATURE_CRL_CONFIG.cert_revocation_check_mode.name,
+        )
+        if (
+            FEATURE_CRL_CONFIG.cert_revocation_check_mode
+            != CertRevocationCheckMode.ENABLED
```
sfc-gh-fpawlowski: We validate crl if its not Enabled? 🤔 
Is it the other way around? If yes, then are we missing tests on the main branch that would cover this case?

src/snowflake/connector/aio/_session_manager.py
```diff
@@ -90,6 +112,24 @@ async def connect(
...
+    def validate_crl(self, protocol: ResponseHandler, req: ClientRequest):
+        # Resolve CA file path from environment variables or use certifi default
+        cafile_for_ctx = resolve_cafile({"ca_certs": certifi.where()})
+        crl_validator = CRLValidator.from_config(
+            FEATURE_CRL_CONFIG,
+            get_current_session_manager(),
+            trusted_certificates=load_trusted_certificates(cafile_for_ctx),
+        )
+        sll_object = protocol.transport.get_extra_info("ssl_object")
+        if not crl_validator.validate_connection(sll_object):
```
sfc-gh-fpawlowski: Do we wish to have a Jira for GA Launch checklist to implement validate_connection in async manner as well (and add that Jira number here in a comment)?
Downloading large CRL files could be unnecessarily blocking the customer's runtime (if there are parallel coroutines running in Python other then this Snowflake Driver instance).

src/snowflake/connector/session_manager.py
```diff
@@ -266,7 +266,7 @@ def get(
         url: str,
         *,
         headers: Mapping[str, str] | None = None,
-        timeout: int | None = 3,
+        timeout: int | tuple[int, int] | None = 3,
```
sfc-gh-fpawlowski, line 269: Was this missed in async version or unnecessary?

src/snowflake/connector/ssl_wrap_socket.py
```diff
@@ -141,11 +147,24 @@ def reset_current_session_manager(token) -> None:
 def inject_into_urllib3() -> None:
     """Monkey-patch urllib3 with PyOpenSSL-backed SSL-support and OCSP."""
     log.debug("Injecting ssl_wrap_socket_with_ocsp")
-    connection_.ssl_wrap_socket = ssl_wrap_socket_with_ocsp
+    connection_.ssl_wrap_socket = ssl_wrap_socket_with_cert_revocation_checks
+
+
+def load_trusted_certificates(cafile: str | None) -> list[x509.Certificate]:
```
sfc-gh-fpawlowski, line 154: nit: Names of few functions (e.g. load_trusted_certificates, resolve_cafile), as well as logic in crl.py was changed here.
If we were to keep changes in synch part of code as little as possible on the asyncio branch for the final merge, those refactoring changes should have been separated into the main branch, and only then cherrypicked.
Since we are already finishing tho, I treat this as a nit nice to have, but not obligatory.

test/integ/test_crl.py
```diff
@@ -0,0 +1,189 @@
...
+    [
+        ("DISABLED", True, True),  # DISABLED mode always succeeds
+        ("DISABLED", False, True),  # DISABLED mode always succeeds
+        ("ADVISORY", True, True),  # ADVISORY mode is lenient
+        ("ADVISORY", False, True),  # ADVISORY mode is lenient
+        ("ENABLED", True, True),  # ENABLED with allow_without_crl should succeed
+        ("ENABLED", False, True),  # ENABLED might succeed if certs have valid CRL URLs
+    ],
+)
+def test_crl_validation_modes_parametrized(
```
sfc-gh-fpawlowski, line 115: Should have this test run for async connection as well?

test/integ/test_crl.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _ignore_deprecation_warnings():
+    """Fixture to handle deprecation warnings in all tests in this module."""
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", CryptographyDeprecationWarning)
+        yield
+
+
```
sfc-gh-fpawlowski, line 24: Do you think it may be useful to similar have tests (as in this file) - to make sure crl settings work with async connection as well?

test/unit/test_telemetry.py
```diff
@@ -409,6 +409,16 @@ def get_mocked_telemetry_connection(telemetry_enabled: bool = True) -> Mock:
     mock_connection.is_closed = False
     mock_connection.socket_timeout = None
     mock_connection.messages = []
+    mock_connection.crl_cache_validity_hours = None
```
sfc-gh-fpawlowski, line 412: Was this missed?

src/snowflake/connector/aio/_session_manager.py
```diff
@@ -90,6 +112,24 @@ async def connect(
                 protocol._snowflake_ocsp_validated = True
         return connection
 
+    def validate_crl(self, protocol: ResponseHandler, req: ClientRequest):
+        # Resolve CA file path from environment variables or use certifi default
+        cafile_for_ctx = resolve_cafile({"ca_certs": certifi.where()})
```
sfc-gh-fpawlowski, line 120: Will this cover all env vars being used to pass certs location? 
In requests.request we have this parameter passed later to conn.ca_certs if specified.

                verify = (
                    os.environ.get("REQUESTS_CA_BUNDLE")
                    or os.environ.get("CURL_CA_BUNDLE")
                    or verify
                )
I see that we use resolve_cafile which parses env vars, but should we support CURL_CA_BUNDLE as well there? Also do we maintain the same hierarchy - is certify.where used when no env vars are provided or other way around?

src/snowflake/connector/aio/_session_manager.py
sfc-gh-turbaszek: Fixed, it was changed to enable CRL by default 🙈

src/snowflake/connector/ssl_wrap_socket.py
sfc-gh-turbaszek: Personally I see those changes as part of asyncio implementation so imo it's fine to have them in aio branch

test/unit/test_telemetry.py
sfc-gh-turbaszek: Existing async test works as expected, tbh I don't think this change adds any value to the sync tesdt.

src/snowflake/connector/aio/_session_manager.py
sfc-gh-turbaszek: Should we add `os.environ.get("CURL_CA_BUNDLE")` to `resolve_cafile` implementation to keep the logic in single place?

src/snowflake/connector/aio/_session_manager.py
sfc-gh-turbaszek: Thanks, I didn't know about it

test/integ/test_crl.py
sfc-gh-turbaszek: Added

test/integ/test_crl.py
sfc-gh-turbaszek: Added

src/snowflake/connector/session_manager.py
sfc-gh-turbaszek: I will call it unnecessary. This change was applied in single place while `timeout: int | None = 3,` is used in other places in `session_manager.py`

src/snowflake/connector/aio/_session_manager.py
sfc-gh-turbaszek: Good point, added

src/snowflake/connector/aio/_session_manager.py
sfc-gh-fpawlowski: Thats one thing - add in implementation. Having tests for passing certs in such way maybe a JIRA to be created - since we support it by our code but don't test it. 

Also the mentioned hierarchy should be tested imo - if certify.where() overrides the env vars or is it the other way around. I have concerns this may differ between async and sync code.

sfc-gh-turbaszek: The async code tries to duplicate sync version:
```
    # Ensure CA bundle default if not provided
    if not params.get("ca_certs"):
        params["ca_certs"] = certifi.where()

    cafile_for_ctx = resolve_cafile(params)
```
Created [SNOW-2681235](https://snowflakecomputing.atlassian.net/browse/SNOW-2681235) for CURL_CA_BUNDLE support. For now, I would suggest to keep this out of aio equation.

sfc-gh-fpawlowski:
Fyi: for our tests most SSL checks will happen in synch init_test_schema fixture - which is autoused for the whole testing session. Just so we are aware we need really good tests for that as async code isn't using it under the hood in tests.

src/snowflake/connector/aio/_connection.py
```diff
@@ -1017,6 +1019,7 @@ async def connect(self, **kwargs) -> None:
         else:
             self.__config(**self._conn_parameters)
 
+        self._crl_config: CRLConfig = CRLConfig.from_connection(self)
```
sfc-gh-fpawlowski, line 1040: Is this value ever used in the final CRL setup?

src/snowflake/connector/network.py
```diff
@@ -338,6 +339,13 @@ def __init__(
             self._connection._ocsp_response_cache_filename if self._connection else None
         )
 
+        # CRL mode (should be DISABLED by default)
+        ssl_wrap_socket.FEATURE_CRL_CONFIG = (
```
sfc-gh-fpawlowski, line 343: Do we set config from connection for the async code as well? If not we may be still not testing this enough - all our integ tests of crls seem to expect them to be passing - so if our config isn't propagated correctly -we still wouldn't detect it as there is no test mocking actual cert deprication for async code iiuc

src/snowflake/connector/aio/_session_manager.py
sfc-gh-fpawlowski: Sure, fyi below is how I was able to understand the urllib3 flow for ca_certs - since we do not have adapters for aiohttp those steps we may want to replicate ourselves.

Specifically the fact, that 2 env vars overwrite the certify.where result.
But if we want to do that in the follow-up Jira I am okay with that.


Priority | Source | File:Line | Usage
-- | -- | -- | --
1 | session.get(verify="/path") | sessions.py:777 | Request-level override
2 | session.verify = "/path" | sessions.py:426+777 | Session-level default
3 | REQUESTS_CA_BUNDLE env | sessions.py:769 | Only if verify=True/None
4 | CURL_CA_BUNDLE env | sessions.py:770 | Only if verify=True/None
5 | certifi.where() no 1 | adapters.py:295 | Via DEFAULT_CA_BUNDLE_PATH
6 | certifi.where() no 2 | ssl_wrap_socket.py:176 | Direct call fallback
7 | SSL_CERT_FILE env | ssl_wrap_socket.py:62 | Via resolve_cafile()

src/snowflake/connector/aio/_connection.py
sfc-gh-turbaszek: Yes... synchronous `SnowflakeRestful` updates global `FEATURE_CRL_CONFIG`. And this will work also for async version because we inherit from `SnowflakeRestfulSync`
```
ssl_wrap_socket.FEATURE_CRL_CONFIG = (
            CRLConfig.from_connection(self._connection)
            if self._connection
            else ssl_wrap_socket.DEFAULT_CRL_CONFIG
        )
```

test/helpers.py
```diff
@@ -311,3 +311,31 @@ def _arrow_error_stream_random_input_test(use_table_iterator):
...
     assert len(result_array) == 0
+
+
+def create_mock_auth_body():
+    ocsp_mode = Mock()
+    ocsp_mode.name = "ocsp_mode"
+    session_manager = Mock()
+    session_manager.clone = lambda max_retries: "session_manager"
+
+    return Auth.base_auth_data(
```
sfc-gh-fpawlowski, line 324: We may be lacking import of this Auth class

test/helpers.py
```diff
@@ -311,3 +311,31 @@ def _arrow_error_stream_random_input_test(use_table_iterator):
     # error instance users get should be the same
     assert len(exception_result)
     assert len(result_array) == 0
+
+
+def create_mock_auth_body():
```
sfc-gh-fpawlowski, line 318: Is this used anywhere?

sfc-gh-turbaszek: Yes, there are some usages

PR #2629, opened 2025-11-04
SNOW-2483517 sproc continuous integration

sfc-gh-zyao:

- Provide a set of build scripts, for Snowflake-internal CI/CD, to run tests against Python Sproc given that we will soon run Python sprocs with snowflake-connector-python as well
  - It auto runs the validation script (`ci/anaconda/validate_deps_sync.py`) to make sure the build recipe we use for CI/CD matches setup.py and setup.cfg
  - for internal reviewers: this is the pipeline side change: https://github.com/snowflakedb/jenkins_utils/pull/22284

   Fixes SNOW-2483517

### DIFF
--- a/ci/anaconda/validate_deps_sync.py
+++ b/ci/anaconda/validate_deps_sync.py
@@ -8,10 +8,13 @@ Exit behavior:
 - If there is a diff: print the diff and exit 1.
 """
 
+import configparser
 import re
 import sys
 from pathlib import Path
 from typing import Dict, Iterable, List, Tuple
+
+import yaml
 
 
 def repo_root() -> Path:
@@ -77,35 +80,21 @@ def get_setup_install_requires(cfg_path: Path) -> List[str]:
     Returns:
       List of strings in the form "<name> <spec>" where spec may be empty.
     """
-    lines = cfg_path.read_text(encoding="utf-8").splitlines()
-    in_options = False
-    in_install = False
+    parser = configparser.ConfigParser(interpolation=None)
+    parser.read(cfg_path, encoding="utf-8")
+    if not parser.has_section("options"):
+        raise RuntimeError(f"Missing [options] section in {cfg_path}")
+    if not parser.has_option("options", "install_requires"):
+        raise RuntimeError(f"Missing install_requires under [options] in {cfg_path}")
+    raw_value = parser.get("options", "install_requires")
     deps: List[str] = []
-
-    for line in lines:
-        stripped = line.strip()
-        if stripped.startswith("[options]"):
-            in_options = True
-            in_install = False
+    for line in raw_value.splitlines():
+        item = line.strip()
+        if not item or item.startswith("#"):
             continue
-        if in_options and stripped.startswith("[") and "]" in stripped:
-            # Left [options]
-            in_options = False
-            in_install = False
-        if in_options and stripped.startswith("install_requires"):
-            in_install = True
-            continue
-        if in_install:
-            # Stop when dedented (not an indented list item)
-            if not re.match(r"\s{4,}\S", line):
-                in_install = False
-                continue
-            item = stripped
-            if not item or item.startswith("#"):
-                continue
-            name, spec = split_requirement(item)
-            if name and name != "python":
-                deps.append(f"{name} {spec}".strip())
+        name, spec = split_requirement(item)
+        if name and name != "python":
+            deps.append(f"{name} {spec}".strip())
     return deps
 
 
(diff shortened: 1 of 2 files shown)


ci/anaconda/recipe/meta.yaml
```diff
@@ -0,0 +1,81 @@
+{% set name = "snowflake-connector-python" %}
```
sfc-gh-sfan, line 1: Is this mostly from conda-forge?

ci/anaconda/validate_deps_sync.py
```diff
@@ -0,0 +1,270 @@
...
+            item = stripped
+            if not item or item.startswith("#"):
+                continue
+            name, spec = split_requirement(item)
+            if name and name != "python":
+                deps.append(f"{name} {spec}".strip())
+    return deps
+
+
+def compare_deps(setup_deps: Iterable[str], meta_deps: Iterable[str]) -> str:
```
sfc-gh-sfan, line 173: Would this be fragile if we have dependencies whose name is different between conda and pip? I guess at that time we could introduce a mapping, not something to worry about now

ci/anaconda/validate_deps_sync.py
```diff
@@ -0,0 +1,270 @@
...
+            item = stripped
+            if not item or item.startswith("#"):
+                continue
+            name, spec = split_requirement(item)
+            if name and name != "python":
+                deps.append(f"{name} {spec}".strip())
+    return deps
+
+
+def get_meta_run_requirements(meta_path: Path) -> List[str]:
```
sfc-gh-sfan, line 101: Would `PyYAML` make this processing easier?

ci/anaconda/validate_deps_sync.py
```diff
@@ -0,0 +1,270 @@
...
+                continue
+            if re.match(r"-\s+", stripped):
+                item = stripped[1:].strip()
+                name, spec = split_requirement(item)
+                if name and name != "python":
+                    deps.append(f"{name} {spec}".strip())
+    return deps
+
+
+def get_setup_extra_requires(cfg_path: Path, extra: str) -> List[str]:
```
sfc-gh-sfan, line 141: Would `configparser` make it easier?

ci/anaconda/validate_deps_sync.py
```diff
@@ -0,0 +1,270 @@
+"""Validate conda run requirements match setup.cfg install_requires.
+
+Note that it assumes default requirements to be install_requires + boto. If that assumption is no longer true,
```
sfc-gh-sfan, line 3: For my curiosity, why do we assume required + boto?

ci/anaconda/recipe/meta.yaml
sfc-gh-zyao: Err no, unfortunately it is hand-written by myself... I referred to what we have for sproc connector and wrote this one. (But I guess the sproc connector meta.yaml is essentially based on conda-forge, so the answer should be yes?)

ci/anaconda/validate_deps_sync.py
sfc-gh-zyao: Sure. It is the current behavior in setup.py https://github.com/snowflakedb/snowflake-connector-python/blob/11803c559622deb9eeadd541bb128bde2f8d7378/setup.py#L190-L193

ci/anaconda/validate_deps_sync.py
sfc-gh-zyao: Yeah, for names that are too different between pip and conda, this current version does not work. But your idea of using mapping should solve the problem decently. As of now, we do take care of one name difference. `charset-normalizer` in conda, versus `charset_normalizer` in pip. And we achieve this by always normalize `_` in name to `-`

ci/anaconda/validate_deps_sync.py
sfc-gh-zyao: It helps to get rid of that "tedious" for loop. Meanwhile it makes the script no longer vanilla Python though. Which one shall we prefer? (`configparser` on the other hand, is available to vanilla python)

sfc-gh-zyao: Actually in `run.sh` I can simply create a virtual env and pip install pyyaml, do the validation, and then get rid of the virtual env. So let me use PyYAML to avoid the hassles of re-inventing the wheel

ci/anaconda/recipe/meta.yaml
sfc-gh-sfan: Let's make sure this is consistent with conda-forge. It would also be useful to check with Anaconda and see how the recipe looks like for testing purpose.

ci/anaconda/validate_deps_sync.py
sfc-gh-sfan: Yeah if it is not a lot of effort to install a package, it would hopefully make it eaiser.

sfc-gh-zyao: Done

ci/anaconda/recipe/meta.yaml
sfc-gh-zyao: As of now I see conda-forge is quite outdated (https://anaconda.org/conda-forge/snowflake-connector-python/) (Its latest version is still 3.17.4, but we already have 3.18.0, and 4.0.0 released after that. And 4.0.0 has quite some dependency set update)

sfc-gh-zyao: Let me initiate a separate talk with folks from Anaconda, and see how we want to align this build. Any concern if we move forward with current version of test build scripts, and then later revise it after discussion with Anaconda, if discrepancy exists?

sfc-gh-turbaszek (approved):

ci/anaconda/recipe/meta.yaml
sfc-gh-sfan: Yea we could revise later. Not a big deal

PR #2633, opened 2025-11-06
NO-SNOW: Stop CRL cache background task when closing conn

sfc-gh-turbaszek:

   Fixes #NNNN

### DIFF
--- a/test/unit/test_connection.py
+++ b/test/unit/test_connection.py
@@ -955,40 +955,8 @@ def test_connect_metadata_preservation():
     # Should have parameters like account, user, password, etc.
 
 
-def test_connections_registry(mock_post_requests):
-    """Test that connections are properly tracked in the _ConnectionsPool."""
-    from snowflake.connector.connection import _connections_registry
-
-    # Get initial connection count
-    initial_count = _connections_registry.get_connection_count()
-
-    # Create and connect first connection
-    conn1 = fake_connector()
-    assert (
-        _connections_registry.get_connection_count() == initial_count + 1
-    ), "Connection count should increase by 1 after creating a connection"
-
-    # Create and connect second connection
-    conn2 = fake_connector()
-    assert (
-        _connections_registry.get_connection_count() == initial_count + 2
-    ), "Connection count should increase by 2 after creating two connections"
-
-    # Close first connection
-    conn1.close()
-    assert (
-        _connections_registry.get_connection_count() == initial_count + 1
-    ), "Connection count should decrease by 1 after closing a connection"
-
-    # Close second connection
-    conn2.close()
-    assert (
-        _connections_registry.get_connection_count() == initial_count
-    ), "Connection count should return to initial count after closing all connections"
-
-
 @mock.patch("snowflake.connector.connection.CRLCacheFactory")
-def test_connections_registry_stops_crl_task_if_empty(crl_mock, mock_post_requests):
+def test_connections_registry_lifecycle(crl_mock, mock_post_requests):
     """Test the individual methods of _ConnectionsPool."""
     from snowflake.connector.connection import _ConnectionsRegistry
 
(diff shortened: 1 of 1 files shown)


src/snowflake/connector/crl.py
```diff
@@ -63,7 +63,7 @@ class CRLConfig:
     crl_cache_dir: Path | str | None = None
     crl_cache_removal_delay_days: int = 7
     crl_cache_cleanup_interval_hours: int = 1
-    crl_cache_start_cleanup: bool = False
+    crl_cache_start_cleanup: bool = True
```
sfc-gh-fpawlowski, line 66: Hey, this change has caused the past issue with Jenkins:
https://github.com/snowflakedb/snowflake-connector-python/pull/2572/files

Do we believe that periodic cleanup was the cause of errors there? We should confirm that not to introduce a regression

src/snowflake/connector/connection.py
```diff
@@ -1157,6 +1158,9 @@ def close(self, retry: bool = True) -> None:
         # unregister to dereference connection object as it's already closed after the execution
         atexit.unregister(self._close_at_exit)
         try:
+            # Stop CRL-related background process
+            CRLCacheFactory.stop_periodic_cleanup()
```
sfc-gh-fpawlowski: What if 2 connections were created in the same scope, then one of them is closed? Will this stop periodic cleanups for the other connection as well?

sfc-gh-fpawlowski: Shouldn't this work like a weak pointer - only closing of the last connection would cause stopping the periodic cleanup

src/snowflake/connector/crl.py
sfc-gh-turbaszek: > CRL cache tests were spawning additional process which was never killed and caused Jenkins build to become idle

Setting this to True replicates the behavior locally. Adding the cleanup to connection close solves the issue locally so I believe it does it also on Jenkins.

src/snowflake/connector/connection.py
```diff
@@ -2537,3 +2540,53 @@ def _detect_application() -> None | str:
             return "jupyter_notebook"
         if "snowbooks" in sys.modules:
             return "snowflake_notebook"
+
+
+class _ConnectionsPool:
```
sfc-gh-fpawlowski: nit: Won't the name ConnectionsPool confusing in the future - knowing in most drivers its a factory for connections basically? Maybe the discussed earlier name like _ConnectionsRegistry or smth similar may be less conflicting?

src/snowflake/connector/connection.py
```diff
@@ -2537,3 +2540,53 @@ def _detect_application() -> None | str:
...
+        with self._lock:
+            self._connections.discard(connection)
+            logger.debug(
+                f"Connection {id(connection)} removed from pool. Total connections: {len(self._connections)}"
+            )
+
+            if len(self._connections) == 0:
+                # If no connections left then stop CRL background task
+                # to avoid script dangling
+                CRLCacheFactory.stop_periodic_cleanup()
```
sfc-gh-fpawlowski: nit: we could probably have this in a separate function like _handle_last_connection_removed or _handle_registry_emptied / _finish / _cleanup ?
So that in the future its a clear, one place for adding next operations on this 'hook' - emptying the registry.

sfc-gh-fpawlowski (approved):
Left 2 nits.

src/snowflake/connector/connection.py
sfc-gh-turbaszek: I agree, had the same doubt when thinking about the name

sfc-gh-turbaszek: Changed

src/snowflake/connector/connection.py
sfc-gh-turbaszek: Done

PR #2657, opened 2025-11-19
SNOW-2465273: Add core library import

sfc-gh-turbaszek:

   Fixes #NNNN


src/snowflake/connector/__init__.py
```diff
@@ -48,6 +50,14 @@
...
 setup_external_libraries()
 
+# Load the core library - failures are captured in core_loader and don't prevent module loading
+try:
+    core_loader.load()
+except Exception:
+    # Silently continue if core loading fails - the error is already captured in core_loader
+    # This ensures the connector module loads even if the minicore library is unavailable
+    pass
+
```
sfc-gh-turbaszek, line 57: This is probably redundant because `load` method is already big `try / except`

src/snowflake/connector/_utils.py
```diff
@@ -96,3 +101,74 @@ def get_application_path() -> str:
...
+            self._error = None
+        except Exception as err:
+            self._error = err
+
+    def load(self):
+        """Spawn a separate thread to load the minicore library (non-blocking)."""
+        if self._is_core_disabled():
+            self._error = "mini-core-disabled"
+            return
+        self._error = "still-loading"
```
sfc-gh-pczajka, line 276: Could we wrap it into an error / change class definition to `Exception | str | None`?

src/snowflake/connector/_utils.py
```diff
@@ -96,3 +101,74 @@ def get_application_path() -> str:
...
+            self._register_functions(core)
+            self._version = core.sf_core_full_version()
+            self._error = None
+        except Exception as err:
+            self._error = err
+
+    def load(self):
+        """Spawn a separate thread to load the minicore library (non-blocking)."""
+        if self._is_core_disabled():
+            self._error = "mini-core-disabled"
```
sfc-gh-pczajka, line 274: Could we wrap it into an error / change class definition to Exception | str | None?

src/snowflake/connector/_utils.py
sfc-gh-turbaszek: Can you add more context why? `_is_core_disabled` is imo safe on its own. All getters should be safe to use. The only one potentially causing the issue is `get_core_version` but bytes decoding is in try/except already.

ci/download_minicore.py
```diff
@@ -0,0 +1,246 @@
...
+
+def detect_arch() -> str:
+    """Detect the CPU architecture."""
+    machine = platform.machine().lower()
+    if machine in ("x86_64", "amd64"):
+        return "x86_64"
+    elif machine in ("aarch64", "arm64"):
+        return "aarch64"
+    elif machine in ("i686", "i386", "x86"):
+        return "i686"
```
sfc-gh-dheyman, line 66: do we support x86 minicore in python driver?

src/snowflake/connector/_utils.py
sfc-gh-pczajka: Just keeping types clean, it won't cause problems now as all reads are later casted to str anyway.
The field definition promises Error, not Error or str, I'd expect that content is raiseable if exists.

PR #2673, opened 2025-11-24
CRL validation fixes

sfc-gh-pczajka:
Note for reviewer - all below changes are enclosed in separate commits, ~200 lines each

* [x] the driver must enforce a reasonable maximum size limit for the download
* [x] the file permissions for the CRL file must be set to be writable only by the user running the driver
* [x] the cache directory must have similarly restrictive permissions
* [x] the driver must check that the new CRL is more recent than any cached version

### DIFF
--- a/src/snowflake/connector/crl_cache.py
+++ b/src/snowflake/connector/crl_cache.py
@@ -366,9 +366,11 @@ class CRLFileCache(CRLCache):
 
                 # Write to file with secure permissions (owner read/write only)
                 # Using os.open with 0o600 ensures the file is created with secure permissions
-                fd = os.open(
-                    crl_file_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600
-                )
+                flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+                if IS_WINDOWS:
+                    # flag necessary for writing binary data to file on Windows OS
+                    flags |= os.O_BINARY
+                fd = os.open(crl_file_path, flags, 0o600)
                 try:
                     os.write(fd, crl_data)
                 finally:


src/snowflake/connector/crl_cache.py
```diff
@@ -248,6 +265,44 @@ def _get_crl_file_lock(self, crl_cache_file: Path) -> BaseFileLock:
             timeout=self._cache_file_lock_timeout,
         )
 
+    def _check_permissions(
+        self, path: Path, resource_type: str, expected_perms: str
+    ) -> None:
+        """
+        Check that a CRL cache resource has secure permissions (owner-only access).
```
sfc-gh-jkasten, line 272: nit /optional: This goes past the requirement.
```
the file permissions for the CRL file must be set to be writable only by the user running the driver
```
I suspect that not being able to read the cache might make debugging more difficult, but you would be a better judge of that.

(i.e. another user can't see what CRL is being used and what certs are listed as revoked)

src/snowflake/connector/crl_cache.py
```diff
@@ -248,6 +265,44 @@ def _get_crl_file_lock(self, crl_cache_file: Path) -> BaseFileLock:
...
+
+        Note: This check is only performed on Unix-like systems. Windows file
+        permissions work differently and are not checked.
```
sfc-gh-jkasten, line 275: How is Windows being handled?

src/snowflake/connector/crl.py
```diff
@@ -550,15 +568,99 @@ def _fetch_crl_from_url(self, crl_url: str) -> bytes | None:
...
+                        return None
+                    chunks.append(chunk)
+
+            return b"".join(chunks)
         except Exception:
             # CRL fetch or parsing failed
             logger.exception("Failed to download CRL from %s", crl_url)
             return None
 
+    def _get_crl_last_update(
```
sfc-gh-jkasten: no action required - FWIW, there is no `lastUpdate` field in an RFC 5280 CRL. I am sure this is referring to `thisUpdate`. I see this is what `Cryptography` gives you.

src/snowflake/connector/crl.py
```diff
@@ -550,15 +568,99 @@ def _fetch_crl_from_url(self, crl_url: str) -> bytes | None:
...
+            chunks = []
+            total_size = 0
+            for chunk in response.iter_content(chunk_size=8192):
+                if chunk:
+                    total_size += len(chunk)
+                    if total_size > self._crl_download_max_size:
+                        logger.warning(
+                            "CRL from %s exceeded maximum size limit during download (%d bytes > %d bytes)",
+                            crl_url,
+                            total_size,
```
sfc-gh-jkasten: nit/optional: This total_size is a bit misleading as it will always be slightly higher than the max allowed. I would probably just drop it. We don't know how large the file being downloaded was and we are going to encourage users to try to figure out how to increase the max by 8KB.

src/snowflake/connector/crl.py
sfc-gh-jkasten: outside the scope of this PR: We should really be rejecting the CRL if it doesn't have a parseable `next_update` field. This is a violation of both the RFC and the governing policy documents. Allowing it can only lead to unexpected behavior in other places.

https://datatracker.ietf.org/doc/html/rfc5280#:~:text=When%20CRLs%20are%20issued%2C%0A%20%20%20the%20CRLs%20MUST%20be%20version%202%20CRLs%2C%20include%20the%20date%20by%20which%20the%20next%0A%20%20%20CRL%20will%20be%20issued%20in%20the%20nextUpdate%20field

https://github.com/cabforum/servercert/blob/main/docs/BR.md#72-crl-profile

src/snowflake/connector/crl_cache.py
```diff
@@ -200,29 +203,43 @@ class CRLFileCache(CRLCache):
...
     def _ensure_cache_directory_exists(self) -> None:
-        """Create the cache directory if it doesn't exist."""
+        """Create the cache directory if it doesn't exist with secure permissions."""
         try:
-            self._cache_dir.mkdir(parents=True, exist_ok=True)
+            # Create directory with secure permissions (owner read/write/execute only)
+            self._cache_dir.mkdir(parents=True, exist_ok=True, mode=0o700)
+
+            # Verify directory permissions (if it already existed)
+            if not self._unsafe_skip_file_permissions_check:
```
sfc-gh-jkasten, line 236: I assume that this is acting to protect existing users who upgrade their python library, but already have the directory? What is the migration plan out of curiosity?

src/snowflake/connector/crl.py
```diff
@@ -550,15 +568,99 @@ def _fetch_crl_from_url(self, crl_url: str) -> bytes | None:
...
+                        "Invalid Content-Length header for %s: %s",
+                        crl_url,
+                        content_length,
+                    )
+
+            # Stream the content and check size as we download
+            chunks = []
+            total_size = 0
+            for chunk in response.iter_content(chunk_size=8192):
+                if chunk:
```
sfc-gh-turbaszek: `if not chunk: continue` may reduce the indentation

src/snowflake/connector/crl_cache.py
sfc-gh-pczajka: I think the idea behind the requirement is "no one but the owner can modify the file" to secure from attack basing on modifying cached info trusted by the driver. The function checks that no one except the owner (current user if the file can be opened) have any permissions on the file, it does not check owner's permission set

src/snowflake/connector/crl_cache.py
sfc-gh-pczajka: We had similar discussion with Snowflake CLI config file :D
Default cache location is under user's home directory, which should have secure permissions limited to user and admins only. This was accepted by security in CLI's case.
```
    if platform.system() == "Windows":
        return (
            _get_windows_home_path()
            / "AppData"
            / "Local"
            / "Snowflake"
            / "Caches"
            / "crls"
        )
```

src/snowflake/connector/crl.py
sfc-gh-pczajka: :+1:

src/snowflake/connector/crl.py
sfc-gh-pczajka: :+1:
Added, this is a small change in newly added logic - IMO can be included in this PR

src/snowflake/connector/crl_cache.py
sfc-gh-pczajka: "unsafe_skip_file_permissions_check" flag is meant to be used only by users who can't control the permission on the system (for example CI/CD with different users on the same filesystem, there were use cases like this in Snowflake CLI)

The rest of users will receive "file X have incorrect permissions, run chmod 600 X" error

src/snowflake/connector/crl.py
sfc-gh-pczajka: :+1:

sfc-gh-turbaszek (approved):

src/snowflake/connector/crl_cache.py
sfc-gh-jkasten: Correct - I thought that not giving read access to other users might inhibit debugging, but it isn't that big of a deal.

PR #2678, opened 2025-11-26
Pczajka fix aio wif

sfc-gh-pczajka:
This PR:
 * adds aioboto and aiobotocore dependency if [aio] extra is specified, if SNOWFLAKE_NO_BOTO flag is not provided (checked by extra tests in test installation github check)
 * fixes bug in AWS WIF async implementation
 * asjusts async tests to the fix

### DIFF
--- a/.github/workflows/check_installation.yml
+++ b/.github/workflows/check_installation.yml
@@ -16,8 +16,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
 
 jobs:
-  test-installation:
-    name: Test Boto Dependency
+  test-installation-boto:
+    name: Test boto dependency
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/check_installation.yml
```diff
@@ -54,3 +54,44 @@ jobs:
...
+      - name: Test installation with [aio] (should include aioboto)
+        shell: bash
+        run: |
+          python -m venv test_no_boto_env
+          source test_no_boto_env/bin/activate
+
+          python -m pip install '.[aio]'
+
+          # Check that aioboto3 and aiobotocore are installed
+          pip freeze | grep aioboto || exit 1  # aioboto3 and aiobotocore should be installed
```
sfc-gh-turbaszek, line 78: That's not how it will work. To install boto on aio you need to do `pip install '.[aio, aioboto]'`

setup.py
```diff
@@ -186,12 +186,15 @@ class SetDefaultInstallationExtras(egg_info):
 
     def finalize_options(self):
         super().finalize_options()
-
         # if not explicitly excluded, add boto dependencies to install_requires
         if not SNOWFLAKE_NO_BOTO:
             boto_extras = self.distribution.extras_require.get("boto", [])
             self.distribution.install_requires += boto_extras
 
+            if "aio" in self.distribution.extras_require:
```
sfc-gh-turbaszek, line 194: As far as I debugged this, it does not work as you would expect. `aio` is always in `self.distribution.extras_require` as this is representation of what's defined in `setup.cfg`

setup.py
```diff
@@ -186,12 +186,15 @@ class SetDefaultInstallationExtras(egg_info):
...
+                aioboto_extras = self.distribution.extras_require.get("aioboto", [])
+                self.distribution.extras_require["aio"] += aioboto_extras
```
sfc-gh-turbaszek, line 196: This causes aioboto to be added every time, even if non-aio installation is being done.

PR #2689, opened 2025-12-02
NO-SNOW jira creation GH Action fix

sfc-gh-dszmolka:
### Descripton
Still no code was changed. Continuation of https://github.com/snowflakedb/snowflake-connector-python/pull/2687 ; closure works now but for some reason the `EOF` is made part of the JSON which is of course then invalid.

Tested the change locally, worked, so hopefully will work on GH too.


sfc-gh-fpawlowski (approved):
lgtm : )

PR #2718, opened 2025-12-16
Prepare 4.2.0 release

sfc-gh-pczajka:
This PR:
 * disables AIO in current shape (install extension is no longer available, importing aio raises ImportError):
 ```
Cell In[2], line 1
----> 1 from snowflake.connector.aio import connect

File ~/Snowflake/snowflake-connector-python/src/snowflake/connector/aio/__init__.py:4
      1 from __future__ import annotations
      3 # temporarily disable aio
----> 4 raise ImportError("aio is not supported in this version of the connector")
      7 from functools import wraps
      8 from typing import Any, Coroutine, Generator, Protocol, TypeVar, runtime_checkable

ImportError: aio is not supported in this version of the connector
 ```
 
  * disable aio-connected CICD workflows
  * adjusts release notes

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -447,7 +447,7 @@ jobs:
             junit.py${{ env.shortver }}-lambda-ci-dev.xml
 
   test-aio:
-    if: false  # temporarily disable aio
+    if: false  # temporarily disable aio - SNOW-2905263
     name: Test asyncio ${{ matrix.os.download_name }}-${{ matrix.python-version }}-${{ matrix.cloud-provider }}
     needs: build
     runs-on: ${{ matrix.os.image_name }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build_test.yml
```diff
@@ -447,6 +447,7 @@ jobs:
             junit.py${{ env.shortver }}-lambda-ci-dev.xml
 
   test-aio:
+    if: false  # temporarily disable aio
```
sfc-gh-fpawlowski: Maybe lets do ticket in asyncio PrPr - "Revisit Asyncio product decision" or smth like that - and link it here so ppl in the future know why it is disabled

sfc-gh-fpawlowski (approved):
Left 1 nit, overall lgtm 👍

PR #2722, opened 2025-12-18
SNOW-2690900: isolate-server_session_keep_alive-condition

sfc-gh-fpawlowski:
Added new INFO log (we have 4 in basic flow, this 5th will happen only when param server_session_keep_alive is set to True).

Optimised conditions placement 
- possible decrease in time overhead for the whole session from 1% to 10% when firing 100 async queries.
100 queries + close:
  server_session_keep_alive=False: 33.28s
  server_session_keep_alive=True:  30.04s
  → True is faster by 3.25s (9.8%)
  
- possible decrease in time overhead for only the connection.close() from 40% to 60% when firing 100 async queries.
  100 queries, then CLOSE ONLY:
  server_session_keep_alive=False: 0.504s
  server_session_keep_alive=True:  0.214s
  → True is faster by 0.290s (57.5%)
  
  
  Tested manually - no standard benchmarking.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -13,7 +13,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 - v4.2.0(January 07,2026)
   - Added `SnowflakeCursor.stats` property to expose granular DML statistics (rows inserted, deleted, updated, and duplicates) for operations like CTAS where `rowcount` is insufficient.
   - Added support for injecting SPCS service identifier token (`SPCS_TOKEN`) into login requests when present in SPCS containers.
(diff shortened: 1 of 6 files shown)


src/snowflake/connector/connection.py
```diff
@@ -1197,17 +1197,26 @@ def close(self, retry: bool = True) -> None:
...
+                    )
+                    self.rest.delete_session(retry=retry)
+                else:
+                    logger.debug(
+                        "There are {} async queries still running, not deleting session".format(
+                            len(self._async_sfqids)
+                        )
                     )
+            else:
+                logger.info(
```
sfc-gh-fpawlowski, line 1214: Important to make sure such communication at the INFO level is smth we want.

PR #2723, opened 2025-12-18
Fix MaxOS build on Jenkins

sfc-gh-pczajka:
This PR fixes failing MacOS build due to leftover CRL cache directory from old versions of the connector


sfc-gh-fpawlowski (approved):
Wonderful! 🚢

PR #2738, opened 2026-01-13
[SNOW-2853586] Add secondary_roles connection parameter

sfc-gh-jole:
This adds support for specifying secondary roles mode at connection time, allowing clients to control secondary role activation without relying on user's `DEFAULT_SECONDARY_ROLES` settings. GS back-end server changes [here](https://github.com/snowflake-eng/snowflake/pull/370852)

Supported values:
- ALL: Activate all directly granted roles as secondary roles
- NONE: Do not activate any secondary roles
- Otherwise, use the user's DEFAULT_SECONDARY_ROLES setting, we default to this for unsupported value / no value

Changes:
- Add secondary_roles to `DEFAULT_CONFIGURATION` in `connection.py`
- Add `SECONDARY_ROLES` to auth request body when specified
- Add `SECONDARY_ROLES` to authentication request key whitelist
- Add unit tests for the new parameter

Example usage:
```
conn = snowflake.connector.connect( account='myaccount', user='myuser', secondary_roles='NONE' )
```

This change is fully backward compatible:
- When not specified, existing behavior is unchanged
- No modifications to existing parameters or behavior

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -8,13 +8,16 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 - v4.2.1(TBD)
+  - Ensured proper list conversion - the converter runs to_snowflake on all list elements.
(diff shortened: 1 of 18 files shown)


sfc-gh-jole:

PR #2757, opened 2026-02-02
Exchange vendored urllib to with newest 2.6.3

sfc-gh-pczajka:
Change vendored urllib3 to the newest version. Semi-automatic change:
* Make LLM to list differences between 2.5.0 and vendored
* Ask the same LLM to apply these differences to 2.6.3
* Ask different LLM to compare git diff to differences between not patched 2.5.0 and 2.6.3


sfc-gh-fpawlowski (approved):
LGTM since only flaky tests are failing

Issue #1868, opened 2024-01-31
SNOW-1025489: inconsistent timestamp downscaling


jwyang-qraft:
### Python version

Python 3.11.5 (main, Sep 11 2023, 13:54:46) [GCC 11.2.0]

### Operating system and processor architecture

Linux-5.4.0-165-generic-x86_64-with-glibc2.31

### Installed packages

```python
numba==0.58.1
numpy @ file:///work/mkl/numpy_and_numpy_base_1682953417311/work
pandas==2.1.4
python-dateutil @ file:///tmp/build/80754af9/python-dateutil_1626374649649/work
pytz==2022.7.1
requests==2.31.0
snowballstemmer @ file:///tmp/build/80754af9/snowballstemmer_1637937080595/work
snowflake-connector-python==3.7.0
snowflake-sqlalchemy==1.5.1
SQLAlchemy==1.4.50
tqdm==4.66.1
```

### What did you do?

```python
TIMESTAMP_NTZ(9) column with values that overflow int64 with ns precision (eg. '9999-12-31 00:00:00.000')
```

### What did you expect to see?

I tried to fetch a column with TIMESTAMP_NTZ(9) dtype and the max datetime is '9999-12-31 00:00:00.000' and minimum is '1987-01-30 23:59:59.000'.

I get following error when I select from that column.

      File "/home/jwyang/anaconda3/lib/python3.11/site-packages/snowflake/connector/result_batch.py", line 79, in _create_nanoarrow_iterator
        else PyArrowTableIterator(
             ^^^^^^^^^^^^^^^^^^^^^
      File "src/snowflake/connector/nanoarrow_cpp/ArrowIterator/nanoarrow_arrow_iterator.pyx", line 239, in snowflake.connector.nanoarrow_arrow_iterator.PyArrowTableIterator.__cinit__
      File "pyarrow/table.pxi", line 4116, in pyarrow.lib.Table.from_batches
      File "pyarrow/error.pxi", line 154, in pyarrow.lib.pyarrow_internal_check_status
      File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
      
      pyarrow.lib.ArrowInvalid: Schema at index 2 was different:
      DT: timestamp[us]
      vs
      DT: timestamp[ns]

[...]

PR #2758, opened 2026-02-02
SNOW-1025489 enforce ms scale for timestamps

sfc-gh-pczajka:
This PR adds `force_microseconds_precision` parameter to `fetch_pandas*` and `fetch_arrow*` methods, and adds test case replicating issue #1868


src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowTableIterator.hpp
```diff
@@ -49,6 +50,8 @@ class CArrowTableIterator : public CArrowIterator {
   /** local time zone */
   char* m_timezone;
   const bool m_convert_number_to_decimal;
+  /** force microsecond precision for timestamps to ensure consistent schema */
+  const bool m_force_microsecond_precision;
```
sfc-gh-turbaszek, line 54: Let's add lebel to PR to mark it as something to be added in UD

sfc-gh-turbaszek (approved):

Issue #2746, opened 2026-01-24
SNOW-3020359: Automatically expand tilde (~) to $HOME in private_key_file path


panasenco:
### What is the current behavior?

Currently [connection.py](https://github.com/snowflakedb/snowflake-connector-python/blob/d82300f8c46d65f9ab13f9450562a2f544820c9d/src/snowflake/connector/connection.py#L175) opens the private_key_file directly without expanding `~`.

### What is the desired behavior?

`snowflake-connector-python` should expand `~` to the user's home directory just like `snowflake-cli` does.

I believe the easiest way is to use `os.path.expanduser` like snowflake-cli does.

### How would this improve `snowflake-connector-python`?

A `connections.toml` that contains a private_key_file with a `~` in it works with `snowflake-cli`, but breaks with `snowflake-connector-python`. This leads to confusion and frustration.

This also breaks dev container workflows, because I need to hardcode an absolute path like `/home/user/.snowflake/key.p8` to work with `snowflake-connector-python`, but this path won't exist in my dev container.

### References and other background

`snowflake-cli` [snow_connector.py](https://github.com/snowflakedb/snowflake-cli/blob/6856431422db630e4378e7079b4d1a30d2a7432a/src/snowflake/cli/_app/snow_connector.py#L335) wraps `private_key_file` in [SecurePath](https://github.com/snowflakedb/snowflake-cli/blob/d6f14681e46ecaf6207923d71834d8686bbdb0c4/src/snowflake/cli/api/secure_path.py#L41), which uses `os.path.expanduser()` to expand the path.

sfc-gh-dszmolka:
hi, thanks for raising awareness to this gap, the team will review. if it's within your possibilities, you can also submit a PR for the change

edit: and of course in the meantime the workaround is to provide an absolute path to the file

sfc-gh-dszmolka:
fix attempt PR is up (https://github.com/snowflakedb/snowflake-connector-python/pull/2759)

sfc-gh-dszmolka:
change is now merged and will be part of the next release. will keep this thread posted.

sfc-gh-dszmolka:
released with Snowflake Python driver v4.3.0

PR #2759, opened 2026-02-02
SNOW-3020359 expand tilde to key file path in connection

sfc-gh-dszmolka:

   This fix aims to address issue https://github.com/snowflakedb/snowflake-connector-python/issues/2746, where we do not expand tilde (`~`) in Connection for `private_key_file` configuration, thus the driver cannot read the key file for keypair auth in such cases. 

   Using `pathlib.Path(path_to_key).expanduser()`, which should work on both Unix and Windows bases systems, and even work without `HOME` envvar set (and for Linux, read it from `/etc/passwd`)
   
   Besides adding an automated test, also manually tested on Linux and Windows (latter both with `\\` and `/` style path separator)

### DIFF
--- a/src/snowflake/connector/util_text.py
+++ b/src/snowflake/connector/util_text.py
@@ -311,8 +311,10 @@ def expand_tilde(path_to_expand: Any) -> Any:
             if isinstance(path_to_expand, str)
             else path_to_expand
         )
-    except RuntimeError:
+    except Exception as e:
         # user home could not be resolved
-        _logger.debug("User home could not be determined, not expanding tilde.")
+        _logger.debug(
+            "User home could not be determined, not expanding tilde. Exception: %s", e
+        )
 
     return path_to_expand


sfc-gh-dszmolka:

src/snowflake/connector/util_text.py
```diff
@@ -301,3 +302,17 @@ def get_md5_for_integrity(text: str | bytes) -> bytes:
...
+
+
+def expand_tilde(path_to_expand: Any) -> Any:
+    try:
+        path_to_expand = (
+            str(Path(path_to_expand).expanduser())
+            if isinstance(path_to_expand, str)
+            else path_to_expand
+        )
+    except RuntimeError:
```
sfc-gh-turbaszek: Should we catch any error to be fully safe?

sfc-gh-dszmolka: you're right, i should have done like this from the start. now changed to generic exception handling (and logging it)

PR #2782, opened 2026-03-05
NO-SNOW: Fix Azure GET returning UPLOADED instead of DOWNLOADED status

sfc-gh-turbaszek:
Fix Azure GET returning UPLOADED instead of DOWNLOADED status
In finish_download(), calling get_file_header() to retrieve encryption
metadata has a side effect: the Azure implementation unconditionally sets
meta.result_status to UPLOADED. This overwrites the DOWNLOADED status
that was just set, causing GET operations on Azure to report incorrect
status.

Fix by skipping the redundant get_file_header() call when encryption
metadata was already populated by prepare_download(), and restoring the
correct status in the fallback path.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -9,6 +9,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 # Release Notes
 - Upcoming Release
   - Fixed default `crl_download_max_size` to be 20MB instead of 200MB, as the previous value was set too high and could cause out-of-memory issues.
+  - Fixed a bug where Azure GET commands would incorrectly set the file status to UPLOADED instead of preserving the DOWNLOADED status during metadata retrieval.
 
 - v4.3.0(February 12,2026)
   - Ensured proper list conversion - the converter runs to_snowflake on all list elements.


src/snowflake/connector/storage_client.py
```diff
@@ -394,9 +394,10 @@ def finish_download(self) -> None:
                 # preserve the idea of getting metadata in the first place.
                 # One example of this is the utils that use presigned url
                 # for upload/download and not the storage client library.
-                if meta.presigned_url is not None:
+                if meta.presigned_url is not None and self.encryption_metadata is None:
                     file_header = self.get_file_header(meta.src_file_name)
                     self.encryption_metadata = file_header.encryption_metadata
+                    meta.result_status = ResultStatus.DOWNLOADED
```
sfc-gh-fpawlowski: Is this really needed - we seem to set it always in line 386 above anyway?

src/snowflake/connector/storage_client.py
```diff
@@ -394,9 +394,10 @@ def finish_download(self) -> None:
                 # preserve the idea of getting metadata in the first place.
                 # One example of this is the utils that use presigned url
                 # for upload/download and not the storage client library.
-                if meta.presigned_url is not None:
+                if meta.presigned_url is not None and self.encryption_metadata is None:
```
sfc-gh-fpawlowski: Do we know what changed on the backend that we had to add it here? Should we add some comment to document why this decision was made / why was this if extended?

sfc-gh-fpawlowski (commented):
LGTM

PR #2829, opened 2026-03-26
SNOW-3289793: add timestamped logging to build_darwin.sh to trace aarch64 hang

sfc-gh-fpawlowski:
## Summary
- Adds a `log()` helper that prefixes every message with `date '+%Y-%m-%d %H:%M:%S'`
- Logs host info (`uname -a`) and Python availability (`which`, `--version`) at the start of each loop iteration
- Wraps each major step with before/after timestamps: venv creation, `pip install`, and `python -m build --wheel`

## Why
The `BuildPyConnector-Macaarch64` job on the new `mac_m1_v2` Jenkins labels hangs silently inside `./ci/build_darwin.sh` and times out after 25 minutes with no output indicating _where_ it hangs. The Intel (`Mac64`) job works fine. This PR adds enough logging to pinpoint the stuck step on the next run.

## Test plan
- [ ] Trigger `BuildPyConnector-Macaarch64` on the new `mac_m1_v2` labels against this branch
- [ ] Check the console output for the last timestamped `[Info]` line before the timeout — that identifies the hanging step

### DIFF
--- a/ci/build_darwin.sh
+++ b/ci/build_darwin.sh
@@ -9,10 +9,19 @@ THIS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 CONNECTOR_DIR="$(dirname "${THIS_DIR}")"
 DIST_DIR="$CONNECTOR_DIR/dist"
 
+# Print a timestamped info message
+log() {
+    echo "[$(date '+%Y-%m-%d %H:%M:%S')] $*"
+}
+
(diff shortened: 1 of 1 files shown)


sfc-gh-fpawlowski:
> Does it include a credential detector to prevent the exposure of sensitive logs? Is it in pair with the current 'echo' behaviour?

No - those are simple logs fully controlled by us. Just to make CI script's steps easier to monitor. I would say we do not need such safety net here.

PR #2836, opened 2026-03-30
SNOW-3241482: use login shell and pyenv local for macOS managed node migration

sfc-gh-fpawlowski:
## Summary
- Change shebang from `#!/bin/bash -e` to `#!/bin/bash -e -l` in `ci/build_darwin.sh` and `ci/test_darwin.sh`
  - Login shell (`-l`) sources `~/.bash_profile` where pyenv is initialized system-wide on the new managed nodes — no manual `PYENV_ROOT`/`PATH`/`eval` needed
- Add `pyenv local <matched-version>` inside the build loop in `build_darwin.sh` to select the exact installed Python (e.g. `3.9` → `3.9.21`)

## Context
- ARM64 builds hang on `mac_m1_v2` because bash CI scripts run as non-interactive, non-login shells — they skip `~/.bash_profile` and therefore pyenv is never initialized
- x86 worked on `mac_x86_v2` without changes because Homebrew installed Python directly into `/usr/bin` on that image
- Other drivers (LibSnowflakeClient, ODBC) have already successfully migrated to `mac_m1_v2`
- The corresponding Jenkins label changes (`mac_m1` → `mac_m1_v2`, `mac_x86` → `mac_x86_v2`) will be done in a separate PR in `jenkins_utils`

## Test plan
- [ ] Trigger `BuildPyConnector-Macaarch64` from this branch on `mac_m1_v2` label — verify it builds wheels for Python 3.9-3.14 without hanging
- [ ] Trigger `BuildPyConnector-Mac64` from this branch on `mac_x86_v2` label — verify it builds wheels
- [ ] Verify wheels are uploaded to S3 with correct architecture suffixes (`arm64.whl` / `x86_64.whl`)
- [ ] Verify at least one downstream `RT-PyConnector*` test job passes on each architecture

### DIFF
--- a/ci/build_darwin.sh
+++ b/ci/build_darwin.sh
@@ -38,7 +38,7 @@ for PYTHON_VERSION in ${PYTHON_VERSIONS}; do
 
     # Select the matching pyenv-installed version (e.g. 3.9 -> 3.9.21)
     if command -v pyenv &> /dev/null; then
-        PYENV_MATCH=$(pyenv versions --bare | grep "^${PYTHON_VERSION}" | head -1)
+        PYENV_MATCH=$(pyenv versions --bare | grep "^${PYTHON_VERSION//./\\.}" | head -1)
         if [ -n "$PYENV_MATCH" ]; then
             pyenv local "$PYENV_MATCH"
             log "[Info] pyenv local set to $PYENV_MATCH"


ci/build_darwin.sh
```diff
@@ -36,6 +36,15 @@ for PYTHON_VERSION in ${PYTHON_VERSIONS}; do
 
     log "[Info] ===== Starting build for Python ${PYTHON_VERSION} ====="
 
+    # Select the matching pyenv-installed version (e.g. 3.9 -> 3.9.21)
+    if command -v pyenv &> /dev/null; then
+        PYENV_MATCH=$(pyenv versions --bare | grep "^${PYTHON_VERSION}" | head -1)
```
sfc-gh-dheyman: We should escape the `.` in `^${PYTHON_VERSION}`, because it will be treated as a wildcard. This still will work, but worth noting.

sfc-gh-fpawlowski: Nice catch, added 👍

PR #2843, opened 2026-04-07
Fix client_prefetch_threads=0 normalization

### LINKED ISSUE #2841: SNOW-3324331: client_prefetch_threads validation can bypass the lower bound due to ordering of casting and bounds checks
While looking at `_validate_client_prefetch_threads()` in `src/snowflake/connector/connection.py`, I noticed that bounds checking is performed before the value is converted to an integer, and then the value is unconditionally reassigned using `int(self.client_prefetch_threads)` afterward.

Because of this ordering, fractional values between `0` and `1` can bypass the lower-bound check.

For example, if `0.5` is provided:
- The bounds check evaluates `0.5 <= 0` as false, so no correction is applied  
- The value is then converted using `int(0.5)`, which results in `0`  
- The final stored value becomes `0`, even though the intended minimum is `1`

In practice, this can happen when values come from configuration files, environment variables, or computed expressions (for example, scaling based on CPU count), where non-integer values may be passed unintentionally.

The issue is that conversion and validation are happening in the wrong order, and the value is reassigned after the bounds checks have already run. A more robust approach would be to read the raw value once, convert it once, then apply bounds checks before storing the final result.

I’d be happy to open a PR with a fix and accompanying tests if this approach sounds reasonable.

abdulselamadillmohammed:

   Fixes #2841

This change makes the existing defensive validation path for client_prefetch_threads internally consistent when a non-positive value reaches it.

In the sync path, _validate_client_prefetch_threads() previously read through the defaulting property, which meant a raw value of 0 could resolve to DEFAULT_CLIENT_PREFETCH_THREADS instead of following the existing lower-bound correction. This PR updates the validator to normalize from the raw backing value, preserve None as the default case, and then apply the existing bounds logic before storing the final value.

I also included corresponding async updates so sync and async behavior stay aligned. In particular, async now validates direct assignment through the setter and continues to validate explicitly provided client_prefetch_threads values during connection setup.

The PR adds targeted sync and async unit tests covering:

- connection construction with client_prefetch_threads=0
- direct assignment of client_prefetch_threads=0

N/A


sfc-gh-turbaszek (approved):

sfc-gh-turbaszek (approved):

abdulselamadillmohammed:
@sfc-gh-turbaszek I merged current upstream/main and resolved the DESCRIPTION.md conflict. The affected sync and async unit suites pass locally (116 passed, 1 skipped), along with pre-commit checks. GitHub now shows five workflows awaiting maintainer approval. Could you please approve the workflows and re-enable auto-merge when appropriate? Thanks!

PR #2863, opened 2026-05-04
SNOW-3472780 Azure Impersonation Support for WIF Authentication

sfc-gh-yuzzhang:

     SNOW-3472780: Azure WIF impersonation support
     Adds support for Azure managed identity (MI) to impersonate a service principal (SP) when authenticating to Snowflake via Workload Identity Federation.
     Changes:                                                                                                                                                           
      - wif_util.py: new get_azure_sp_token_via_impersonation(), updated create_azure_attestation() to accept impersonation_path                                         
      - connection.py: AZURE added to the list of providers that support workload_identity_impersonation_path                                                            
      - Unit tests and WIF integration test added 
      - Update DESCRIPTION.md

    When workload_identity_impersonation_path=[<sp_client_id>] is set, the driver fetches an MI token with audience api://AzureADTokenExchange, exchanges it for an SP token via the Entra ID token endpoint, and sends the SP token to Snowflake. Only single-hop is supported (Azure doesn't allow chained federation). Gated behind SNOWFLAKE_ENABLE_AZURE_WIF_IMPERSONATION=true pending full release.                        

Test passed in vm:
Screenshot 2026-05-05 at 11 10 23 AM


src/snowflake/connector/connection.py
```diff
@@ -1613,14 +1613,32 @@ def __open_connection(self):
...
                         self,
                         None,
                         ProgrammingError,
                         {
-                            "msg": "workload_identity_impersonation_path is currently only supported for GCP and AWS.",
+                            "msg": "workload_identity_impersonation_path is currently only supported for GCP, AWS, and AZURE.",
+                            "errno": ER_INVALID_WIF_SETTINGS,
+                        },
+                    )
+                if (
```
sfc-gh-xizhao: Just to confirm, we don't need to implement the async part right?

test/unit/test_auth_workload_identity.py
```diff
@@ -559,3 +560,33 @@ def test_explicit_azure_uses_explicit_client_id_if_set(
     auth_class.prepare(conn=None)
 
     assert fake_azure_metadata_service.requested_client_id == "custom-client-id"
+
+
+def test_azure_impersonation_raises_error_if_multi_hop(monkeypatch):
```
sfc-gh-xizhao: Do we have coverage for happy path?

src/snowflake/connector/wif_util.py
```diff
@@ -343,17 +344,69 @@ def create_gcp_attestation(
...
 def create_azure_attestation(
     snowflake_entra_resource: str,
     session_manager: SessionManager | None = None,
+    impersonation_path: list[str] | None = None,
 ) -> WorkloadIdentityAttestation:
     """Tries to create a workload identity attestation for Azure.
 
     If the application isn't running on Azure or no credentials were found, raises an error.
     """
+    if impersonation_path:
```
sfc-gh-xizhao, line 401: Do we need param protection here as well?

src/snowflake/connector/wif_util.py
```diff
@@ -343,17 +344,69 @@ def create_gcp_attestation(
...
 
 
+def get_azure_sp_token_via_impersonation(
+    mi_token: str,
+    sp_client_id: str,
+    snowflake_entra_resource: str,
+    session_manager: SessionManager,
+) -> str:
+    """Exchanges a managed identity token for a service principal token via the Entra ID token endpoint."""
+    tenant_id = jwt.decode(mi_token, options={"verify_signature": False}).get("tid")
```
sfc-gh-xizhao, line 356: So we assume the MI token’s tid claim is the tenant for the SP token endpoint. Is cross-tenant federation expected?
Might worth documenting and testing coverage for this contract.

test/unit/test_auth_workload_identity.py
sfc-gh-yuzzhang: yes, it's in CI end2end test: test_should_authenticate_with_impersonation

src/snowflake/connector/wif_util.py
sfc-gh-yuzzhang: I moved the feature flag check to connection.py before reaching wif_util.py because I thought wif_util.py should only handle the actual attestation logic, not feature gating.

src/snowflake/connector/connection.py
sfc-gh-yuzzhang: I'll double check. I assumed async wasn't needed here since we didn't handle it for the AWS outbound token feature either, but I may be wrong. Do we have a general policy on when WIF providers should support async?

src/snowflake/connector/wif_util.py
sfc-gh-yuzzhang: Thanks! Azure requires the MI and the app registration to be in the same tenant, so tid from the MI token is always the right tenant for the token exchange. I  added a code comment to call this out and also updated the design doc. 
I think the happy path CI test covers this because the token exchange would fail if tid were wrong.

sfc-gh-xizhao (approved):
LGTM, thanks

src/snowflake/connector/connection.py
```diff
@@ -1613,14 +1613,32 @@ def __open_connection(self):
...
+                    self._workload_identity_impersonation_path
+                    and self._workload_identity_provider == AttestationProvider.AZURE
+                    and os.environ.get(
+                        "SNOWFLAKE_ENABLE_AZURE_WIF_IMPERSONATION", "false"
```
sfc-gh-pczajka: Hiding this behind feature flag makes little sense in connector as a client-side library. As the change does not break current behavior, let's remove the flag requirement.

src/snowflake/connector/wif_util.py
```diff
@@ -343,17 +344,79 @@ def create_gcp_attestation(
...
     session_manager: SessionManager | None = None,
+    impersonation_path: list[str] | None = None,
 ) -> WorkloadIdentityAttestation:
     """Tries to create a workload identity attestation for Azure.
 
     If the application isn't running on Azure or no credentials were found, raises an error.
     """
+    if impersonation_path:
+        if (
+            os.environ.get("SNOWFLAKE_ENABLE_AZURE_WIF_IMPERSONATION", "false").lower()
```
sfc-gh-pczajka: As in another comment, let's remove the flag requirement

test/unit/test_connection.py
```diff
@@ -709,11 +707,35 @@ def test_workload_identity_impersonation_path_errors_for_unsupported_providers(
...
             in str(excinfo.value)
         )
 
 
+@pytest.mark.parametrize(
+    "env_value",
+    [None, "false"],
+    ids=["env_var_not_set", "env_var_false"],
+)
+def test_azure_impersonation_errors_if_not_enabled(monkeypatch, env_value):
```
sfc-gh-pczajka: Unit test coverage is significantly lighter than AWS/GCP:

**Azure impersonation unit tests:**

- ✓ Multi-hop rejection
- ✓ Missing `tid` claim error
- ✓ Feature flag disabled error
- ✗ No happy path (successful MI → SP token exchange)
- ✗ No Entra ID API failure handling
- ✗ No missing `access_token` in response

Could you please add missing scenarios?  


**Recommendation:** Add a unit test that mocks the Entra ID token exchange (similar to the GCP pattern) to validate:

1. MI token is requested with `api://AzureADTokenExchange` audience
2. POST to [`https://login.microsoftonline.com/`](https://login.microsoftonline.com/)`{tenant_id}/oauth2/v2.0/token` with correct payload
3. SP token is extracted and returned
4. Final attestation contains the SP token

This would catch regressions without requiring Azure infrastructure.

src/snowflake/connector/wif_util.py
```diff
@@ -343,17 +344,79 @@ def create_gcp_attestation(
...
+            url=f"https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token",
+            data={
+                "grant_type": "client_credentials",
+                "client_id": sp_client_id,
+                "client_assertion_type": "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
+                "client_assertion": mi_token,
+                "scope": f"{snowflake_entra_resource}/.default",
+            },
+        )
+        res.raise_for_status()
```
sfc-gh-boler, line 376: It would helpful to return error details from the response instead of immediately raising ProgrammingError.

src/snowflake/connector/aio/_wif_util.py
```diff
@@ -225,17 +228,80 @@ async def create_gcp_attestation(
...
+            url=f"https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token",
+            data={
+                "grant_type": "client_credentials",
+                "client_id": sp_client_id,
+                "client_assertion_type": "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
+                "client_assertion": mi_token,
+                "scope": f"{snowflake_entra_resource}/.default",
+            },
+        )
+        content = await res.content.read()
```
sfc-gh-boler, line 258: Error handling in the sync variant is different. Let's keep them consistent.

sfc-gh-boler (changes requested):

src/snowflake/connector/connection.py
sfc-gh-yuzzhang: Thanks! But we'd prefer to keep this behind a flag for now，this is a new behavior and we want users to explicitly opt in before it becomes the default.

src/snowflake/connector/aio/_wif_util.py
sfc-gh-boler: It seems that `response_data = json.loads(response_text)` should go before `raise_for_status`. Otherwise an exception will be raised before we get to read `response_text`.

src/snowflake/connector/wif_util.py
sfc-gh-boler: See my followup comment in _wif_util.py.

sfc-gh-boler (changes requested):
I second Patryk's question about the necessity of gating this behind `SNOWFLAKE_ENABLE_AZURE_WIF_IMPERSONATION`. If it's needed, then when is the gate going to be removed?

sfc-gh-yuzzhang:
> I second Patryk's question about the necessity of gating this behind `SNOWFLAKE_ENABLE_AZURE_WIF_IMPERSONATION`. If it's needed, then when is the gate going to be removed?

Thanks for the feedback! Our plan is to keep the flag while this feature is in PrPr for the Python driver only, and remove it once it's GA across all drivers.

src/snowflake/connector/aio/_wif_util.py
```diff
@@ -283,6 +346,11 @@ async def create_azure_attestation(
             errno=ER_WIF_CREDENTIALS_NOT_FOUND,
         )
 
+    if impersonation_path:
+        jwt_str = await get_azure_sp_token_via_impersonation(
```
sfc-gh-boler, line 350: Do we want to verify whether we have a v1.0 token?

sfc-gh-boler (approved):
lgtm

src/snowflake/connector/aio/_wif_util.py
sfc-gh-yuzzhang: Thanks! We tested this on both vm and ci test. Even with the v2.0 endpoint, the resulting SP token is ver: 1.0 with sub == oid. So we don't think validation is necessary on the driver side.

sfc-gh-pczajka (approved):

PR #2864, opened 2026-05-07
Fix write_pandas temp stage name collisions in forked processes (SNOW-3481510)

sfc-gh-pczajka:
## Summary
Fixed write_pandas / create_dataframe failures caused by temp stage name collisions (002002 (42710): Object 'SNOWPARK_TEMP_STAGE_<X>' already exists) by switching the random name generator from random.choice (Mersenne Twister) to secrets.choice (CSPRNG).
The Mersenne Twister PRNG shares state across fork(), so forked processes (e.g. Notebook kernels) produce identical "random" name sequences, making collisions deterministic rather than probabilistic.
This is a single-line change in _utils.py (from random import choice → from secrets import choice) with no API or format changes.
## Context
In Notebooks, every time a user displays a pandas DataFrame the kernel calls session.create_dataframe(pandas_df) which triggers write_pandas → _create_temp_stage. A single session may invoke this path many times, and Notebook kernels commonly fork from a shared parent process. Because random.choice uses the Mersenne Twister PRNG (which inherits state across fork), child processes generate the exact same sequence of stage names, causing CREATE TEMPORARY STAGE to fail when the name already exists.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -8,6 +8,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 - Upcoming Release
+  - Fixed `write_pandas` temp stage name collisions (SNOW-3481510). The old PRNG could produce identical name sequences in forked processes (e.g. Notebook kernels), causing `CREATE TEMPORARY STAGE` to fail with "Object already exists".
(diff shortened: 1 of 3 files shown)


sfc-gh-mcepiga (approved):
lgtm

PR #2871, opened 2026-05-12
SNOW-3526469: prune foreign-arch minicore binaries at build_py

sfc-gh-olorek:
Please answer these questions before submitting your pull requests. Thanks!

1. What GitHub issue is this PR addressing? Make sure that there is an accompanying issue to your PR.

   Fixes #2870

2. Fill out the following pre-review checklist:

   - [x] I am adding a new automated test(s) to verify correctness of my new code
   - [ ] I am adding new logging messages
   - [ ] I am adding a new telemetry message
   - [ ] I am modifying authorization mechanisms
   - [ ] I am adding new credentials
   - [ ] I am modifying OCSP code
   - [ ] I am adding a new dependency

3. Please describe how your code solves the related issue.

   **Problem.** Since 4.0.0 the sdist ships pre-built minicore binaries for every supported platform under `src/snowflake/connector/minicore/`. `_CoreLoader._get_platform_subdir` only ever loads the native one, so the other 7 platforms' binaries are dead weight inside every install. Packaging audits that reject foreign-arch binaries (e.g. `brew audit`) fail, blocking the Homebrew release of `snowflake-cli` 3.17.0 ([Homebrew/homebrew-core#282333](https://github.com/Homebrew/homebrew-core/pull/282333)) — the first CLI release to pull `snowflake-connector-python >= 4.0.0`. Same class of problem will hit conda-forge, nixpkgs, distro maintainers, and internal packagers.

   **Fix.** Hook setuptools' `build_py` command to strip non-matching `minicore/<platform>/` subdirs from the built distribution. This mirrors the platform-specific cleanup that `ci/build_linux.sh` already runs before wheel builds, but runs for every `build_py` invocation — so wheels, `pip install sdist`, `python -m build`, and downstream packager builds all produce a clean single-platform layout.

   The sdist itself keeps all 8 platforms (single source of truth). The `_minicore_native_subdir` helper in `setup.py` mirrors `_CoreLoader._get_platform_subdir` so build-time and runtime detection always agree (enforced by `test_pruner_agrees_with_runtime_loader`).

   **Verified locally** by building the sdist, `pip install`-ing it, and confirming only the native `linux_aarch64_glibc/` remains in the installed tree (and the sdist still ships all 8 platforms).

   Tracked internally as SNOW-3526469.

PR #2872, opened 2026-05-12
SNOW-3526469: prune foreign minicore binaries at sdist build, not install

### LINKED ISSUE #2870: SNOW-3526564: sdist installs minicore binaries for every platform, breaking downstream packagers (Homebrew)
## Summary

Starting with `snowflake-connector-python` 4.0.0, the sdist (`snowflake_connector_python-*.tar.gz`) ships pre-built minicore binaries for **every** supported platform under `src/snowflake/connector/minicore/`:

- `aix_ppc64/`
- `linux_aarch64_glibc/libsf_mini_core.so`
- `linux_aarch64_musl/libsf_mini_core.so`
- `linux_x86_64_glibc/libsf_mini_core.so`
- `linux_x86_64_musl/libsf_mini_core.so`
- `macos_aarch64/libsf_mini_core.dylib`
- `macos_x86_64/libsf_mini_core.dylib`
- `windows_x86_64/sf_mini_core.dll`

When a downstream packager builds from sdist (pip install, Homebrew, conda-forge, nixpkgs, distro maintainers), all of these end up in the installed tree. `src/snowflake/connector/_utils.py::_CoreLoader._get_platform_subdir` only ever loads the one matching the current platform — the rest are dead weight and, worse, trip packaging audits that reject foreign-arch binaries.

## Impact

**Homebrew `snowflake-cli` 3.17.0 cannot be released.** The brew PR [Homebrew/homebrew-core#282333](https://github.com/Homebrew/homebrew-core/pull/282333) fails `brew audit` on all bottle builds (Linux arm64, Linux x86_64, macOS 14-arm64, macOS 15-arm64, macOS 26-arm64):

```
* Binaries built for a non-native architecture were installed into snowflake-cli's prefix.
  The offending files are:
    …/libexec/lib/python3.13/site-packages/snowflake/connector/minicore/linux_x86_64_glibc/libsf_mini_core.so  (x86_64)
Error: 1 failed step!
brew audit --formula snowflake-cli --online --git --skip-style
```

The bump to `snowflake-cli` 3.17.0 is what pulled `snowflake-connector-python` from 3.18.0 → 4.4.0 into the formula, which is the first version to ship the minicore blobs in the sdist.

Same class of problem will hit any downstream packager that runs similar audits (conda-forge, nixpkgs, distro packages, corporate internal packaging).

## Root cause

`MANIFEST.in` includes the full `minicore/` tree in the sdist:

```
recursive-include src/snowflake/connector/minicore *.so *.dll *.dylib *.a *.h *.lib
```

`ci/build_linux.sh` strips the non-native dirs before building platform-specific wheels, but that cleanup only runs in this repo's own CI wheel build. Anyone consuming the sdist gets all 8 platforms' binaries copied into their install prefix.

## Expected behavior

Installing `snowflake-connector-python` on any single target platform should leave only that platform's minicore binary in the installed tree.

## Proposed fix

Hook setuptools' `build_py` command to strip non-matching `minicore/<platform>/` directories from the built distribution so the sdist → install path produces a clean single-platform layout (matching what `ci/build_linux.sh` already does for wheels). The sdist itself keeps every platform as a single source of truth.

Tracked internally as SNOW-3526469.

sfc-gh-olorek:

   Fixes #2870. Alternative to #2871.

   **Problem.** Since 4.0.0 the sdist ships prebuilt minicore binaries for every supported platform under `src/snowflake/connector/minicore/`. The runtime loader (`_CoreLoader._get_platform_subdir`) only ever opens one, so the other 7 are dead weight. Downstream packaging audits that reject foreign-arch binaries (e.g. `brew audit`) fail — this blocked the Homebrew release of `snowflake-cli` 3.17.0 ([Homebrew/homebrew-core#282333](https://github.com/Homebrew/homebrew-core/pull/282333)), the first CLI release pulling `snowflake-connector-python >= 4.0.0`. Same class of problem will hit conda-forge, nixpkgs, and distro maintainers.

   **Why this PR, not #2871.** #2871 treated the symptom: a `build_py` hook pruned non-native subdirs on every install. That copy of `_CoreLoader._get_platform_subdir` had to be kept in sync with the runtime via a test, and it ran on every wheel build and end-user install. This PR tracks the root cause — the binaries should never have been in the sdist in the first place — and fixes it at the single source: the `sdist` command.

   **Fix.** Override `setuptools.command.sdist.make_release_tree` to remove foreign-arch subdirs from the staged sdist tree before the tarball is written. The git tree keeps all 8 binaries (wheel CI consumes them directly via `ci/build_linux.sh`, unchanged). Install-time is untouched for wheels and sdists alike.

   The command fails fast if the host platform is unrecognised or its minicore subdir is missing, so a silently-broken sdist never ships. `check-manifest` is configured with an `ignore` rule for `src/snowflake/connector/minicore/**` because the intentional asymmetry between git and the sdist would otherwise trip the pre-commit hook.

   **Verified locally** by running `python setup.py sdist`, confirming only the native `linux_aarch64_glibc/libsf_mini_core.so` ended up in the tarball (git tree still has all 8), then `pip install`-ing the tarball into a clean venv and confirming the installed tree contains only the host binary.

   Tracked internally as SNOW-3526469.

### DIFF
--- a/DESCRIPTION.md
+++ b/DESCRIPTION.md
@@ -8,6 +8,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 - v4.5.0(May 12,2026)
+  - Fixed sdist to ship only the host platform's minicore binary (SNOW-3526469). Previous 4.x sdists bundled every supported platform's `.so`/`.dylib`/`.dll`, breaking downstream packagers (e.g. Homebrew) whose audits reject foreign-arch binaries. Wheels are unaffected — they were already single-platform.
   - Fixed `write_pandas` temp stage name collisions (SNOW-3481510). The old PRNG could produce identical name sequences in forked processes (e.g. Notebook kernels), causing `CREATE TEMPORARY STAGE` to fail with "Object already exists".
   - Fixed a security bug in Okta SAML authentication where `_is_prefix_equal()` compared `url1`'s port against itself instead of `url2`'s port, allowing an attacker to redirect credentials to a different port on the same hostname. Also fixed the default port fallback to use `int` instead of `str` for correct comparison when one URL omits the port.
   - Fixed `executemany` with `paramstyle="pyformat"` to correctly locate the VALUES clause using a balanced-parentheses parser instead of a greedy regex. This fixes incorrect behaviour with nested function calls such as SQLAlchemy `@compiles VARIANT` patterns (e.g. `PARSE_JSON(%(col)s)`) and subquery-form INSERTs (SNOW-298756).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,4 +27,9 @@ repair-wheel-command = ""
 archs = ["AMD64"]
 
 [tool.check-manifest]
+# Prebuilt minicore binaries live in git for every supported platform so
+# wheel CI can consume them directly. setup.py's HostOnlyMinicoreSdist
+# strips foreign-arch subdirs from the sdist. Tell check-manifest not to
+# flag the resulting asymmetry between the git tree and the sdist tarball.
+ignore = ["src/snowflake/connector/minicore/**"]
 ignore-bad-ideas = ["src/snowflake/connector/minicore/**"]
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,14 @@
 #!/usr/bin/env python
 
 import os
+import platform
+import shutil
 import sys
 import warnings
 
 from setuptools import Extension, setup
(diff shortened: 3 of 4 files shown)


sfc-gh-olorek:
Closing: this approach doesn't fix the Homebrew audit. The sdist is built on a Linux release bot, so my custom sdist command would ship a Linux .so — brew's macOS audit sees a foreign-arch binary regardless of whether it's 1 or 8 of them. #2871's install-time approach is the correct phase because only at install time is the target platform knowable.

PR #2877, opened 2026-05-19
SNOW-3351450: emit client_connection_identifier_shape in-band telemetry

sfc-gh-boler:
## Summary

Capture the shape of connection-identifier inputs (which of `account` / `region` / `host` the caller supplied, and whether the raw account string carries dot/dash markers) at the earliest parse point in `SnowflakeConnection.__config`, then emit a single in-band `client_connection_identifier_shape` telemetry record per successful login alongside the existing `_log_minicore_import` / `_log_nanoarrow_import` hooks. Mirrors the gosnowflake reference implementation in [snowflakedb/gosnowflake#1797](https://github.com/snowflakedb/gosnowflake/pull/1797).

Time-boxed data collection ahead of the Northstar migration. The emission honors the post-login `CLIENT_TELEMETRY_ENABLED` parameter (automatic via ``self.telemetry_enabled`` inside ``_log_telemetry``) and a local `SF_TELEMETRY_DISABLE_CONNECTION_SHAPE=true` env-var kill switch (case-insensitive). No hostname or account value is included in the payload — only five stringified booleans describing intent.

Removal is tracked in SNOW-3548350 (target: 2026-11-30); all new code is marked with a `TODO(SNOW-3548350)` so future grep finds it.

## Changes

- **New** `src/snowflake/connector/connection_identifier_shape.py`: ``ConnectionIdentifierShape`` dataclass + pure ``record_input_shape(kwargs)`` helper. Captures the dash signal only from the *account portion* of the raw account string (the part before any first dot), so region-tail dashes (the `-east-` in `"myacct.us-east-1"`) aren't miscounted as `account_org_provided`.
- **Capture site** at the very top of ``SnowflakeConnection.__config``, idempotent so a re-config with normalized values cannot overwrite the original user intent. Runs before the ``setattr`` loop, before ``construct_hostname`` for host inference, and before ``parse_account`` strips the ``.global`` suffix.
- **Emission site** mirrors ``_log_minicore_import`` exactly: ``_log_connection_identifier_shape()`` is invoked from ``__init__`` right after the existing minicore/nanoarrow hooks (sync path) and from ``aio._connection.SnowflakeConnection.connect`` right after ``_log_telemetry_imported_packages()`` (async path). The async class has its own ``async def _log_connection_identifier_shape`` override because ``_log_telemetry`` is overridden as an awaitable there; calling the sync emit from the async path would return an unawaited coroutine.
- **TelemetryField** picks up one new event type (``CONNECTION_IDENTIFIER_SHAPE = "client_connection_identifier_shape"``) and five payload keys (``KEY_ACCOUNT_PROVIDED`` etc.) so the wire format matches the Go reference byte-for-byte.
- **DESCRIPTION.md** entry under ``Upcoming Release`` mirroring the gosnowflake CHANGELOG voice.

## Test plan

- New `test/unit/test_connection_identifier_shape.py` — 13 table-driven cases exercising ``record_input_shape`` over the truth table (locator, dotted account, org-prefixed account, region-dash isolation, host-only, all-three, empty/None/non-string defensive paths, irrelevant-kwargs isolation).
- New `test/unit/test_connection_identifier_shape_telemetry.py` — 16 cases exercising the emission method: payload field check, skip when shape unset / attribute missing, env-switch kill (case-insensitive ``true`` only), env-switch off-by-default for every other value including former truthy aliases (``1`` / ``yes``) so the new contract is unambiguous.
- 643 existing unit tests still pass locally; the 3 failures + 27 errors observed are pre-existing arrow-stream-extension / wiremock requirements unrelated to this change.

### DIFF
--- a/src/snowflake/connector/_connection_identifier_shape.py
+++ b/src/snowflake/connector/_connection_identifier_shape.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python
+"""Capture user-supplied connection-identifier provenance for in-band telemetry.
+
+The shape captured here is consumed by ``SnowflakeConnection._log_connection_identifier_shape``
+and emitted as a single ``client_connection_identifier_shape`` telemetry event
+per successful login. The capture function inspects the raw kwargs passed to
+``SnowflakeConnection.__config`` before any normalization (host inference,
+account stripping of ``.global``, region extraction from dotted account) runs,
+so the shape reflects user intent rather than the final post-normalization
+state of the connection.
+
+Removal of this module and the emission it backs is tracked in SNOW-3548350
+(target: 2026-11-30).
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Mapping
+
+from .telemetry import TelemetryField
+
+
+@dataclass(frozen=True)
+class ConnectionIdentifierShape:
+    """Provenance of connection-identifier fields the user supplied.
+
+    All fields describe what the user supplied at the moment of input — they
+    reflect intent, not the final post-normalization state of the connection.
+
+    - ``account_provided``: the user explicitly set the ``account`` parameter
+      (via ``connect(account=...)``, kwargs, or ``connections.toml`` merged
+      into kwargs before ``__config`` runs).
+    - ``account_with_region``: the raw account string the user typed contained
+      a dot (e.g. ``"myacct.us-east-1"``), signaling the deprecated
+      ``account.region`` embedded form. Set only on the raw input.
+    - ``account_org_provided``: the raw account string carried a dash in its
+      account portion (e.g. ``"myorg-myacct"``), signaling the org-prefixed
+      form. Region-portion dashes (e.g. the ``-east-`` in
+      ``"myacct.us-east-1"``) are intentionally not counted; only the portion
+      before the first ``.`` is examined.
+    - ``region_provided``: the user explicitly set the ``region`` parameter as
+      a distinct kwarg. A region embedded inside a dotted account string is
+      NOT ``region_provided``; that's ``account_with_region``.
+    - ``host_provided``: the user explicitly set the ``host`` parameter.
+    """
+
+    account_provided: bool = False
+    account_with_region: bool = False
+    account_org_provided: bool = False
+    region_provided: bool = False
+    host_provided: bool = False
+
+
+def _is_user_supplied_string(value: Any) -> bool:
+    """A kwarg counts as user-supplied iff it's a non-empty string. Non-string
+    truthy values (e.g. an accidentally-passed ``True`` / ``int``) are not
+    treated as provided here — the regular ``__config`` validation will warn
+    or error on them later, and shape capture is best kept conservative."""
+    return isinstance(value, str) and value != ""
+
+
+def record_input_shape(kwargs: Mapping[str, Any]) -> ConnectionIdentifierShape:
+    """Capture the connection-identifier shape from the raw kwargs that
+    ``SnowflakeConnection.__config`` receives.
+
+    Must be invoked before any normalization (the ``setattr`` loop in
+    ``__config``, the ``construct_hostname`` call for host inference, or any
+    ``parse_account`` stripping) — otherwise inferred values are
+    indistinguishable from user-supplied ones and ``host_provided`` /
+    ``account_provided`` are no longer trustworthy.
+    """
+    account = kwargs.get("account")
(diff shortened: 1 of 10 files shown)


sfc-gh-boler:
Pushed a CI fix for the two failing required checks:

1. **`Check linting` (black)** — black 24.10.0 (the pinned hook version) wanted to reformat `connection.py`, `aio/_connection.py`, and `test/unit/test_connection_identifier_shape_telemetry.py`. Re-ran the formatter; no functional changes.

2. **`Check for corresponding async changes`** — the gate flagged the two new sync-side unit-test files under `test/unit/` with no counterpart under `test/unit/aio/`. Added `test/unit/aio/test_connection_identifier_shape_telemetry_async.py` mirroring the sync emission tests, but using `AsyncMock` + `@pytest.mark.asyncio` to exercise the async override in `aio/_connection.py::_log_connection_identifier_shape`. Did not add a separate `test_connection_identifier_shape_async.py`: `record_input_shape` is a pure function with no sync/async branching, so the shape-truth-table is identical in both worlds and adding a clone would just be visual noise — the gate is satisfied by any file under `test/unit/aio/`.

All 45 unit tests (sync + async) pass locally. No production-code changes in this push.

sfc-gh-boler:
Follow-up CI fix: `test/integ/test_connection.py::test_disable_telemetry` was counting telemetry batch entries (5 with the cursor.execute path, 3 without). Adding the new `client_connection_identifier_shape` event bumps both counts by one. Updated assertions to 6 and 4 respectively and left `TODO(SNOW-3548350)` comments next to the bumped constants so they get reverted in the same removal pass as the emission.

sfc-gh-boler:
Pushed review-round-1 fixes addressing reviewer findings on cross-driver consistency with gosnowflake (the spec).

## Changes

- **`account_org_provided` logic aligned to Go** — `record_input_shape` now gates the dot-split on `dot_index > 0` (matching gosnowflake's `recordAccountShape` in `internal/config/dsn.go`). Pathological leading-dot inputs like `.us-east-1` now correctly report `account_org_provided=true` / `account_with_region=false` rather than the previous (Python-only) `false` / `true` interpretation. The dash search runs over the full string when there is no real account/region split, mirroring the Go reference byte-for-byte.
- **`SF_TELEMETRY_DISABLE_CONNECTION_SHAPE` parsing tightened** — dropped the `.strip()` to match gosnowflake's `strings.EqualFold(os.Getenv(...), "true")` and the JDBC `equalsIgnoreCase` siblings. Whitespace-wrapped `" true "` no longer disables the emission; only an exact case-insensitive `"true"` does. Sync and async branches both updated, with a parity comment pointing at the Go counterpart.
- **`build_shape_telemetry_message(shape)` extracted** — the wire-format payload is now a single pure builder in `connection_identifier_shape.py`. Sync (`SnowflakeConnection._log_connection_identifier_shape`) and async (`aio.SnowflakeConnection._log_connection_identifier_shape`) both call through it, so the five payload keys stay byte-identical across the two code paths.
- **Test coverage expanded** — added `leading_dot_account_does_not_split` and `trailing_dot_account_with_empty_region` cases that pin the dash-rule edges; added two builder tests that lock the lowercase-boolean wire format; updated env-switch parametrize tables to assert \` true \`, \`true \`, \` true\` no longer disable. Sync + async test files mirror each other.

All 53 unit tests in `test/unit/test_connection_identifier_shape*.py` + `test/unit/aio/test_connection_identifier_shape_telemetry_async.py` pass locally. \`black\` + \`isort\` both clean.

src/snowflake/connector/_connection_identifier_shape.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python
```
sfc-gh-turbaszek, line 1: Please make this module private by prefixing file name with `_`. This logic is probably not intended for end users.

sfc-gh-turbaszek (approved):
One comment about module name otherwise looks good

src/snowflake/connector/_connection_identifier_shape.py
sfc-gh-boler: Done — renamed `connection_identifier_shape.py` → `_connection_identifier_shape.py` and updated all four import sites (`connection.py`, `aio/_connection.py`, the two unit tests, and the async unit test) plus the unit-test docstring. The module attribute on the connection instance (`self._connection_identifier_shape`) was already private, so the naming is now consistent end-to-end. All 53 unit tests still pass; pushed as 51186739.

PR #2881, opened 2026-05-20
Enable free-threaded (no-GIL) CPython support for the Arrow iterator

sfc-gh-tycole:
## Summary
Makes the connector's compiled Arrow iterator safe to import on free-threaded CPython (3.13t / 3.14t) without re-enabling the GIL.

1. Declares nanoarrow_arrow_iterator free-threading-compatible via Cython (Py_MOD_GIL_NOT_USED);
2. bumps build-time Cython to >=3.1.
3. Serializes Python object caches in the Arrow iterator C++ backend with std::call_once — required because PyGILState_Ensure is not a mutex on no-GIL builds. 
4. Opts into cibuildwheel free-threaded wheel builds and adds a FREETHREADING=1 conda recipe path for cp314t artifacts.
5. Adds subprocess import tests that assert GIL state after first import (warning check on 3.13+, disabled/enabled checks where applicable).
Adds PyPI classifier Programming Language :: Python :: Free Threading :: 2 - Beta.

## Test plan

- [x] Local build verified: `python3 setup.py build_ext --inplace` succeeds with only pre-existing `-Wunused-const-variable` warnings in bundled flatcc-generated `nanoarrow_ipc.c`.
- [x] Rebuilt extension imports cleanly under CPython 3.12 with zero warnings.
- [x] New regression test skips cleanly on 3.12 with the expected reasons (`sys._is_gil_enabled` requires 3.13+).
- [ ] CI matrix exercises all three regression-test branches on 3.13 / 3.14 / 3.14t (requires a 3.14t entry in `generated_pr_matrix.json`, tracked separately).


sfc-gh-tycole:

test/unit/test_freethreading_import.py
```diff
@@ -0,0 +1,133 @@
...
+        f"`# cython: freethreading_compatible = True` (Cython) or the "
+        f"equivalent Py_mod_gil slot (hand-written extensions).\n"
+        f"--- subprocess stderr ---\n{proc.stderr}\n"
+        f"--- subprocess stdout ---\n{proc.stdout}"
+    )
+
+
+@pytest.mark.skipif(
+    not _HAS_GIL_INTROSPECTION or sys._is_gil_enabled(),
+    reason="requires a free-threaded interpreter with the GIL disabled "
```
sfc-gh-fpawlowski: We have such checks for multiple tests - can we unify them - maybe in some fixture?
Also can we unify messages in "reason" for tests that are skipped for the same reason?

DESCRIPTION.md
```diff
@@ -15,6 +15,8 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
   - Removed pandas upper bound dependency constraint on the `[pandas]` extra to allow installation of pandas 3.0.0 and later.
   - Fixed S3 storage client to correctly handle 307/308 (method-preserving) and 301/302 (GET/HEAD only) redirects by disabling automatic redirect following and re-signing each request with AWS SigV4 credentials for the redirect target. The region is updated from the `x-amz-bucket-region` response header on each redirect. Redirects are capped at 5 hops.
   - Added support for Python 3.14t (free-threaded). Note: Python 3.14t CI testing excludes `win_arm64` (no `cryptography` wheels available) and `mitmproxy` proxy tests on all platforms (transitive dependencies `aioquic`/`pylsqpack` lack free-threaded-compatible wheels).
+  - Fixed Arrow iterator compatibility with free-threaded Python (3.13t/3.14t). Importing the connector on a `--disable-gil` interpreter no longer re-enables the GIL, and concurrent Arrow result conversion is now thread-safe.
+  - Published free-threaded (`cp314t`) wheels and conda packages alongside the existing 3.14 builds.
```
sfc-gh-fpawlowski: Maybe lets not mention publishing in this PR as it does not add it - we will add such entry in the PR after extending Jenkins CI/CD with 3.14t publishing.

pyproject.toml
```diff
@@ -4,14 +4,23 @@ requires = [
     # and may be stricter than the version required in `setup.py`
     "setuptools>=40.6.0",
     "wheel",
-    "cython",
+    # cython>=3.1 is required for the `freethreading_compatible` directive
+    # used by src/snowflake/connector/nanoarrow_cpp/ArrowIterator/nanoarrow_arrow_iterator.pyx
+    "cython>=3.1",
```
sfc-gh-fpawlowski, line 9: Do we suppose customers building from source are rare enough for us to be able to bump this without breaking their envs (in case they have cython pinned down)?

Cause cython 3.0.0 is quite recent after all.

sfc-gh-fpawlowski (commented):
LGTM, left few nits and one discussion point

sfc-gh-fpawlowski:
One more thing - this is the first time python driver will be able to process data in such a concurrent manner. Therefore, I think we are currently lacking - due to the previous, synchronous nature of the driver with GIL - any proper stress tests to at least flag if there are data races in critical user journeys.

I think adding such tests should be a release gate for this feature - since otherwise we would just hope it won't cause CDIs, simply because it has built correctly.

test/unit/test_freethreading_import.py
sfc-gh-pczajka: How about requiring cython>=3.1 only on python 3.14?

sfc-gh-tycole: Sure - Added fixture for GIL/NoGIL

DESCRIPTION.md
sfc-gh-tycole: Yes that is good call.

pyproject.toml
sfc-gh-tycole: Yeah I would think since cython is only build time dependency, I wouldn't think it would present much of an issue. Although if you guys have different insight from previous upgrades then we could reconsider. 

Regarding above comment (I think meant here) from @sfc-gh-pczajka: 
> How about requiring cython>=3.1 only on python 3.14?

I don't think that is possible given its required for a pyx free threading flag. 

From some searching we could "Cynthonize" the pyx/cpp to get rid of Cython dependency altogether into native modules but that might be a bit of a maintainability burden.

sfc-gh-tycole:
> any proper stress tests to at least flag if there are data races in critical user journeys.

@sfc-gh-fpawlowski Great call. There is some int test concurrency coverage but it was a bit patchy. Extended the concurrency integration tests to include the fetch path, which I believe is primary extent of changes. Also included concurrency variants for the local/e2e stress tests and verified it looks green.

test/conftest.py
```diff
@@ -24,6 +25,36 @@
...
 )
 
+# ``sys._is_gil_enabled`` was added in Python 3.13; on older interpreters there
+# is no notion of a free-threaded build, so any GIL-state-specific test is
+# skipped.
+_HAS_GIL_INTROSPECTION = hasattr(sys, "_is_gil_enabled")
+
+
+@pytest.fixture
+def gil_enabled():
```
sfc-gh-fpawlowski: nit: if we wanted to express the intent name like "require_gil_enabled / require_enabled_gil" could fit well

test/integ/test_concurrent_converter.py
```diff
@@ -0,0 +1,61 @@
...
+        select seq4() as n,
+               dateadd(day, seq4(), to_date('2020-01-01')) as d,
+               time_from_parts(mod(seq4(), 24), mod(seq4(), 60), mod(seq4(), 60)) as t,
+               (seq4() / 1000.0)::decimal(18, 3) as dec,
+               'row_' || seq4()::string as s
+        from table(generator(rowcount => {rowcount}))
+        order by n
+    """
+    for _ in range(3):
+        with conn_cnx(client_prefetch_threads=4) as cnx:
```
sfc-gh-fpawlowski: nit: could we use some constant for this level of concurrency in chunk download == 4?

test/integ/test_concurrent_fetch.py
```diff
@@ -0,0 +1,84 @@
...
+def _barrier_map(fn, n=N_THREADS):
+    """Run ``fn(i, barrier)`` on ``n`` threads released together; collect results.
+
+    The barrier maximizes the chance the threads collide (e.g. on a cold converter
+    cache) versus starting staggered like ``ThreadPoolExecutor.map``. Returns
+    (results, errors) with per-thread exceptions captured.
+    """
+    results: list = [None] * n
+    errors: list = [None] * n
+    barrier = threading.Barrier(n)
```
sfc-gh-fpawlowski, line 35: That's really nice idea : )

test/integ/test_concurrent_converter.py
```diff
@@ -0,0 +1,61 @@
...
+            rows = cnx.cursor().execute(query).fetchall()
+        assert len(rows) == rowcount
+        assert rows[0][0] == 0 and rows[-1][0] == rowcount - 1
+        first = rows[0]
+        assert isinstance(first[1], datetime.date)
```
sfc-gh-fpawlowski: These `isinstance` checks confirm type but not value. If `DateConverter::initPyDatetimeDate` returned a corrupted reference — e.g. two threads both passing the empty check before `call_once` serialises them — the result would still be a `datetime.date`, just the wrong one.

`test_concurrent_fetch.py` has the right shape: fetch a reference result once (single-threaded, no prefetch), then assert `rows == reference` on each concurrent iteration. Worth applying the same pattern here so the test actually guards the `call_once` change.

test/integ/test_concurrent_converter.py
```diff
@@ -0,0 +1,61 @@
...
+               dateadd(second, seq4(), '2020-01-01'::timestamp_ntz) as ts
+        from table(generator(rowcount => {rowcount}))
+        order by n
+    """
+    with conn_cnx(numpy=True, client_prefetch_threads=4) as cnx:
+        rows = cnx.cursor().execute(query).fetchall()
+    assert len(rows) == rowcount
+    assert rows[0][0] == 0 and rows[-1][0] == rowcount - 1
+    first = rows[0]
+    assert isinstance(first[0], np.int64)
```
sfc-gh-fpawlowski, line 78: Same point as the test above: `isinstance(first[0], np.int64)` survives a data-corrupting race. Add a single-threaded reference fetch and assert the concurrent result equals it.

test/unit/test_detect_platforms.py
```diff
@@ -487,3 +489,53 @@ def test_platform_detection_suppresses_all_library_logs(
...
+                with _suppress_platform_detection_logs():
+                    pass
+
+        threads = [threading.Thread(target=hammer) for _ in range(16)]
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
+
+        for name in _LOGGERS_TO_SUPPRESS:
```
sfc-gh-fpawlowski: `== logging.NOTSET` implicitly assumes every logger in `_LOGGERS_TO_SUPPRESS` starts at `NOTSET`. If a prior test left, say, `botocore.utils` at `WARNING`, the implementation would correctly restore it to `WARNING` — but this assertion would fail spuriously, making "race bug still present" indistinguishable from "another test changed a logger level".

There's also a related gap: the `finally` only restores `target`, leaving the other loggers in `_LOGGERS_TO_SUPPRESS` without save/restore, which can leak state to later tests.

A more robust pattern:

```python
originals = {n: logging.getLogger(n).level for n in _LOGGERS_TO_SUPPRESS}
try:
    # ... hammer ...
    for name, lvl in originals.items():
        assert logging.getLogger(name).level == lvl, f"{name} not restored"
finally:
    for name, lvl in originals.items():
        logging.getLogger(name).setLevel(lvl)
```

sfc-gh-fpawlowski (commented):
Left few more comments; apart from them looks good

test/conftest.py
sfc-gh-tycole: Agree - Updated

test/integ/test_concurrent_converter.py
sfc-gh-tycole: Yeah that is fair, changed to reference check similar to other instance.

test/unit/test_detect_platforms.py
sfc-gh-tycole: Added test fix - Thanks for patch!

src/snowflake/connector/nanoarrow_cpp/Logging/logging.hpp
```diff
@@ -30,6 +31,9 @@ class Logger {
 
  private:
   py::UniqueRef m_pyLogger;
+  // Guards once-only init of m_pyLogger; on free-threaded builds (3.13t+)
```
sfc-gh-turbaszek:
```suggestion
  // Guards once-only init of m_pyLogger; on free-threaded builds (3.14t+)
```

src/snowflake/connector/platform_detection.py
```diff
@@ -33,6 +34,11 @@
     "urllib3.connectionpool",
 ]
 
+# Guards the process-global logger-level mutation against concurrent connections.
+_suppress_lock = threading.Lock()
+_suppress_depth = 0
```
sfc-gh-turbaszek, line 39: Will the depth be different from 0 or 1?

sfc-gh-turbaszek (approved):
Minor comment, otherwise looks reasonable

sfc-gh-fpawlowski (approved):
Lgtm, rebased on top of the fixed main branch

src/snowflake/connector/platform_detection.py
sfc-gh-tycole: Yes on concurrent calls, it is basically a reference count to reset the loggers to original level when the last one is done.

PR #2889, opened 2026-05-25
NO-SNOW: drop EOL Python 3.9 support

sfc-gh-pczajka:
## Summary
- Drop support for Python 3.9. Minimum supported version is now Python 3.10.
- Update `setup.cfg`, `tox.ini`, GitHub workflows, CI scripts, Docker images, prober matrix, and docs accordingly.
- Migrate the FIPS testing image from CentOS 8 (EOL, no `python3.10+` package) to Rocky Linux 9; FIPS coverage now runs on Python 3.11.
- Remove the runtime `sys.version_info < (3, 10)` guard in `aio/_session_manager.py` (now structurally enforced by `python_requires`).

## Why
PR #2888 (urllib3 2.7.0 bump) is blocked because upstream urllib3 dropped 3.9 and now imports `ssl.VERIFY_X509_PARTIAL_CHAIN` unconditionally — a constant added in CPython 3.10. Rather than maintaining ad-hoc back-ports across future urllib3 bumps, drop 3.9 here. #2888 stacks on top.

## Notes
- PR test matrix stays at Python 3.10 + 3.14 only. The workflow injects `(manylinux_x86_64, 3.11)` into the build matrix so the FIPS-required wheel is built on PR runs without expanding the test matrix.
- `test/aiodep/unsupported_python_version.py` is now an orphan (its only invoker, the `[testenv:aio-unsupported-python]` env, is removed). Left in place for a follow-up cleanup.


.github/workflows/build_test.yml
```diff
@@ -323,7 +325,7 @@ jobs:
         shell: bash
 
   test-fips:
-    name: Test FIPS linux-3.9-${{ matrix.cloud-provider }}
+    name: Test FIPS linux-3.11-${{ matrix.cloud-provider }}
```
sfc-gh-fpawlowski, line 328: Not sure so leaving as smth to confirm - was there a reason FIPS job was on the lowest supported version? I thought we did that based on the fact that newer versions may have some issues patched that the older ones face so we needed to check on the lowest to know if we are FIPS compliant

sfc-gh-fpawlowski (approved):
Looks good 👍

PR #2892, opened 2026-05-28
SNOW-3392651: replace URL_VALIDATOR regex with urllib.parse in is_valid_url

sfc-gh-turbaszek:
## Summary

- Replaces the hand-rolled `URL_VALIDATOR` regex in `url_util.py` with `urllib.parse.urlparse` 
- Removes the `URL_VALIDATOR` constant entirely (no other callers)
- Incidentally fixes a secondary bug where the port sub-pattern used `(0-9)` (literal string match) instead of `[0-9]` (character class)

## Test plan

- [ ] Existing `test/unit/test_url_util.py` tests pass (verified locally: 3/3 passed)
- [ ] Manual verification: valid HTTP/HTTPS URLs return `True`, non-HTTP and malformed URLs return `False`


DESCRIPTION.md
```diff
@@ -8,6 +8,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
 
 # Release Notes
 - v4.6.0(May 28,2026)
+  - Improved URL validation reliability by replacing the hand-rolled regex in `is_valid_url()` with `urllib.parse.urlparse` (SNOW-3392651).
```
sfc-gh-fpawlowski, line 12: Should it be added as new release notes, not to the old ones?

src/snowflake/connector/url_util.py
```diff
@@ -11,21 +11,17 @@
...
-        url: the URL that needs to be validated
-
-    Returns:
-        true/ false depending on whether the URL is valid or not
-    """
-    return bool(URL_VALIDATOR.match(url))
+    """Confirms if the provided URL is a valid HTTP/HTTPS URL."""
+    if not isinstance(url, str):
+        return False
+    if any(c < "\x20" for c in url):
```
sfc-gh-fpawlowski: Just out of curiosity - could it break the urlparse itself or would urlparse not catch those as well?

test/unit/test_url_util.py
sfc-gh-fpawlowski: nit: Would be cool with pytest mark parametrize

test/unit/test_url_util.py
```diff
@@ -25,6 +25,22 @@ def test_url_validator():
...
+    assert not is_valid_url("https://\x00evil.com")
+    assert not is_valid_url("https://evil.com/path\x01")
+    assert not is_valid_url("https://evil.com\nnewline")
+
+    # embedded credentials are accepted (urlparse handles them safely)
+    assert is_valid_url("https://user:pass@host.snowflakecomputing.com")
+
+    # IPv4 and IPv6 literals
+    assert is_valid_url("https://192.168.1.1/path")
+    assert is_valid_url("https://[::1]:8080/path")
```
sfc-gh-fpawlowski: nit: Leaving here just a few more cases that we could add to this set if we wanted to extend it; not a must have - leaving for your decision in case you needed more examples
```

# missing / unsupported scheme
assert not is_valid_url("ftp://evil.com")
assert not is_valid_url("javascript:alert(1)")
assert not is_valid_url("mailto:test@example.com")
assert not is_valid_url("//evil.com/path")
assert not is_valid_url("host.snowflakecomputing.com")

# empty / malformed host
assert not is_valid_url("https://")
assert not is_valid_url("https:///path")
assert not is_valid_url("https://:443")
assert not is_valid_url("https://exa mple.com")
assert not is_valid_url("https://.com")
assert not is_valid_url("https://com")

# whitespace tricks
assert not is_valid_url(" https://evil.com")
assert not is_valid_url("https://evil.com ")
assert not is_valid_url("\thttps://evil.com")
assert not is_valid_url("https://evil.com\r\nHost: attacker.com")

# ports
assert is_valid_url("https://example.com:443/path")
assert is_valid_url("https://example.com:0/path")      # if your validator allows urlparse-valid ports
assert not is_valid_url("https://example.com:65536")
assert not is_valid_url("https://example.com:-1")
assert not is_valid_url("https://example.com:abc")

# localhost / internal names
assert is_valid_url("https://localhost")
assert is_valid_url("https://localhost:3000/callback")
assert is_valid_url("https://intranet/path")

# query / fragment handling
assert is_valid_url("https://example.com/path?x=1&y=2")
assert is_valid_url("https://example.com/path#fragment")
assert is_valid_url("https://example.com/path?next=https://evil.com")

# case / normalization
assert is_valid_url("HTTPS://EXAMPLE.COM/Path")
assert is_valid_url("https://example.com.")
assert is_valid_url("https://xn--bcher-kva.example")   # punycode IDN

# malformed IPv literals
assert not is_valid_url("https://[::1")
assert not is_valid_url("https://::1/path")
assert not is_valid_url("https://[gggg::1]/path")

# repeated dots / odd but parseable hostnames
assert not is_valid_url("https://example..com")
assert is_valid_url("https://sub.sub2.example.com/path")

# percent-encoded dangerous chars
assert is_valid_url("https://example.com/%00")
assert is_valid_url("https://example.com/%0a")

```

sfc-gh-fpawlowski (approved):
Lgtm; left few comments

DESCRIPTION.md
sfc-gh-turbaszek: Good catch — moved the entry under a new "Upcoming Release" section in 235d290.

src/snowflake/connector/url_util.py
sfc-gh-turbaszek: Good question — `urlparse` would not catch them. Verified locally on Python 3.10: `urlparse("https://\x00evil.com")` returns `scheme="https", netloc="\x00evil.com"` (NUL byte preserved); inputs containing `\n`, `\r`, or `\t` are silently stripped per WHATWG URL parsing rules — e.g. `urlparse("https://evil.com\nnewline").netloc == "evil.comnewline"` and `urlparse("\thttps://evil.com").scheme == "https"`. Both behaviours would let smuggled control characters slip through, so the explicit pre-check is necessary.

test/unit/test_url_util.py
sfc-gh-turbaszek: Done in 235d290 — split into `test_url_validator_accepts` / `test_url_validator_rejects` with `pytest.mark.parametrize`.

test/unit/test_url_util.py
sfc-gh-turbaszek: Thanks for the list! In 235d290 I folded in the cases that the current implementation already handles correctly: unsupported schemes (`ftp`, `javascript`, `mailto`, scheme-relative `//`, no scheme), `https://` and `https:///path` (empty host), leading whitespace / `\t` / CRLF header injection, malformed IPv6 with missing closing bracket, query/fragment handling, case + punycode normalization, and percent-encoded path bytes.

I left out the cases that would require tightening the validator beyond the ReDoS-fix scope of this PR — `urlparse` is intentionally permissive and currently accepts `https://:443`, `https://exa mple.com`, `https://.com`, `https://com`, `https://example.com:65536`, `:-1`, `:abc`, `https://::1/path`, `https://[gggg::1]/path`, `https://example..com`, and trailing whitespace. These are real gaps but fall under stricter host/port validation, which is a separate concern from replacing the regex; happy to file a follow-up if you want stricter semantics.

PR #2886, opened 2026-05-21
Skip offline_access scope for Snowflake custom OAuth

tanyasri:
## Description

Fixes # 2885.

`AuthByOAuthBase.__init__` unconditionally appends `offline_access` to
the OAuth scope when `refresh_token_enabled=True`. This is correct for
OIDC-compliant external IdPs but breaks Snowflake custom OAuth, whose
authorization server rejects `offline_access` with:

> The requested scope is invalid.

Per Snowflake's [custom OAuth docs](https://docs.snowflake.com/en/user-guide/oauth-custom),
the scope used to request offline access is `refresh_token`, not
`offline_access`.

## Fix

Append `offline_access` only when the token endpoint is **not** a
Snowflake host (`*.snowflakecomputing.com` / `*.snowflakecomputing.cn`)
**and** the caller hasn't already put `refresh_token` in scope.

External IdP behavior is unchanged.

## Tests

Added `test/unit/test_oauth_base_scope.py` covering:
- External IdP + refresh enabled → `offline_access` appended (existing behavior)
- External IdP + empty scope → `offline_access` appended
- Snowflake custom OAuth (`.com` and `.cn`) → not appended
- Caller-supplied `refresh_token` → not appended
- `refresh_token_enabled=False` → scope unchanged

PR #2894, opened 2026-05-28
Fix OAuth infinite loop bug

sfc-gh-mborins:
## Summary

Fixes infinite loop when OAuth tokens expire, where users get stuck in repeated macOS Keychain prompts and browser auth flows that never complete.

## Root Cause

1. `prepare()` reads expired token from keychain on every call
2. Snowflake server rejects with `ReauthenticationRequest`
3. `reauthenticate()` calls `conn.authenticate_with_retry(self)`
4. This calls `_authenticate()` → `prepare()` again
5. `prepare()` reads same expired token from keychain → **LOOP**

Additionally, the old code pattern of calling `token_cache.remove()` then `token_cache.store()` destroyed macOS Keychain Access Control Lists (ACL), causing repeated permission prompts.

## Solution

**Core principle:** Read from keychain ONCE per connection, write ONCE if tokens updated.

### 1. Break the Infinite Loop

- **`reauthenticate()` calls `_request_tokens()` directly** instead of `conn.authenticate_with_retry()` 
- This breaks the loop: no longer cycles back through `prepare()` → `_authenticate()` → `reauthenticate()` → `prepare()`

### 2. One-Time Token Loading

- **New method:** `_load_tokens_from_cache()` loads both access + refresh tokens at connection start
- **Modified:** `prepare()` calls `_load_tokens_from_cache()` exactly once per connection
- **Prevents repeated keychain reads:** `_tokens_loaded_from_cache` flag ensures single load

### 3. Preserve macOS Keychain ACL

- **New method:** `_store_tokens()` - **only calls `store()`, never `remove()`**
- **Why this matters:** Calling `remove()` then `store()` destroys the Keychain ACL entry
- **Result:** No more repeated keychain permission prompts on macOS
- **All token updates** go through `_store_tokens()`: initial auth, refresh, and reauthentication

### 4. Cleanup

- **Removed deprecated methods:** `_pop_cached_access_token()`, `_pop_cached_refresh_token()`, `_reset_access_token()`, `_reset_refresh_token()`, `_reset_cached_token()`
- **Removed flag-based workarounds:** Replaced scattered `_tried_cache_this_session` checks with clean state machine

### 5. Conditional offline_access Scope

- Only added for external IdPs (Okta, Azure AD, etc.)
- NOT added when Snowflake is the OAuth provider
- Snowflake returns refresh tokens via `OAUTH_ISSUE_REFRESH_TOKENS` security integration setting, not the `offline_access` scope

## Files Changed

- **`src/snowflake/connector/auth/_oauth_base.py`** (166 lines changed)
  - Added `_load_tokens_from_cache()` - one-time token loading
  - Added `_store_tokens()` - ACL-preserving storage
  - Modified `prepare()` - uses `_load_tokens_from_cache()`
  - Modified `reauthenticate()` - calls `_request_tokens()` directly (breaks loop)
  - Modified `_do_refresh_token()` - uses `_store_tokens()`
  - Removed 6 deprecated methods

- **`src/snowflake/connector/auth/oauth_code.py`** (3 lines changed)
  - Pass `is_snowflake_as_idp` to parent `__init__()` for conditional `offline_access`

- **`test/unit/test_oauth_infinite_loop_fix.py`** (375 lines, new file)
  - 11 comprehensive unit tests

## Testing

✅ **TestOAuthTokenCacheLoading** (2 tests)
- Verifies `_load_tokens_from_cache()` loads exactly once per connection
- Verifies both access and refresh tokens are loaded

✅ **TestOAuthTokenCacheStorage** (2 tests)  
- **Verifies `_store_tokens()` NEVER calls `remove()`** (ACL preservation)
- Verifies both in-memory and cache are updated

✅ **TestOAuthReauthenticateNoLoop** (2 tests)
- Verifies `reauthenticate()` calls `_request_tokens()` directly, NOT `prepare()`
- Verifies refresh token is used from memory, not re-read from cache

✅ **TestOAuthOfflineAccessScope** (3 tests)
- Verifies `offline_access` NOT added for Snowflake as IdP
- Verifies `offline_access` IS added for external IdPs
- Verifies `offline_access` NOT added when refresh tokens disabled

✅ **TestOAuthPrepareUsesCache** (2 tests)
- Verifies `prepare()` uses cached tokens without calling `_request_tokens()`
- Verifies `prepare()` requests new tokens when cache empty

✅ **Integration test** validates end-to-end flow with cached tokens

## Breaking Changes

**None.**
- Defaults unchanged (`client_store_temporary_credential`, `oauth_enable_refresh_tokens` remain `False`)
- Fix works regardless of whether caching is enabled
- Existing tokens in keychain remain compatible

### Before Fix
```
First connection:  2+ keychain prompts, 1 browser prompt, connects
Second connection: 2+ keychain prompts, 1 browser prompt, INFINITE LOOP
```

### After Fix
```
First connection:  1 keychain prompt (if unlocked: 0), 1 browser prompt, connects
Second connection: 0 keychain prompts, 0 browser prompts, connects
Expired token:     1-2 keychain reads (access + refresh), 0-1 browser prompts
```

## Test Plan

```python
import snowflake.connector

# First connection (browser auth)
    account="...",
    user="...",
    authenticator="oauth_authorization_code",
    client_store_temporary_credential=True,  # Must opt-in
    oauth_enable_refresh_tokens=True,  # Must opt-in
)
conn.close()

# Second connection (cached token, no browser, no loop)
conn2 = snowflake.connector.connect(...)
conn2.close()
```

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


sfc-gh-turbaszek:
To have a full solution we need https://github.com/snowflakedb/snowflake-connector-python/pull/2886 in

sfc-gh-turbaszek (approved):

PR #2909, opened 2026-06-23
SNOW-3665226: prevent fully-qualified DDL from polluting schema/database cache

sfc-gh-turbaszek:
## Summary
- `finalDatabaseName`/`finalSchemaName` in the server query response reflects the *object's* schema for fully-qualified DDL statements, not the session's current schema — causing silent cache corruption when a session has no schema set
- Guard the `_database`/`_schema` cache update: only update when the cache already has a value (legitimate refresh), or when the SQL is a `USE` statement (the only operation that explicitly switches session context)
- Same fix applied to both `connection.py` (sync) and `aio/_connection.py` (async)

## Test plan
- [x] New integration test `test_fqn_ddl_does_not_pollute_schema_cache` reproduces the bug (fails before fix, passes after)
- [x] Existing schema/connection integration tests remain green

Fixes SNOW-3665226


src/snowflake/connector/connection.py
```diff
@@ -1989,10 +1989,13 @@ def cmd_query(
             ret["data"] = {}
         if _update_current_object:
             data = ret["data"]
+            _is_context_switch = sql.strip().upper().startswith("USE ")
```
sfc-gh-pczajka: Nit (readability): consider `is_context_switch` without the leading underscore. In this file the `_` prefix is reserved for internal-knob *parameters* (`_no_results`, `_update_current_object`, `_no_retry`) and module-private helpers; every plain local here (`data`, `ret`, `client`, `url_parameters`) is unprefixed, so a `_`-prefixed local reads like one of those params. Applies to all three occurrences here.

src/snowflake/connector/connection.py
```diff
@@ -1989,10 +1989,13 @@ def cmd_query(
...
             if "finalDatabaseName" in data and data["finalDatabaseName"] is not None:
-                self._database = data["finalDatabaseName"]
+                if self._database is not None or _is_context_switch:
+                    self._database = data["finalDatabaseName"]
             if "finalSchemaName" in data and data["finalSchemaName"] is not None:
-                self._schema = data["finalSchemaName"]
+                if self._schema is not None or _is_context_switch:
+                    self._schema = data["finalSchemaName"]
```
sfc-gh-pczajka: Optional (simplicity): the nested `if` collapses to a single condition, and `data.get(...) is not None` folds the `"x" in data and data["x"] is not None` pair:

```suggestion
            if data.get("finalDatabaseName") is not None and (
                self._database is not None or _is_context_switch
            ):
                self._database = data["finalDatabaseName"]
            if data.get("finalSchemaName") is not None and (
                self._schema is not None or _is_context_switch
            ):
                self._schema = data["finalSchemaName"]
```

(A loop/helper over all four fields would obscure the deliberate db/schema-vs-warehouse/role asymmetry, so keeping these explicit is right.)

test/integ/test_connection.py
```diff
@@ -2685,3 +2685,46 @@ def test_empty_result_stats(conn_cnx):
...
+                cur.execute(f"CREATE OR REPLACE TEMP VIEW {view_name} AS SELECT 1 AS x")
+            finally:
+                cur.execute(f"DROP VIEW IF EXISTS {view_name}")
+
+            # Cache must still be None — the session schema never changed.
+            assert cnx._schema is None, (
+                f"connector cache was polluted: _schema={cnx._schema!r}, "
+                f"but no USE SCHEMA was issued"
+            )
+            assert cnx._database is not None  # database was set at connect time
```
sfc-gh-pczajka: This assertion is effectively a tautology: `_database` is set at connect time and the object here lives in the *same* database, so it stays non-`None` whether or not the guard works. The `_database` half of the bug (which this PR also fixes) is therefore unasserted. Suggest capturing `db_before = cnx._database` before the `CREATE` and asserting `cnx._database == db_before` afterwards — ideally with an object in a *different* database so a regression actually flips the value.

test/integ/test_connection.py
```diff
@@ -2685,3 +2685,46 @@ def test_empty_result_stats(conn_cnx):
...
+            # Touch a fully-qualified object — this must not change the session context.
+            try:
+                cur.execute(f"CREATE OR REPLACE TEMP VIEW {view_name} AS SELECT 1 AS x")
+            finally:
+                cur.execute(f"DROP VIEW IF EXISTS {view_name}")
+
+            # Cache must still be None — the session schema never changed.
+            assert cnx._schema is None, (
+                f"connector cache was polluted: _schema={cnx._schema!r}, "
+                f"but no USE SCHEMA was issued"
```
sfc-gh-pczajka: Nit: this string has no placeholders, so the `f` prefix is dead (Ruff F541).

```suggestion
                "but no USE SCHEMA was issued"
```

test/integ/test_connection.py
```diff
@@ -2685,3 +2685,46 @@ def test_empty_result_stats(conn_cnx):
...
+    """SNOW-3665226: fully-qualified DDL must not mutate the connector's cached
+    current schema / database when the session has no schema set.
+
+    Repro: open a connection without a schema, execute a DDL that references a
+    fully-qualified object (db.schema.name).  The connector was updating its
+    _schema/_database cache from the server's finalSchemaName/finalDatabaseName
+    response fields, which reflect the *object's* schema — not the session's
+    current schema — causing get_current_schema() to diverge from
+    SELECT CURRENT_SCHEMA() on the server.
+    """
```
sfc-gh-pczajka: Nit (simplicity): the root-cause detail is already in the changelog and the ticket; the docstring can be tighter:

```suggestion
    """SNOW-3665226: a fully-qualified DDL must not populate the connector's
    cached _schema/_database when the session has none set (finalSchemaName/
    finalDatabaseName reflect the referenced object, not the session context).
    """
```

DESCRIPTION.md
```diff
@@ -16,6 +16,7 @@ Source code is also available at: https://github.com/snowflakedb/snowflake-conne
...
   - Added native AKS (Azure Kubernetes Service) workload identity support. When running on AKS with workload identity configured, the connector automatically uses `WorkloadIdentityCredential` to authenticate via the injected service account credentials. OIDC backward compatibility is also supported.
   - Added the `workload_identity_aws_use_outbound_token` connection option (default `false`) to opt into AWS WIF JWT attestation via STS `GetWebIdentityToken` instead of the default SigV4 `GetCallerIdentity` method.
+  - Fixed a bug where executing a fully-qualified DDL statement (e.g. `CREATE VIEW db.schema.obj`) on a session with no current schema would silently populate the connector's cached `_schema`/`_database`, causing `get_current_schema()` to diverge from the server's `CURRENT_SCHEMA()` and Snowpark temp-object qualification to target the wrong schema (SNOW-3665226).
```
sfc-gh-pczajka: Nit (simplicity): one 60-word sentence — splitting cause from effect reads easier:

```suggestion
  - Fixed a bug where a fully-qualified DDL statement (e.g. `CREATE VIEW db.schema.obj`) on a session with no current schema would populate the connector's cached `_schema`/`_database` from the referenced object's namespace. This made `get_current_schema()` diverge from the server's `CURRENT_SCHEMA()` and mis-qualified Snowpark temp objects (SNOW-3665226).
```

sfc-gh-pczajka (commented):
Code-quality sweep (non-blocking) — the fix itself is correct and well-scoped. A few readability/simplicity nits inline, plus one real coverage gap: the `_database` half is unasserted (see comment on the test).

sfc-gh-pczajka (approved):

PR #2912, opened 2026-06-24
SNOW-3680325: fix python-connector.log rotation on Windows

sfc-gh-pczajka:
create_log() called logging.basicConfig(filename=...) in addition to attaching a TimedRotatingFileHandler to each easy-logging logger. The basicConfig FileHandler (1) had no rotation and duplicated every record in the default (unmasked) format, and (2) kept the log file open, so on Windows the midnight rename in TimedRotatingFileHandler.doRollover() was blocked by exclusive file locking and rotation silently failed.

create_log() also runs on every connection, so each connect() stacked another TimedRotatingFileHandler on the same file, compounding the duplication and open-handle problem.

Drop basicConfig and use a single TimedRotatingFileHandler shared across the easy-logging loggers. Handler registration is now idempotent, so repeated connections do not re-stack handlers.


src/snowflake/connector/log_configuration.py
```diff
@@ -42,21 +47,29 @@ def parse_config_file(self, skip_config_file_permissions_check: bool = False):
...
+        # A single TimedRotatingFileHandler is shared across all easy-logging
+        # loggers, keeping exactly one open handle on the log file so rotation
+        # works on Windows. create_log() runs on every connection, so skip any
+        # logger that already has a rotating handler for this file to avoid
+        # stacking handlers (SNOW-3680325).
+        handler = None
+        for logger_name in EASY_LOGGING_LOGGERS:
+            logger = logging.getLogger(logger_name)
+            logger.setLevel(level)
+            if any(
```
sfc-gh-fpawlowski: `create_log()` runs on every `connect()`. How does this registration behave if two connections are opened concurrently (e.g. a connection-pool warm-up, or multiple threads each constructing a connection)? The "does this logger already have a rotating handler for this file?" check and the `addHandler` below don't look synchronized — could both calls observe "no handler" and each attach its own `TimedRotatingFileHandler`, leaving two handlers/open handles on the same file again? Is there a test that exercises concurrent `create_log()` calls?

src/snowflake/connector/log_configuration.py
```diff
@@ -42,21 +47,29 @@ def parse_config_file(self, skip_config_file_permissions_check: bool = False):
...
+                isinstance(h, TimedRotatingFileHandler)
+                and getattr(h, "baseFilename", None) == log_file_path
+                for h in logger.handlers
+            ):
+                continue
+            if handler is None:
```
sfc-gh-fpawlowski, line 86: What happens on a subsequent `create_log()` if only *some* of the three loggers already hold the shared handler (e.g. one was detached/reset elsewhere) and the others don't? Does the lazy `handler is None` creation then build a second handler instance and attach it to the remaining loggers — i.e. two distinct handlers on the same file? Is that partial-state case covered by a test?

src/snowflake/connector/log_configuration.py
```diff
@@ -42,21 +47,29 @@ def parse_config_file(self, skip_config_file_permissions_check: bool = False):
...
+                handler = TimedRotatingFileHandler(log_file_path, when="midnight")
+                handler.setLevel(level)
```
sfc-gh-fpawlowski: Since registration is idempotent, if a later connection's config specifies a different `level`, does the reused handler's level get updated — or does it stay at the level captured when it was first created (here)? Is it tested that changing the configured level between `create_log()` calls is actually reflected in what reaches `python-connector.log`?

src/snowflake/connector/log_configuration.py
```diff
@@ -42,21 +47,29 @@ def parse_config_file(self, skip_config_file_permissions_check: bool = False):
...
+
+        # A single TimedRotatingFileHandler is shared across all easy-logging
+        # loggers, keeping exactly one open handle on the log file so rotation
+        # works on Windows. create_log() runs on every connection, so skip any
+        # logger that already has a rotating handler for this file to avoid
+        # stacking handlers (SNOW-3680325).
+        handler = None
+        for logger_name in EASY_LOGGING_LOGGERS:
+            logger = logging.getLogger(logger_name)
+            logger.setLevel(level)
```
sfc-gh-fpawlowski: `logger.setLevel(level)` is re-applied to `botocore`/`boto3` on every connection — is it intended that this overrides a level the user may have set on those loggers themselves? Should the level only be (re)applied when the handler is first attached?

src/snowflake/connector/log_configuration.py
```diff
@@ -42,21 +47,29 @@ def parse_config_file(self, skip_config_file_permissions_check: bool = False):
...
-                    os.path.join(self.path, LOG_FILE_NAME), when="midnight"
-                )
-                ch.setLevel(logging.getLevelName(self.level))
-                ch.setFormatter(
-                    SecretDetector(
-                        "%(asctime)s - %(threadName)s %(filename)s:%(lineno)d - %(funcName)s() - %(levelname)s - %(message)s"
-                    )
-                )
-                logger.addHandler(ch)
+        if not self.save_logs:
```
sfc-gh-fpawlowski, line 58: With `logging.basicConfig(...)` removed, is it intended that the root logger level is no longer set and that records from libraries *other* than `snowflake.connector`/`botocore`/`boto3` (which previously propagated into `python-connector.log` via the root handler) are no longer captured? If so, is that behavior change called out in the release note / docs for users who relied on it?

src/snowflake/connector/log_configuration.py
```diff
@@ -42,21 +47,29 @@ def parse_config_file(self, skip_config_file_permissions_check: bool = False):
...
+            if any(
+                isinstance(h, TimedRotatingFileHandler)
+                and getattr(h, "baseFilename", None) == log_file_path
+                for h in logger.handlers
+            ):
+                continue
+            if handler is None:
+                handler = TimedRotatingFileHandler(log_file_path, when="midnight")
+                handler.setLevel(level)
+                handler.setFormatter(SecretDetector(LOG_FORMAT))
```
sfc-gh-fpawlowski, line 88: Now that every record funnels through this one shared handler — is it verified by a test that everything written to `python-connector.log` actually goes through `SecretDetector`? e.g. is there a case asserting that a credential-shaped value is masked in the file (and written exactly once), so the masking guarantee can't silently regress if the handler/formatter wiring changes later?

sfc-gh-fpawlowski (commented):
Thanks for consolidating this onto a single shared rotating handler with idempotent registration — clean shape, and it clearly addresses the duplication and Windows rotation. A few questions about edge cases and test coverage, left inline.

src/snowflake/connector/log_configuration.py
sfc-gh-fpawlowski: To sharpen this one: the bigger compatibility angle is callers who relied on the old `basicConfig(filename=...)` configuring the **root** logger — their own application loggers (and anything else propagating to root) used to be captured into `python-connector.log`, and the root level was set for them. After this change only `snowflake.connector`/`botocore`/`boto3` are written. Could that break setups/pipelines that enabled `save_logs` and then relied on the file containing their app's logs (or on the root level being set)? If that's an intended scope reduction, would it be worth a migration note, plus a test pinning exactly which loggers are (and are not) written to the file now?

src/snowflake/connector/log_configuration.py
sfc-gh-pczajka: Kept intentional. Easy logging is a config-file-driven feature whose explicit job is to apply the configured `level` to the connector and its AWS deps on each connection, so the config should win — and this matches the prior behavior (the old code also called `setLevel` every time). Switching to "only on first attach" would mean the configured level silently stops applying on reconnect, which is more surprising. The level is now also kept in sync on handler reuse, covered by `test_create_log_reflects_changed_level_on_reuse`.

src/snowflake/connector/log_configuration.py
sfc-gh-pczajka: Yes, this is an intended scope reduction. Capturing the root logger via `basicConfig` is exactly what produced the double-logging: a `snowflake.connector` record was written once by its own handler and again by the root `FileHandler`. You can't both capture-everything-via-root and not-duplicate with this setup, so scoping to `snowflake.connector`/`botocore`/`boto3` is the fix rather than an accident.

Called it out as a **Behavior change** callout in the release notes, and pinned the new scope with `test_create_log_only_writes_easy_logging_loggers` (asserts connector/botocore records reach the file and an unrelated library logger does not). If we ever want to keep capturing all libraries without duplication, the way to do it would be to attach the single shared handler to the root logger only (and drop it from the three named loggers, since they propagate up) — happy to do that in a follow-up if there's appetite, but it's a larger change than this fix intends.

src/snowflake/connector/log_configuration.py
sfc-gh-pczajka: Good catch — the check and `addHandler` weren't synchronized. Wrapped the whole find-or-create-then-attach in a module-level `_handler_lock`, so two concurrent `create_log()` calls can no longer both observe "no handler" and each attach their own. Added `test_create_log_concurrent_calls_attach_single_handler`, which fires 8 threads through a `threading.Barrier` into `create_log()` and asserts exactly one shared handler instance ends up on each logger.

src/snowflake/connector/log_configuration.py
sfc-gh-pczajka: Addressed. Instead of lazily creating a handler when a given logger has none, `create_log()` now searches all three loggers for an existing rotating handler for this file and reuses that instance, constructing a new one only if none exists anywhere. So the partial-state case (handler detached from some loggers but not others) reattaches the same instance rather than creating a second handler on the same file. Covered by `test_create_log_reuses_handler_for_partial_state`, which detaches the handler from `boto3` only, calls `create_log()` again, and asserts all three loggers share the original instance.

src/snowflake/connector/log_configuration.py
sfc-gh-pczajka: Fixed. `handler.setLevel(level)` is now applied on every `create_log()` call (not just at creation), so a later connection that changes the configured level updates the reused handler. `test_create_log_reflects_changed_level_on_reuse` sets INFO, then ERROR, and asserts both the shared handler and the loggers reflect ERROR after the second call.

src/snowflake/connector/log_configuration.py
sfc-gh-pczajka: Added `test_create_log_masks_secrets_in_file`: it logs `password=SuperSecret123!` through `snowflake.connector`, flushes, and asserts the raw secret is absent from the file and `password=****` appears exactly once — pinning both the masking guarantee and write-once behavior so they can't silently regress if the handler/formatter wiring changes later.

sfc-gh-fpawlowski (approved):
lgtm

PR #2914, opened 2026-06-25
SNOW-3675580: make WIF CI source-fetch repo-agnostic; drop flaky RT T…

sfc-gh-pczajka:
…est stage

Mirror of snowflake-connector-python-private#24.

ci/test_wif.sh fetched the connector source on the remote WIF workers from a hardcoded snowflakedb/snowflake-connector-python URL, anonymously. Make it repo-agnostic and authenticated so the identical script works in both the public repo and the private mirror:

- Derive owner/repo from GIT_URL (scmInfo) / git remote; each repo fetches itself.
- Fetch via the authenticated GitHub API tarball by commit SHA (api.github.com/repos/$SLUG/tarball/$REF), which survives the codeload redirect that strips the Authorization header; extract with --strip-components=1.
- Thread REPO_SLUG / REF / GITHUB_TOKEN through ssh/docker/kubectl exec.

Jenkinsfile:
- Capture env.GIT_URL in the checkout stage.
- Bind jenkins-snowflakedb-github-app (GITHUB_TOKEN) to the Test WIF stage.
- Drop the RT regression "Test" parallel branch (downstream RT-PyConnector*-PC jobs, chronically Unstable due to infra flakiness; covered by GitHub Actions).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,6 +10,7 @@ timestamps {
       println("${scmInfo}")
       env.GIT_BRANCH = scmInfo.GIT_BRANCH
       env.GIT_COMMIT = scmInfo.GIT_COMMIT
+      env.GIT_URL = scmInfo.GIT_URL
     }
 
     stage('Authenticate Artifactory') {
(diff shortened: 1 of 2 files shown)


sfc-gh-fpawlowski (approved):
LGTM

PR #2916, opened 2026-06-29
SNOW-3650888: Upgrade python prober images

sfc-gh-pczajka:
Upgrade python prober images

- Switch base to ubuntu:26.04 (current LTS, multi-year security support).
- Add `apt-get upgrade -y` so base packages that aren't explicitly reinstalled still receive security fixes on each rebuild.


sfc-gh-fpawlowski (approved):
lgtm

PR #2917, opened 2026-06-29
SNOW-3677751 Add connection option to WIF outbound token

sfc-gh-rsavenok:
Matches Node.js implementation - https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1437

* env variable is kept for backward compatibility
* A new connection option is added
* A proper e2e test is added


sfc-gh-pczajka (approved):
LGTM

PR #2920, opened 2026-07-01
SNOW-3665226: fix FQN DDL cache test to assert unchanged instead of None

sfc-gh-pczajka:
## Description

Post-merge CI on `main` is failing on the newly-added test `test/integ/test_connection.py::test_fqn_ddl_does_not_pollute_schema_cache` across **all** integ jobs (aws/azure/gcp × py3.10–3.14):

```
test/integ/test_connection.py:2707: in test_fqn_ddl_does_not_pollute_schema_cache
    assert cnx._database is None
E   AssertionError: assert 'TESTDB_PYTHON' is None
```

### Root cause

The failing assertion is the **first sanity check**, before any feature logic runs. It was introduced by commit `7150c5b` ("make FQN DDL test assert None cache is not overridden") in #2909, which changed the test to connect with `database=None` and assert `cnx._database is None`.

That assumption only holds when the connecting user has **no server-side default namespace**. All CI test accounts have a database-only default namespace (`TESTDB_PYTHON`), so during authentication the login response's `sessionInfo.databaseName` is `TESTDB_PYTHON` and the connector caches it (`auth/_auth.py`), making `cnx._database` non-`None` at connect time — regardless of the connector fix.

The connector fix in #2909 itself is fine; only the test's precondition is wrong.

### Fix

Revert the test to the environment-agnostic **"unchanged"** style (effectively reverting `7150c5b`):

- connect with `schema=None` only,
- capture the connect-time `_database` (`db_before`),
- run the fully-qualified DDL,
- assert `_schema` stays `None` and `_database == db_before`.

This still exercises the regression — a fully-qualified DDL must not pollute the schema/database cache — without depending on the account having no default namespace.

### DIFF
--- a/test/integ/test_connection.py
+++ b/test/integ/test_connection.py
@@ -2696,20 +2696,14 @@ def test_fqn_ddl_does_not_pollute_schema_cache(conn_cnx, db_parameters):
     schema = db_parameters["schema"].upper()
     view_name = f'"{database}"."{schema}"."SNOW_3665226_REPRO_VIEW"'
 
-    # Open a session with NO database and NO schema so both caches start as None.
-    # The FQN object below lives in `database`.`schema`; without the guard the
-    # connector would overwrite the None caches with the object's namespace. We
-    # start _database as None too (not just _schema) so its "None is not
-    # overridden" path is actually exercised rather than a tautology.
-    with conn_cnx(database=None, schema=None) as cnx:
+    # Open a session with NO schema so the cache starts as None.
+    with conn_cnx(schema=None) as cnx:
         with cnx.cursor() as cur:
-            # Sanity-check: cache and server agree (both empty) before the test.
-            assert cnx._database is None
+            # Sanity-check: cache and server agree before the test.
             assert cnx._schema is None
-            server_db_before, server_schema_before = cur.execute(
-                "SELECT CURRENT_DATABASE(), CURRENT_SCHEMA()"
-            ).fetchone()
-            assert server_db_before is None
+            db_before = cnx._database
+            assert db_before is not None  # database was set at connect time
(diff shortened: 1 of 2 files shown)


test/integ/test_connection.py
```diff
@@ -2696,20 +2696,14 @@ def test_fqn_ddl_does_not_pollute_schema_cache(conn_cnx, db_parameters):
     schema = db_parameters["schema"].upper()
     view_name = f'"{database}"."{schema}"."SNOW_3665226_REPRO_VIEW"'
 
-    # Open a session with NO database and NO schema so both caches start as None.
-    # The FQN object below lives in `database`.`schema`; without the guard the
-    # connector would overwrite the None caches with the object's namespace. We
-    # start _database as None too (not just _schema) so its "None is not
```
sfc-gh-fpawlowski, line 2702: Can we do mocked test at least for scenario that we deleted - the one where DB has no default value? Real integ test would detect changes for scheme and unit mocked test could detect connectors change of the behaviour in terms of database

sfc-gh-fpawlowski (approved):
LGTM, left one comment

test/integ/test_connection.py
sfc-gh-pczajka: Good call — added a mocked unit test in `18db3785`: `test_fqn_ddl_does_not_pollute_none_database_cache`. It drives `cmd_query` with a controlled response and asserts the client-side guard directly for the no-default-database case: (1) an FQN DDL (non-`USE`) must not overwrite a `None` cache, (2) a `USE` context switch still populates it, and (3) once a context is set, `final*Name` values refresh it. Confirmed it fails when the guard is reverted to the pre-fix logic.

PR #2973, opened 2026-07-22
snow-3784431: Part 2 - token cache key v2 fixup

sfc-gh-mhofman:
This is a follow-up change to #3030 and should be merged/released altogether.
  
## This PR
- Moves `token_type` out of the hashed `keyData` into the key prefix:
  `SnowflakeTokenCache.v2.<TokenType>.<sha256hex>`
- Flow-specific `keyData`:
  - OAuth (`OauthAccessToken`, `OauthRefreshToken`, `DpopBundledAccessToken`): 4 fields — `idp`, `role`, `snowflake`, `username`
  - MFA / ID token (`MfaToken`, `IdToken`): 2 fields — `snowflake`, `username` only (`idp`/`role` absent — not part of those auth calls)
- `TokenType` enum values → PascalCase (`MfaToken`, `OauthAccessToken`, …) matching the cross-driver contract
- `normalize_url` / `normalize_identifier` → lowercase instead of uppercase
- `TokenKey.idp` and `role` → optional (default `""`); MFA/ID call sites in `_auth.py` omit them
- `_LEGACY_TOKEN_TYPE_VALUES` added so backward migration still emits the old `SCREAMING_SNAKE_CASE` type strings
## After both PRs
- Key format: `SnowflakeTokenCache.v2.<TokenType>.<sha256hex(compact-sorted-json)>`
- Collisions impossible across accounts (different `idp`), roles, token types, and flows
- MFA/ID and OAuth produce structurally different keys (different prefix + different field set)
- Existing cache entries from both prior layouts transparently migrated on first access
## Test plan
- `test/unit/test_token_cache_key.py` — 32 tests (normalization, two cross-driver golden hash vectors, structural MFA assertions, dimension isolation)
- `test/unit/test_auth_mfa.py` — MFA token cache round-trip
- `test/unit/test_linux_local_file_cache.py` — file cache tests (Linux only)


sfc-gh-mhofman:
Moved to internal repo.

PR #2979, opened 2026-07-27
Recover from stale cached OAuth access tokens without a browser

sfc-gh-turbaszek:
## Summary

OAuth authorization-code (and client-credentials) connections cache access/refresh tokens when `client_store_temporary_credential=True`. Two gaps made a *cached* token fail hard instead of recovering — the observable symptom was `250001 (08001) ... Invalid OAuth access token`, only worked around by removing `client_store_temporary_credential=True`.

### Gap A — proactive silent refresh in `prepare()`
`prepare()` only short-circuited on a cached **access** token. A fresh process holding only a valid **refresh** token went straight to the interactive browser flow. It now attempts a silent refresh with the cached refresh token *before* falling back to interactive auth. Gated on `refresh_token_enabled`, so behavior is unchanged when refresh tokens are off. The async path inherits this via the sync `prepare()`.

### Gap B — reauthenticate on *invalid* (not just expired) tokens
The connector only reauthenticated on GS code `390318` (access token **expired**). A token rejected as **invalid** — GS code `390303`, `"Invalid OAuth access token"` (confirmed in `gs_error_messages.properties`) — fell through to a hard `250001` failure, and the poisoned token stayed cached so **every** subsequent run failed.

- `390303` now triggers reauthentication in both sync and async auth paths, excluding OAuth v1.0 (`AuthByOAuth`), which cannot renew a user-supplied token.
- `reauthenticate()` now evicts the stale access token from the cache on terminal failure via a **lone `remove()`** — never a remove-then-store, which would destroy the macOS Keychain ACL. On success, `_store_tokens()` overwrites the stale token ACL-safely.

## Tests
- **Unit**: proactive refresh (`prepare()`), invalid-token error handling (`390303` -> `ReauthenticationRequest`), access-token eviction on reauth failure / no-eviction on success.
- **Wiremock integration**: no-access-token -> silent refresh; `390303` -> refresh -> retry-login recovery.

All OAuth/auth suites green locally.

## Notes for reviewers
- Async `prepare()`/refresh runs blocking on the event loop — pre-existing (`SNOW-2324426`), not addressed here.
- No JIRA ticket attached; happy to add one.


test/unit/test_oauth_token.py
```diff
@@ -475,6 +475,138 @@ def test_oauth_code_successful_refresh_token_flow(
     assert new_refresh_token == "refresh-token-123"
 
 
+@pytest.mark.skipolddriver
+@patch("snowflake.connector.auth._http_server.AuthHttpServer.DEFAULT_TIMEOUT", 30)
+def test_oauth_code_invalid_access_token_triggers_refresh(
```
sfc-gh-fpawlowski, line 480: Is the same scenario tested / confirmed not-occurring in the UD?

sfc-gh-fpawlowski (approved):
LGTM, left similar UD - related note

test/unit/test_oauth_token.py
sfc-gh-turbaszek: UD backport label added to check in future

PR #2982, opened 2026-07-30
SNOW-3839943: Scale Azure PUT chunk size to respect the 50,000-block limit

sfc-gh-turbaszek:
## Description

Large single-file `PUT` uploads to **internal Azure stages** could fail because the multipart chunk size was hardcoded at **4 MB** (`AZURE_CHUNK_SIZE`) with no upper bound on the number of blocks. Azure Block Blob caps a blob at **50,000 blocks**, so a ~300 GB file needs ~78,600 blocks at 4 MB — over the limit, so the upload **fails outright** rather than just running slowly.

S3 already guards against its own 10,000-part limit via `_chunk_size_calculator`, which scales the part size up for large files. Azure had no equivalent — it just passed the flat `AZURE_CHUNK_SIZE` constant.

### Changes
- Generalize `_chunk_size_calculator` to be **cloud-aware** — parameterized by `default_chunk_size`, `max_parts`, `min_part_size`, and `max_object_size`, all defaulting to the existing S3 values so the S3 call site is behaviorally unchanged.
- Use it for the Azure client (`max_parts = AZURE_MAX_BLOCKS = 50000`) so the chunk size scales up for very large files.
- Raise the default Azure chunk size from **4 MB → 8 MB**, consistent with `S3_DEFAULT_CHUNK_SIZE`, for better and more predictable throughput.
- Add `AZURE_MAX_BLOCKS`, `AZURE_MAX_BLOCK_SIZE`, `AZURE_MAX_OBJECT_SIZE` constants.

### Resulting behavior (Azure)
| File size | Chunk | Blocks | Within 50k? |
|---|---|---|---|
| 300 GB | 8 MB | 38,400 | ✅ (was ~78,600 at 4 MB → **failed**) |
| 400 GB | 8 MB | 50,000 | ✅ |
| 1 TB | ~20 MB (scaled) | 50,000 | ✅ |

S3 path is unchanged (8 MB default, scales to keep ≤ 10,000 parts).

## Testing
`test/unit/test_compute_chunk_size.py`:
- `test_check_chunk_size` — S3 calculator (updated for new error-message wording).
- `test_check_azure_chunk_size` — Azure calculator math: default kept for normal/300 GB files, scaling for very large files, over-ceiling `ValueError`.
- `test_azure_client_wired_with_scaled_chunk_size` — **wiring**: builds the Azure branch of `_create_file_transfer_client` with a mocked `src_file_size` and asserts the constructed `SnowflakeAzureRestClient.chunk_size` scales so the block count stays ≤ 50,000. No network / no live account / no giant file.

> Note: a true 300 GB+ Azure e2e upload is impractical to run in CI (cost/time; Azure integ tests are gated behind `@pytest.mark.azure`). The unit tests cover the deterministic parts — the calculator math and the wiring.

## Scope
This is a pure internal fix — **no public API change**, consistent with how the Go and C/C++ (ODBC) drivers already handle Azure. The secondary ask in the ticket (skipping the SHA256 digest when `overwrite=True`) is intentionally **out of scope** here — the digest also feeds server-side `sfc-digest` integrity, so it needs separate investigation (overlaps SNOW-3766021).

Fixes SNOW-3839943.

### DIFF
--- a/src/snowflake/connector/file_transfer_agent.py
+++ b/src/snowflake/connector/file_transfer_agent.py
@@ -21,6 +21,8 @@ from .compat import IS_WINDOWS
 from .constants import (
     AZURE_CHUNK_SIZE,
     AZURE_FS,
+    AZURE_MAX_BLOCKS,
+    AZURE_MAX_OBJECT_SIZE,
     CMD_TYPE_DOWNLOAD,
     CMD_TYPE_UPLOAD,
     GCS_FS,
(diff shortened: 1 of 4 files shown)


src/snowflake/connector/file_transfer_agent.py
```diff
@@ -186,24 +188,31 @@ def _update_progress(
     return progress == 1.0
 
 
-def _chunk_size_calculator(file_size: int) -> int:
-    # S3 has limitation on the num of parts to be uploaded, this helper method recalculate the num of parts
-    if file_size > S3_MAX_OBJECT_SIZE:
-        # check if we don't exceed the allowed S3 max file size 5 TiB
+def _chunk_size_calculator(
```
sfc-gh-fpawlowski, line 191: Do we have such logic mirrored into UD Azure put gets? Is there a ticket for that?

sfc-gh-fpawlowski (approved):
LGTM, but please make sure UD won't end up out of sync.

src/snowflake/connector/file_transfer_agent.py
sfc-gh-turbaszek: No idea that's why I added backport label

PR #2991, opened 2026-08-05
Fix OAuth IdP detection breaking for uppercase account names

sfc-gh-mborins:
## Summary

- `urlparse().hostname` always returns lowercase, but `self.host` preserves the account name's original casing (e.g. `"PM.snowflakecomputing.com"`). The strict `==` comparison in `_is_snowflake_as_idp` introduced in abbca73b therefore returned `False` for any account with uppercase letters.
- This caused `_eligible_for_default_client_credentials` to fall through to `_validate_client_credentials_present`, which raised `251013: Oauth code flow requirement 'client_id' is empty` even for vanilla Snowflake-as-IdP connections that don't need a `client_id`.
- Fix: `host.lower()` before the comparison — RFC 1123 hostnames are case-insensitive and `urlparse` already normalises to lowercase.

## Test plan

- [ ] New parametrize cases in `test_eligible_for_default_client_credentials_via_constructor` covering `PM.snowflakecomputing.com` (all-caps) and `MyAccount.snowflakecomputing.com` (mixed-case) — both assert `expected_local=True`
- [ ] `pytest test/unit/test_auth_oauth_auth_code.py::test_eligible_for_default_client_credentials_via_constructor` — all 10 cases pass
- [ ] Manual: `snowflake.connector.connect(connection_name="pm")` with `OAUTH_AUTHORIZATION_CODE` and no `oauth_client_id` connects successfully

.... Generated with [Cortex Code](https://docs.snowflake.com/en/user-guide/cortex-code/cortex-code)


sfc-gh-mborins:
The two remaining failures are unrelated to this PR:

- **AI Code Review: security-bot** — Jenkins check, no actionable output visible from outside the internal network.
- **Test macosx_x86_64-3.14-azure** — `ERROR at setup of test_large_query_through_proxy[explicit_args]`: `TimeoutError: WiremockClient did not respond within 12 seconds`. WireMock failed to start on the runner before the test could run. Same infrastructure flake hit macosx_x86_64-3.10 in the previous run (`test_oauth_code_token_request_error`). No test assertion failures in either case; all tests that actually executed passed.

sfc-gh-turbaszek (approved):

PR #2992, opened 2026-08-05
Migrate packaging metadata from setup.cfg to pyproject.toml

sakshamgoyal-01:
> Disclosure:
> Human directed, AI assisted PR

## Summary
- Move packaging metadata from `setup.cfg` into `pyproject.toml` (PEP 621 + setuptools).
- Declare the nanoarrow extension via `[tool.setuptools.ext-modules]` instead of `cythonize()`.
- Resolve version from `snowflake.connector.version.__version__`.
- Make `boto3`/`botocore` opt-in via `.[boto3]` (remove default install and `SNOWFLAKE_NO_BOTO`).

`setup.py` keeps only platform-specific `build_ext` / `build_py` logic. The `if True:` wrapper around `MyBuildExt` preserves the old indentation so the review diff stays easy to read.

## Test plan
- [x] `python ci/anaconda/validate_deps_sync.py`
- [x] Wheel build includes `nanoarrow_arrow_iterator.cpython-*.so`
- [x] `SNOWFLAKE_DISABLE_COMPILE_ARROW_EXTENSIONS=1` omits that compiled module
- [ ] CI installation / build matrix


sakshamgoyal-01:

PR #3027, opened 2026-09-04
SNOW-4058589: retry SSLError "Unexpected EOF" during TLS handshake

sfc-gh-dszmolka:
## Problem
`SnowflakeRestful._request_exec` only retries an `SSLError` when its `repr()` contains `ECONNRESET`. A TLS handshake aborted by the peer surfaces (via pyOpenSSL) as `SSLError("bad handshake: SysCallError(-1, 'Unexpected EOF')")`, which does not match — so it is treated as non-retryable and raised as `OperationalError` (errno 250003).

## Impact
When the peer closes the connection mid-TLS-handshake — a transient, retryable condition — the request fails immediately instead of being retried, surfacing as a hard connection error. Affects both the sync and async request paths.

We got a customer complaint due to this. 

## Fix
Retry when `"Unexpected EOF"` is present in `repr(se)`, alongside the existing `ECONNRESET` check, in `network.py` (sync) and `aio/_network.py` (async).

## Tests
- Added `test_sslerror_with_unexpected_eof_retries` (sync + async), asserting `RetryRequest`.
- Renamed the negative test to `test_sslerror_without_econnreset_or_unexpected_eof_does_not_retry`.
- All `test_retry_network` unit tests pass (43, sync + async).

### DIFF
--- a/src/snowflake/connector/aio/_network.py
+++ b/src/snowflake/connector/aio/_network.py
@@ -816,7 +816,7 @@ class SnowflakeRestful(SnowflakeRestfulSync):
             finally:
                 raw_ret.close()  # ensure response is closed
         except (aiohttp.ClientSSLError, aiohttp.ClientConnectorSSLError) as se:
-            if is_econnreset_exception(se):
+            if is_econnreset_exception(se) or "Unexpected EOF" in repr(se):
                 raise RetryRequest(se.os_error)
             msg = f"Hit non-retryable SSL error, {str(se)}.\n{_CONNECTIVITY_ERR_MSG}"
             logger.debug(msg)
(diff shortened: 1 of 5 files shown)


src/snowflake/connector/aio/_network.py
```diff
@@ -816,7 +816,7 @@ async def _request_exec(
             finally:
                 raw_ret.close()  # ensure response is closed
         except (aiohttp.ClientSSLError, aiohttp.ClientConnectorSSLError) as se:
-            if is_econnreset_exception(se):
+            if is_econnreset_exception(se) or "Unexpected EOF" in repr(se):
```
sfc-gh-turbaszek: Let's make it into `is_unexpected_eof_exception` to keep it aligned

sfc-gh-dszmolka: done!

sfc-gh-turbaszek (approved):

PR #3035, opened 2026-09-15
SNOW-4109042: fail instead of silently returning an incomplete result set

sfc-gh-turbaszek:

   SNOW-4109042 (no public GitHub issue; reported via support)

## The problem

A customer draining a result set with `cursor.fetchmany(100)` received fewer rows than the query produced, with **no exception raised**. Server-side row counts were confirmed (954 produced / 875 received, and 520 / 456) and `ERROR_CODE` was `NULL` on both queries.

Two independent places in the connector turn a failure into a normal end-of-results:

**A downloaded chunk that is shorter than advertised just stops producing rows.** `fetchone`/`fetchmany`/`fetchall`/iteration all treat a stopped iterator as "no more rows", so the caller silently gets a truncated result set. Nothing ever compared the rows yielded against the row count the back-end declared.

**`_fetchone()` caught every `TypeError` and returned `None`:**

```python
except TypeError as err:
    if self._result_state == ResultState.DEFAULT:
        raise err
    else:
        return None
```

`_result_state` is `VALID` during normal fetching, so any `TypeError` from the download/parse chain became `None`, which the fetch methods report as end-of-results. Only one case should be quiet here: `next()` on a `None` iterator after `reset()`.

## The fix

`ResultBatch` now compares a downloaded chunk against the row count the back-end declared for it and raises `OperationalError` (new errno `252013`, `ER_INCOMPLETE_RESULT_CHUNK`) when they disagree. JSON chunks are checked right after parsing; Arrow chunks are counted as they are consumed, so the check fires exactly where the premature EOF used to happen. Local chunks are skipped, since their row count is derived from the data itself rather than reported by the server.

`_fetchone()` now only reports end-of-results when there is no result iterator at all; everything else propagates. The existing `TypeError("'NoneType' object is not an iterator")` behaviour for a never-executed cursor is unchanged (covered by `test_empty_execution`), as is returning no rows after `reset()`.

Both fixes are mirrored in the async cursor and async result batch.

## Tests

- `test_fetchmany_short_remote_chunk_is_not_silent_eof` drives the real connector through Wiremock against a result set that reproduces the ticket's numbers: `total=954`, a remote chunk declaring `rowCount=879` whose gzip body holds only 800 rows. Before the fix `fetchmany(100)` returned 875 rows and stopped; now it raises. A live warehouse cannot produce a truncated chunk, so Wiremock is the only way to cover this end to end.
- `test_short_remote_chunk_raises_instead_of_ending_iteration` covers the batch layer directly, with `test_complete_remote_chunk_iterates_without_error` guarding the normal path.
- `test_fetchmany_typeerror_during_iteration_is_not_eof` pins that an error mid-iteration reaches the caller, and `test_fetchmany_on_reset_cursor_still_reports_end_of_results` pins the one case that must stay quiet.

## Note on the Wiremock fixtures

`select_large_request_successful.json` and `select_qrmk_only_successful.json` declared `rowCount: 4096` per chunk while their stub bodies hold a single row, and `total: 50000` against a one-row inline rowset. The new check correctly flagged them, so their declared counts and totals are now self-consistent. The tests using them (`test_proxies.py`, `test_result_batch.py` and the async twins) assert batch counts rather than row counts and are unaffected; all 26 tests in `test_proxies.py` pass.


sfc-gh-pbulawa:
## Why this commit

Commit `8374f0d8` added the `252013` machinery but validates **downloaded chunks only**. I verified against the six affected production queries that it catches **none** of them, because the rows were lost from the **inline first chunk** of the `query-request` response.

In both reported cases where the client-side count is known it equals `total − first_chunk_rows` exactly. Job telemetry confirms that the result files in storage are complete — the loss is in the HTTP response, after a successful write.

### Why the client was silent

`create_batches_from_response` derives the inline batch's row count by subtraction from server metadata:

```python
first_chunk_len = data.get("total", 0)
for c in rest_of_chunks:
    first_chunk_len -= c.rowcount
```

`ArrowResultBatch.from_data` stores that as the batch's `rowcount`, so on the Arrow path it is **server metadata, not data-derived**. An empty `rowsetBase64` is zero bytes, which nanoarrow reads as a clean end-of-stream, and a stopped iterator is reported by the fetch methods as a normal end of results.

The previous docstring justified exempting local chunks with "a local chunk's `rowcount` is derived from the data itself." That is true for JSON and **inverted for Arrow**, which is exactly why the production cases slipped through.

## What changed

**1. The inline first chunk is now row-counted** (`result_batch.py`). Two exemptions had to go, and the first short-circuited before the second was ever consulted:

- `_iter_checked_rows` returned `rows` unwrapped when `_remote_chunk_info is None`
- `_check_rowcount` early-returned on the same condition

**2. Guard is now `rowcount <= 0` rather than local-vs-remote.** When a response carries no `total`, the subtraction above yields zero or a negative number while the inline rowset genuinely has rows. Enforcing there would fail healthy queries — this is the one real false-positive hazard the old exemption was accidentally covering, and it is the only thing that needs guarding.

**3. New result-set-level check** (`result_set.py`): `check_total_rowcount()` compares rows actually produced against the declared `total` once iteration reaches exhaustion. This is the **only** defence on the JSON path, where `JSONResultBatch.from_data` takes `rowcount` from `len(rowset)` — an empty inline rowset is a self-consistent zero-row batch with no mismatch to detect at batch level.

**4. DML is excluded from the total check** (`cursor.py`, `aio/_cursor.py`): DML reuses `total` for rows *affected*, so an `INSERT` touching 500 rows returns one summary row. Passing `None` for DML avoids a guaranteed false positive on every DML.

**5. Mirrored on the async path** (`aio/_result_set.py`): `generator()` became a counting wrapper around the renamed `_generate()`, and `fetch_all_data()` checks the assembled list.

### Partial consumption cannot false-positive

Both checks sit **past the last `yield`** of a generator. Code after the final `yield` runs only when the consumer drives the generator to `StopIteration`. A consumer that stops early — `fetchmany(10)`, `break`, `cursor.reset()`, `close()`, or just dropping the reference — gets `GeneratorExit` thrown *at* the `yield`, which propagates out of the loop and never reaches the comparison. This is language semantics rather than bookkeeping, and two tests pin it.

Table and pandas units are exempt at the wrapping site: they yield whole tables, not rows, so counting yields would compare "2 tables" against "1000 rows."

## Verification

Same script driving the real response shape (empty inline chunk + healthy remote chunk) through `create_batches_from_response` → `ResultSet` → `fetchmany(100)`, run against both builds:

| | parent `8374f0d8` | this commit |
|---|---|---|
| Arrow, 954/79 | silent, **875 of 954** | `OperationalError` 252013 |
| Arrow, 36/36 | silent, **0 of 36** | `OperationalError` 252013 |
| JSON, all | silent, loses `total − first` | 252013 via result-set check |

The "before" column reproduces the ticket exactly on both formats.

**19 new offline unit tests** (67 passed in `test_result_batch.py` + `test_cursor.py`, up from 48):

- `test_empty_inline_arrow_chunk_raises` — parametrized over all logged pairs
- `test_short_inline_arrow_chunk_raises` — short, not just empty
- `test_empty_inline_json_chunk_is_caught_by_the_total_check` — 875 of 954, every batch self-consistent
- `test_empty_inline_arrow_chunk_with_healthy_remote_chunk_is_loud` — end to end
- Regressions that must stay quiet: `test_complete_result_set_does_not_raise`, `test_empty_result_set_does_not_raise`, `test_result_set_without_declared_total_does_not_raise`, `test_inline_arrow_chunk_without_declared_total_does_not_raise`, `test_fetchmany_stopping_early_does_not_raise`, `test_breaking_out_of_cursor_iteration_does_not_raise`, `test_partially_consumed_inline_arrow_chunk_does_not_raise`, `test_result_batches_are_not_row_counted_against_the_total`, `test_complete_inline_arrow_chunk_iterates_without_error`, `test_empty_inline_json_chunk_is_self_consistent`

Wider run: `test/unit/` gives 1694 passed with 17 failures that reproduce identically on the parent commit (local Arrow extension ABI mismatch, not a regression).

## Reviewer notes

- **`total` is trusted as an exact promise.** The supporting argument is that `total` is already exposed as `cursor.rowcount`, so it would already be misreporting if it could overstate. Reasoned but not proven for `SHOW`/`DESC`, `RESULT_SCAN`, and multi-statement `nextset()`. Note `describe()` never iterates, so a wrong `total` there cannot trigger the check.
- **`get_result_batches()` consumers** get the batch-level checks but no total check — no single object sees the whole set. Pinned by a test so it is a deliberate choice.
- A non-integer `total` would raise `TypeError` from the `>=` comparison rather than `252013`. Not defended against, since `cursor.rowcount` would already be broken in that scenario.

PR #3036, opened 2026-09-16
Preserve Decimal values as numeric literals in client-side binding

### LINKED ISSUE #2455: converter.py: parameter binding for decimal.Decimal produces incorrect SQL
### Python version

3.11.6

### Operating system and processor architecture

macOS-15.5-arm64-arm-64bit

### Installed packages

```python
snowflake-connector-python==3.15.0
snowflake-sqlalchemy==1.7.3
```

### What did you do?

```python
engine = sa.create_engine(...)
with engine.connect() as conn:
    result = conn.execute(sa.select(sa.literal(decimal.Decimal('1.3'))))
    resval = result.fetchone()
    assert resval == '1.3' # True - the SQL executed will be `SELECT '1.3'`, so it will return a string literal.
```

### What did you expect to see?

I expect this query to be executed as `SELECT 1.3`, not `SELECT '1.3'`. The bind parameter value of `decimal.Decimal('1.3')` is not being correctly rendered as a [NUMBER contstant/literal](https://docs.snowflake.com/en/sql-reference/data-types-numeric#numeric-constants). If the `decimal.Decimal` value was being correctly rendered during client-side parameter binding, then a numeric value (either `float` or, preferably `decimal.Decimal`) would be returned, not a `str`.

### Can you set logging to DEBUG and collect the logs?

```bash
import logging
import os

for logger_name in ('snowflake.connector',):
    ch = logging.StreamHandler()
    ch.setLevel(logging.DEBUG)
    logger.addHandler(ch)
```

Shubham-Padkonde:
1. Issue: Fixes #2455.

2. Pre-review checklist:
- [x] New automated tests
- [ ] New logging messages
- [ ] New telemetry messages
- [ ] Authorization changes
- [ ] New credentials
- [ ] OCSP changes
- [ ] New dependency

3. Finite Decimal parameters were converted to strings and then quoted, so Decimal('1.3') produced a string literal. Preserve finite Decimal values until quoting and use their exact decimal representation, without converting through float. Keep non-finite values quoted and server-side binding unchanged. Includes mixed-list, precision, scale, negative-zero and special-value tests plus release notes.

Validation: six regression cases fail before the fix; all 38 converter tests pass afterward. Black, isort and flake8 checks passed. No live Snowflake integration tests were run.

Prepared with Codex assistance.


Shubham-Padkonde:
Merged current main and resolved the release-note conflict by keeping the Decimal binding fix under the current unreleased section alongside the upstream entries. All 38 converter tests pass on Windows/Python 3.13 (`pytest test/unit/test_converter.py --noconftest`); this isolated converter run omits the integration fixtures and reports three missing-plugin configuration warnings. No live Snowflake tests were run. `git diff --check` passes. Prepared with Codex assistance.

